PR #2, opened 2021-04-29
Fix/restrict inputs types for client models

dcfidalgo:
Closes https://github.com/recognai/__rubrix/issues/13

Added change to server side + test.


dcfidalgo:
@frascuchon I removed the test for nested inputs, my understanding was that we do not support these for now, right?

PR #4, opened 2021-04-30
Add active learning tutorial

dcfidalgo:
Ok, this should be good to review.
Here is the link to the build docs to help you with the review process:
https://docs.rubrix.ml/en/feat-active_learning_tutorial/tutorials/05-active_learning.html
This version of the docs is hidden, so you just can access it via the link above.


dvsrepo (changes requested):
This is good to go!

If possible I would replace the screenshot because of the weird grey lines which I could not reproduce on Firefox, and also mention the new Refresh button (if useful for seeing only the remaining default records).

btw, you know if @leiyre has tried to reproduce this glitch with the grey bars?

dcfidalgo:
Ok, made the screenshot with Chrome and added a sentence explaining the "Refresh" button.
https://docs.rubrix.ml/en/feat-active_learning_tutorial/tutorials/05-active_learning.html

>btw, you know if @leiyre has tried to reproduce this glitch with the grey bars?

Nope, did you manage to reproduce this @leiyre ? If Dani was not able to reproduce it, maybe it is a Linux thing (if this makes any sense)??

dvsrepo (approved):

PR #24, opened 2021-05-10
Snapshot dataset name aligned

ignacioct:
As we discussed. I couldnt find any tests regarding snapshots, so I didn't run anything, but it looks good to me.


frascuchon (approved):
💯

PR #35, opened 2021-05-14
Fix Filters selector options filter

### LINKED ISSUE #33: Bug: Search in "Annotated as" not working properly
The search in the "Annotated as" filter does not work properly.

leiyre:
fix #33

### DIFF
--- a/frontend/components/commons/filters/SelectFilter.vue
+++ b/frontend/components/commons/filters/SelectFilter.vue
@@ -93,7 +93,7 @@ export default {
       }
       let filtered = Object.fromEntries(
         Object.entries(options).filter(([id, value]) =>
-          id.match(text.toLowerCase())
+          id.toLowerCase().match(text.toLowerCase())
         )
       );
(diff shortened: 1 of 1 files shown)


frontend/components/commons/filters/SelectFilter.vue
```diff
@@ -93,7 +93,7 @@ export default {
       }
       let filtered = Object.fromEntries(
         Object.entries(options).filter(([id, value]) =>
-          id.match(text.toLowerCase())
+          id.toLowerCase().match(text.toLowerCase())
```
frascuchon, line 96: @leiyre `text` is a selected option? In that case, could we just match the original values?

```javascript 
id.match(text)
```

leiyre: @frascuchon,`text` is any of available options, if we want case sensitive we can use `id.match(text)`. Is it ok case sensitive?

PR #46, opened 2021-05-18
title and video changed on index.rst

ignacioct:
@Amelie-V requested the following changes:

"First steps with Rubrix" instead of "Rubrix Docs"
Rubrix video in full width.
She confirmed the video proportions already.


dcfidalgo (approved):
Nice!!

PR #52, opened 2021-05-19
README update

### LINKED ISSUE #39: Readme update
Hi! I think we've moved enought in the documentation process to update our beautiful README. In my opinion, two main tasks have to be adressed:

- Decide if we change the actual content of the readme (many of this things are now on separate documentation section).
- Complete what's left (some code blocks are placeholders, for example)

So we can discuss here how to approach this. My proposal:

1. First paragraph and quick links stays, but quick links could maybe be at the bottom of the readme.
2. Use cases look cool there, but I think design principles is info overloading. We can either move it to an existing doc page (Introduction feels right to me).
3. Main concepts have their own place now to be. We could maintain a copy here too, I'm down to anything here.
4. Get started coul be a little bit reduced, and placed before main concepts. We would reference the full instructions doc page. I see this **get started** as a **I want to test Rubrix in less than a minute to test something out**, and the docs **get started** as a **that was cool! something cool enough for me I even made it to the documentation**.
5. Supported and planned tasks feels right to me here, we give a glance at we have done and what are we working on from the readme. 

Tell me what do you think 🤠

ignacioct:
Closes #39 , implementing what we talked about there.


dcfidalgo (approved):
Perfecto! thanks!

PR #56, opened 2021-05-19
image urls changed

### LINKED ISSUE #43: URL not found for images
When clicking on the images in the docs you get to a not found url. There are several solutions:

1. remove the ability to click on the image.
2. maybe there's an sphinx-contrib for enlarging/full-screen the image.
3. Fix the path

I don't know which one is easier to do, maybe 2 is the most difficult, I let you decide. In general I would avoid sending the users to other urls and leaving the docs, even if it's our image

ignacioct:
Closes #43


dcfidalgo (approved):
Nice!! Thanks a lot

PR #74, opened 2021-05-21
feat(middleware): starlette middleware for rubrix logging

frascuchon:
The rubrix log middleware tries to simplify workflow for http model inference with rubrix records logging.

Notes:

- The default inputs/outputs -> record mapping is for text classification with a fixed  `labels` - `probabilities` format
- Mappings are customizable, but a inputs/outputs norrmalization depending on model task should be a better option
- The rubrix log is a non-blocking process, so response latency should not be affected

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -22,6 +22,7 @@ install_requires =
   # Client
   pandas >=1.0.0,<2.0.0 # For data loading
   pydantic ~= 1.8.1
+  starlette >=0.13.0,<1.0.0
 
 [options.packages.find]
 where = src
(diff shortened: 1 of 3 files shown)


dcfidalgo (approved):
LGTM!! But there is a good chance that i still don't fully understand the use case :)

frascuchon:
> LGTM!! But there is a good chance that i still don't fully understand the use case :)

The main idea of use case is avoid call rubrix log when you have a mlmodel running a a web service for data inference.

ignacioct (approved):
It also looks good to me, nice work!

PR #80, opened 2021-05-31
Cookbook/CheatSheet addition

ignacioct:
Hi! This is more or less the cookbook. There are still a few missing pieces

- [x] Deciding if we make both NER and POS examples on Spacy.
- [x] I've made both NER and POS examples for Stanza and Flair, for us to have variety. Now we can have them all, or keep just one. I think this decisition will be aligned with the previous. 

For the rest, take a look at the text and snippets. Looking forward for your feedback 😄


ignacioct:
Merging now, I've checked all typos. If anyone escaped to me, we will se it and address it in a separate issue.

PR #83, opened 2021-06-02
feat(text-classification): allow pass string or string lists as inputs

frascuchon:


ignacioct (approved):
Looks good to me

src/rubrix/client/models.py
```diff
@@ -89,7 +89,7 @@ class TextClassificationRecord(BaseModel):
             The timestamp of the record.
     """
 
-    inputs: Dict[str, Union[str, List[str]]]
+    inputs: Union[Union[str, List[str], Dict[str, Union[str, List[str]]]]]
```
dcfidalgo: I think there is one `Union` too many
```suggestion
    inputs: Union[str, List[str], Dict[str, Union[str, List[str]]]]
```

dcfidalgo (approved):
LGTM, just a small inline suggestion.

PR #127, opened 2021-06-11
Allow refresh query if exploration mode

leiyre:
Show refresh button


frontend/components/commons/GlobalActions.vue
```diff
@@ -76,6 +81,10 @@ export default {
       type: Object,
       required: true,
     },
+    annotationEnabled: {
```
frascuchon, line 84: passing dataset to component should be enough for calculate this param. I will create an tech-debt issue for refactoring components and dataset getter fields.

frascuchon (approved):

PR #148, opened 2021-06-18
Fix/init error

### LINKED ISSUE #147: Missleading error in rb.init()
Let's avoid 
`raise Exception("Unidentified error, it should not get here.")`

Showing info about the error response (status, body, etc.)

ignacioct:
Closes #147 

Alligned the error message with errors from rb.delete


frascuchon (approved):
Great!

PR #192, opened 2021-07-05
Annotation guide

### LINKED ISSUE #172: Annotation Guide using Misogyny Use Case
This guide will serve as a guide for:

- Training an AMI system with IberEval 2018's dataset
- Annotating and exporting annotations
- Retraining 
- Inter-Annotator Agreement

To simplify the system for the guide, we will exclude Subtask 3 (target) and make Multiclass, but not Multilabel, of Subtask 2 (category). We also need to make the datasers public, the initial and the annotated ones, to make the guide replicable.

Even though it is not a tutorial, it is intended to use runnable code as much as possible.

ignacioct:
Closes #172 

Still a draft, looking for feedback @dvsrepo.


dvsrepo (changes requested):
cool!

One minor suggestion:

"We wanted not only to enlarge our numbers but to place ourselves inside the Deep Learning pipeline, as women and men in the middle, analyzing the output of our first models, searching their weaknesses, and trying to enforce them. We truly believe in Data Science as a non-linear process in which monitoring obtained models and iterating through them for improvement is the right thing to do."

I would simplify this paragraph, a few suggestions:

- We wanted not only to enlarge our numbers ... --> Besides improving our model's performance we wanted to explore a development lifecycle with Rubrix, where models can be improved over time, with different roles participating... 
- as women and men in the middle --> humans in the loop
- ....them for improvement is the right thing to do --> ...for improvement is key 
- We truly believe in Data Science as a non-linear process --> We believe data science is an iterative process.
-

ignacioct:
@leiyre already has the file, so I'm closing it now.

frascuchon:
So, can we delete the branch?

ignacioct:
> So, can we delete the branch?

Yes, I've saved the notebook just in case.

ignacioct:
Done!

PR #197, opened 2021-07-08
build: apply eslint before build statics

frascuchon:
A pre-commit hook will be applied in future versions, but for now, these changes are quite enough


frontend/models/Notifications.js
```diff
@@ -5,7 +5,7 @@ class Notification extends Model {
 
   static fields() {
     return {
-      id: this.uid(() => uuid()),
+      id: this.uid(),
```
frascuchon: I think we must keep as is

frontend/components/commons/filters/SelectFilter.vue
```diff
@@ -122,7 +122,7 @@ export default {
         return options;
       }
       let filtered = Object.fromEntries(
-        Object.entries(options).filter(([id, value]) =>
+        Object.entries(options).filter((id) =>
```
frascuchon: This  change breaks entry decomposition and should be kept

frontend/components/commons/Metadata.vue
```diff
@@ -91,7 +91,7 @@ export default {
       let meta = { ...this.metadata };
       meta[key] ? (meta[key] = undefined) : (meta[key] = value);
       this.metadata = Object.fromEntries(
-        Object.entries(meta).filter(([key, value]) => value)
+        Object.entries(meta).filter((value) => value)
```
frascuchon: This change breaks entry decomposition ...

leiyre (approved):

PR #214, opened 2021-07-14
Feat: similarity vector

dcfidalgo:
This PR adds a `record_embedding` parameter to the `TextClassificationRecord` model, to be used in cosine-similarity searches.
It also adds a check to the `rb.log` method, but I feel like we should catch the error code of the API in this case, rather than checking this explicitly before.

This PR is a work in progress, the server side, as well as the UI component is still missing.

### DIFF
--- a/src/rubrix/client/__init__.py
+++ b/src/rubrix/client/__init__.py
@@ -94,26 +94,21 @@ class RubrixClient:
         """
         Register a set of logs into Rubrix
 
-        Parameters
-        ----------
-        records:
-            The data records list.
-        name:
-            The dataset name
-        tags:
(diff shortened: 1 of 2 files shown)


frascuchon:
`dense_vector` is a free x-pack extension. Elasticsearch OpenDistro does not support this kind of field but `knn_vector`.

The way we must apply similarity search will depends on which es distro rubrix is using. This became a bit tedious if we want maintain both.

We must make a decision of which es version we will support first.

dcfidalgo:
I will close this PR. Like @frascuchon said, we first must have a discussion about the ES distribution.

PR #241, opened 2021-07-26
Tutorial: 🧐 Using Rubrix to find label errors with cleanlab 

dcfidalgo:
This PR adds a small tutorial about finding label errors with cleanlab and rubrix.
You can have a look at it here: https://docs.rubrix.ml/en/docs-find_label_errors_tutorial/tutorials/07-find_label_errors.html
It's a rather short tutorial (which is not necessarily a drawback), but I think it is quite straightforward in its structure ... maybe a little diagram would be a nice addition, but not sure what to put.


dvsrepo (changes requested):
Looks really good to me! We could add a diagram on a second review.

Before merging:

- do a quick proof reading, for example persuasive > pervasive.
- normalize the section about Rubrix setup (see for example the latest fine-tune huggingface tutorial)

dvsrepo:
👍

PR #251, opened 2021-07-29
Remove snapshots

dcfidalgo:
This PR addresses #238 . There are two small issues:

1. we extensively use this feature in the kglab tutorial (not addressed in this PR)
2. our [concepts diagram](https://docs.rubrix.ml/en/master/getting_started/concepts.html#rubrix-data-model) refers to snapshots 

Maybe it would be better to leave this feature for now and just replace it with the new one when ready, @frascuchon @dvsrepo what do you think?


frascuchon:
We could:

1. Change the kglab tutorial by using `rb.load` and apply filters to dataframe for obtain the same results, and
2. Remove all snapshot references in documentation.


@dcfidalgo Could you please address 1. 
@dvsrepo could you upload diagram. Then, I can revise and update documentation.

frascuchon:
The kglab tutorial is already updated

frascuchon:
> The kglab tutorial is already updated

Also de data model diagram

PR #254, opened 2021-07-30
Supported Tasks Example Article

### LINKED ISSUE #253: Example of Supported Tasks
An article with explicit examples about all supported tasks, along with a revision of the supported task list.

ignacioct:
Closes #253


ignacioct:
@dvsrepo @Amelie-V  do we really have the textual similarity task supported? In the case of the paraphrasing style of MRPC datasets, it is a yes, because they work as normal labels with a confidence value. But I've also seen some researches using models to predict the raw value of closeness between two sentences (what I did in my last commit), and that is usually what they call Semantic Textual Similarity. 

For that case, I think we have no support. I was able to introduce tuples like ("similarity": 0.5) to Rubrix, but they cannot be annotated (you cannot manually insert a value). It feels a little bit hacky, maybe we could add the similarity task to the roadmap and give it special support. 

My suggestion: change the name of the currently supported task to Paraphrasing Identification or something like that, and add Semantic Textual Similarity (following, for example, this information https://paperswithcode.com/task/semantic-textual-similarity) to our roadmap.

dvsrepo:
Semantic textual similarity can be represented in different ways, and we do
not support all. For the examples, we were refering to the mrpc dataset,
which is a form of semantic textual similarity:

https://paperswithcode.com/task/semantic-textual-similarity

So, let's use mrpc for this

docs/guides/tasks_examples.ipynb
```diff
@@ -0,0 +1,286 @@
...
+    "rb.log(\n",
+    "    records=records,\n",
+    "    name=\"textual-similarity\",\n",
+    "    tags={\n",
+    "        \"task\": \"similarity\",\n",
+    "        \"dataset\": \"assin\",\n",
+    "    },\n",
+    ")"
+   ],
+   "outputs": [
```
frascuchon: We can  just remove this cell output

ignacioct: I'm planning on removing all cell outputs on the final commit

ignacioct:
> Semantic textual similarity can be represented in different ways, and we do not support all. For the examples, we were refering to the mrpc dataset, which is a form of semantic textual similarity: https://paperswithcode.com/task/semantic-textual-similarity So, let's use mrpc for this El lun., 2 ago. 2021 13:37, Ignacio Talavera ***@***.***> escribió:
> […](#)

Done! Thank you so much

ignacioct:
@dvsrepo about Relationship Extraction, I think it fits better on Token Classification. There are three main datasets available at HuggingFace, https://huggingface.co/datasets?filter=task_ids:other-other-relation-extraction, and they represent a relation between two entities/tokens inside the phrase. 

I can pick any of these datasets and apply some kind of zero-shot or NER-like classification.

dvsrepo:
This we can skip for now as it's not really supported, it could be done as
a mix of ner and text classification.

ignacioct:
> This we can skip for now as it's not really supported, it could be done as a mix of ner and text classification. El mar., 3 ago. 2021 18:09, Ignacio Talavera ***@***.***> escribió:
> […](#)

Should I move it to pending tasks in the Task list?

dvsrepo:
Yes

ignacioct:
Just two to go now (key-phrase and slot-filling)! I've made a quick grammatical correction, so the things written here should be close to final. 

For the graph neural network example, I wasn't able to find a self-contained recipe that doesn't involve running several cells and explaining them. I decided to directly link the tutorial, which is self-contained as it is. However, I can structure a recipe and use it, let me know your opinion.

ignacioct:
I have doubts about the two missing tasks. I will go one by one.

In keyphrase extraction, I've followed Paco Nathan's https://spacy.io/universe/project/spacy-pytextrank tutorial, adapting it to spaCy 3. However, I don't know how to integrate as TokenClassification Record. We obtain a huge list of important words, their score, the number of appearances on the text, etc. How shall I upload them? I thought of tuples (ranking, start chat, end char) but I think it's kind of hacky, and there are going to be a lot of entities. The spacy cells are committed if you wanna try them.

For slot-filling, I'm not finding examples on Hugging or spaCy. I see it as an extension of NER, so it will probably reuse code if we stick to spaCy. I will continue to research, but if you have ideas, they are more than welcome.

frascuchon (approved):

dvsrepo:
Hi Ignacio, skip and remove keyphrase extraction. For slot filling, use our
biome text tutorial.

ignacioct:
I've just uploaded the Slot Filling part. I will go seeking for typos and checking the style now, but the content should be now finished. 

Take into account that, for the slot filling part, several instances of the test set give alignment problems in Rubrix. Taking a slice o [0:10] none of them appears, but if you try the whole dataset, they might appear. I've not been able to track its cause., but given its spontaneous nature, I think the problem is in the dataset.

Thanks all for your support on this guide 😊.

docs/getting_started/supported_tasks.rst
```diff
@@ -24,7 +24,7 @@ Rubrix is flexible with input and output shapes, which means you can model many
 * `Key phrase extraction <https://paperswithcode.com/task/keyword-extraction>`_
```
dvsrepo: Very good job @ignacioct . I'd remove this from here, I think I added this accidentally.

dvsrepo (approved):

docs/getting_started/supported_tasks.rst
ignacioct: Done!

PR #263, opened 2021-08-05
feat(app): copy datasets feature

frascuchon:
This PR addresses discussions from [this issue](https://github.com/recognai/rubrix/issues/136)

Dataset copy internally uses  elasticsearch [index clone](https://www.elastic.co/guide/en/elasticsearch/reference/7.13/indices-clone-index.html), that requires source index in read-only mode. Logging records into a dataset that is being copied could lost data. 

The clone index feature is include since version 7.4.x, so rubrix instances using older elasticsearch versions won't support this new feature


src/rubrix/__init__.py
```diff
@@ -115,6 +115,20 @@ def log(
         records=records, name=name, tags=tags, metadata=metadata, chunk_size=chunk_size
     )
 
+def copy(dataset:str, dest:str):
+    """Copies a dataset an all its tags and metadata into another dataset
```
dcfidalgo:
```suggestion
    """Creates a copy of a dataset including its tags and metadata
```

src/rubrix/__init__.py
```diff
@@ -115,6 +115,20 @@ def log(
...
+
+    Args:
+        dataset:
+            The source dataset
```
dcfidalgo:
```suggestion
            Name of the source dataset
```

src/rubrix/server/datasets/api.py
```diff
@@ -183,3 +182,38 @@ def open_dataset(
...
     service.open_dataset(name, owner=current_user.current_group)
+
+
+@router.put(
+    "/{name}:copy",
+    operation_id="copy_dataset",
+    response_model=Dataset,
+    response_model_exclude_none=True,
+)
+def open_dataset(
```
dcfidalgo: Shouldn't this be called `copy_dataset` or something like this?

src/rubrix/server/datasets/api.py
```diff
@@ -183,3 +182,38 @@ def open_dataset(
...
+    name: str,
+    copy_request: CopyDatasetRequest,
+    service: DatasetsService = Depends(create_dataset_service),
+    current_user: User = Security(auth.get_user, scopes=[]),
+) -> Dataset:
+    """
+    Closes a dataset. This operation will releases backend resources
```
dcfidalgo: Is this the right doc string? If so, maybe the method should be called `close_dataset`?

src/rubrix/__init__.py
```diff
@@ -115,6 +115,20 @@ def log(
         records=records, name=name, tags=tags, metadata=metadata, chunk_size=chunk_size
     )
 
+def copy(dataset:str, dest:str):
```
dcfidalgo: I think I prefer something more explicit for the `dest` parameter. Maybe `name_of_copy`, I think most of the time you will use this method without kwargs, that is simply `rb.copy("dataset", "dataset_copy")`. I would also provide a default value `None`, in which case the name of the copy would be something like `{source_name}_copy`.

src/rubrix/__init__.py
```diff
@@ -115,6 +115,20 @@ def log(
...
+    """Copies a dataset an all its tags and metadata into another dataset
+
+    Args:
+        dataset:
+            The source dataset
+        dest:
+            The destination dataset name
```
dcfidalgo:
```suggestion
        name_of_copy:
            Name of the copied dataset
```

tests/client/test_client.py
```diff
@@ -241,6 +242,29 @@ def test_delete_dataset(monkeypatch):
         rubrix.load(name=dataset_name)
 
 
+def test_dataset_copy(monkeypatch):
```
dcfidalgo, line 241: What happens if the name of the copy already exists? Is it worth testing?

dcfidalgo (commented):
Nice, LGTM! Just a few minor inline comments.

src/rubrix/server/datasets/api.py
frascuchon: yes, done

src/rubrix/__init__.py
frascuchon: done

frascuchon: > I would also provide a default value None, in which case the name of the copy would be something like {source_name}_copy

I prefer to keep as is. Making this default behaviour and so own could bring extra logic. What about if already exists a `{source_name}_copy` dataset? client could be asign a new `{source_name}_copy2` name, and so on.

Once this method is validated and we're sure it will be part of client API for a while, we can improve with proposal like yours.

tests/client/test_client.py
frascuchon: An exception should be launched as default behaviour. I will test the case.

In future method versions, we can include  boolean `force_recreate`, or `overrides`, to configure default behaviours in those cases.

dcfidalgo (approved):
Perfecto

Issue #129, opened 2021-06-11
Document elasticsearch field for Elasticsearch DSL and Kibana


dvsrepo:
Add this somewhere in the docs (probably in the Rubrix UI guide):

// Commons search fields
`annotated_as`
`annotated_by`
`event_timestamp`
`id`
`last_updated`
`metadata.*`
`multi_label`
`predicted`
`predicted_as`
`predicted_by`
`status`
`words`
// Text classification
`inputs.*`
`score`
// Token classification
`tokens`

frascuchon:
Including a breaf description of query string format and reserved characters and importance of escape them. 
https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html#_reserved_characters

See #211

dcfidalgo:
In my opinion, the default behavior (without the need of prepending anything) of the search field should allow for:
- searching for literal strings
- searching for stop words (i think this can be very confusing)
- no need to escape characters

I am not sure if this was the behavior we had a while ago, but I think it was more similar to this one. Maybe one could add a little checkbox that allows one to switch on the advanced ES search options, just an idea.

frascuchon:
> In my opinion, the default behavior (without the need of prepending anything) of the search field should allow for:
> 
> * searching for literal strings
> * searching for stop words (i think this can be very confusing)
> * no need to escape characters
> 
> I am not sure if this was the behavior we had a while ago, but I think it was more similar to this one. Maybe one could add a little checkbox that allows one to switch on the advanced ES search options, just an idea.

Several things here.

* About stopwords, we could include a multiple search, include fields with stopwords filters and text without them, but giving more relevance to fields without stopwords (more general results and better behaviour for common searches)

    ```
    search: example  in [field_with_stopwords*2, field_without_stopwords]
    ```
    where `*2` means that matched results in `fields_with_stopwords` will be an `x2` in search score, what's mean that they will be more relevant in results

* I agree with confusing behaviour using "special" characters and stop words, but I would like to reduce the search logic and user interaction. and enabling/disabling an advance mode doesn't make me confortable with. 

    Elasticsearch query string operations are quite powerful and could upgrade data exploration from conventional search text, even for default search (without prepending).

    Searching `women OR man` (equivalent to `women || man`)  or `!king` could be search queries that a standard user could easily do, if we provided the concise documentation for that.

[...]

PR #269, opened 2021-08-11
docs(ui): include special chars query reference

frascuchon:
see #129


frascuchon:
@dvsrepo  I will include this PR in next release. We can discuss better content after

PR #275, opened 2021-08-16
refactor(search): simplify query text search model

frascuchon:
Simplify `query_text` input for search, accepting only string values


dcfidalgo (approved):
LGTM, as far as i understand ... :)

PR #276, opened 2021-08-16
Docs: Add advanced setup guides

dcfidalgo:
This PR adds a section to the docs called *Advanced setup guides*. It includes the setup with vanilla docker, running on AWS, configuring ES with other applications, and the users management. I left various references to the advanced guides where it made sense.
See the docs here: https://docs.rubrix.ml/en/feat-revise_docs/getting_started/advanced_setup_guides.html


docs/getting_started/advanced_setup_guides.rst
```diff
@@ -0,0 +1,273 @@
...
+
+
+.. _users-management:
+
+Users management
+----------------
+
+The Rubrix server allows you to manage various users, which helps you to keep track of the annotation agents.
+
+Default user credentials
```
dvsrepo: This should be indicated in other places too (maybe only the username and password): readme, first steps and setup and installation?)

docs/getting_started/advanced_setup_guides.rst
```diff
@@ -0,0 +1,273 @@
...
+   :$2y$05$T5mHt/TfRHPPYwbeN2.q7e11QqhgvsHbhvQQ1c/pdap.xPZM2axje
+
+then, configure the environment variable omitting the first ``:`` character:
+
+.. code-block:: shell
+
+    export RUBRIX_LOCAL_AUTH_DEFAULT_PASSWORD="<generated_user_password>"
+
+
+Create a user db file
```
dvsrepo: All this section still seems a series of steps to follow but I think is not the case. For example, you dont need to override the default user name if you want to add more users or you don t need to use the docker compose if you launch the python server directly. I'd recommend more explicit headings, like  how to add new users, how to override the default user, and then subsections: launching the server, using docker, etc.

dvsrepo (changes requested):
Looks good! Only some comments to tackle

frascuchon:
Very nice!! 

I'm just missing docs, like @dvsrepo mentions, when an user wants just create a users team, without advanced configuration stuff

docs/getting_started/advanced_setup_guides.rst
dcfidalgo: Done

docs/getting_started/advanced_setup_guides.rst
dcfidalgo: Done, I restructured this part a bit: https://docs.rubrix.ml/en/feat-revise_docs/getting_started/advanced_setup_guides.html#users-management

dcfidalgo:
> Very nice!!
> 
> I'm just missing docs, like @dvsrepo mentions, when an user wants just create a users team, without advanced configuration stuff

Something like this? https://docs.rubrix.ml/en/feat-revise_docs/getting_started/advanced_setup_guides.html#how-to-add-new-users

dvsrepo:
💯 very good job!

dvsrepo (approved):

frascuchon:
Yinial!!!

PR #278, opened 2021-08-18
feat(task): new text2text task api and client

frascuchon:
This is a WIP PR providing a task (just in API) for text2text tasks (machine translation, sumarization,...). Once this PR is discussed and approved, we can make efforts to bring the integration with client and UI.

The api record data model proposed is (common record properties are omitted):
```yaml
---
title: Text2TextAnnotation
description: |-
  Annotation class for text2text tasks
required:
- agent
- text
- score
type: object
properties:
  agent:
    title: Agent
    type: string
  text:
    title: Text
    type: string
  score:
    title: Score
    maximum: 1
    minimum: 0
    type: number
---
title: Text2TextRecord
description: The output text2text task record
required:
- text
type: object
properties:
  text:
    title: Text
    type: string
  prediction:
    "$ref": "#/components/schemas/Text2TextAnnotation"
  annotation:
    "$ref": "#/components/schemas/Text2TextAnnotation"
```

For search queries, `predicted_as` and `annotated_as` fields are removed and 2 new fields are available for text search on those fields `text_predicted` and `text_annotated` .

### DIFF
--- a/src/rubrix/__init__.py
+++ b/src/rubrix/__init__.py
@@ -28,6 +28,16 @@ _client: Optional[
 ] = None  # Client will be stored here to pass it through functions
 
 
+def _client_instance() -> RubrixClient:
+    """Checks module instance client and init if not initialized."""
+
+    global _client
+    # Calling a by-default-init if it was not called before
+    if _client is None:
+        init()
+    return _client
+
+
 def init(
     api_url: Optional[str] = None,
     api_key: Optional[str] = None,
@@ -135,16 +145,6 @@ def copy(dataset: str, name_of_copy: str):
         >>> df = rb.load("new_dataset")
     """
     _client_instance().copy(source=dataset, target=name_of_copy)
-
-
-def _client_instance() -> RubrixClient:
-    """Checks module instance client and init if not initialized."""
-
-    global _client
-    # Calling a by-default-init if it was not called before
-    if _client is None:
-        init()
-    return _client
 
 
 def load(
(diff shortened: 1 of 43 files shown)


dvsrepo (commented):
This looks awesome!!!

My only doubt is about having the score as a required field in the Annotation model. I am not sure this is something that's available for every text to text model. Maybe @dcfidalgo can do some experimentation with Huggingface summarisation and translation models. I guess for some models you could have a score.

frascuchon:
> This looks awesome!!!
> 
> My only doubt is about having the score as a required field in the Annotation model. I am not sure this is something that's available for every text to text model. Maybe @dcfidalgo can do some experimentation with Huggingface summarisation and translation models. I guess for some models you could have a score.

We could include `0.0` as default value? or should be better a `None` value on those cases? I'm not sure to figure out the better approach about the use cases

dvsrepo:
I'm not sure either, maybe even a default of 1.0? What does @dcfidalgo think? I'd say None is the less desirable. We can in any case review this when we do some experiments

frascuchon:
For text classification we're using `1.0` as default value. We can follow the same approach

dcfidalgo:
It seems that in the HF transformers library you can get a score for each generated output sequence, but I am not sure how this score is computed and what its range is. In my translation tests, it had always negative scores, but not sure if this is representative ... I think we can leave our range from 0 to 1 since it is a relative score, but we should clearly state, that 1 means best, and 0 worst, and keep 1 as default.

frascuchon:
I've used `1.0` as default value for score. I'm still keeping score range between `0` and `1`

PR #288, opened 2021-08-27
refactor(panel): load side bar task comp. dynamically

frascuchon:


frascuchon:
Already integrates in master from PR #293

PR #300, opened 2021-09-02
docs: Document Elasticsearch fields

dvsrepo:


dvsrepo:
see: https://rubrix.readthedocs.io/en/docs-es_fields_documentation/reference/rubrix_webapp_reference.html

docs/reference/rubrix_webapp_reference.rst
```diff
@@ -1,7 +1,7 @@
 .. _webapp_reference:
 
-Rubrix UI
-=========
+Rubrix webapp UI
+================
```
dcfidalgo: I think it is more common to write "web app" (separated), a quick google search prefers this notation. Maybe we could drop Rubrix (it's kind of obvious) and just write:
```suggestion
Web App UI
==========
```

dcfidalgo (approved):
LGTM, just a small inline comment. I pushed one commit renaming the `Python Client API` heading to simply `Python Client`. I really like the structure of the [wandb reference](https://docs.wandb.ai/ref) section, really simple and concise.

PR #307, opened 2021-09-07
[UI] Review sort component

### LINKED ISSUE #302: Review sort component
Only some minor improvements:

- The component should work as the filters: indicating the number of active sorting fields: for example (1) if one field is selected for sorting.
- The dropdown list elements should have the hand pointer on hover, not the arrow pointer (maybe we could review all other dropdowns, I remember having the same issue with another dropdown.
-  What happens when we have a long list of fields? If possible I'd enable search as well for this dropdown.
- What happens when the users define a metadata field with the same name as one of the reserved fields? e.g., `metadata.Status` (I think this is possible @frascuchon ?) In this case we'll have two Status on the list.
- Add more top and bottom padding to the dropdown box. 

As a broader comment, our URL params look a bit ugly with the encoding, ideally we should have clear text for the param values,  I'm not sure if this is something we could fix with nuxtjs config or other solution.

leiyre:
fix #302


frascuchon:
@leiyre is this PR ready or still in progress?

leiyre:
@frascuchon we can merge it

frascuchon (approved):
cool!

PR #311, opened 2021-09-09
UI/fix pagination component

### LINKED ISSUE #308: Review pagination component

leiyre:
fix #308


frascuchon (approved):
Nice work!!

arrow keys for pagination is very cool.

Just minimal proposal that could be discussed on #308 

- Use a "Loading animation" when page changes. Long query responses could works fine with for records refresh. 
- "Record: x of y" message when records per page = 1 instead of "Records: x-x of y"
- Limit token classification records per page to 20. When 50, page becomes useless

PR #318, opened 2021-09-10
Fix/explain legend styles

leiyre:


frascuchon:
Maybe Help button should be previous to annotation actions? @Amelie-V @leiyre ??
Screenshot 2021-09-10 at 16 28 57

frascuchon:
Once changes proposed in #302 are made:

> Group Sort feature with other filters + add icon

we can place Help button in sort button old place

frascuchon (approved):

PR #327, opened 2021-09-13
Refactor: 'confidence' to 'score'

### LINKED ISSUE #322: chore: Replace confidence by score everywhere
There are still places where we use confidence instead of the more general `score`. Coordinate with Paco if the API needs to be changed.

dcfidalgo:
This PR changes the naming from 'confidence' to 'score'. Closes #322 

@dvsrepo Could you quickly approve the changes in the tutorials, i tried to replace all references to 'confidence' with either 'score' or 'prediction score'.
@frascuchon I think we still have to change the static json files in the server folder, right?
@leiyre To be honest i am not sure what has to be done for the web app to remove all 'confidence' references ... :)


frascuchon:
I've made changes to frontend part with backward compatibility (allow to see current datasets).

@leiyre please verify i'm not missing something.

Thanks!

PR #378, opened 2021-09-23
feat(UI): new text2text UI

leiyre:


frascuchon (commented):
License header should be include as part of generated .js files.

PR #384, opened 2021-09-27
feat(UI): format quantities locale dependant

### LINKED ISSUE #363: [next-release] Use thousand separator big numbers
We should review which numbers larger than 1000 we have in the UI and add a decorator to show thousand/million separators.

For example:

- total number of records
- total annotations
- annotations by label

See:

Screenshot 2021-09-17 at 15 05 14

leiyre:
fixes #363

### DIFF
--- a/frontend/plugins/filters.js
+++ b/frontend/plugins/filters.js
@@ -1,7 +1,12 @@
 import Vue from 'vue'
 
-var numeral = require("numeral");
-
 Vue.filter("formatNumber", function (value) {
-  return numeral(value).format("0,0"); // displaying other groupings/separators is possible, look at the docs
+  const locale = [
+    ...(navigator.languages || []),
+    navigator.language,
+    navigator.browserLanguage,
+    navigator.userLanguage,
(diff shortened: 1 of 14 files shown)


frontend/plugins/filters.js
```diff
@@ -0,0 +1,7 @@
+import Vue from 'vue'
+
+var numeral = require("numeral");
+
+Vue.filter("formatNumber", function (value) {
+  return numeral(value).format("0,0"); // displaying other groupings/separators is possible, look at the docs
```
frascuchon: We should accept language dependant formatting. See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat

frascuchon (changes requested):

frontend/plugins/filters.js
leiyre: I have included format based on locale

frascuchon (approved):

PR #385, opened 2021-09-27
Feat: score per entity

dcfidalgo:
WIP, tests are still missing.
@frascuchon but maybe you could have a quick look if I am going in the right direction with the server stuff.

### DIFF
--- a/src/rubrix/client/__init__.py
+++ b/src/rubrix/client/__init__.py
@@ -378,7 +378,7 @@ class RubrixClient:
             status=record.get("status"),
             metadata=record.get("metadata") or {},
             prediction=[
-                (entity["label"], entity["start"], entity["end"])
+                (entity["label"], entity["start"], entity["end"], entity["score"])
                 for entity in record["prediction"]["entities"]
             ]
             if record.get("prediction")
@@ -408,10 +408,13 @@ class RubrixClient:
             "status": record.status,
         }
         if record.prediction is not None:
-            entities = [
-                {"label": pred[0], "start": pred[1], "end": pred[2]}
-                for pred in record.prediction
-            ]
+            entities = []
+            for pred in record.prediction:
+                try:
+                    ent = {"label": pred[0], "start": pred[1], "end": pred[2], "score": pred[3]}
+                except IndexError:
+                    ent = {"label": pred[0], "start": pred[1], "end": pred[2]}
+                entities.append(ent)
             model_dict["prediction"] = {
                 "agent": record.prediction_agent or RubrixClient.MACHINE_NAME,
                 "entities": entities,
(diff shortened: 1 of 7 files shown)


src/rubrix/client/__init__.py
```diff
@@ -405,10 +405,13 @@ def _token_classification_client_to_sdk(
...
-                {"label": pred[0], "start": pred[1], "end": pred[2]}
-                for pred in record.prediction
-            ]
+            entities = []
+            for pred in record.prediction:
+                try:
+                    ent = {"label": pred[0], "start": pred[1], "end": pred[2], "score": pred[3]}
+                except IndexError:
+                    ent = {"label": pred[0], "start": pred[1], "end": pred[2]}
+                entities.append(ent)
```
frascuchon:
```suggestion
                ent = {"label": pred[0], "start": pred[1], "end": pred[2], "score": pred[3]} 
                    if len(pred) == 4 
                    else  {"label": pred[0], "start": pred[1], "end": pred[2]}
                entities.append(ent)
```

src/rubrix/server/tasks/token_classification/api/model.py
```diff
@@ -172,10 +175,9 @@ def annotated_as(self) -> List[str]:
 
     @property
     def scores(self) -> List[float]:
-        """Values of prediction scores"""
         if not self.prediction:
             return []
-        return [self.prediction.score]
+        return [ent.score for ent in self._predicted_entities()]
```
frascuchon: Here, we lost record level scores information.

- Could we have both score levels, at record and entity? 
- Could we use both in same exploration?
- Should use entities scoring in conjuntion with its labels for filtering?  (Give me records with entities labelled as `?` and a low score (under 0.5))

dvsrepo:
That's a  very good catch!

I'm not sure what's the best option. I'd say that entity-level scores will
be more valuable and we'd like to have the type of queries you are
suggesting (label+score). Maybe if users want a record-level score they can
put that into the metadata, and we work on providing specialized support
for entity-level score filtering and querying.

what do you think @dcfidalgo and ***@***.****


*Daniel Vila Suero*
CEO & Co-founder

Tel: +34 679120675

www.recogn.ai

<http://www.recogn.ai/>
[image: Recognai Twitter] <https://twitter.com/recogn_ai> [image: Recognai
Linkedin] <https://www.linkedin.com/company/recognai/>

dcfidalgo:
In a first version I actually completely removed the record level score, since we are not using it right now. I think we introduced this to store a score returned by a CRF for a set of entities, but I agree with Dani, in my opinion, entity-level scores will be more valuable. Using the metadata to store this information, I think is a valid way to go, it's basically the same for the loss.

frascuchon (approved):
great!

PR #390, opened 2021-09-29
fix(UI): highlight predicted label using predicted_as info

### LINKED ISSUE #388: [UI] Exploring records in Text classification always should mark predicted labels
In example, 

both labels should be marked. 

Screenshot 2021-09-29 at 13 04 54

leiyre:
fix #388

### DIFF
--- a/frontend/components/text-classifier/results/RecordTextClassification.vue
+++ b/frontend/components/text-classifier/results/RecordTextClassification.vue
@@ -34,7 +34,7 @@
         @annotate="onAnnotate"
         @edit="onEdit"
       />
-      <ClassifierExplorationArea v-else :labels="predictionLabels" />
+      <ClassifierExplorationArea v-else :labels="predictionLabels" :predicted-as="record.predicted_as" />
     </div>
     <div v-if="!annotationEnabled" class="record__labels">
       <LabelPill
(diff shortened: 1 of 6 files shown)


frontend/components/text-classifier/results/RecordTextClassification.vue
```diff
@@ -34,7 +34,7 @@
         @annotate="onAnnotate"
         @edit="onEdit"
       />
-      <ClassifierExplorationArea v-else :labels="predictionLabels" />
+      <ClassifierExplorationArea v-else :labels="predictionLabels" :predicted-as="record.predicted_as" />
     </div>
     <div v-if="!annotationEnabled" class="record__labels">
```
frascuchon, line 38: I'm a bit confused here. Regarding the PR changes, the `ClassifierExplorationArea` component handles already a LabelPill for label rendering. 

What this extra `LabelPill` section does? Is shown in explore views for records with annotation but making some tests, I cannot render the component in view.

frascuchon:
I would like to include in this PR changes in `ClassifierExplorationArea` and (maybe) `LabelPill` to accept in props only the record, and calculate required properties internally. It's easy to get lost in another way

frontend/components/text-classifier/results/RecordTextClassification.vue
leiyre: @frascuchon There are two label pills that were visually the same before, those of the predictions, in white with a grey border and the small grey ones that indicate the annotation on the right side of the record.

leiyre: Maybe the small pills in grey should be another component, `tag` for example

frascuchon (approved):

PR #412, opened 2021-10-04
feat(server): configure dataset metrics

frascuchon:
This PR allows configure metrics for datasets.

The simple metric configuration just give a name, id and stored elasticsearch field and generate the apropiate metrics by regarding field data type.

Also, you can define a metric `spec` for advanced users where you can setup an elasticsearch aggregation for more complex metrics calculation.

This feature is EXPERIMENTAL and could change in future.


dcfidalgo (approved):

dvsrepo (approved):
LGTM

Only thing would be to add something in the docs (cc @dcfidalgo) , at least some sort of example, or is it too early and end-users (logging data, using the webapp or Kibana) are not concerned?

frascuchon:
> LGTM
> 
> Only thing would be to add something in the docs (cc @dcfidalgo) , at least some sort of example, or is it too early and end-users (logging data, using the webapp or Kibana) are not concerned?

Changes are visible only through API. Since we have no API documentation, I prefer to keep this hidden until we'll provide some client integrations, for example

dvsrepo:
perfect!

*Daniel Vila Suero*
CEO & Co-founder

Tel: +34 679120675

www.recogn.ai

<http://www.recogn.ai/>
[image: Recognai Twitter] <https://twitter.com/recogn_ai> [image: Recognai
Linkedin] <https://www.linkedin.com/company/recognai/>

Issue #429, opened 2021-10-06
[Server]: Limit search pagination to 10000 


frascuchon:
See performance issues [here](https://www.elastic.co/guide/en/elasticsearch/reference/current/paginate-search-results.html)

The idea is just set a validation rule for [`PaginationParams.from`](https://github.com/recognai/rubrix/blob/master/src/rubrix/server/tasks/commons/api/model.py#L86) attribute

krishnajalan:
can I pick this issue?

dvsrepo:
Definitely! Don't hesitate to ask questions!

PR #435, opened 2021-10-07
fix(server): Limit search pagination to 10000

krishnajalan:
Implements #429


frascuchon:
Great !! Thanks @krishnajalan !

PR #452, opened 2021-10-11
Implement TokenClassification part of the custom SDK

dcfidalgo:
This PR implements the TokenClassification part of our custom SDK. It copies the same approach as for the TextClassification part.
**PR #453 should be first merged into this one, before merging both of them into master**.


frascuchon (commented):
As same as the other PR, we could optimize code generation by using dataset task as for methods generalization, centralizing api calls in one single point instead of per-task points.

dcfidalgo:
Good point! For the current implementation, I followed a bit the approach of the automatically generated SDK. If you don't mind I would prefer to have a first working solution this way, and then open another PR to implement your suggestion, and maybe other improvements I have in mind.

PR #453, opened 2021-10-12
Add Text2Text custom SDK

### LINKED ISSUE #454: Custom SDK does not allow uploading records with `event_timestamp`
Due to the custom SDK in the current master, logging `TextClassificationRecord`s with event_timestamps will fail. PR #453 fixes this bug.

dcfidalgo:
This PR adds the Text2Text part to our custom SDK. It also includes an important bug fix that prevents logging records with an `event_timestamp`. **Keep in mind that this PR merges into #452.** Closes #454

### DIFF
--- a/src/rubrix/client/sdk/text2text/api.py
+++ b/src/rubrix/client/sdk/text2text/api.py
@@ -43,7 +43,7 @@ def bulk(
         headers=client.get_headers(),
         cookies=client.get_cookies(),
         timeout=client.get_timeout(),
-        json=json_body.dict(exclude_unset=True),
+        json=json_body.dict(by_alias=True),
     )
 
     return build_bulk_response(response)
--- a/src/rubrix/client/sdk/text2text/api.py
+++ b/src/rubrix/client/sdk/text2text/api.py
@@ -1,0 +1,69 @@
+#  coding=utf-8
+#  Copyright 2021-present, the Recognai S.L. team.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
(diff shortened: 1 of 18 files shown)


src/rubrix/client/sdk/text2text/api.py
```diff
@@ -0,0 +1,69 @@
...
+    Text2TextRecord,
+)
+
+
+def bulk(
+    client: AuthenticatedClient,
+    name: str,
+    json_body: Text2TextBulkData,
+) -> Response[Union[BulkResponse, ErrorMessage, HTTPValidationError]]:
+    url = "{}/api/datasets/{name}/Text2Text:bulk".format(client.base_url, name=name)
```
frascuchon, line 39: Could we generalize this kind of methods by parameterizing the url endpoint with records task?

For example:

```python
def bulk(....):
  url = "{}/api/datasets/{name}/{task}:bulk".format(client.base_url, name=name, task=json_body.task)
  ...
```

src/rubrix/client/sdk/text2text/api.py
```diff
@@ -0,0 +1,69 @@
...
+    return build_bulk_response(response)
+
+
+def data(
+    client: AuthenticatedClient,
+    name: str,
+    request: Optional[Text2TextQuery] = None,
+    limit: Optional[int] = None,
+) -> Response[Union[List[Text2TextRecord], HTTPValidationError, ErrorMessage]]:
+    url = "{}/api/datasets/{name}/Text2Text/data".format(client.base_url, name=name)
```
frascuchon, line 58: Same comment as before

PR #468, opened 2021-10-14
feat(server): generate automatic task dataset metrics endpoints 

frascuchon:
This PR includes some logic for automatic generate API endpoints for dataset metrics calculation


dcfidalgo (commented):
Nice!! Looks very good, my only comment would be to try to remove the warnings by CodeCov.

frascuchon:
> Nice!! Looks very good, my only comment would be to try to remove the warnings by CodeCov.

Done!

Issue #455, opened 2021-10-13
Allow disabling ES index template creation by environment var


frascuchon:
Include a [API environment variable](https://github.com/recognai/rubrix/blob/master/src/rubrix/server/commons/settings.py#L25) `RUBRIX_DISABLE_ES_INDEX_TEMPLATE_CREATION` allowing advanced users create its own es [index settings and mappings](https://github.com/recognai/rubrix/blob/master/src/rubrix/server/tasks/commons/dao/dao.py#L123).

See #443

torkashvand:
I can work on this issue. :)

frascuchon:
Great @torkashvand !!!

Thanks. Go ahead with the solution.

torkashvand:
check this [PR](https://github.com/recognai/rubrix/pull/469)

frascuchon:
Thanks @torkashvand !!

Issue #443, opened 2021-10-08
Any plan to support no-whitespace language?


faisalron:
I am planning to use rubrix for Japanese text data.
The search functionality doesn't seem to work well on this language.
I think it's better if we can customize the tokenizer used in elasticsearch instead of hardcoded "whitespace" tokenizer.

frascuchon:
Hi @faisalron 

Thanks for your comment.

Customize elasticsearch index settings and mappings is a  feature that we have in our roadmap in a medium term, but we have some tech debt and considerations that we must aboard before apply these customizations. 

Anyway, we could make some minimal changes to allow elasticsearch "out-of-rubrix" customizations by disabling automatic index template creation [here](https://github.com/recognai/rubrix/blob/master/src/rubrix/server/tasks/commons/dao/dao.py#L123)

We could introduce an environment variable in order to enable/disable automatic index template creation, given more control to experimented users that would like to customize some elasticsearch aspect. 

Would you like to aboard that? In that case, I can open an issue including some code guides for that.

frascuchon:
Hi @faisalron 

You can try configure your own elasticsearch index templates and disable automatic index template crreation in rubrix using `DISABLE_ES_INDEX_TEMPLATE_CREATION` env var

faisalron:
@frascuchon Thank you so much for your quick response.
Let me try it and I'll let you know if anything comes up.

faisalron:
@frascuchon I tried to work with the env var and it works perfectly. Thanks

frascuchon:
Great @faisalron 

We would appreciate a little tutorial explaining an example of use with this elasticsearch customization from rubrix. Would you like to try it?

faisalron:
Hi @frascuchon 

> We would appreciate a little tutorial explaining an example of use with this elasticsearch customization from rubrix. Would you like to try it?

Sure, I would love to if you don't mind.

However, I'm sorry the other day I closed the issue, because I think it worked fine, but it turns out that my index template is overwritten because of this line hasn't been modified yet.
https://github.com/recognai/rubrix/blob/master/src/rubrix/server/tasks/commons/dao/dao.py#L121

frascuchon:
It should works now from master. Could you check it @faisalron ?

dvsrepo:
Hi @faisalron  , did you manage to take a look at this? Let us know if you need support

frascuchon:
closing since no activity

PR #469, opened 2021-10-14
Allow disabling ES index template creation by environment var

torkashvand:
allowing advanced users to create their own es index settings and mappings.

See [#443](https://github.com/recognai/rubrix/issues/443) and [ #455](https://github.com/recognai/rubrix/issues/455)


frascuchon (approved):
Great @torkashvand 

Thanks for the work

PR #470, opened 2021-10-14
refactor(client): remove old sdk files

frascuchon:


dcfidalgo (changes requested):
Nice!! I think we can also remove two dependencies from the `setup.cfg` related to the old SDK:
- `attrs`
- `python-dateutils`

Another comment would be to have a look at the coverage warnings. The rest LGTM!

dcfidalgo (approved):

PR #471, opened 2021-10-15
fix(UI): allow remove sort filter

leiyre:
- allow remove sort filter
- replace button text "apply" by  "filter"


frascuchon (approved):

frascuchon:
In my opinion, The `Filter` action shouldn't be applied to Sort component.

PR #472, opened 2021-10-15
chore(CI): codecov config only for CI builds

frascuchon:
This PR configures code coverage generation for codecov only in CI builds.

This changes allow to debug tests internally without comment `addons` parameter in `pyproject.toml` file and running tests with code coverage from PyCharm (https://www.jetbrains.com/help/pycharm/running-test-with-coverage.html)


dcfidalgo (approved):
:+1: much better this way!

PR #501, opened 2021-10-23
Change dictionary key in RubrixLogHTTPMiddleware from probabilities into scores

iakhil:
Made the said change in the corresponding test as well.


frascuchon:
Thanks @iakhil !

PR #503, opened 2021-10-25
Feat: weak supervision

dcfidalgo:
The first implementation of the `WeakLabels` and the `Rule`.

Missing:
- [x] Example in the `Rule` doc string
- [x] `WeakLabels.summary` method
- [x] `WeakLabels.filter_records` method


src/rubrix/weaksupervision/text_classification/rule.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def apply(self, dataset: str):
+        """Apply the rule to a dataset and save matching ids of the records.
+
+        Args:
+            dataset: The name of the dataset
+        """
+        client = rb._client_instance()._client
+        response = data(client=client, name=dataset, request=self._query)
+        _check_response_errors(response)
```
frascuchon: could we move the data or `scan_data` method to rubrix client logic? Even if is a private method, al api interaction should be defined as the client logic.

src/rubrix/client/__init__.py
```diff
@@ -215,7 +209,8 @@ def load(
         name: str,
         ids: Optional[List[Union[str, int]]] = None,
         limit: Optional[int] = None,
-    ) -> pandas.DataFrame:
+        return_pandas: bool = True,
```
frascuchon: should be `as_pandas` a better naming?

```python
rb.load(..., as_pandas=False)
```

src/rubrix/weaksupervision/text_classification/rule.py
dcfidalgo: :+1: I thought about adding a `query_or_ids` parameter to the `rubrix.load` method, then I can use it here as well.

src/rubrix/client/__init__.py
dcfidalgo: :+1: yes, much better.

src/rubrix/labeling/text_classification/rule.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def apply(self, dataset: str):
+        """Apply the rule to a dataset and save matching ids of the records.
+
+        Args:
+            dataset: The name of the dataset.
+        """
+        records = rb.load(name=dataset, query=self._query, as_pandas=False)
+
+        self._matching_ids = [record.id for record in records]
```
frascuchon: This could help performance for large query results
```suggestion
        self._matching_ids = { record.id : b'\xff' fo record in records }
```

src/rubrix/labeling/text_classification/rule.py
```diff
@@ -0,0 +1,71 @@
...
+
+        Raises:
+            `RuleNotAppliedError` if the rule was not applied to the dataset before.
+        """
+        if self._matching_ids is None:
+            raise RuleNotAppliedError(
+                "Rule was still not applied. Please call `self.apply(dataset)` first."
+            )
+
+        if record.id in self._matching_ids:
```
frascuchon:
```suggestion
        if self._matching_ids.get(record.id):
```

src/rubrix/labeling/text_classification/weak_labels.py
```diff
@@ -0,0 +1,395 @@
...
+        label2int: Optional[Dict[str, int]] = None,
+    ):
+        self._rules = rules
+        self._dataset = dataset
+
+        # load records and check compatibility
+        self._records: List[TextClassificationRecord] = load(
+            dataset, query=query, ids=ids, as_pandas=False
+        )
+        if self._records[0].multi_label:
```
frascuchon, line 85: Could we add a default query filter to dataset discarding multi-label fields and avoid this weak check?

```python
query = query or "" +  " AND multi_label: false"
```

src/rubrix/labeling/text_classification/weak_labels.py
```diff
@@ -0,0 +1,395 @@
...
+            (len(self._records), len(self._rules)), dtype=np.short
+        )
+        annotation_array = np.empty(len(self._records), dtype=np.short)
+        _label2int = {"None": -1} if label2int is None else label2int
+
+        for n, record in tqdm(
+            enumerate(self._records), total=len(self._records), desc="Applying rules"
+        ):
+            # First: fill annotation array
+            annotation = record.annotation or "None"
```
frascuchon: Could be a problem define this default `"None"`  if some labelling function return `"None"`  as the matched label?

src/rubrix/labeling/text_classification/weak_labels.py
```diff
@@ -0,0 +1,395 @@
...
+                    ) from error
+                # we already have "None" -> we need to subtract 1
+                _label2int[annotation] = len(_label2int) - 1
+                annotation = _label2int[annotation]
+
+            annotation_array[n] = annotation
+
+            # Second: fill weak label matrix
+            for m, rule in enumerate(self._rules):
+                weak_label = rule(record) or "None"
```
frascuchon: ditto

dcfidalgo:
Ok, this should be ready to review.

In the end, I called the "bucket" feature `WeakLabels.filter_records`. Originally the idea for this feature comes from snorkel's `get_label_buckets` method, hence the initial name. But I think `filter_records` is more explicit in our case, in the end, you filter records by their weak labels, or rules. However, I am not 100% convinced of this name, so if you have suggestions please let me know @dvsrepo @frascuchon .

You can check out the API documentation [here](https://rubrix.readthedocs.io/en/feat-weak_supervision/reference/python/index.html).

src/rubrix/labeling/text_classification/weak_labels.py
```diff
@@ -0,0 +1,395 @@
...
+        """
+        return self._matrix[self._annotation_array == self._label2int["None"]]
+
+    def test_matrix(self) -> np.ndarray:
+        """Returns the part of the weak label ``self.matrix`` that has corresponding annotations.
+
+        Returns:
+            The test weak label matrix.
+        """
+        return self._matrix[self._annotation_array != self._label2int["None"]]
```
frascuchon: Discussing about annotated data  in rubrix with @dvsrepo, this filter could not be a validated test set. I mean
we could have annotated data in rubrix but still non validated, so, this matrix could be confusing for the whole process.

What do you think @dvsrepo @dcfidalgo ?

src/rubrix/labeling/text_classification/weak_labels.py
```diff
@@ -0,0 +1,395 @@
...
+        Args:
+            exclude_missing_annotations: If True, excludes missing annotations,
+                that is all entries with the ``self.label2int["None"]`` integer.
+
+        Returns:
+            The annotation array of integers.
+        """
+        if not exclude_missing_annotations:
+            return self._annotation_array
+        return self._annotation_array[self._annotation_array != self._label2int["None"]]
```
frascuchon: ditto

frascuchon (commented):
Some comments that we should discuss

dcfidalgo:
After discussing it with @frascuchon, we will go for `WeakLabels.show_records` for now, instead of `WeakLabels.filter_records`.

src/rubrix/labeling/text_classification/rule.py
dcfidalgo: Thanks for the suggestion! Done!

src/rubrix/labeling/text_classification/rule.py
dcfidalgo: done

src/rubrix/labeling/text_classification/weak_labels.py
dcfidalgo: After discussing it, we decided to leave this as is and implement a check to only allow datasets of one "multi_label" type: either the whole dataset is multi_label True or the whole dataset is multi_label False.

src/rubrix/labeling/text_classification/weak_labels.py
dcfidalgo: Done, now the abstention label is `None`, not the string "None".

src/rubrix/labeling/text_classification/weak_labels.py
dcfidalgo: Done

src/rubrix/labeling/text_classification/weak_labels.py
dcfidalgo: For now, we will leave these shortcuts for the most common use case, that is annotated records are actually validated. The shortcuts are explicitly explained in the docstrings. But the user can always create their own masks for the label matrix/annotations:
```python
maks = [rec.status == "Validated" for rec in weak_labels.records]
weak_labels.matrix[mask]
```

dcfidalgo:
@dvsrepo added the rule names to the `summary` and `show_records` methods. Feel free to change the example in the docstring yourself, or add a suggestion/comment to this PR, as you prefer.

dvsrepo (approved):
💯

PR #505, opened 2021-10-26
feat(UI): new search box behaviour

### LINKED ISSUE #459: [UX] Search box improvement ¨Next Release¨
We are looking for a solution to make more efficient the use of the search box especially when user is doing several researches in a row.

Solutions to discuss:

- When user clear the last character into the search box, refresh automatically the search.
- When user clear the last character, maintain the cross "X" (rather than showing again "the Search icon" + "placeholder text field") and allow new search when click on the cross "X" is done.
-  .. suggestions

leiyre:
fix #459


frascuchon (approved):
Great!!

PR #516, opened 2021-10-27
feat(client): using task metrics

frascuchon:
With this PR you can cool things like:

```python
from rubrix.metrics.text_classification import f1

results = f1(dataset)
results.visualize() # you must install plotly before
# or give raw data
results.data
```

### DIFF
--- a/docs/reference/python/python_metrics.rst
+++ b/docs/reference/python/python_metrics.rst
@@ -1,0 +1,23 @@
+.. _python_metrics:
+
+Metrics (Experimental)
+======================
+
+Here we describe the available of Rubrix:
+
+- Text classification: Metrics for text classification
+- Token classification: Metrics defined for token classification
+
+Text classification
+-------------------
+
+.. automodule:: rubrix.metrics.text_classification.metrics
+   :members:
+
+Token classification
+--------------------
+
+.. automodule:: rubrix.metrics.token_classification.metrics
+   :members:
+
+
--- a/docs/reference/python/python_metrics.rst
+++ b/docs/reference/python/python_metrics.rst
@@ -3,10 +3,10 @@
 Metrics (Experimental)
 ======================
 
-Here we describe the available of Rubrix:
+Here we describe the available metrics in Rubrix:
 
 - Text classification: Metrics for text classification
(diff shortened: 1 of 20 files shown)


dvsrepo (changes requested):
Very cool! My only request would be to add the relevant user-facing classes and methods in the API reference:

https://github.com/recognai/rubrix/blob/master/docs/reference/python_client.rst

frascuchon:
> Very cool! My only request would be to add the relevant user-facing classes and methods in the API reference:
> 
> https://github.com/recognai/rubrix/blob/master/docs/reference/python_client.rst

Take a look to new toc structure. Now, there is a global python section including client and metrics (and monitoring and labeling in a future)

docs/reference/python/python_metrics.rst
```diff
@@ -0,0 +1,23 @@
+.. _python_metrics:
+
+Metrics (Experimental)
+======================
+
+Here we describe the available of Rubrix:
```
dvsrepo: Here we describe the available metrics in Rubrix:

docs/reference/python/python_metrics.rst
```diff
@@ -0,0 +1,23 @@
...
+
+- Text classification: Metrics for text classification
+- Token classification: Metrics defined for token classification
```
dvsrepo: - Token classification: Metrics for token classification

dvsrepo (changes requested):

dvsrepo:
> > Very cool! My only request would be to add the relevant user-facing classes and methods in the API reference:
> > https://github.com/recognai/rubrix/blob/master/docs/reference/python_client.rst
> 
> Take a look to new toc structure. Now, there is a global python section including client and metrics (and monitoring and labeling in a future)

Looks perfect! I've just suggested two edits

docs/reference/python/python_metrics.rst
frascuchon: Done!

docs/reference/python/python_metrics.rst
frascuchon: Done!

dvsrepo (approved):
💯 !

PR #518, opened 2021-10-28
fix(UI): enable shortcut pagination only for global body focus

### LINKED ISSUE #511: [UI]: shortcut pagination must be disabled on editing search box text
When users are editing the text in search box could use arrows moving between characters. This behavior brings in conflicts with shortcut pagination

leiyre:
fix #511


frascuchon (approved):
🎉

PR #528, opened 2021-10-29
refactor: remove query_inputs alias for text-classification

### LINKED ISSUE #510: Unify `query_text` parameter for the `TextClassificationQuery` model
At the moment, the `TextClassificationQuery` model (server and sdk) is the only *Query* model that has an alias. We should remove it to align it with the other *Query* models.

frascuchon:
closes #510


dcfidalgo (approved):
Perfect! No UI changes necessary?

frascuchon:
> Perfect! No UI changes necessary?

UI was already normalized using query_text, so nothing needed from that part

PR #529, opened 2021-10-29
Docs: Adds metrics guide

dvsrepo:


dvsrepo:
Hi @dcfidalgo and @frascuchon. It would be great to be to visualize Plotly plots in the docs. However, there seems to be a long pending issue with the interplay between rtd theme/sphinx and the plotly.js bundle. You can see the console error here:

https://rubrix.readthedocs.io/en/docs-metrics_guide/guides/metrics.html

Screenshot 2021-11-02 at 15 06 15

The main resources that I found for trying to solve this issue are (I've tried two things with no luck, see commits):

https://github.com/readthedocs/sphinx_rtd_theme/issues/788

https://stackoverflow.com/questions/46269345/embed-plotly-graph-in-a-sphinx-doc

@dcfidalgo could you take a quick look when you have some free slot?

dcfidalgo:
@frascuchon I see some new commits, did you manage to solve this issue?

frascuchon:
All related work will be done in #544

Issue #526, opened 2021-10-29
Single Text classifier : Remove label selections


Amelie-V:
- [x] Permit to remove all selection
- [x] When there is no annotation and nore prediction show "Validate" button disabled

frascuchon:
A missing task here the behavour when validate records from global actions. These kind of records should not be selectables for validate as global action. Just for annotate as..

We must discussed about that @Amelie-V @leiyre @dcfidalgo @dvsrepo

PR #530, opened 2021-10-29
fix(server): validated records for single text-class must include annotations

frascuchon:


src/rubrix/server/tasks/text_classification/api/model.py
```diff
@@ -134,11 +134,27 @@ class CreationTextClassificationRecord(BaseRecord[TextClassificationAnnotation])
...
+    @classmethod
+    def _check_annotation_integrity(
+        cls,
+        annotation: TextClassificationAnnotation,
+        multi_label: bool,
+        status: TaskStatus,
+    ):
+        if status == TaskStatus.validated and not multi_label:
+            assert (
+                len(annotation.labels) > 0
```
dcfidalgo: I think this fails if `annotation is None` which could be the case I think, not sure though ...
Could you add the test case when the status is set to `validated` but `annotation is None`?

dcfidalgo (commented):
:+1: Just one inline comment/doubt.

src/rubrix/server/tasks/text_classification/api/model.py
frascuchon: Fixed! Thanks!

frascuchon:
See #526

dcfidalgo (approved):

Issue #525, opened 2021-10-29
Question: remove record attributes should be allowed ?


frascuchon:
In rubrix, after logging some records, you can change record attributes but you cannot remove them.

For example, you can change a record prediction but you cannot remote it. Should we support this kind of updates? Is this behavior a bug?

By other hand, this new behaviour could difficult partial updates, and you should send the whole data record each time

@dvsrepo @dcfidalgo

dcfidalgo:
As discussed with @frascuchon on slack, I would prefer the behavior of replacing the records, rather than partially updating them. Until now I supposed that this was the case, so I never really had the need for partial updates of a record. 
Often my flow is loading records from Rubrix, that is most of the time I already start with something. So having to provide the full record when you want to update it, does not feel like a big penalty to me.

PR #533, opened 2021-11-01
feat(server): remove partial update

frascuchon:
This PR includes changes discussed on #525 

Records could be updated by deleting some fields like annotations or predictions. 

With these changes, records must be totally sent when logging for updates


dcfidalgo (approved):
Nice, LGTM!

PR #538, opened 2021-11-02
feat(client): add zero-shot classifier monitor

frascuchon:


src/rubrix/monitoring/model_monitor.py
```diff
@@ -3,16 +3,16 @@
...
 def monitor(task_model: Any, dataset: str, sample_rate: float = 0.3):
     if isinstance(task_model, Language):
         return ner_monitor(task_model, dataset=dataset, sample_rate=sample_rate)
-    if isinstance(task_model, TextClassificationPipeline):
-        return classifier_monitor(task_model, dataset=dataset, sample_rate=sample_rate)
+    if isinstance(task_model, Pipeline):
+        return huggingface_monitor(task_model, dataset=dataset, sample_rate=sample_rate)
 
     warnings.warn(
         "The provided task model is not supported by monitoring module. "
```
dcfidalgo: In the case of HF pipelines other than text classification and zero-shot, the warning below is not triggered even if the `task_model` is not monitored in the end.

dcfidalgo (commented):
Nice, LGTM! Just a minor inline comment

src/rubrix/monitoring/model_monitor.py
frascuchon: I've made some changes covering unsupported models in a more general way. Please, take  a  look

src/rubrix/monitoring/model_monitor.py
```diff
@@ -3,19 +3,31 @@
...
+                task_model, dataset=dataset, sample_rate=sample_rate
+            )
+        if model_monitor is None:
+            raise ModelNotSupportedError()
+        return model_monitor
+    except ModelNotSupportedError:
+        warnings.warn(
+            "The provided task model is not supported by monitoring module. "
+            "Predictions won't be logged into rubrix"
+        )
```
dcfidalgo: I think the try-except block is not necessary here:
```suggestion
    model_monitor = None
    if isinstance(task_model, Language):
        model_monitor = ner_monitor(
            task_model, dataset=dataset, sample_rate=sample_rate
        )
    elif isinstance(task_model, Pipeline):
        model_monitor = huggingface_monitor(
            task_model, dataset=dataset, sample_rate=sample_rate
        )
        
    if model_monitor is None:
        warnings.warn(
            "The provided task model is not supported by monitoring module. "
            "Predictions won't be logged into rubrix"
        )
        return task_model
        
    return model_monitor
```

dcfidalgo (approved):
LGTM! Just a minor suggestion.

src/rubrix/monitoring/model_monitor.py
frascuchon: done

PR #543, opened 2021-11-03
feat(UI): disable "Validate" with records without prediction & annotation

### LINKED ISSUE #526: Single Text classifier : Remove label selections

leiyre:
fix #526

### DIFF
--- a/frontend/components/text-classifier/results/RecordTextClassification.vue
+++ b/frontend/components/text-classifier/results/RecordTextClassification.vue
@@ -39,6 +39,7 @@
           v-if="record.status !== 'Validated'"
           class="button-primary"
           @click="onValidate(record)"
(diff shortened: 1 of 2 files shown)


frontend/components/text-classifier/results/RecordTextClassification.vue
```diff
@@ -39,6 +39,7 @@
           v-if="record.status !== 'Validated'"
           class="button-primary"
           @click="onValidate(record)"
+          :disabled="!record.annotation && !record.prediction"
```
frascuchon: Just for single text classification

frascuchon (changes requested):

frascuchon (approved):

PR #544, opened 2021-11-03
feat: review client metrics

frascuchon:
Some review of token classification metrics and the metrics documentation guide updated.

@dvsrepo It would be nice to review also the metrics names and descriptions [here](https://github.com/recognai/rubrix/blob/master/src/rubrix/server/tasks/text_classification/metrics.py#L76) and [here](https://github.com/recognai/rubrix/blob/master/src/rubrix/server/tasks/token_classification/metrics.py#L282 )

See [here](https://github.com/recognai/rubrix/pull/529#issuecomment-957661549) for doc generation problems


frascuchon:
@dvsrepo I finally fix documentation to enable show plotly figures. Maybe it would be nice include this as part of release. 

We can skip metrics guides an just include the metrics review

Tell me what do you think before create the new release branch

dvsrepo:
> @dvsrepo I finally fix documentation to enable show plotly figures. Maybe it would be nice include this as part of release.
> 
> We can skip metrics guides an just include the metrics review
> 
> Tell me what do you think before create the new release branch

Done!

I've also reviewed the guide and I'd be happy if we would include it in the release too. Only thing would be to run all cells to display the graphs (and make sure I haven't destroyed anything :-)). Also, I let you decide whether to include it in the index.rst, my view is that is harmless

frascuchon:
Great. Ready to merge

PR #561, opened 2021-11-04
[Fix] Docs: tab titles

### LINKED ISSUE #549: versions for stable documentation are wrong
Stable and master versions titles in read the docs shows dirty or wrong version numbers.

Maybe this could help (from [setuptools-scm readme](https://github.com/pypa/setuptools_scm/#usage-from-sphinx))

> ## Usage from Sphinx
> It is discouraged to use setuptools_scm from Sphinx itself, instead use importlib.metadata after editable/real installation:
> 
> 
> ```python
> # contents of docs/conf.py
> from importlib.metadata import version
> release = version('myproject')
> # for example take major/minor
> version = '.'.join(release.split('.')[:2])
> ```
> 
> The underlying reason is, that services like Read the Docs sometimes change the working directory for good reasons and using the installed metadata prevents using needless volatile data there.

dcfidalgo:
Closes #549 

The whole thing is a bit more complicated, but I think for now we can live with this "hacky" solution.

Maybe this should go into the 0.6 release branch, so the stable tab title is fixed? @frascuchon

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -16,17 +16,25 @@
 
 
 # -- Project information -----------------------------------------------------
+import os
+
 import rubrix as rb
 
 project = "Rubrix"
 copyright = "2021, Recognai"
 author = "Recognai"
 
-# The full version, including alpha/beta/rc tags
-release = rb.__version__
+# Normally the full version, including alpha/beta/rc tags.
+# But since rtd changes some files during the build process, the scm version becomes dirty,
+# so we just want to display the major and minor version in the tab title of the browser.
+release = ".".join(rb.__version__.split(".")[0:2])
+
+# If on the master branch, set release and version to 'master'
+if "master" in os.popen("git branch --show-current").read():
+    release = "master"
 
 # The short X.Y version
-version = ".".join(rb.__version__.split(".")[0:2])
+version = release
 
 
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -16,17 +16,25 @@
...
 project = "Rubrix"
 copyright = "2021, Recognai"
 author = "Recognai"
 
-# The full version, including alpha/beta/rc tags
-release = rb.__version__
+# Normally the full version, including alpha/beta/rc tags.
+# But since rtd changes some files during the build process, the scm version becomes dirty,
+# so we just want to display the major and minor version in the tab title of the browser.
+release = ".".join(rb.__version__.split(".")[0:2])
```
frascuchon, line 30: Could this brings conflicts in documentation if we publish a version patch (0.6.1, 0.6.2,.. for example)?

dcfidalgo: I think the only conflict will be in the title of the tab, for both versions (0.6.1, 0.6.2) it will state 0.6.
The rest should not be affected. Haven't tested it though.

frascuchon: Ok. We can hide patch version for documentation

frascuchon (approved):

PR #562, opened 2021-11-04
Remove dynamic metadata, move it to `setup.cfg`

dcfidalgo:
No need to have our package metadata duplicated, moved everything to the static `setup.cfg`.

Also, set the flag of using scm in the `pyproject.toml`. pip 21.3 now supports editable installs with `pyproject.toml`, so at some point, we might want to get rid of the `setup.py` file.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,6 +2,8 @@
 requires = ["setuptools", "wheel", "setuptools_scm[toml]"]
 build-backend = "setuptools.build_meta"
 
+[tool.setuptools_scm]
+
 [tool.pytest.ini_options]
 log_format = "%(asctime)s %(name)s %(levelname)s %(message)s"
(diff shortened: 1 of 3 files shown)


frascuchon:
Do the changes affect to info published in [pypi](https://pypi.org/project/rubrix/)?

dcfidalgo:
It should not. It is [recommended](https://packaging.python.org/tutorials/packaging-projects/#configuring-metadata) to put all "static" metadata into `setup.cfg`, which is basically a [configuration file](https://setuptools.pypa.io/en/latest/userguide/declarative_config.html) for setuptools. So instead of providing the metadata directly to the `setup(...)` call, we use the cfg file now. We already used the `setup.cfg` file for all the options basically, now we also use it for the metadata.

frascuchon (approved):
so, go ahead!

PR #577, opened 2021-11-05
Fix: `rb.load` for ids with mixed types

### LINKED ISSUE #564: [Bug] `rb.load` fails when record ids are str and int
Fails because it cannot sort them:
```python
TypeError                                 Traceback (most recent call last)
<ipython-input-40-8967ccc9dc03> in <module>
----> 1 rb.load("weak_supervision_yt")

~/recognai/rubrix/src/rubrix/__init__.py in load(name, query, ids, limit, as_pandas)
    192         >>> dataframe = rb.load(name="example-dataset")
    193     """
--> 194     return _client_instance().load(
    195         name=name, query=query, limit=limit, ids=ids, as_pandas=as_pandas
    196     )

~/recognai/rubrix/src/rubrix/client/rubrix_client.py in load(self, name, query, ids, limit, as_pandas)
    267         _check_response_errors(response)
    268 
--> 269         records_sorted_by_id = sorted(
    270             [record.to_client() for record in response.parsed], key=lambda x: x.id
    271         )

TypeError: '<' not supported between instances of 'int' and 'str'
```

dcfidalgo:
Fixes #564 , by sorting always as str type.

### DIFF
--- a/src/rubrix/client/rubrix_client.py
+++ b/src/rubrix/client/rubrix_client.py
@@ -266,9 +266,12 @@ class RubrixClient:
 
         _check_response_errors(response)
 
-        records_sorted_by_id = sorted(
-            [record.to_client() for record in response.parsed], key=lambda x: str(x.id)
-        )
+        records = [sdk_record.to_client() for sdk_record in response.parsed]
+        try:
+            records_sorted_by_id = sorted(records, key=lambda x: x.id)
+        # record ids can be a mix of int/str -> sort all as str type
+        except TypeError:
+            records_sorted_by_id = sorted(records, key=lambda x: str(x.id))
 
         if as_pandas:
             return pandas.DataFrame(map(lambda r: r.dict(), records_sorted_by_id))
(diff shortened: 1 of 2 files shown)


src/rubrix/client/rubrix_client.py
```diff
@@ -267,7 +267,7 @@ def load(
         _check_response_errors(response)
 
         records_sorted_by_id = sorted(
-            [record.to_client() for record in response.parsed], key=lambda x: x.id
+            [record.to_client() for record in response.parsed], key=lambda x: str(x.id)
```
frascuchon: This change penalizes sorts with homogeneous ids. (Sorting integers [1,2,3,11] by string representation results in a wrong sort)

Could we just avoid ordering in these cases?

dcfidalgo: :+1: I added a try-except block that first tries to sort by int if possible, if not sorts all by str.

My initial motivation of sorting was actually not the absolute order, but that all `rb.load` calls returned the records in the same relative order. But if we can provide a nice int order of the ids (when possible), even better.

frascuchon: great !

PR #599, opened 2021-11-10
[Docs] Make building the docs faster

dcfidalgo:
This PR adds a specific conda env for building the docs, to make the process faster. It also fixes the check if the build is made for the master version of the docs.

It also does some minor cleanup.


frascuchon (approved):
👍

PR #600, opened 2021-11-10
feat(client): compute metrics with query filter

### LINKED ISSUE #581: Allow filtered dataset metrics.
Include basic query string filtering for dataset metrics

```python
from rubrix.metrics.token_classification import entity_labels

entity_labels(name="spacy_sm_wnut17", query="metadata.split: test").visualize()
```

frascuchon:
Fixes #581

### DIFF
--- a/src/rubrix/client/sdk/metrics/api.py
+++ b/src/rubrix/client/sdk/metrics/api.py
@@ -1,5 +1,5 @@
 from functools import lru_cache
-from typing import Any, Dict, List, Union
+from typing import Any, Dict, List, Optional, Union
 
 import httpx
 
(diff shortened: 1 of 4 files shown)


src/rubrix/client/sdk/metrics/api.py
```diff
@@ -47,7 +48,7 @@ def calculate_metric(
         headers=client.get_headers(),
         cookies=client.get_cookies(),
         timeout=client.get_timeout(),
-        json={},  # TODO: parameterize !!!
+        json={"query_text": query},
```
dcfidalgo, line 51: does this fail if `query=None` ?

dcfidalgo (approved):
LGTM! Just one little question.

src/rubrix/client/sdk/metrics/api.py
frascuchon: API accepts these kind of request. For query=None, no filters at all will be applied

dcfidalgo: :+1:

Issue #410, opened 2021-10-04
[Python client] Display dataset URL after rb.log()


dvsrepo:
Currently if a user logs some records using `rubrix.log` there's no indication of the dataset URL where the records have been logged.

The idea would be to print a INFO log message using the `_LOGGER` object with the url to the dataset, something like: `f"Your Rubrix dataset is available at {dataset_url}"`. In principle the info for dataset URL can be retrieved from `self._client`, together with `name` parameter (e.g., http://localhost:6900/dataset_name) and the log message should be added here https://github.com/recognai/rubrix/blob/ac7bb730483a2ceb30aa0b37444a406afe2fefd9/src/rubrix/client/__init__.py#L216.

frascuchon:
Fixed in #609

PR #609, opened 2021-11-11
Add tqdm to `rb.log`

dcfidalgo:
This PR adds a progress bar to the `rb.log` method, and also prints out the URL where the dataset can be found.
I left a todo to improve our logging policy in the library, for now, it is a simple `print` ...

### DIFF
--- a/src/rubrix/client/rubrix_client.py
+++ b/src/rubrix/client/rubrix_client.py
@@ -111,7 +111,7 @@ class RubrixClient:
 
     def log(
         self,
-        records: Iterable[Record],
+        records: Union[Record, Iterable[Record]],
         name: str,
         tags: Optional[Dict[str, str]] = None,
         metadata: Optional[Dict[str, Any]] = None,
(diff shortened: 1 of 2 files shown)


src/rubrix/client/rubrix_client.py
```diff
@@ -200,6 +205,12 @@ def log(
             processed += response.parsed.processed
             failed += response.parsed.failed
 
+            progress_bar.update(len(chunk))
+        progress_bar.close()
+
+        # TODO: improve logging policy in library
+        print(f"Records logged to {self._client.base_url + '/' + name}")
```
dvsrepo: Can we add the number of records: print(f"{len(processed} records logged to {self._client.base_url + '/' + name}")

dvsrepo (changes requested):

src/rubrix/client/rubrix_client.py
dcfidalgo: por supuesto, let's just keep in mind that this info is also present in the `BulkResponse`.

frascuchon:
Closes #410 ?

dcfidalgo:
Yesss, closes #410 !

dvsrepo (approved):

PR #619, opened 2021-11-11
Typo fix in 05-active_learning.ipynb

rasbt:
Minor typo fix, Scitkit-learn -> Scikit-learn


dvsrepo:
Good catch, thanks so much @rasbt !

Issue #616, opened 2021-11-11
[Scrolling] Expected record view state IS lost after scroll 


frascuchon:
![CPT2111111623-1024x598](https://user-images.githubusercontent.com/2518789/141323781-d7dca5c7-9096-4b0e-98ea-00c0cdb67238.gif)

For text2text, record view state is lost after scrolling the page. It's an expected behavior? Should we keep the same state when scrolling @Amelie-V @leiyre @dvsrepo @dcfidalgo ?

leiyre:
I think it should be maintained after scrolling. I doubt if it should also be maintained when we switch between explore/annotate view.

PR #623, opened 2021-11-12
[UI] Text2Text: keep state after scrolling

leiyre:
see #616

### DIFF
--- a/frontend/components/text2text/results/Text2TextList.vue
+++ b/frontend/components/text2text/results/Text2TextList.vue
@@ -39,10 +39,10 @@
               <re-button
                 v-if="hasAnnotationAndPredictions && !editionMode"
                 class="button-clear"
-                @click="changeVisibleSentences"
+                @click="changeSentencesOrigin"
                 >{{
                   sentencesOrigin === "Annotation"
-                    ? annotationEnabled ? `View predictions (${predictionsLength})` : `Back to predictions (${predictionsLength})`
+                    ? annotationEnabled ? `View predictions (${predictions.length})` : `Back to predictions (${predictions.length})`
                     : annotationEnabled ? "Back to annotation" : "View annotation"
                 }}</re-button
               >
@@ -159,6 +159,9 @@
 import "assets/icons/pencil";
 export default {
   props: {
+    recordViewSettings: {
+      type: Object,
+    },
     predictions: {
       type: Array,
       required: true
@@ -166,10 +169,6 @@ export default {
     annotations: {
       type: Array,
       required: true
-    },
-    sentencesOrigin: {
-      type: String,
-      default: undefined
     },
     status: {
       type: String,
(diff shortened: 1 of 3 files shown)


frontend/components/text2text/results/Text2TextList.vue
```diff
@@ -185,15 +184,41 @@ export default {
...
+      if (oldValue !== newValue) {
+        this.$emit('update-view-config', 'newSentence', this.newSentence)
+      }
+    },
+    sentencesOrigin(oldValue, newValue) {
+      if (oldValue !== newValue) {
+        this.$emit('update-view-config', 'sentencesOrigin', this.sentencesOrigin)
+      }
+    },
+  },
```
frascuchon, line 220: Could we avoid defining watchers if we use directly the recordViewSettings attributes?

frontend/components/text2text/results/Text2TextList.vue
```diff
@@ -228,6 +255,28 @@ export default {
...
+        } else if (this.predictions.length) {
+          this.sentencesOrigin = "Prediction";
+        }
+      } else {
+        if (this.predictions.length) {
+          this.sentencesOrigin = "Prediction";
+        } else if (this.annotations.length) {
+          this.sentencesOrigin = "Annotation";
+        }
+      }
```
frascuchon, line 273: This could be the default recordsViewSettings value in `models/Text2TextRecord` class

frontend/components/text2text/results/Text2TextList.vue
```diff
@@ -228,6 +255,28 @@ export default {
...
+    },
+    getRecordViewSetting() {
+      this.recordViewSettings && Object.keys(this.recordViewSettings).forEach(key => {
+        this[key] = this.recordViewSettings[key];
+      })
+    },
```
frascuchon, line 279: if we discard the watchers, this won't make sense. 
 
 Also is not the best naming for the function that returns nothing

frontend/components/text2text/results/Text2TextList.vue
```diff
@@ -262,10 +311,12 @@ export default {
...
       this.$emit('reset-initial-record')
     },
-    changeVisibleSentences() {
+    changeSentencesOrigin() {
       this.itemNumber = 0;
       this.editionMode = false;
-      this.$emit("change-visible-sentences");
+      this.sentencesOrigin !== "Annotation"
+        ? (this.sentencesOrigin = "Annotation")
+        : (this.sentencesOrigin = "Prediction");
```
frascuchon, line 319: This could trigger a recordViewSettings update

frontend/models/Text2Text.js
```diff
@@ -20,10 +20,12 @@ import { BaseRecord, BaseSearchQuery, BaseSearchResults } from "./Common";
 
 class Text2TextRecord extends BaseRecord {
   text;
+  recordViewSettings;
 
-  constructor({ text, ...superData }) {
+  constructor({ text, recordViewSettings, ...superData }) {
     super(superData);
     this.text = text;
+    this.recordViewSettings = recordViewSettings;
```
frascuchon, line 28: we can set here the default value for recordViewSettings.

Better naming could be `viewSettings` or `viewState`  since is already defined in a record

frascuchon (changes requested):

frascuchon:
We will resolve that in #710

PR #636, opened 2021-11-15
[QA] text classification labels

### LINKED ISSUE #630: [QA] Text classification labels
- [x] Show less should not appear when we don't have more labels / or we have extended the list, same for search labels

- [x] Also, I'd try to find another way to do the A/B testing (at least in the master version) as there might be people installing from master : for now Keep option A (bricks) in Master.

- [x] Labels should be order by score (as in the explore mode)

- [x] In "brick-mode", when we select a label, the width should no change even if the score dissapears

- [x] In "fixed-mode" scores should be aligned to the right

Screenshot 2021-11-14 at 14 51 53

- [x] Label Rounded corner: 8px

- [x] Bring Search component closer

- [x]  Show 10 labels max in default view

leiyre:
fix #630

### DIFF
--- a/frontend/components/core/ReTag.vue
+++ b/frontend/components/core/ReTag.vue
@@ -1,5 +1,5 @@
 <template>
-  <p class="tag" :style="{backgroundColor: bgColor}">{{ name }}</p>
+  <p :title="name" class="tag" :style="{backgroundColor: bgColor}">{{ name }}</p>
 </template>
 <script>
 export default {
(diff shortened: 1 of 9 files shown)


frascuchon (commented):
Everything is working now, but a great code refactor should be applied to `ClassifierAnnotationArea.vue`. It's hard to trace the component logic, and take a while to understand different between several max..Labels variables. 

Let's talk tomorrow.

leiyre (commented):
fix #637

frascuchon (approved):

PR #640, opened 2021-11-16
Add F1 metrics to token classification task

### LINKED ISSUE #620: [Metrics] Include F1 metric for token classification

dcfidalgo:
Closes #620

Adds the F1 metric (including precision and recall, averaged and per label) to the Token Classification task. When you have a lot of labels, this results in a lot of metrics, and the visualization could potentially become a bit messy. But with plotly you can always zoom in, so I am not sure if this is really an issue.
Screenshot from 2021-11-16 10-59-10

### DIFF
--- a/src/rubrix/server/tasks/token_classification/metrics.py
+++ b/src/rubrix/server/tasks/token_classification/metrics.py
@@ -1,4 +1,4 @@
-from typing import Any, ClassVar, Dict, List, Tuple
+from typing import Any, ClassVar, Dict, Iterable, List, Set, Tuple
 
 from pydantic import BaseModel, Field
 
@@ -9,6 +9,7 @@ from rubrix.server.tasks.commons.metrics.model.base import (
     BaseTaskMetrics,
     ElasticsearchMetric,
     NestedPathElasticsearchMetric,
+    PythonMetric,
 )
 from rubrix.server.tasks.token_classification import (
     EntitySpan,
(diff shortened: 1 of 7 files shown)


src/rubrix/server/tasks/token_classification/metrics.py
```diff
@@ -160,7 +161,100 @@ def aggregation_result(self, aggregation_result: Dict[str, Any]) -> Dict[str, An
...
+            per_label_metrics.update(
+                {
+                    f"{label}_precision": precision,
+                    f"{label}_recall": recall,
+                    f"{label}_f1": self._safe_divide(
+                        2 * precision * recall, precision + recall
+                    ),
+                }
+            )
+
```
frascuchon, line 206: Not sure if have sense group metrics by kind (micro/macro and per_label) instead set all as flat dictionary:

```python
{
   "micro": { "recall",... "precision"....},
   "macro": { "recall",... "precision"....},
   "per_label": {
      "a": { "recall",... "precision"....},
      "b": { "recall",... "precision"....}
   }
```

what do you think, @dcfidalgo ?

frascuchon (commented):
Great work!!!

Just a comment about metric visualization related to inline comment.

Could you group metrics visualization by kind: `micro`, `macro` and `per-label` bar graphs?

src/rubrix/server/tasks/token_classification/metrics.py
dcfidalgo: This was actually my first schema, but then I thought it would be more practical to flatten it for further usage:
- quicker access: `summary.data["MyLabel_f1"]` versus `summary.data["per_label"]["MyLabel"]["f1"]`
- summary table: `pd.DataFrame(summary.data.values(), index=summary.data.keys())` versus `pd.DataFrame(.........)`

dcfidalgo:
> Could you group metrics visualization by kind: micro, macro and per-label bar graphs?

Yes, I think like this it looks nicer
Screenshot from 2021-11-16 15-14-22

But in terms of practically, I think a table is actually the best "view":
```python
pd.DataFrame(summary.data.values(), index=summary.data.keys())
```
I think I will put this one-liner in the docstring.

dvsrepo:
A related question. For Text Classification, doing some tests yesterday I was missing something that you are adding here for token classification: precision and recall. We could tackle this on a separate PR, but just to make sure we keep a consistent structure for both tasks.

frascuchon:
> A related question. For Text Classification, doing some tests yesterday I was missing something that you are adding here for token classification: precision and recall. We could tackle this on a separate PR, but just to make sure we keep a consistent structure for both tasks.

Yes. F1 for text classification has no recall and prediction related metrics. We can normalize them after this PR.

frascuchon (approved):

dvsrepo:
> > A related question. For Text Classification, doing some tests yesterday I was missing something that you are adding here for token classification: precision and recall. We could tackle this on a separate PR, but just to make sure we keep a consistent structure for both tasks.
> 
> Yes. F1 for text classification has no recall and prediction related metrics. We can normalize them after this PR.

Cool!

PR #641, opened 2021-11-16
NoRecordsFoundError when rb.load results in empty list in WeakLabels

### LINKED ISSUE #617: Improve error message when WeakLabels gets no records
When a query is provided to the `WeakLabels` object, that results in no records, the following error message is raised:
```python
IndexError                                Traceback (most recent call last)
<ipython-input-28-3df95653c46d> in <module>
      3 
      4 # apply the rules to a dataset to obtain the weak labels
----> 5 weak_labels = WeakLabels(
      6     rules=rules,
      7     dataset="weak_supervision_yt",

~/recognai/rubrix/src/rubrix/labeling/text_classification/weak_labels.py in __init__(self, rules, dataset, ids, query, label2int)
     83             dataset, query=query, ids=ids, as_pandas=False
     84         )
---> 85         if self._records[0].multi_label:
     86             raise MultiLabelError(
     87                 "Multi-label text classification is not yet supported."

IndexError: list index out of range
```
I think this can be improved. 

@dvsrepo @frascuchon Do you think it would be useful to raise an error in `rb.load` if the loaded list/dataframe is empty? Or should we tackle this case downstream?

dcfidalgo:
Closes #617 

This PR adds a `NoRecordsFoundError` that is raised if the `rb.load` call results in an empty list/DataFrame. 
I introduced a custom `RubrixClientError` type, maybe we can use it for the other Errors raised by the `RubrixClient` class. improved

### DIFF
--- a/docs/reference/python/python_labeling.rst
+++ b/docs/reference/python/python_labeling.rst
@@ -18,7 +18,7 @@ Labeling tools for the text classification task.
 
 .. automodule:: rubrix.labeling.text_classification.weak_labels
    :members:
-   :exclude-members: WeakLabelsError, MultiLabelError, MissingLabelError
+   :exclude-members: WeakLabelsError, NoRecordsFoundError, MultiLabelError, MissingLabelError
 
 .. automodule:: rubrix.labeling.text_classification.label_models
    :members:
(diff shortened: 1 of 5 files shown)


frascuchon:
It could be an option return an empty dataframe or list when no data found ? I prefer this solution if dataset is already created but no data matches the query.

dcfidalgo:
Ok, I moved the error to the load consumer (the `WeakLabels` in this case).

frascuchon (approved):

PR #642, opened 2021-11-16
fix(metrix): empty metrics visualization

### LINKED ISSUE #638: [Metrics] f1 score visualize fails when empty list results
We should review what happens when we call .visualize in metrics that do not return any value.

Example
```python
from rubrix.metrics.text_classification import f1

f1(name="tuits_en_es").visualize()

# fails because result is MetricSummary(data={})
```
```python
ValueError                                Traceback (most recent call last)
<ipython-input-100-80a9d5dd6989> in <module>
      1 from rubrix.metrics.text_classification import f1
      2 
----> 3 f1(name="tuits_en_es", query="metadata.source:tweets_pos_clean.txt").visualize() #  query="metadata.source:tweets_pos_clean.txt"

/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/rubrix/metrics/models.py in visualize(self)
     13     def visualize(self):
     14         try:
---> 15             return self._build_visualization()
     16         except ModuleNotFoundError:
     17             warnings.warn(

/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/rubrix/metrics/text_classification/metrics.py in <lambda>()
     29     return MetricSummary.new_summary(
     30         data=metric.results,
---> 31         visualization=lambda: helpers.bar(
     32             metric.results,
     33             title=metric.description,

/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/rubrix/metrics/helpers.py in bar(data, title, x_legend, y_legend)
      5     import plotly.graph_objects as go
      6 
----> 7     keys, values = zip(*data.items())
      8     fig = go.Figure(data=go.Bar(y=values, x=keys))
      9     fig.update_layout(

ValueError: not enough values to unpack (expected 2, got 0)
```

frascuchon:
Metrics with empty results will visualizes something like:

Screenshot 2021-11-16 at 13 48 06

Fixes #638

### DIFF
--- a/src/rubrix/metrics/helpers.py
+++ b/src/rubrix/metrics/helpers.py
@@ -1,8 +1,23 @@
 from typing import Dict
 
 
+def empty_visualization(message: str = "No data found"):
+    import plotly.graph_objects as go
+
+    fig = go.Figure()
+    fig.update_layout(
+        xaxis={"visible": False},
+        yaxis={"visible": False},
+        annotations=[{"text": message, "showarrow": False, "font": {"size": 30}}],
+    )
+    return fig
+
+
 def bar(data: dict, title: str = "Bar", x_legend: str = "", y_legend: str = ""):
     import plotly.graph_objects as go
+
+    if not data:
+        return empty_visualization()
 
     keys, values = zip(*data.items())
     fig = go.Figure(data=go.Bar(y=values, x=keys))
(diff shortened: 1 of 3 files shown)


src/rubrix/metrics/helpers.py
```diff
@@ -1,9 +1,24 @@
 from typing import Dict
 
 
+def empty_visualization(message: str = "No data found"):
```
dvsrepo, line 4:
```suggestion
def empty_visualization(message: str = "No data found. Try refining your request"):
```

dvsrepo (changes requested):
Not sure if my suggestion applies, feel free to skip it, No data found is already good.

dvsrepo (approved):
LGTM

PR #651, opened 2021-11-16
refactor(metrics): module shortcut for compute_for and enum def

### LINKED ISSUE #644: [Metrics] Simplify the API usage for `compute_for`
I started testing the metrics for analysing training sets and I'd like to simplify how we set it for `Annotations` or `Predictions`. Currently:

```python
from rubrix.metrics.token_classification import entity_consistency
from rubrix.metrics.token_classification.metrics import Annotations

entity_consistency(name="conll2002_es", mentions=30, threshold=4, compute_for=Annotations).visualize()
```

I'd like at least to be able to do (avoiding `rubrix.metrics.token_classification.metrics`) :
```python
from rubrix.metrics.token_classification import entity_consistency, Annotations
```

I guess that using plain strings like "Annotations" might be problematic right?

frascuchon:
Should fix #644

### DIFF
--- a/src/rubrix/metrics/token_classification/__init__.py
+++ b/src/rubrix/metrics/token_classification/__init__.py
@@ -6,4 +6,7 @@ from .metrics import (
     f1,
     mention_length,
     tokens_length,
+    ComputeFor,
+    Annotations,
+    Predictions,
 )
--- a/src/rubrix/metrics/token_classification/metrics.py
+++ b/src/rubrix/metrics/token_classification/metrics.py
@@ -1,4 +1,5 @@
-from typing import Optional, TypeVar, Union
+from enum import Enum
+from typing import Optional, Union
 
 from rubrix import _client_instance as client
 from rubrix.metrics import helpers
@@ -43,8 +44,13 @@ def tokens_length(
     )
 
 
-Annotations = TypeVar("Annotations")
-Predictions = TypeVar("Predictions")
+class ComputeFor(str, Enum):
+    ANNOTATIONS = "annotations"
+    PREDICTIONS = "predictions"
+
+
+Annotations = ComputeFor.ANNOTATIONS
+Predictions = ComputeFor.PREDICTIONS
 
 _ACCEPTED_COMPUTE_FOR_VALUES = {
     Annotations: "annotated",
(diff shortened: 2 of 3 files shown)


dvsrepo (approved):
🧨

src/rubrix/metrics/token_classification/metrics.py
```diff
@@ -54,16 +60,17 @@ def tokens_length(
 
 def _check_compute_for(compute_for) -> str:
     try:
+        compute_for = compute_for.lower().strip()
         return _ACCEPTED_COMPUTE_FOR_VALUES[compute_for]
     except KeyError:
-        raise f"Wrong compute_for value {compute_for}"
+        raise RuntimeError(f"Wrong value [{compute_for}] for compute_for arg")
```
dcfidalgo: Instead of this function, I would copy the `ExplicitEnum` [by HF](https://github.com/huggingface/transformers/blob/558f8543ba3860c736a7a9a4176ac20f23f9d5a0/src/transformers/file_utils.py#L2063). Then you get this check and a nice error message for free. Then in the methods below you would need to instantiate a `ComputeFor` object, if the input is a string. Let's discuss this via slack maybe.

dcfidalgo (commented):
I would maybe simplify it a bit, see the inline comment. Let's discuss this via slack.

src/rubrix/metrics/token_classification/metrics.py
frascuchon: Done! I've include `_missing_` method just in `ComputeFor` enum class. We can discuss and tackle in a separate PR to use this mechanism for all defined enums in rubrix

src/rubrix/metrics/token_classification/__init__.py
```diff
@@ -6,4 +6,7 @@
     f1,
     mention_length,
     tokens_length,
+    ComputeFor,
+    Annotations,
+    Predictions,
```
dcfidalgo, line 11: I think we can drop these, we already have `ComputeFor`

```suggestion
```

dcfidalgo (approved):
LGTM, just a minor suggestion

PR #655, opened 2021-11-17
fix(search): prevent ignore zeros for aggregation result keys

frascuchon:
This PR fixes `0` key  exclusions for aggregation results


dcfidalgo (approved):
:+1:

Issue #310, opened 2021-09-09
[New Feature]: Team workspaces


Amelie-V:
UX-UI

- Show active session
- Show team's workspaces
- Allow to switch between team's workspace
- Include logout  

Client: (resolved in #661)

- [x] allow configure active workspace in init / environment var 
- [x] allow to change active workspace
- [x] query active workspace

frascuchon:
@dcfidalgo @dvsrepo  Remember: we must define how teams take in place for python client.

PR #661, opened 2021-11-18
feat(client): user workspace management from client

frascuchon:
This PR includes client functions for workspace management for client session.

See #310

### DIFF
--- a/src/rubrix/__init__.py
+++ b/src/rubrix/__init__.py
@@ -57,6 +57,7 @@ def _client_instance() -> RubrixClient:
 def init(
     api_url: Optional[str] = None,
     api_key: Optional[str] = None,
+    workspace: Optional[str] = None,
     timeout: int = 60,
 ) -> None:
     """Init the python client.
@@ -69,6 +70,8 @@ def init(
             it will default to `http://localhost:6900`.
         api_key: Authentification key for the REST API. If `None` (default) and the env variable ``RUBRIX_API_KEY``
             is not set, it will default to `rubrix.apikey`.
+        workspace: Active user workspace used for client session. If `None` (default) and the
+            env variable ``RUBRIX_WORKSPACE`` is not set, it will default to user workspace
         timeout: Wait `timeout` seconds for the connection to timeout. Default: 60.
 
     Examples:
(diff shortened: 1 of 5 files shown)


src/rubrix/__init__.py
```diff
@@ -69,6 +70,8 @@ def init(
             it will default to `http://localhost:6900`.
         api_key: Authentification key for the REST API. If `None` (default) and the env variable ``RUBRIX_API_KEY``
             is not set, it will default to `rubrix.apikey`.
+        workspace: Active user workspace used for client session. If `None` (default) and the
+            env variable ``RUBRIX_WORKSPACE`` is not set, it will default to user workspace
```
dcfidalgo:
```suggestion
        workspace: The workspace to which records will be logged/loaded. If `None` (default) and the 
            env variable ``RUBRIX_WORKSPACE`` is not set, it will default to the private user workspace.
```

src/rubrix/__init__.py
```diff
@@ -86,15 +89,39 @@ def init(
...
 
+def active_workspace() -> str:
+    """
+    Fetch the client session workspace
+
+    Returns
+    -------
+    The active session workspace
+
+    """
```
dcfidalgo:
```suggestion
def get_workspace() -> str:
    """Returns the name of the active workspace for the current client session.

    Returns:
        The name of the active workspace as a string.
    """
```

src/rubrix/__init__.py
```diff
@@ -86,15 +89,39 @@ def init(
...
+    return _client_instance().active_workspace
+
+
+def set_workspace(ws: str) -> None:
+    """Set the client session workspace
+
+    Args:
+        ws: The new workspace
```
dcfidalgo:
```suggestion
    """Sets the active workspace for the current client session.

    Args:
        ws: The new workspace
```

src/rubrix/client/rubrix_client.py
```diff
@@ -75,13 +77,15 @@ def __init__(
...
         api_key: str,
+        workspace: Optional[str] = None,
         timeout: int = 60,
     ):
         """Client setup function.
 
         Args:
             api_url: Address from which the API is serving.
             api_key: Authentication token.
+            workspace: Client session active workspace.
```
dcfidalgo:
```suggestion
            workspace: Active workspace for this client session.
```

dcfidalgo (commented):
Nice! A few naming suggestions, but we can discuss them further.

dcfidalgo (approved):

PR #668, opened 2021-11-19
[UI] Limit pagination in UI

### LINKED ISSUE #466: [UI]: Deep dataset pagination is not allowed
Since elasticsearch has some deep pagination performances limitations (see [here](https://www.elastic.co/guide/en/elasticsearch/reference/7.10/paginate-search-results.html)), we should limit pagination limits according backend limitations.

As a simple approach, a message should be prompted from app requiring some data filtering before deep pagination.

leiyre:
fix #466


frontend/components/commons/results/PaginationEndAlert.vue
```diff
@@ -0,0 +1,34 @@
+<template>
+    <div class="results-pagination-end">
+        <svgicon width="40" height="40" name="pagination-end" />
+        <p>You have reached the current record limit (10,000). To view more records please refine your query</p>
```
frascuchon: Could we use the `maxRecordsLimit` value to build the message?

frascuchon (approved):

PR #693, opened 2021-11-23
[Labeling] Throw error when encountering duplicated rule names

### LINKED ISSUE #669: [Labeling] Trigger a warning when two rules with the same name are provided
Right now, the `WeakLabels.summary` call fails when you provided two rules with the same name to the `WeakLabels.__init__`.
The idea is to issue a warning on init, and add some unique id to the rules used under the hood for the `summary` method.

dcfidalgo:
Closes #669 

In the end, I decided that it would be better to throw an error when encountering duplicated rule names as described in the linked issue. Most of the time when this happens, it's actually a mistake that you want to correct. I could not come up with a use case where you actually prefer a warning and which is not easily fixable in the user's code.

So now, if duplicated rule names are encountered, `WeakLabels` will throw a `DuplicatedRuleNameError` on init, mentioning the duplicated rule names.


dvsrepo (approved):
🥇

PR #694, opened 2021-11-23
[Metrics] Normalize F1 metrics for Text-/TokenClassification

### LINKED ISSUE #648: [Metrics] Add precision and recall to the F1 metric for the text classification task
For the token classification task, we have the F1 score, precision, and recall included in the F1 metric.
For the text classification task only the F1 score. This should be normalized.

dcfidalgo:
Closes #648 

This PR replicates the f1 metric from the token classification task to the text classification task, that is it now includes also recall and precision.


frascuchon (approved):
Yeah!

Issue #698, opened 2021-11-24
[Tutorial] Possible improvement in the "Clean labels using your model loss" tutorial


leireropl:
After running the function for computing the loss in this tutorial, it is required to turn the dataset into a Pandas data frame. When running the code, the following warnings are shown:

**/var/folders/ym/p8klf4tx3cb_7q9vsngj0z4w0000gn/T/ipykernel_16926/2192143421.py:4: UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).**
**input_ids = torch.tensor(batch["input_ids"], device=device)/var/folders/ym/p8klf4tx3cb_7q9vsngj0z4w0000gn/T/ipykernel_16926/2192143421.py:5: UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).**
 **attention_mask = torch.tensor(batch["attention_mask"], device=device)
/var/folders/ym/p8klf4tx3cb_7q9vsngj0z4w0000gn/T/ipykernel_16926/2192143421.py:6: UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).
  labels = torch.tensor(batch["labels"], device=device)**

Therefore, these variables can be changed to:
    **input_ids = batch["input_ids"].clone().detach()**
    **attention_mask = batch["attention_mask"].clone().detach()**
    **labels = batch["labels"].clone().detach()**

Then, the warning disappears.

dcfidalgo:
Done in #714

PR #708, opened 2021-11-24
changed variables to avoid warning

leireropl:
changed variables


leireropl:
Closes #698

dcfidalgo:
I had a more detailed look at this, and with your changes, we lose the feature of being able to use GPUs (the `device` argument is basically for this). I think we do not need to clone those tensors at all, we can just move them to the selected device via the `torch.Tensor.to` method. @leireropl let me know if you need help with this, maybe we can discuss this in slack.

Issue #702, opened 2021-11-24
[Tutorials] Suggestions for the Sentiment Classifier tutorial


leireropl:
There are some things to add that could be interesting:

1. TensorFlow gives some problems when running the code, so maybe we could add a warning saying that if people encounter problems when using TensorFlow, they can change to Pytorch.
2. Jupyter Notebook can encounter due to its widgets. If so, the line **!pip install ipywidgets** can solve this. This can be also added as a warning.
3. The dataset is quite big and it takes a long time to be loaded, so **"split = 'train[0:50]'"** can be added to the banking_ds variable.

dvsrepo:
Regarding the first point, I'd say we should enforce using PyTorch. We typically add the pip install torch at the beginning and we should make sure we're loading the pt versions of transformer

dcfidalgo:
List of actions:
- [ ] Regarding 1, with Dani's comment: Add `%pip install transformers[torch]`
- [ ] Regarding 2: Add `%pip install ipywidgets`
- Regarding 3: Not sure, the 'banking77' data set is about 1MB , which for most normal working connections should be fine in my opinion.

@leireropl If you have some spare time, could you tackle 1 and 2?

PR #711, opened 2021-11-25
Fix/spacy_transformers

### LINKED ISSUE #700: [Tutorials] Issues with transformes in the SpaCy tutorial
After importing space and loading the model, this error may be displayed:

**ValueError: [E002] Can't find factory for 'transformer' for language English (en). This usually happens when spaCy calls `nlp.create_pipe` with a custom component name that's not registered on the current language class. If you're using a Transformer, make sure to install 'spacy-transformers'. If you're using a custom component, make sure you've added the decorator `@Language.component` (for function components) or `@Language.factory` (for class components).**

In order to get rid of it, the following line could be added after importing Rubrix and installing SpaCy datasets:
**%pip install spacy-transformers -f https://download.pytorch.org/whl/torch_stable.html**

There are other suggestions (i.e. **%pip install spacy-transformers**), but sometimes it fails.

leireropl:
Added "%pip install space-transformers" line to avoid errors when loading the model

Closes #700

### DIFF
--- a/docs/tutorials/02-spacy.ipynb
+++ b/docs/tutorials/02-spacy.ipynb
@@ -73,7 +73,9 @@
    },
    "outputs": [],
    "source": [
-    "%pip install datasets spacy~=3.0 protobuf -qqq"
+    "%pip install datasets spacy~=3.0 protobuf -qqq\n",
+    "%pip install spacy-transformers -f https://download.pytorch.org/whl/torch_stable.html\n",
+    "#install this or %pip install spacy-transformers to avoid problems when loading the model"
    ]
   },
   {
--- a/docs/tutorials/02-spacy.ipynb
+++ b/docs/tutorials/02-spacy.ipynb
@@ -73,7 +73,7 @@
    },
    "outputs": [],
    "source": [
-    "%pip install torch datasets spacy[transformers]~=3.0 protobuf -qqq"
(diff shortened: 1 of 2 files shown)


leireropl:
Closes #700

docs/tutorials/02-spacy.ipynb
```diff
@@ -73,7 +73,9 @@
    },
    "outputs": [],
    "source": [
-    "%pip install datasets spacy~=3.0 protobuf -qqq"
+    "%pip install datasets spacy~=3.0 protobuf -qqq\n",
+    "%pip install spacy-transformers -f https://download.pytorch.org/whl/torch_stable.html\n",
+    "#install this or %pip install spacy-transformers to avoid problems when loading the model"
```
dcfidalgo: We should explicitly install PyTorch, and then specify `spacy[transformers]` which automatically installs the spacy-transformers package:
```suggestion
    "%pip install torch datasets spacy[transformers]~=3.0 protobuf -qqq",
```

dcfidalgo (changes requested):
Following Dani's comment in #702, we should enforce the installation of pytorch in the beginning, see inline comment.

dcfidalgo (approved):
Nice! LGTM!

PR #725, opened 2021-11-26
fix(ui): Preserving the annotate/explore state on browser refresh

leiyre:


frontend/database/modules/datasets.js
```diff
@@ -234,7 +234,7 @@ async function _querySearch({ dataset, query, sort, size }) {
 
 async function _paginate({ dataset, size, page }) {
   const pagination = new Pagination({ size, page });
-
+  const viewSettings = DatasetViewSettings.find(dataset.name);
```
frascuchon, line 237: This is not needed if `_updateViewSettings` returns the updated `DatasetViewSettings` object

PR #744, opened 2021-12-02
feat(server): allow handle labeling rules for datasets from API

frascuchon:
This PR includes dataset labeling rules management for API.

- Following the API rule data model:

LabelingRule | Data model for labeling rules creation
-- | --
query* | The rule es query
label* | The linked rule label
description | Description related to rule

- Rules are keyed by query, we cannot include rules with same query

@dcfidalgo @dvsrepo  Feedback welcome

See #745

### DIFF
--- a/src/rubrix/server/datasets/dao.py
+++ b/src/rubrix/server/datasets/dao.py
@@ -14,8 +14,9 @@
 #  limitations under the License.
 
 import json
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Optional, Type, TypeVar
 
+import deprecated
 from fastapi import Depends
 
 from rubrix.server.commons.es_wrapper import ElasticsearchWrapper, create_es_wrapper
@@ -51,8 +52,34 @@ def dataset_records_index(dataset_id: str) -> str:
     return DATASETS_RECORDS_INDEX_NAME.format(dataset_id)
 
 
+BaseDatasetDB = TypeVar("BaseDatasetDB", bound=DatasetDB)
+
+
 class DatasetsDAO:
     """Datasets DAO"""
+
+    _INSTANCE = None
+
+    @classmethod
+    def get_instance(
+        cls, es: ElasticsearchWrapper = Depends(create_es_wrapper)
+    ) -> "DatasetsDAO":
+        """
+        Get or creates the dao instance
+
+        Parameters
+        ----------
+        es:
+            The elasticsearch wrapper dependency>
+
+        Returns
+        -------
+            The dao instance
+
+        """
+        if cls._INSTANCE is None:
+            cls._INSTANCE = cls(es)
+        return cls._INSTANCE
 
     def __init__(self, es: ElasticsearchWrapper):
         self._es = es
@@ -157,6 +184,30 @@ class DatasetsDAO:
         finally:
             self._es.delete_document(index=DATASETS_INDEX_NAME, doc_id=dataset.id)
 
+    def find_by_id(
+        self, dataset_id: str, ds_class: Type[BaseDatasetDB] = DatasetDB
+    ) -> Optional[BaseDatasetDB]:
+        """
+        Finds a dataset db by its id
+
+        Parameters
+        ----------
+        dataset_id:
+            The dataset id
+        ds_class:
+            Class used for data deserialization. Class could be an specialization
+            of DatasetDB. Default ``DatasetDB``
+
+        Returns
+        -------
+            The dataset instance if found. ``None`` otherwise
+        """
+        document = self._es.get_document_by_id(
+            index=DATASETS_INDEX_NAME, doc_id=dataset_id
+        )
+        if document:
+            return self._es_doc_to_dataset(document, ds_class=ds_class)
+
     def find_by_name(self, name: str, owner: Optional[str]) -> Optional[DatasetDB]:
         """
         Finds a dataset by name
@@ -195,7 +246,9 @@ class DatasetsDAO:
         return self._es_doc_to_dataset(document) if document else None
 
     @staticmethod
-    def _es_doc_to_dataset(doc: Dict[str, Any]) -> DatasetDB:
+    def _es_doc_to_dataset(
+        doc: Dict[str, Any], ds_class: Type[BaseDatasetDB] = DatasetDB
+    ) -> BaseDatasetDB:
         """Transforms a stored elasticsearch document into a `DatasetDB`"""
 
         def __key_value_list_to_dict__(
@@ -207,7 +260,7 @@ class DatasetsDAO:
         tags = source.pop("tags", [])
         metadata = source.pop("metadata", [])
 
-        return DatasetDB.parse_obj(
+        return ds_class.parse_obj(
             {
                 **source,
                 "tags": __key_value_list_to_dict__(tags),
@@ -254,6 +307,7 @@ class DatasetsDAO:
 _instance: Optional[DatasetsDAO] = None
 
 
+@deprecated.deprecated(reason="Use ``DatasetsDAO.get_instance`` class method instead")
 def create_datasets_dao(
     es: ElasticsearchWrapper = Depends(create_es_wrapper),
 ) -> DatasetsDAO:
--- a/src/rubrix/server/datasets/dao.py
+++ b/src/rubrix/server/datasets/dao.py
@@ -65,12 +65,12 @@ class DatasetsDAO:
         cls, es: ElasticsearchWrapper = Depends(create_es_wrapper)
     ) -> "DatasetsDAO":
         """
-        Get or creates the dao instance
+        Gets or creates the dao instance
 
         Parameters
         ----------
         es:
-            The elasticsearch wrapper dependency>
+            The elasticsearch wrapper dependency
 
         Returns
         -------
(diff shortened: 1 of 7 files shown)


dvsrepo (approved):
Niiiice. My only questions would be:

- would this accommodate in the future for multilabel rules?
- for token classification besides from the query and the label we will need additional information (if the label should be applied to the matching token, or to the token in the next i positions, etc), just to keep it mind and if something needs to be taken into account in this first design.
- ideally, I'd like to have the agent adding the rule (annotation agent should do it) in the same way we do for annotations in the UI if it's not too much trouble, I foresee rules can be defined much more collaboratively and this info would be useful.
That's all, I like the design a lot!

src/rubrix/server/tasks/text_classification/service/service.py
```diff
@@ -168,29 +203,52 @@ def _check_multi_label_integrity(
...
+            The rule
+
+        """
+        self.__labeling__.add_rule(dataset, rule)
+
+    def delete_labeling_rule(self, dataset: Dataset, rule_query: str):
+        """
+        Deletes a rule from a dataset.
+
+        Nothing happens if rule does exists in dataset.
```
dcfidalgo:
```suggestion
        Nothing happens if the rule does not exist in dataset.
```

src/rubrix/server/datasets/dao.py
```diff
@@ -51,9 +52,35 @@ def dataset_records_index(dataset_id: str) -> str:
...
     """Datasets DAO"""
 
+    _INSTANCE = None
+
+    @classmethod
+    def get_instance(
+        cls, es: ElasticsearchWrapper = Depends(create_es_wrapper)
+    ) -> "DatasetsDAO":
+        """
+        Get or creates the dao instance
```
dcfidalgo:
```suggestion
        Gets or creates the dao instance
```

src/rubrix/server/datasets/dao.py
```diff
@@ -51,9 +52,35 @@ def dataset_records_index(dataset_id: str) -> str:
...
+
+        Parameters
+        ----------
+        es:
+            The elasticsearch wrapper dependency>
```
dcfidalgo:
```suggestion
            The elasticsearch wrapper dependency
```

src/rubrix/server/tasks/text_classification/api/model.py
```diff
@@ -35,6 +35,67 @@
...
 
+class CreateLabelingRule(BaseModel):
+    """
+    Data model for labeling rules creation
+
+    Attributes:
+    -----------
+
+    query:
+        The rule es query
```
dcfidalgo:
```suggestion
        The ES query of the rule
```

src/rubrix/server/tasks/text_classification/api/model.py
```diff
@@ -35,6 +35,67 @@
...
+
+    label: str
+        The linked rule label
```
dcfidalgo:
```suggestion
        The label associated with the rule
```

src/rubrix/server/tasks/text_classification/api/model.py
```diff
@@ -35,6 +35,67 @@
...
+
+    description:
+        Description related to rule
```
dcfidalgo:
```suggestion
        A brief description of the rule
```

src/rubrix/server/tasks/text_classification/api/model.py
```diff
@@ -35,6 +35,67 @@
...
+
+    @validator("query")
+    def strip_query(cls, query: str) -> str:
+        """Remove blank spaces for query"""
+        return query.strip()
+
+
+class LabelingRule(CreateLabelingRule):
+    """
+    Adds read-only attributes to labeling rule
```
dcfidalgo:
```suggestion
    Adds read-only attributes to the labeling rule
```

src/rubrix/server/tasks/text_classification/api/model.py
```diff
@@ -35,6 +35,67 @@
...
+
+    Attributes:
+    -----------
+
+    author:
+        Who has created the rule
+
+    created_at:
+        When rule was created
```
dcfidalgo:
```suggestion
        When was the rule created
```

src/rubrix/server/tasks/text_classification/api/model.py
```diff
@@ -35,6 +35,67 @@
...
+
+class LabelingRule(CreateLabelingRule):
+    """
+    Adds read-only attributes to labeling rule
+
+    Attributes:
+    -----------
+
+    author:
+        Who has created the rule
```
dcfidalgo:
```suggestion
        Who created the rule
```

src/rubrix/server/tasks/text_classification/service/service.py
```diff
@@ -168,29 +203,52 @@ def _check_multi_label_integrity(
...
 
-_instance = None
+        Parameters
+        ----------
+        dataset:
+            The dataset
 
+        Returns
+        -------
+            A list of labeling rules create for a given dataset
```
dcfidalgo:
```suggestion
            A list of labeling rules for a given dataset
```

dcfidalgo (approved):
great, really nice!! Just a few minor inline suggestions. Two questions though, is the `NEW_BASE_ENDPOINT` supposed to replace the `BASE_ENDPOINT` in the long term? The second one, in the client code we try to follow google style docstrings, since they work better together with sphinx when taking type annotations into account. should we try to follow the same style in the server code?

frascuchon:
> great, really nice!! Just a few minor inline suggestions. Two questions though, is the `NEW_BASE_ENDPOINT` supposed to replace the `BASE_ENDPOINT` in the long term? The second one, in the client code we try to follow google style docstrings, since they work better together with sphinx when taking type annotations into account. should we try to follow the same style in the server code?

For the first question, YES. I want to move the api path in the form `/api/datasets/{task}/{name}/...`. The current format was adopted when a dataset could wrap several task records. So, I prefer create new endpoints with this new format. 

For the second I will create a tech debt issue planning docstring migration to google style.

frascuchon:
> Niiiice. My only questions would be:
> 
> * would this accommodate in the future for multilabel rules?

Only text classification with single label are supported as weak supervision paradigm in rubrix. Do we known all we need for design data model for multi-label weak supervision? We can define `label` as  `Union[str, List[str]]` but not sure if we're still missing something else. 

> * for token classification besides from the query and the label we will need additional information (if the label should be applied to the matching token, or to the token in the next i positions, etc), just to keep it mind and if something needs to be taken into account in this first design.

The rule data model only affects to text classification task. The token classification task will have it own labeling rule data model.

> * ideally, I'd like to have the agent adding the rule (annotation agent should do it) in the same way we do for annotations in the UI if it's not too much trouble, I foresee rules can be defined much more collaboratively and this info would be useful.
>   That's all, I like the design a lot!

Described fields are input or modifiable fields. There are also 2 read-only fields for the [API response data model](https://github.com/recognai/rubrix/blob/cfe1cd4bcc1a7d51f6bd7d81e2c3b2f3da762890/src/rubrix/server/tasks/text_classification/api/model.py#L66): `author` and `created_at`

```python
class LabelingRule(CreateLabelingRule):
    """
    Adds read-only attributes to the labeling rule
    Attributes:
    -----------
    author:
        Who created the rule
    created_at:
        When was the rule created
    """

    author: str
    created_at: Optional[datetime] = Field(default_factory=datetime.utcnow)
```

The rule author will be automatically filled with username who makes the creation request. We can normalizes the name and use `agent` but I think `author` match better for rule creation.

dvsrepo:
> > Niiiice. My only questions would be:
> > 
> > * would this accommodate in the future for multilabel rules?
> 
> Only text classification with single label are supported as weak supervision paradigm in rubrix. Do we known all we need for design data model for multi-label weak supervision? We can define `label` as `Union[str, List[str]]` but not sure if we're still missing something else.


Yes, what I mean is that we plan to support it in the near future just want to make sure there are no big issues when we do so, but we need to define how the weak label matrix works, etc. As we did for single label.
> 
> > * for token classification besides from the query and the label we will need additional information (if the label should be applied to the matching token, or to the token in the next i positions, etc), just to keep it mind and if something needs to be taken into account in this first design.
> 
> The rule data model only affects to text classification task. The token classification task will have it own labeling rule data model.


Ok, good
> 
> > * ideally, I'd like to have the agent adding the rule (annotation agent should do it) in the same way we do for annotations in the UI if it's not too much trouble, I foresee rules can be defined much more collaboratively and this info would be useful.
> >   That's all, I like the design a lot!
> 
> Described fields are input or modifiable fields. There are also 2 read-only fields for the [API response data model](https://github.com/recognai/rubrix/blob/cfe1cd4bcc1a7d51f6bd7d81e2c3b2f3da762890/src/rubrix/server/tasks/text_classification/api/model.py#L66): `author` and `created_at`
> 
> ```python
> class LabelingRule(CreateLabelingRule):
>     """
>     Adds read-only attributes to the labeling rule
>     Attributes:
>     -----------
>     author:
>         Who created the rule
>     created_at:
>         When was the rule created
>     """
> 
>     author: str
>     created_at: Optional[datetime] = Field(default_factory=datetime.utcnow)
> ```
> 
> The rule author will be automatically filled with username who makes the creation request. We can normalizes the name and use `agent` but I think `author` match better for rule creation.

Good, I hadn't seen this. Not strong opinion on the naming, although I always prefer to have more aligned names especially when they work so similarly (in the sense that is the username performing the action: annotating or adding a rule).

Great job!!

Issue #752, opened 2021-12-02
[Metadata filters] Filter by metadata at record level fails with no-string values


frascuchon:
Filtering at record level by a integer metadata field value:
![Screenshot 2021-12-02 at 17 34 38](https://user-images.githubusercontent.com/2518789/144464138-87eb02d6-8f88-484b-8abc-ac628b838cfa.png)

1. Doesn't activate filter at global metadata filter level:
![Screenshot 2021-12-02 at 17 34 53](https://user-images.githubusercontent.com/2518789/144464226-fbfa4b42-c97b-4493-ad9a-7a102d982686.png)

2 Raise an error when metadata modal is open
![Screenshot 2021-12-02 at 17 35 07](https://user-images.githubusercontent.com/2518789/144464290-d0ca27ec-67bc-4cd2-899f-fa0e134088bb.png)

PR #753, opened 2021-12-02
feat(search): show boolean values in metadata

### LINKED ISSUE #733: [Metadata] Show boolean values in metadata
It would be really nice if we could show metadata that consists of boolean values in the "Metadata" filter of the web app. I think we can simply treat them as strings there. @leiyre @frascuchon what do you think?

frascuchon:
Fixes #733

Discover an issue #752 related to metadata filtering in UI


dcfidalgo (approved):
:+1:

Issue #745, opened 2021-12-02
[Weak Supervision] Integrate labeling rules management from app


frascuchon:

Amelie-V:
QA

Global

- [x] **P2** Center tooltip
- [x] Metrics: % con 3 decimales. ej: 70.566 (except record, Correct/incorrect Total correct/incorrect)

Create rule:
- [x] Change message when you just save a new rule to: Rule has been saved (mess)
- [x] Reduce empty space between labels and metrics
- [x] Search box : change "Search records" to "Introduce a query" in the third views (explore/annotate/define rules)
- [x] Show query in semi bold and color
- [x] Add edition name (Amelie)
- [x] Tooltip on each metrics 
- [x] Shadow on the WS Area when the search area is actived
- [x] Review button color on grey area (Amélie)
- [x] Add overall metrics 
- [x] Edit name : on hover: text darker, on click: cursor
- [x] When user change the query maintain the previous label selected actived

Rules setting:

- [x] Delete rule doesnt work
- [x] Integrate copy delete modal : 
Permanently delete rule
You are about to delete the rule "earth" from your dataset. This action cannot be undone.
- [x] Create Content "0 results found" (Amélie)
- [x] Create Content "0 rules" (Amélie)
- [x] Change Search dataset to "Search rule by name"
- [x] Put metrics results (ej. 0.325) in bold
- [x] change "C. Annotated" to "Cov. Annot."
- [ ] Add state to Update button
- [x] Fix Update button to the table
- [x] Active sidebar and all interaction in "No rules" page. For now nothing is clicable
- [x] Allow clic on the table when the hand is actived
- [ ] **P2** CenterAdd toast to confirm rule is deleted

dvsrepo:
Create rule:

- [x] If I just saved a rule and I remove the query from the search input, the rule config module still shows the state of my previous saved rule. I should see the initial empty status (the same as when I access the rules definition from the first time)

Screenshot after removing the query from the search box:

<img width="1576" alt="Screenshot 2021-12-20 at 14 28 50" src="https://user-images.githubusercontent.com/1107111/146774418-7f6c29f4-e1d0-4b6b-89fa-440d3b9c4fad.png">

I expect to see: (1) no label selected, (2) stats with `-`, and (3) Save rule button disabled


- [x] When no query is defined (in the search input), I would change the message "Select a label for your query" to "Start by entering a query in the search box" (or something like that). Then, when a query is entered, show the message "Select a label for your query"

[...]

PR #790, opened 2021-12-11
feat(server): labeling rule metrics

frascuchon:
Add labeling rule metrics api endpoint. 

A query and a label must be provided

See #745


dcfidalgo (approved):
LGTM!

PR #794, opened 2021-12-13
new folder and images for spacy tutorial

leireropl:
Actions:

-A new folder for the tutorial has been added,
- Images/videos have been updated,
- Former pictures have been deleted.


dcfidalgo (commented):
Thanks @leireropl ! The new gifs seem to have a much worse resolution compared to the old ones. Would it be possible to improve the resolution for the new ones?

leireropl:
> Thanks @leireropl ! The new gifs seem to have a much worse resolution compared to the old ones. Would it be possible to improve the resolution for the new ones?

Done!

dcfidalgo:
@leireropl Just a minor thing, could you remove the gifs/files we don't need anymore?

frascuchon:
Yesterday I was trying the spacy tutorial and I remark that the executable shell imports rubrix 

Screenshot 2021-12-16 at 11 52 11

If we skip the cell execution (since we can have already installed deps or the notebook kernel is restarted), the record generation will fail (since rubrix was not imported)
Screenshot 2021-12-16 at 11 53 36

Should be better to import rubrix modeul the first time is used in the notebook ?

@dvsrepo @leireropl ??

dcfidalgo:
Yes, definitely a bug, I will push some changes to this PR.

PR #795, opened 2021-12-13
new folder and images for weak supervision tutorial

leireropl:
A new folder has been added, pictures have been transferred/updated.


dcfidalgo (approved):
Nice!

PR #796, opened 2021-12-13
new folder and images for active learning tutorial

leireropl:
I added a new folder and I updated images and videos


dcfidalgo (commented):
Thanks! Could we improve the resolution of the gif? Also, I think the png got somehow cropped, could you maybe check this?

leireropl:
> Thanks! Could we improve the resolution of the gif? Also, I think the png got somehow cropped, could you maybe check this?

I uploaded the original .mp4 file, let me know if this works better! Also, I changed the png- path was originally wrong and I found the correct .png in another folder!

dvsrepo:
> > Thanks! Could we improve the resolution of the gif? Also, I think the png got somehow cropped, could you maybe check this?
> 
> I uploaded the original .mp4 file, let me know if this works better! Also, I changed the png- path was originally wrong and I found the correct .png in another folder!

For GIFs I highly recommend using https://getkap.co/

dcfidalgo:
Thanks, could you also remove the files we don't need anymore?

PR #805, opened 2021-12-14
new folder and videos for model loss tutorial

leireropl:
I added new videos and a folder for them


dcfidalgo:
The `relabeling_ag_news.mp4` does not work for me in the browser, could you check that again? Also in the `ModelLoss_Filter.mp4` you don't really see how to sort the examples by their loss, maybe we could include this part, otherwise this video is not super helpful.

PR #809, opened 2021-12-14
fix(ui): Unique ID for scroll record state 

### LINKED ISSUE #808: [Annotation/search label] keyword stays visible
Se mantiene el estado cuando cambia algo en el filtro de búsqueda.
En el segundo dataset que probé (domain) tenía unos records con el search box rellenado de una palabra buscada en el dataset anterior (banking). Te paso un SS

leiyre:
fix #808

### DIFF
--- a/frontend/components/commons/results/RecordExtraActions.vue
+++ b/frontend/components/commons/results/RecordExtraActions.vue
@@ -53,7 +53,7 @@ export default {
   mixins: [
     IdState({
       // You can customize this
-      idProp: vm => vm.record.id,
+      idProp: vm => `${vm.dataset.name}-${vm.record.id}`,
     }),
   ],
   props: {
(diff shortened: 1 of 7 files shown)


frascuchon:
These changes won't fix linked issue. 

The problem is related to record state when filters are changed. For example:

1. without a filter applied 
Screenshot 2021-12-14 at 18 10 10

2. Filter the record cards labels
Screenshot 2021-12-14 at 18 10 59

3.  Applying a new filter that matches the record with filtered card  labels still is showing filtered card labels, when the filter should be removed (If I understood the issue properly)
Screenshot 2021-12-14 at 18 12 06

Could you please confirm the case @Amelie-V ?

frascuchon (changes requested):
Changes doesn't fix the issue

frascuchon:
> Changes doesn't fix the issue

Okay, i understood nothing. Sorry

frascuchon (approved):
nice !

PR #812, opened 2021-12-15
Fix token alignment on CreationTokenClassificationRecord

ruanchaves:
Changed .lstrip() to .strip() on the check_annotation method of the
CreationTokenClassificationRecord class.

Added unit tests that cover both edge cases: newline followed by spaces
to the left of the annotated entity, and also newline followed by spaces
to the right of the annotated entity.


frascuchon (approved):
Great !!!

PR #818, opened 2021-12-16
draft for the new web app ui reference doc

leireropl:


leireropl:
I noticed that images and links aren't properly set, will fix that later

dcfidalgo:
Closing this, see #827 instead.

Issue #816, opened 2021-12-15
[Workspaces] Include current workspace as part of url


frascuchon:
Including workspaces as part of url in UI brings benefits like sharing datasets between workspace members or simplify user workspaces management.

my url proposal:

`/ws/<current-workspace>/` -> will show workspace datasets
`/ws/<current-workspace>/<dataset>` -> will show workspace dataset 

I've included the `ws` namespace in url path to contextualize info and regarding future features in UI (management other than workspaces, or personal settings...)

Accessing to home url `/` will automatically redirect to the main workspace url view.

What do you think @Amelie-V @leiyre @dvsrepo @dcfidalgo ?

 See also #810  and #736

dcfidalgo:
Sounds good to me! :rocket:

Issue #736, opened 2021-11-30
[Datasets] Datasets with same name in different workspaces are not shown in list


frascuchon:
ONLY FOR superusers who can view all datasets

![Screenshot 2021-11-30 at 12 09 59](https://user-images.githubusercontent.com/2518789/144036778-a66f141d-84df-4121-8390-18052ecea540.png)

PR #820, opened 2021-12-16
feat: Include workspace in url path

### LINKED ISSUE #810: [Users login] last user workspace remains when login without a previous logout
How to reproduce: 

1. Login with a user A 
2. See A datasets 
3. Login in a separate tab (without a logout) a user B
4. User B see datasets from last used workspace for user A or a 403 error is raised (if user B has no access to workspace)

frascuchon:
Closes #810 and #816

Once this PR is merged, old app urls will fail

Not covered here but see #736


leiyre (commented):
:ok_hand:

PR #837, opened 2021-12-20
feat (search): allow to filtering record metrics fields in search

frascuchon:
This PRs includes query param for task search for exclude metrics fields on each record. This can make it lighter in cases where these metrics are not needed

### DIFF
--- a/src/rubrix/server/tasks/commons/dao/dao.py
+++ b/src/rubrix/server/tasks/commons/dao/dao.py
@@ -196,6 +196,7 @@ class DatasetRecordsDAO:
         search: Optional[RecordSearch] = None,
         size: int = 100,
         record_from: int = 0,
+        exclude_fields: List[str] = None,
     ) -> RecordSearchResults:
         """
         SearchRequest records under a dataset given a search parameters.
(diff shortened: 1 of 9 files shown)


src/rubrix/server/tasks/commons/dao/dao.py
```diff
@@ -210,6 +211,8 @@ def search_records(
             Number of records to retrieve (for pagination)
         record_from:
             Record from witch retrieve records (for pagination)
+        exclude_fields:
+            a list of fields to exclude from result source. Wildcard are accepted
```
dcfidalgo:
```suggestion
            a list of fields to exclude from the result source. Wildcards are accepted
```

src/rubrix/server/tasks/commons/dao/dao.py
```diff
@@ -210,6 +211,8 @@ def search_records(
             Number of records to retrieve (for pagination)
         record_from:
             Record from witch retrieve records (for pagination)
```
dcfidalgo:
```suggestion
            Record from which to retrieve the records (for pagination)
```

dcfidalgo (approved):
LGTM, just two fixes for the doc strings.

PR #838, opened 2021-12-20
feat(token class): add metrics at token level

frascuchon:
This PR computes record metrics at token level. capitalness, IOB tags, token length are some of computed metrics values.

** ONLY WORKS AT SERVER API **  We will include this new metrics in client part in another PR.


dcfidalgo (approved):

dcfidalgo:
is this ready to review?

frascuchon:
> is this ready to review?

Some tests are missing but you can start your review

PR #843, opened 2021-12-22
fix(rules): prevent division by zero

frascuchon:
This PR prevent `Division by zero` errors computing rule metrics without annotated records. 

Also includes annotated_records as part of rule api metrics response.


dcfidalgo (approved):
LGTM

PR #852, opened 2021-12-22
Docs: Review hf tutorial

dvsrepo:


dvsrepo:
Hi @leireropl 

I've made a quick review. We are almost there. Only things left would be:

- [ ]  to replace the screenshot with the score filter. Even if it takes a little time, I'd suggest to load at least 500 records to have a prettier score curve (with more values), be sure you select between 0 and 60% as the tutorial text explains, and  just take the screenshot of the score area as follows:

Screenshot 2021-12-22 at 22 57 43

- [ ] Change the title to: "Label your data to fine-tune a classifier with Hugging Face" (*to* instead of *and*)

In general we should make sure the images/videos match the description in the text, that's why I removed some of the screenshots. It's originally my fault as this tutorial is too descriptive

leireropl:
Hello, 
Sorry for the mistakes. Changes are already applied, but I think the image is too big now. Should we fix this?

dvsrepo:
Images can be resized using:

 where you can play with the percentage.

Anyway I've removed the screenshot because I doesn't add much at the end.

Thanks!

PR #853, opened 2021-12-27
fix/loss video updated

leireropl:
Please. leave feedback on the video


dvsrepo:
Hi Leire,

In general I like the video, but if possible I would recommend the following:

Start the video with no pre-validated annotations. I guess you were doing tests before. Either you delete and load again the dataset or deselect the labels. Once you start the video you should: (1) sort by desc loss, (2) start correcting the labels sequentially from the first record, no need to go to the next page. Another option to do this more sequential would be (1) sort, (2) select 1 record per page, (3) start correcting labels and use the right keyboard arrow (->) to go to the next record.

leireropl:
> Hi Leire,
> 
> In general I like the video, but if possible I would recommend the following:
> 
> Start the video with no pre-validated annotations. I guess you were doing tests before. Either you delete and load again the dataset or deselect the labels. Once you start the video you should: (1) sort by desc loss, (2) start correcting the labels sequentially from the first record, no need to go to the next page. Another option to do this more sequential would be (1) sort, (2) select 1 record per page, (3) start correcting labels and use the right keyboard arrow (->) to go to the next record.


Hi Dani,

Thanks for the tip. I have just uploaded a new one, let me know what you think.

dvsrepo:
it looks perfect!

dvsrepo (approved):

PR #856, opened 2021-12-28
updating videos + feedback

leireropl:
This is just to speed up the process- you can give me feedback when you are able to! This is the last video I have recorded. I think it is too long and maybe the window could be better centered. Anyway, I recorded different versions.


dcfidalgo (commented):
Just a small comment from my side, I would move the video from `_static/tutorials/portada.mp4` to simply `_static/portada.mp4` since it is not a video for the tutorials.

leireropl:
> Just a small comment from my side, I would move the video from `_static/tutorials/portada.mp4` to simply `_static/portada.mp4` since it is not a video for the tutorials.

Yes, you are right. I did this and also updated the video to a more recent one!

dcfidalgo:
I will close this PR. Maybe for the github readme and for the title video of the docs, it would be nice to have a longer video showing all Rubrix features. We should come up with a little video script first.

PR #861, opened 2021-12-30
UI: New rule metrics layout

### LINKED ISSUE #860: [rules] New metrics layout

### LINKED ISSUE #867: [Rules] "Delete rule" and "update summary" actions should be consistent
- When delete an already defined rule, the behaviour is a page reloading, redrawing all components.
- When update summary is applied, nothing visible happens. 

In my opinion the visual effect after both actions should be the same, since both actions affect  directly to the rules state.

leiyre:
Closes #860 
Closes #867


dvsrepo:
Looks good!

Some comments:

- [x] There's a resize of the rule labels box when saving a rule (it becomes smaller). We should keep the size of the box as stable as possible. See gif below and next comment.

- [x] If possible it would be great to reduce the "refresh" effect on the rule labels and rule metrics boxes. Ideally those should not be refreshed completely (only the numbers, the labels, etc.) and should keep a stable, fixed size. The record list below of course needs to be refreshed. I don't know if its a lot of work, if it's not too much I would tackle this for the release. See gif below.

- [x] In the label definition box, is the number of records available to show it before selecting a label ( Records 0/11.100
)? If can be known before selecting a label we should show it, otherwise is fine.

review_rules

Amelie-V:
### Define rule
- [ ] Blue block: Global metrics should be bolder
- [ ] Blue blocks: In both view (define rules and rules list) add a bit more margin (30 instead of 20)
- [ ] Define rule area: query name should be larger (like in the last version, (same size as "New Query")
- [ ] Save rule button should be fixed at the bottom (just like the info when rules is already saved)
- [ ] The goal is to make fit 3 lines of labels and then show : "+ X labels"

Amelie-V:
### Define rule
- [x] Blue block: Global metrics should be bolder (check UI)
- [x] Blue blocks: In both view (define rules and rules list) add a bit more margin (30 instead of 20)
- [x] Define rule area: query name should be larger (like in the last version, (same size as "New Query")
- [x] Save rule button should be fixed at the bottom (just like the info when rules is already saved)
- [ ] The goal is to fit 3 lines of labels and then show : "+ X labels" (just in case, i cant test it)
- [ ] Spaces between search box and the 2 blocks should be 10px (just like between the two blocks)

dvsrepo:
@leiyre I've reviewed all raised points and it's looking good!!

Testing I have found these two issues:

- [x] Avg. precision and total incorrect/correct suddenly become 0 both in rule definition and rule management views (see gif)
- [x] When a rule is saved, you are able to change the label but clicking on save label has no effect.


review_rules2

frontend/components/text-classifier/labeling-rules/RuleDefinition.vue
```diff
@@ -72,12 +95,16 @@ export default {
       await this.dataset.viewSettings.enableRulesSummary();
     },
     async updateRule() {
-      this.refresh++;
       await this.$fetch();
     },
+    updateLabel(label) {
+      this.selectedLabels = label;
```
frascuchon: changing the selected label should change the rule metrics?

frontend/components/text-classifier/labeling-rules/RuleDefinition.vue
```diff
@@ -54,10 +69,18 @@ export default {
     query() {
       return this.dataset.query.text;
     },
+    activeLabel() {
+      return this.selectedLabels
+        ? this.selectedLabels[0]
```
frascuchon, line 74: `selectedLabel` is updated as the label value in `updateLabel` method. Should be considered still as an array ?

leiyre:
@Amelie-V 
> * [ ]  The goal is to fit 3 lines of labels and then show : "+ X labels" (just in case, i cant test it)

The width is dynamic, it's better to define a fixed number of elements. By default is a max of 10

> * [ ]  Spaces between search box and the 2 blocks should be 10px (just like between the two blocks)

The spacing is common to all views, the bottom margin of the seacrhbox is 14px, if we change it to 10px it will affect all views. it's ok?

frascuchon:
Comment from my part:

The message "The query is already saved as a  rule" when a rule is already stored it could be confusing if we only change the label:

Screenshot 2022-01-03 at 11 40 05


Screenshot 2022-01-03 at 11 40 12

In both cases the query remains the same. As suggestion (but sure I'm not an expert) keeping the save rule button enabled/disabled depending on the rule was already created. Or maybe change the action name (Create Rule/ Save rule)

@Amelie-V ?

dvsrepo:
I agree. The message is confusing and the interaction too. 

In my opinion:

1. Improve the message to: This query with this label are already saved as rule.
2. Showing disabled button is a good idea, but I would keep the message otherwise it will still be confusing.
3. The message after saving the rule shouldnt be the message above, but: The rule has been saved.

frontend/components/text-classifier/labeling-rules/RuleDefinition.vue
leiyre: I think yes, it should be treated as an array with a single element, maybe in the future with multiple elements.

I changed label => labels

frontend/components/text-classifier/labeling-rules/RuleDefinition.vue
leiyre: yes, enables the communication of the current label to the metrics component

frascuchon:
> 3. The message after saving the rule shouldnt be the message above, but: The rule has been saved.

We have already a global mechanism to notify completed actions. I prefer show this message as a toast notification.

frascuchon:
It could be nice to include some color schema to highlight the best rule metrics.  The greater value for precision, the better, or the lower value for incorrect records, the better too. This could help to users to analyze rule

frascuchon:
Affected records count (`Records: ...`) should be shown only if we have label info. Otherwise, it's quite confusing: 

Screenshot 2022-01-03 at 14 47 25

The same could be applied to coverage and annotated coverage (Giving 0% of coverage while rule query returns results is confusing)

frontend/components/text-classifier/labeling-rules/RulesManagement.vue
```diff
@@ -1,46 +1,60 @@
...
-        @sort-column="onSortColumns"
-        @onActionClicked="onActionClicked"
-        @close-modal="closeModal"
-      />
-      <re-button
-        v-if="formattedRules.length"
-        class="button-primary"
-        @click="updateSummary()"
-        >Update Summary</re-button
+    <div v-else-if="!$fetchState.error">
```
frascuchon, line 4: What happens if something went wrong? Is there any error handing defined?

frontend/components/text-classifier/labeling-rules/RulesMetrics.vue
```diff
@@ -0,0 +1,340 @@
...
+        <div
+          class="rule-metrics__item"
+          v-for="(field, index) in placeholderFields"
+          :key="index"
+        >
+          <p class="metric__title">{{ field }}</p>
+          <p class="metric__placeholder">-</p>
+        </div>
+      </template>
+      <template v-else-if="!$fetchState.error">
```
frascuchon, line 16: ditto

frontend/components/text-classifier/labeling-rules/RulesManagement.vue
leiyre: At view level we do not have anything that refers to the error, it could be contemplated. Right now, I think it shows the toast with the error and the component will be shown empty

leiyre:
fix #867

frascuchon (approved):
Closing this PR and all pending tasks must be transcribed as separate issues

PR #866, opened 2022-01-03
chore: expose load_rules from base module

### LINKED ISSUE #854: [rules] Expose load_rules at the parent module
I'd like the import to look like this:

`from rubrix.labeling.text_classification import load_rules`

Currently is: `from rubrix.labeling.text_classification.rule import load_rules`

frascuchon:
Closes: #854

### DIFF
--- a/src/rubrix/labeling/text_classification/__init__.py
+++ b/src/rubrix/labeling/text_classification/__init__.py
@@ -15,5 +15,5 @@
 
 from .label_errors import find_label_errors
 from .label_models import FlyingSquid, Snorkel
-from .rule import Rule, RuleNotAppliedError, load_rules
+from .rule import Rule, load_rules
 from .weak_labels import WeakLabels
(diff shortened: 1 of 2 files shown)


src/rubrix/labeling/text_classification/__init__.py
```diff
@@ -15,5 +15,5 @@
 
 from .label_errors import find_label_errors
 from .label_models import FlyingSquid, Snorkel
-from .rule import Rule
+from .rule import Rule, RuleNotAppliedError, load_rules
```
dcfidalgo: I would prefer to keep the imports of the errors at the respective module level, just to make them more explicit and keep the namespace `rubrix.labeling.text_classification` a bit cleaner (the errors of the `label_errors`, `weak_labels` and `label_models` modules are also not exposed in this namespace).
```suggestion
from .rule import Rule, load_rules
```

dcfidalgo (commented):
I was about to tackle this, you're always faster :) Just one inline comment!

src/rubrix/labeling/text_classification/__init__.py
frascuchon: It's ok for me. Just a  consideration. I think should depends on number of error classes we have for this module. Maybe keeping error in a centralized `errors.py` module at `text_classification` module level, should simplify to understard what kind of errors are tackled in weak supervision.

PR #878, opened 2022-01-05
refactor(rules): using labeling rules management with vuex store

frascuchon:
This PR changes app interactions for labeling rules management. Rules and metrics are stored in vuex store and components use the store for communication. Reduce drastically the number of calls done to api server and improve overall performance.

@leiyre take a look carefully since a lot of changes was made.


leiyre (approved):
Much better, please check my latest changes

frontend/components/text-classifier/labeling-rules/RuleDefinition.vue
```diff
@@ -44,66 +34,82 @@
...
-      dataset: this.dataset,
-      query: this.query,
-    });
+    if (this.rules === undefined) {
+      await this.dataset.refreshRules();
+    }
+    if (!this.hasMetrics) {
+      await this.dataset.refreshRulesMetrics();
+    }
+    if (!this.currentRule && this.query) {
```
frascuchon, line 58: Could we do this logic inside query watcher or directly calling the `updateCurrentRule` method?

frascuchon (commented):
- [x] We must compute number of affected records when a query is defined(before select label)
- [x] The update metrics button should refresh dataset rules too (getch rules changes from server in order to compute metrics properly)

frontend/components/text-classifier/labeling-rules/RuleDefinition.vue
leiyre: I'm not sure, fetch actions must be triggered on refresh and also the first time before the query change

leiyre (approved):

PR #882, opened 2022-01-07
fix(UI): Remove case sensitive when filtering labels

### LINKED ISSUE #877: [Annotate] Labels are not found when using cased searches
When searching for a label in the "Annotate" mode (TextClassification), uppercased letters do not seem to work.
For example, when searching for the label `Physics`, it will only show results when I search for `phys`, but not for `Phys`.

leiyre:
fix #877


dcfidalgo (approved):
:+1: LGTM

PR #886, opened 2022-01-07
UI: Improve code snippet for empty workspace

### LINKED ISSUE #881: [Datasets page] Improve code snippet for empty worskpace
The current code snippet has caused several confusions for first-time users. The main issue is that is focused on more complex installations (several workspaces, custom API keys) and that is misaligned with our docs (where we don't use the `rb.init()` method). When we start having more complex installations the docs will recommend to set ENVs for the api key and url anyway.

I'd recommend to have this code snippet instead:

```python
import rubrix as rb

# If you are using the default installation you can skip this line
rb.set_workspace("${ws}")

rb.log(
  rb.TextClassificationRecord(inputs="my first rubrix example"), 
  name='example-dataset'
)
```

Another option is to keep the rb.init but commented out with a message like this:

`# If you have configured your own API url and key, run this line`

leiyre:
fix #881

### DIFF
--- a/frontend/components/commons/datasets-list/DatasetsEmpty.vue
+++ b/frontend/components/commons/datasets-list/DatasetsEmpty.vue
@@ -19,7 +19,6 @@
 
 <script>
 import "assets/icons/datasets-empty";
-import { workspaceHome } from "@/models/Workspace";
 export default {
   props: {
     workspace: {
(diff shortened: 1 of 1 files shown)


dvsrepo (approved):
💯

PR #888, opened 2022-01-07
(UI) Rules summary metrics format

### LINKED ISSUE #879: [Rule metrics] Table summary and rule metrics formats
Just a suggestion that can be discussed.

The precision, coverage and annotated coverage for table summary could be formatted in same way that rule metrics does:

Screenshot 2022-01-06 at 12 09 00

Screenshot 2022-01-06 at 12 09 08

leiyre:
fix #879


frascuchon (approved):
Great!

PR #892, opened 2022-01-07
fix(rules): prevent lost rules with dataset updates

frascuchon:
This PR fixes problem with lost dataset rules after annotation (or even logging new records). Found after intensive @dcfidalgo  feature testing. Thanks @dcfidalgo !

This must be included to the very first release


dcfidalgo (approved):
:100:

PR #904, opened 2022-01-10
UI: drop access to rule name field

### LINKED ISSUE #903: [Define rules] drop access to rule name field
- For query view, show query as rule
- For rules management table, show query instead of name

leiyre:
fix #903

### DIFF
--- a/frontend/components/text-classifier/labeling-rules/RuleLabelsDefinition.vue
+++ b/frontend/components/text-classifier/labeling-rules/RuleLabelsDefinition.vue
@@ -74,7 +74,7 @@
       v-if="
         currentRule &&
         selectedLabels.includes(currentRule.label) &&
-        currentRule.description === query
+        currentRule.description === currentRule.query
       "
     >
       {{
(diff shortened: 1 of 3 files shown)


frontend/components/text-classifier/labeling-rules/RuleLabelsDefinition.vue
```diff
@@ -74,13 +74,13 @@
       v-if="
         currentRule &&
         selectedLabels.includes(currentRule.label) &&
-        currentRule.description === description
+        currentRule.description === query
```
frascuchon: could just use `currentRule.query` instead ?

frontend/components/text-classifier/labeling-rules/RuleLabelsDefinition.vue
```diff
@@ -193,24 +189,19 @@ export default {
         await this.updateRule({
           dataset: this.dataset,
           label: this.selectedLabels[0],
-          description: this.description,
+          description: this.query,
```
frascuchon: Maybe we can just stop storing description?

frontend/components/text-classifier/labeling-rules/RuleLabelsDefinition.vue
```diff
@@ -193,24 +189,19 @@ export default {
...
         });
       } else {
         await this.defineRule({
           dataset: this.dataset,
           label: this.selectedLabels[0],
-          description: this.description,
+          description: this.query,
```
frascuchon: ditto

frascuchon (changes requested):

frascuchon (approved):

PR #907, opened 2022-01-10
feat(rules): allow compute partial query rule metrics

frascuchon:
rule label is not required when rule is not saved. This will compute only query coverages.

Also set precision to `None` when annotation coverage is 0.

See #895

### DIFF
--- a/src/rubrix/server/tasks/text_classification/service/service.py
+++ b/src/rubrix/server/tasks/text_classification/service/service.py
@@ -336,8 +336,10 @@ class TextClassificationService:
             The provided rule query. If already created in dataset, the ``label``
             param will be omitted
         label:
-            Label used for rule metrics. If not provided, defined in stored rule will be used
-            if rule is found, else related label metrics won't be computed
+            Label used for rule metrics. If not provided and no rule was stored with provided query,
+            no precision will be computed. Otherwise, the store rule label will be used for metrics
+            computation
+
 
         Returns
         -------
--- a/src/rubrix/server/tasks/text_classification/service/service.py
+++ b/src/rubrix/server/tasks/text_classification/service/service.py
@@ -337,6 +337,7 @@ class TextClassificationService:
             param will be omitted
         label:
             Label used for rule metrics. If not provided, defined in stored rule will be used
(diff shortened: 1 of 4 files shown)


src/rubrix/server/tasks/text_classification/service/service.py
```diff
@@ -337,6 +337,7 @@ def compute_rule_metrics(
             param will be omitted
         label:
             Label used for rule metrics. If not provided, defined in stored rule will be used
+            if rule is found, else related label metrics won't be computed
```
dcfidalgo: Not sure if I understand this, maybe we can rewrite this?
and do we still need this first `if label is None` check in [line 349](https://github.com/recognai/rubrix/pull/907/files#diff-4e658dba045835ec32c8251b1f51f608b56971dfc55d40d368e559f595e76f5eR351-R355), since `self.__labeling__.compute_rule_metrics` can now handle `label=None`?

dcfidalgo (commented):
LGTM, just one inline comment.

src/rubrix/server/tasks/text_classification/service/service.py
frascuchon: There are 2 cases when label is not provided:

1. Query exists as rule. The label will be obtain from stored label. It's why the line 349 make sense.
2. Query does not exist. No label will be used.

src/rubrix/server/tasks/text_classification/service/service.py
```diff
@@ -336,7 +336,10 @@ def compute_rule_metrics(
             The provided rule query. If already created in dataset, the ``label``
             param will be omitted
         label:
-            Label used for rule metrics. If not provided, defined in stored rule will be used
+            Label used for rule metrics. If not provided and no rule was stored with provided query,
+            no precision will be computed. Otherwise, the store rule label will be used for metrics
+            computation
+
```
frascuchon:
```suggestion
            Label used for the rule metrics. If not provided and no rule was stored with the 
            provided query, no precision will be computed. 
            Otherwise, the label from the stored rule will be used to compute the metrics.
```

PR #911, opened 2022-01-10
fix(security): providing default workspace should pass check

### LINKED ISSUE #906: [Dataset Copy] rb.copy does not work explicitly specifying my workspace
Cannot create a copy when specifying my workspace. 
This works:
```python
rb.copy("ds", "ds2")
```
this does not, no error is thrown, no copy is created:
```python
rb.copy("ds", "ds2", workspace="david")
# or
rb.set_workspace("recognai")
rb.copy("ds", "ds2", workspace="david")
```

frascuchon:
closes #906

### DIFF
--- a/src/rubrix/server/security/model.py
+++ b/src/rubrix/server/security/model.py
@@ -85,7 +85,7 @@ class User(BaseModel):
             The original workspace name if user belongs to it
 
         """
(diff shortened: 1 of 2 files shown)


dcfidalgo (approved):
LGTM

PR #939, opened 2022-01-11
[WeakSupervision] Change load_rules import path in guide and tutorial

dcfidalgo:
Changes the import path of the `load_rules` functions in the weak supervision guide and tutorial.

See #854


dvsrepo (approved):
💯

PR #946, opened 2022-01-13
build(#937): remove big files in the docs from sdist build

### LINKED ISSUE #937: Build source distribution
Using `python -m  build` as default will create a both, binary and source distribution. The binary one is okey but the source distro include doc files and resultant package it's too heavy to post to pypi.

Any way to configure which files we want to include in source distr?

For now, source distribution is omitted

dcfidalgo:
Closes #937 
This PR adds a `MANIFEST.in` file, that removes some parts from the docs for the sdist build. Prior to that the `tar.gz` file was too big to be uploaded to pypi. Now it's around 3MB. Using setuptools_scm, there is no other way than using a manifest file.

Actually, we need the sdist build to publish our package in conda. @frascuchon Can we add a sdist to an already published pypi package?

This PR also includes some cleanup in the docs folder.


frascuchon:
> Actually, we need the sdist build to publish our package in conda. @frascuchon Can we add a sdist to an already published pypi package?

Not using the ci process. Maybe we can upload manually using [`twine`](https://twine.readthedocs.io/en/stable/)


Just for testing, could you format the PR title using the conventional commits format?

frascuchon (approved):
Please, use conventional commits format for PR title

dcfidalgo:
@frascuchon done

PR #986, opened 2022-01-17
fix(#974): display the dropdown in the last record of the scroll

### LINKED ISSUE #974: [Annotation] Not able to annotate token classification records
When I have only one record (or the last record in a list), I cannot annotate the records properly, see screenshot.
Happens to me in Firefox and Chrome:
Screenshot from 2022-01-14 13-49-11

leiyre:
fix #974
This PR enables the dropdown overlay on the last record in the token classifier scroll

### DIFF
--- a/frontend/components/commons/results/ResultsList.vue
+++ b/frontend/components/commons/results/ResultsList.vue
@@ -219,14 +219,19 @@ export default {
     min-height: 140px;
   }
 }
-.scroller {
-  padding-bottom: 61px;
-  .fixed-header & {
-    padding-bottom: 200px;
-  }
-}
 </style>
 <style lang="scss">
+.vue-recycle-scroller__item-wrapper {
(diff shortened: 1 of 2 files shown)


dcfidalgo:
:+1: Now it's possible to annotate the last record as well. Two minor issues, which are maybe a bit more difficult to fix:
- it looks a bit strange that the scroll does not stop with the last record
- a long list of labels is still cropped, see the screenshot
Screenshot from 2022-01-17 17-59-53
t

leiyre:
> it looks a bit strange that the scroll does not stop with the last record

yes, is a bit difficult manage dropdown in the infinite scroll, I will think about how to improve it

> a long list of labels is still cropped, see the screenshot

I fixed this for large screens

dcfidalgo (approved):
Nice! :100:

PR #987, opened 2022-01-17
feat(#735): add warning when agent but no prediction/annotation is provided

### LINKED ISSUE #735: [Log] Provided prediction agent without prediction is not registered
Since prediction agent is only stored if prediction info is provided, logging records from client with agent and no prediction info will produce records without the agent info.

A warning/error should be raised for client users. Same problem with annotations

dcfidalgo:
Closes #735 
This PR adds a warning when a `*_agent` is provided but no corresponding `annotation`/`prediction`.
Also, some refactoring is done, introducing a base class for the records in which common checks are performed.

### DIFF
--- a/src/rubrix/client/models.py
+++ b/src/rubrix/client/models.py
@@ -21,7 +21,7 @@ import datetime
 import warnings
 from typing import Any, Dict, List, Optional, Tuple, Union
 
-from pydantic import BaseModel, Field, validator
+from pydantic import BaseModel, Field, root_validator, validator
 
 from rubrix._constants import MAX_KEYWORD_LENGTH
 from rubrix.server.commons.helpers import limit_value_length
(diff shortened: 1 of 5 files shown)


src/rubrix/client/models.py
```diff
@@ -27,6 +27,74 @@
...
+    metadata: Dict[str, Any] = Field(default_factory=dict)
+    status: Optional[str] = None
+    event_timestamp: Optional[datetime.datetime] = None
+    metrics: Optional[Dict[str, Any]] = None
+
+    @validator("metadata", pre=True)
+    def check_value_length(cls, metadata):
+        return _limit_metadata_values(metadata)
+
+    def __init__(self, *args, **kwargs):
```
frascuchon: Could we use a pydantic `@rootValidator` instead of override the `init` method? See https://pydantic-docs.helpmanual.io/usage/validators/#root-validators

dcfidalgo: Yes, much better with `root_validators`, thanks for the tip!

frascuchon (approved):
👍

Issue #951, opened 2022-01-13
[Labeling Rules] Filter uncovered records


frascuchon:
It could be nice to explore/annotate records that are not covered by any defined rule. Some kind of filter that we can leave activated while navigating between modes

Amelie-V:
#991

Amelie-V:
- [x] add checkbox into annotation Filter

frascuchon:
More specific copy text for checkbox: "Show only not covered records by rules"

@dcfidalgo @dvsrepo @Amelie-V @leiyre ? Wha do you think?

PR #991, opened 2022-01-17
feat(#951): new uncovered_by_rules records filter

frascuchon:
This PR includes at API level a filter param for text classification to enable only uncovered  records in search results (in addition with other search parameters)

See #951


src/rubrix/server/tasks/text_classification/api/model.py
```diff
@@ -489,7 +489,12 @@ class TextClassificationQuery(BaseModel):
     status: List[TaskStatus] = Field(default_factory=list)
     predicted: Optional[PredictionStatus] = Field(default=None, nullable=True)
 
-    def as_elasticsearch(self) -> Dict[str, Any]:
+    only_uncovered: bool = Field(
+        default=False,
+        description="If enabled, filter records that are not affected by defined rules",
+    )
+
+    def as_elasticsearch(self, rules: List[LabelingRule]) -> Dict[str, Any]:
```
dcfidalgo: Maybe I would provide a default here, just in case `service.py` is not the only place where `as_elasticsearch` is called, and to not break calls of this method.
```suggestion
    def as_elasticsearch(self, rules: Optional[List[LabelingRule]] = None) -> Dict[str, Any]:
```

src/rubrix/server/tasks/text_classification/api/model.py
```diff
@@ -489,7 +489,12 @@ class TextClassificationQuery(BaseModel):
     status: List[TaskStatus] = Field(default_factory=list)
     predicted: Optional[PredictionStatus] = Field(default=None, nullable=True)
 
-    def as_elasticsearch(self) -> Dict[str, Any]:
+    only_uncovered: bool = Field(
```
dcfidalgo: nitpicking, the `only_uncovered` is missing in the docstring.

dcfidalgo (commented):
Nice! Just a question, did you consider making `rules` an attribute of the `TextClassificationQuery`? Seems more natural to me, but maybe there are obstacles I don't see.

src/rubrix/server/tasks/text_classification/api/model.py
frascuchon: better using `uncovered_by_rules`  as part of query param

src/rubrix/server/tasks/text_classification/api/model.py
frascuchon: done

src/rubrix/client/sdk/text_classification/models.py
```diff
@@ -150,6 +150,11 @@ class TextClassificationQuery(BaseModel):
     status: List[TaskStatus] = Field(default_factory=list)
     predicted: Optional[PredictionStatus] = Field(default=None, nullable=True)
 
+    uncovered_by_rules: List[str] = Field(
+        default_factory=list,
+        description="List of rule queries that WILL NOT cover the result records",
```
dcfidalgo:
```suggestion
        description="List of rule queries that WILL NOT cover the resulting records",
```

src/rubrix/server/tasks/text_classification/api/model.py
```diff
@@ -473,6 +473,8 @@ class TextClassificationQuery(BaseModel):
         List of task status
     predicted: Optional[PredictionStatus]
         The task prediction status
+    uncovered_by_rules:
+        List of rule queries filtering records uncovered by these rules.
```
dcfidalgo:
```suggestion
        Only return records that are NOT covered by these rules.
```

src/rubrix/server/tasks/text_classification/api/model.py
```diff
@@ -489,6 +491,11 @@ class TextClassificationQuery(BaseModel):
     status: List[TaskStatus] = Field(default_factory=list)
     predicted: Optional[PredictionStatus] = Field(default=None, nullable=True)
 
+    uncovered_by_rules: List[str] = Field(
+        default_factory=list,
+        description="List of rule queries that WILL NOT cover the result records",
```
dcfidalgo:
```suggestion
        description="List of rule queries that WILL NOT cover the resulting records",
```

dcfidalgo (approved):
:100: Just a few inline suggestions

PR #1000, opened 2022-01-19
docs: update video for spacy ner tutorial

leireropl:
New video


dcfidalgo:
Thanks @leireropl reading through the tutorial, I realized that we do not annotate anything, and it's much more about the exploration part. So I would propose a video in the line of:
- start in the explore mode with the cursor in the middle, and the metrics window closed
- open the Stats metrics, show a bit "Predicted as" and "Annotated as" mentions
- go to the filters (either Predictions or Annotations) and filter for some entity in the "Predicted as/Annotated as" section.

leireropl:
> Thanks @leireropl reading through the tutorial, I realized that we do not annotate anything, and it's much more about the exploration part. So I would propose a video in the line of:
> 
> * start in the explore mode with the cursor in the middle, and the metrics window closed
> * open the Stats metrics, show a bit "Predicted as" and "Annotated as" mentions
> * go to the filters (either Predictions or Annotations) and filter for some entity in the "Predicted as/Annotated as" section.

You are right! My bad, I simply replicated the former video... Let's see if this one is better!

dcfidalgo (approved):
perfect :100: thanks a lot!

Issue #994, opened 2022-01-18
[Install] update the installation docs recomending the python -m rubrix.server way   


dcfidalgo:
The idea is following default installation:
```bash
# install server dependencies
pip install rubrix[server]

# if no ES running
docker run ...

# run web app
python -m rubrix.server
```

advantage: 
- without docker-compose: easier to update the rubrix version
- with ES container running in the background, easier to avoid data loss
- we advocate only one way to start the web app, docker-compose is an extra that can still be used by "experts"

disadvantage:
- we include the server dependencies in the default installation, installation is not as lightweight as before

note:
- in the conda package we need to update the dependencies

dcfidalgo:
@frascuchon @dvsrepo

dvsrepo:
I like it a lot this way. Also, it's a good opportunity to have a very clear documentation for installation section

frascuchon:
We can include also the basics of howto launch with [`uvicorn` directly](https://fastapi.tiangolo.com/deployment/manually/#run-the-server-program)

PR #1014, opened 2022-01-21
chore(#994): simplify server start

frascuchon:
This PR changes how to launch rubrix server to:

```bash
python -m rubrix
```

The old command (`python -m rubrix.server`) is also available 

Also, include setup info about uvicorn setup. Basically, you can launch rubrix server using uvicorn:

```bash
uvicorn rubrix:app
```

### DIFF
--- a/docs/getting_started/setup&installation.rst
+++ b/docs/getting_started/setup&installation.rst
@@ -84,12 +84,29 @@ This method may be preferred if you
 
 .. code-block::
 
-   python -m rubrix.server
+   python -m rubrix
 
 By default, the Rubrix server will look for your Elasticsearch endpoint at ``http://localhost:9200``.
 But you can customize this by setting the ``ELASTICSEARCH`` environment variable.
 
 **If you are already running an Elasticsearch instance for other applications and want to share it with Rubrix**, please refer to our :ref:`advanced setup guide <configure-elasticsearch-role-users>`.
+
+Uvicorn
+"""""""
+
+Since, rubrix server is built using fastapi, you can launch rubrix server using **uvicorn** ASGI server:
+
+.. code-block::
+
+   uvicorn rubrix:app
+
+*(for older rubrix version you should launch as)*
+
+.. code-block::
+
+   uvicorn rubrix.server.server:app
+
+See more details `here <https://fastapi.tiangolo.com/deployment/manually/#run-a-server-manually-uvicorn>`_
 
 3. Start logging data
 ---------------------
--- a/docs/getting_started/setup&installation.rst
+++ b/docs/getting_started/setup&installation.rst
@@ -94,7 +94,7 @@ But you can customize this by setting the ``ELASTICSEARCH`` environment variable
 Uvicorn
 """""""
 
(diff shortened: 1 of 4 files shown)


docs/getting_started/setup&installation.rst
```diff
@@ -84,13 +84,30 @@ This method may be preferred if you
...
 
 By default, the Rubrix server will look for your Elasticsearch endpoint at ``http://localhost:9200``.
 But you can customize this by setting the ``ELASTICSEARCH`` environment variable.
 
 **If you are already running an Elasticsearch instance for other applications and want to share it with Rubrix**, please refer to our :ref:`advanced setup guide <configure-elasticsearch-role-users>`.
 
+Uvicorn
+"""""""
+
+Since, rubrix server is built using fastapi, you can launch rubrix server using **uvicorn** ASGI server:
```
dcfidalgo:
```suggestion
Since the Rubrix server is built on fastapi, you can launch it using **uvicorn**:
```

src/rubrix/__init__.py
```diff
@@ -38,13 +38,29 @@
...
+    from rubrix.server.server import app
+except ModuleNotFoundError as ex:
+    module_name = ex.name
+
+    def fallback_app(*args, **kwargs):
+        raise RuntimeError(
+            "\n"
+            f"Cannot start rubrix server. Some dependencies was not found:[{module_name}].\n"
+            "Please, install missing modules or reinstall rubrix with server extra deps:\n"
+            "pip install rubrix[server]"
```
dcfidalgo, line 59: Do we have to define the extra `module_name` variable?

```suggestion
            "\n"
            f"Cannot start the Rubrix server due to following missing dependencies:[{ex.name}].\n"
            "You can install the missing dependencies with the command:\n"
            "pip install rubrix[server]"
```

dcfidalgo (approved):
:100: Just a few inline suggestions. also, the setup&installation docs will be updated soon due to #994.

dcfidalgo:
@frascuchon ah, I forgot, if we want to deprecate the `python -m rubrix.server` command in the future, I would add a `FutureWarning`.

dvsrepo:
One additional comment, which is a feedback we get often. I don't know where is the best place but I would like to have a very clear message when some starts the server and the Elasticsearch endpoint is not available/running, etc, I don't know if we check that at some point but I recall the message it's a standard trace. We should be very informative at this point, explaining:

1. We checked the elastic search endpoint at {ES_URL} and is not responding/available..
2. Please make sure your Elasticsearch is running, or (check the user roles config) <- whatever message it makes sense.

Ideally this should be applicable to other scenarios when the ES instance is not responding --> including a trace in the server process.

I think this is key to help our users troubleshoot the issues with Elastic which is frequently the only bottleneck for getting started.

I understand this should go on another issue/PR

frascuchon:
> Member

Yes, i prefer include in a separate PR. Could you create a issue to track the change @dvsrepo ?

frascuchon:
> @frascuchon ah, I forgot, if we want to deprecate the `python -m rubrix.server` command in the future, I would add a `FutureWarning`.

Done

PR #1017, opened 2022-01-23
[Weak supervision] Simplified version of the skweak tutorial.

### LINKED ISSUE #942: [weak supervision] Feedback on skweak tutorial
Hi @ruanchaves, I finally found a free slot to go through the tutorial. Here's a detailed review:

1. I'd try to streamline the preprocessing part. Maybe create a preprocessed dataset and move the preprocessing part to an appendix? The in the preprocessing section we just describe the data format.
2. As a minor comment for the tag_set we could use `dataset["train"].features["ner_tags"].feature.names`.
3. We strip BI tags, I would add a short sentence saying why we need to do this.
4.I'd try to log the conll train split (without tags) in Rubrix right at the start, so you can test some of the heuristics (for example the company ones, I guess that as conll is an English corpus we could remove `sa` and `ag`, and maybe even find some other suffix that work more or less well?). Also in terms of the workflow and support by Rubrix I believe that including this step adds some value, as you are able to verify your hypotheses by directly searching.
5.  Question: In the gazetteers part, if the conclusion after exploring is that only the wiki gazetteer is working, we should explain what to do, do we don't make use of the other gazetteers? 

The most fundamental issue that I see is: *using conll for evaluation and only defining lfs and aggregating on companies and orgs*. I believe this comes from the fact that we are reusing much from the original skweak tutorial which actually builds a NER model with those entities only. This is likely one reason the final models are underperforming. More specifically:

1. Our `hmm_model = CustomHMM("hmm", ["COMPANY", "OTHER_ORG"])` ignores `PER` and other `MISC` entities from the pre-trained model lf. I have checked this after the first Rubrix logging, you can see there are no `PER`, etc. In any case, if we don't add lfs for at least PER and maybe other MISC, it won't make much sense because the only supervision source will come from the spaCy model.
2. If I understand correctly, we then train a spaCy model with mainly the `hmm` labels, falling back to the pre-trained spaCy annotations if we have no ents for a record. So the only source of PER and other MISC is this fraction of the records (those with no annotation from the hmm) and as said earlier only from this one source of supervision.
3. Another general thing: NER training sets and models largely benefit from unlabelled examples, this is examples where there shouldn't be any entity. I'd like to make sure we are training the end models with examples with no entities.
4. More of a question: couldn't we do the MISC alignment before defining the labelling functions, instead of before training?

I guess that's it. Recap:

1. Streamline the pre-processing part.
2. Make use of Rubrix search capabilities from the start to define the heuristics (we can later in the same PR provide a video)
3. If we decide to evaluate on conll we should include heuristics for PER and MISC, and include them in the HMM aggregation process. Other option is focus only on improving the performance for the ORG label, and thus: 1) evaluate ORG performance with the pre-trained spaCy model (by using only that entity label from conll), 2) build an skweak training set with lfs and the pre-trained model, and 3) evaluate ORG performance with models trained on this dataset (with the hope they improve over the pre-trained model baseline).

We can go through this tomorrow.

ruanchaves:
This is an initial draft of the simplified version of the skweak tutorial. These are the next steps before we have a working version:

- Save the preprocessed dataset to the Rubrix repository and turn the Preprocessing section into an appendix.
- Visualize the data and improve the regular expressions. 
- Introduce a baseline.
- Show how the rules can be created with Rubrix through gifs / screenshots.
- Add clear descriptions to each section. 

Closes #942


dvsrepo (approved):
I've just finished reviewing and it's looking really great!! Good job.

The only thing which could be improved is the quality of the screenshots, but maybe it's better to merge this right now so we're able to promote it this week. Let's see if we go back and review the imgs or do a short video.

dvsrepo (approved):
💯 much better now 🚀

dcfidalgo:
Great job Ruan! Just a few minor comments:
- not sure if the following is really necessary in the setup section, did you encounter problems with the installation of skweak otherwise?
  > %pip install -U pip setuptools wheel -qqq

- the second example in the docstring of the `sports_match_detector` heuristic would not match, maybe use another, or just the first one
- Following could be a bit misleading in our case since we only have one label:
  > As we have implemented several independent labeling functions, the annotations produced by our heuristic rules may overlap and even conflict with each other.
For those cases, skweak includes an aggregation model called majority voter. It considers each labeling function as a voter and outputs the most frequent label.

  Maybe one could rephrase it, or add a short sentence explaining that in our case the majority voter is really just a simple aggregator. Though, I'm not sure how it deals with partial overlaps ...

Also, we should create a reminder issue, that once we release the new NER improvements in the web app, we should simplify the "Logging to Rubrix" section, by using the `annotation` argument for the gold labels. I would also put this section after the "Aggregation" to simply log the final single set of annotations and compare it to the gold labels.

Out of curiosity: did you get any improvements by balancing the training set, that is having the same amount of records with and without annotations?

Again, great job, and after addressing the minor issues above, I would merge this PR.

ruanchaves:
> * not sure if the following is really necessary in the setup section, did you encounter problems with the installation of skweak otherwise?
>   > %pip install -U pip setuptools wheel -qqq

Removed.
These are real dependencies but most environments already come with these libs installed.

> * the second example in the docstring of the `sports_match_detector` heuristic would not match, maybe use another, or just the first one

Fixed. 

>   Maybe one could rephrase it, or add a short sentence explaining that in our case the majority voter is really just a simple aggregator. Though, I'm not sure how it deals with partial overlaps ...

Fixed. 
 
> Out of curiosity: did you get any improvements by balancing the training set, that is having the same amount of records with and without annotations?

The current version of the notebook is balanced: it has the same of amount of records with and without annotations on the training set.

This is what happens if you don't balance the training set:

image

The F score will drop down from 43.67% to 15.28%, which is 22.02% below the baseline.

dcfidalgo:
> This is what happens if you don't balance the training set:

Wow, ok thanks for the clarification!

Will merge this now, an awesome tutorial!

dcfidalgo:
@frascuchon arg, just noticed after the merge that I did not update the title of this PR... can we correct this?

PR #1024, opened 2022-01-24
fix(#1008): set the event_timestamp when annotating TextClassifier

### LINKED ISSUE #1008: [Annotation] Annotating TextClassificationRecords sets the event_timestamp to None
When annotating text classification records via the web app, the event_timestap is lost.
```python
rec = rb.TextClassificationRecord(inputs="test", prediction=[("a", 0.5), ("b", 0.5)], event_timestamp=datetime.now())
rb.log(rec, "test_event_timestamp")
rb.load("test_event_timestamp")   # event_timestamp is fine
# annotate the record in the web app
rb.load("test_event_timestamp")   # event_timestamp is None
```

leiyre:
fix #1008

### DIFF
--- a/frontend/models/TextClassification.js
+++ b/frontend/models/TextClassification.js
@@ -26,13 +26,11 @@ class TextClassificationRecord extends BaseRecord {
     explanation,
     multi_label,
     predicted_as,
-    event_timestamp,
     ...superData
   }) {
     super(superData);
     this.inputs = inputs;
     this.explanation = explanation;
-    this.event_timestamp = event_timestamp,
     this.multi_label = multi_label;
     this.predicted_as = predicted_as;
   }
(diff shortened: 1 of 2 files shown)


frontend/models/TextClassification.js
```diff
@@ -26,11 +26,13 @@ class TextClassificationRecord extends BaseRecord {
     explanation,
     multi_label,
     predicted_as,
+    event_timestamp,
```
frascuchon: This will happen for all tasks. Could you include this change at main `BaseRecord` level?

frascuchon (changes requested):

frontend/models/TextClassification.js
leiyre: sure!

frascuchon (approved):

PR #1030, opened 2022-01-25
fix(#1023): handle elasticsearch connection problems on server startup

### LINKED ISSUE #1023: [Errors] Provide clear messages about Elasticsearch status when running the server
I don't know where is the best place but I would like to have a very clear message when some starts the server and the Elasticsearch endpoint is not available/running, etc, I don't know if we check that at some point but I recall the message it's a standard trace. We should be very informative at this point, explaining:

We checked the elastic search endpoint at {ES_URL} and is not responding/available..
Please make sure your Elasticsearch is running, or (check the user roles config) <- whatever message it makes sense.
Ideally this should be applicable to other scenarios when the ES instance is not responding --> including a trace in the server process.

I think this is key to help our users troubleshoot the issues with Elastic which is frequently the only bottleneck for getting started.

frascuchon:
Closes #1023

### DIFF
--- a/src/rubrix/server/server.py
+++ b/src/rubrix/server/server.py
@@ -99,7 +99,7 @@ def configure_app_startup(app: FastAPI):
         except elasticsearch.exceptions.ConnectionError as error:
             raise ConfigError(
                 f"Your Elasticsearch endpoint at {settings.elasticsearch} is not available or not responding."
-                "\nPlease make sure your Elasticsearch is running or you have access to elasticsearch and "
+                "\nPlease make sure your Elasticsearch instance is launched and correctly running and you have the necessary access permissions and"
                 "Once you have verified this, restart the Rubrix server."
                 f"\nError detail: [{error}]"
             )
--- a/src/rubrix/server/server.py
+++ b/src/rubrix/server/server.py
@@ -20,10 +20,11 @@ This module configures the global fastapi application
 import os
 from pathlib import Path
 
+import elasticsearch
 from brotli_asgi import BrotliMiddleware
 from fastapi import FastAPI
 from fastapi.middleware.cors import CORSMiddleware
-from pydantic import ValidationError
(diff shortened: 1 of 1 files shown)


src/rubrix/server/server.py
```diff
@@ -86,12 +88,21 @@ def configure_app_statics(app: FastAPI):
...
+            es_wrapper = create_es_wrapper()
+            dataset_records: DatasetRecordsDAO = DatasetRecordsDAO(es_wrapper)
+            datasets: DatasetsDAO = DatasetsDAO.get_instance(
+                es_wrapper, dataset_records
+            )
+            datasets.init()
+            dataset_records.init()
+        except elasticsearch.exceptions.ConnectionError as error:
+            raise ConfigError(
+                f"We checked the elastic search endpoint at {settings.elasticsearch} and is not responding/available..."
```
dvsrepo:
```suggestion
                f"Your Elasticsearch endpoint at {settings.elasticsearch} is not available or not responding."
```

src/rubrix/server/server.py
```diff
@@ -86,12 +88,21 @@ def configure_app_statics(app: FastAPI):
...
+                "\nPlease make sure your Elasticsearch is running or you have access to elasticsearch and "
```
dvsrepo:
```suggestion
                "\nPlease make sure your Elasticsearch instance is launched and correctly running and you have the necessary access permissions and"
```

src/rubrix/server/server.py
```diff
@@ -86,12 +88,21 @@ def configure_app_statics(app: FastAPI):
...
+                "restart rubrix."
```
dvsrepo:
```suggestion
                "Once you have verified this, restart the Rubrix server."
```

dvsrepo (changes requested):

dvsrepo (approved):
💯

Issue #924, opened 2022-01-11
[Validation errors] Improve error information when logging data


frascuchon:
Some info like which record cause the error could help to resolve it.

dvsrepo:
As a reminder, we had this related discussion in the early days.

https://github.com/recognai/rubrix/issues/191

frascuchon:
For now, a error raise during a data logging returns (from API server) a response like:

```json
{
  "detail": [
    {
      "loc": [
        "body",
        "records",
        0,
        "annotation",
        "labels",
        0,
        "score"
      ],
      "msg": "ensure this value is less than or equal to 1.0",
      "type": "value_error.number.not_le",
      "ctx": {
        "limit_value": 1
      }
    },
    {
      "loc": [
        "body",
        "records",
        1,
        "prediction",
        "labels",
        0,
        "score"
      ],
      "msg": "ensure this value is less than or equal to 1.0",
      "type": "value_error.number.not_le",
      "ctx": {
        "limit_value": 1
      }
    }
  ]
}
```

This response is generated by the default fastapi error handler is shared by ALL pydantic validation errors in app.

As you see, some info about which record is failing is included here (at least the list index). But logging with data chunk this won't be enough.

1. We could customize the response but it could be dangerous since this change affect to ALL pydantic validations (over list, dict, subclasses or primitive data)

2. We could keep this as is and try to make the client error more human readable. I mean, parse these responses in client, and contextualize the error parsing with client action that triggered the api call. With this approach, we could link list index with corresponding client record.

3. Another solution to improve the error messages from client could be replicate the record validation.

Open discussion

dcfidalgo:
I think I would vote for option 2, if at any time there are more clients than the python one, then maybe option 1.
Not a big fan of option 3.

frascuchon:
Option 3 is interesting from an efficiency point of view. Since with this option the requests and checks that are going to fail are downloaded to the server.

Maybe if it would be interesting to have some validations replicated on the client

For example, if we include a limit on the number of metadata fields that a dataset supports, it would be good practice to throw an error (or at least a warning) if records are being created on the client that exceed that limit.

dcfidalgo:
[...]

PR #1031, opened 2022-01-25
refactor(#924): normalize API error responses

frascuchon:
This PR takes cares about api error responses. Normalizes the responses following the structure:
```json
{
   "detail": { 
       "code": "rubrix.api.error::SomeErrorType",
       "params": {
           "msg": "Some error description"
       }
   } 
}
```

The `code` attribute is always a string value, the `params` section could change depending on the error type (the code), giving more contextual details about error.

Only Generic error (http status 5xx) won't include detailed info since could be compromise the infrastructure security


dcfidalgo (approved):
Nice!! :100:

PR #1032, opened 2022-01-25
Fix(#1027): Improve client models by reordering fields + forbidding extra args

### LINKED ISSUE #1027: [Records] TextClassificationRecord does not raise error passing
This code:
```python
import rubrix as rb

rb.TextClassificationRecord("this is a multi label record", multilabel=True)
```
should raise some `KeyError` validation.

This should be a common behaviour for all client record instances

dcfidalgo:
Closes #1027 

With this PR the client models raise an error when extra arguments are provided. The error message is not the clearest though, but i did not find a way to modify this. @frascuchon Do you know how to customize validation error messages in pydantic?

It also fixes the order of the fields in the models. This is important when converting the records into a DataFrame/Dataset, so that when displaying a DataFrame the order of the columns is more or less meaningful.

### DIFF
--- a/src/rubrix/client/models.py
+++ b/src/rubrix/client/models.py
@@ -74,6 +74,9 @@ class _RootValidators(BaseModel):
 
         return values
 
+    class Config:
+        extra = "forbid"
+
 
 class BulkResponse(BaseModel):
     """Summary response when logging records to the Rubrix server.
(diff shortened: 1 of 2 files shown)


frascuchon (approved):
You should raise your own exceptions. See [here](https://pydantic-docs.helpmanual.io/usage/models/#custom-errors)

I think we should make some effort to parse errors from pydantic validation, and make them human readable

PR #1034, opened 2022-01-25
refactor(#924): better client error management

frascuchon:
This PR changes how client handles incoming API errors.

The idea is to contextualize incoming errors by defining a client error hierarchy, including an operation context giving more details about inputs that raised the error.

Not totally cover solution discussed in #924 but it will help to understand the api error reason. Even more in combination with changes in #1031


dcfidalgo (approved):
:100:

PR #1059, opened 2022-01-28
fix(cleanlab): set cleanlab n_jobs=1 as default

frascuchon:
This  PR disable default parallel processing for `cleanlab.pruning.get_noise_indices` that could bring problems or make systems inestables.

Most of the times github actions are failing because of this.

See https://github.com/recognai/rubrix/runs/4982205560?check_suite_focus=true as example

Include also commit 6f62dc5

### DIFF
--- a/src/rubrix/labeling/text_classification/label_errors.py
+++ b/src/rubrix/labeling/text_classification/label_errors.py
@@ -41,7 +41,7 @@ def find_label_errors(
     records: List[TextClassificationRecord],
     sort_by: Union[str, SortBy] = "likelihood",
     metadata_key: str = "label_error_candidate",
-    n_jobs: int = 1,
+    n_jobs: Optional[int] = 1,
     **kwargs,
 ) -> List[TextClassificationRecord]:
     """Finds potential annotation/label errors in your records using [cleanlab](https://github.com/cleanlab/cleanlab).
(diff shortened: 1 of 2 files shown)


src/rubrix/labeling/text_classification/label_errors.py
```diff
@@ -41,6 +41,7 @@ def find_label_errors(
     records: List[TextClassificationRecord],
     sort_by: Union[str, SortBy] = "likelihood",
     metadata_key: str = "label_error_candidate",
+    n_jobs: int = 1,
```
dcfidalgo:
```suggestion
    n_jobs: Optional[int] = 1,
```

src/rubrix/labeling/text_classification/label_errors.py
```diff
@@ -55,6 +56,8 @@ def find_label_errors(
             - "prediction": sort the returned records by the probability of the prediction (highest probability first)
             - "none": do not sort the returned records
         metadata_key: The key added to the record's metadata that holds the order, if ``sort_by`` is not "none".
+        n_jobs : Number of processing threads used by multiprocessing. Default 1, what it's mean no parallel processing
+            is applied. Set to None to use the number of processing threads on your CPU
```
dcfidalgo:
```suggestion
        n_jobs : Number of processing threads used by multiprocessing. If None, uses the number of threads
            on your CPU. Defaults to 1, which removes parallel processing. 
```

dcfidalgo (approved):
:100: Nice catch! Just a few inline suggestions

Issue #945, opened 2022-01-13
[Search] Improve and normalizes the search data model


frascuchon:
Things to keep in mind:

- Normalize text inputs fields: `text`, `inputs`, `words` must be normalized and use a common pattern for all tasks
- Several es analyzers for text fields: `standard` and `whitespace`(?) for fine tuning searches. Default as `standard`
- What about text fields in metadata ? For now, only terms queries are supported. It's mean that metadata fields with large content are not enabled to be queries as full text search.
- Created indices should contain mapping info only for its fields. A text classification index should not include mapping info for tokens or text predicted (text2text).
- Review filter fields and align with UI names (if any)
- What about nested fields? like token or metrics info for token classification, or label and its score for text classification. As default, query string dsl does not support nested queries, but it could be nice include some minimal support for that kind of queries.

@dvsrepo @dcfidalgo Anything to include here?

## Tasks

To achieve to do the work, we need tackle following tasks (that will be created as separated issues and linked here)

1. [Datasets] Avoid using global template for all indices
2. [Datasets] Dataset migration mechanisms for each release
3. [Datasets] New es document model per task with backward compatibility fields
4. [Datasets] Apply migration to new es doc model 
5. [Datasets] Build searches and aggregations using new doc model

dcfidalgo:
Looks complete to me, just to elaborate on a few things:

>     * Normalize text inputs fields: `text`, `inputs`, `words` must be normalized and use a common pattern for all tasks

The idea here is, that `TextClassification` changes from `inputs` to `text`, right? In the end, did you find a solution for having `text.context` and `text.exact`, for example, the latter being the field analyzed with the whitespace analyzer?
And `words` includes stopwords and is for the keyword list, right?

>     * Several es analyzers for text fields: `standard` and `whitespace`(?) for fine tuning searches. Default as `standard`
 
Yes, I think whitespace is good for now.

>     * What about text fields in metadata ? For now, only terms queries are supported. It's mean that metadata fields with large content are not enabled to be queries as full text search.

For me it sounds reasonable not to support full text search in metadata for now

>     * Created indices should contain mapping info only for its fields. A text classification index should not include mapping info for tokens or text predicted (text2text).
> 
>     * Review filter fields and align with UI names (if any)
> 
>     * What about nested fields? like token or metrics info for token classification, or label and its score for text classification. As default, query string dsl does not support nested queries, but it could be nice include some minimal support for that kind of queries.

[...]

PR #1075, opened 2022-02-01
refactor(#945): normalizes input and output record fields

frascuchon:
This PR normalized incoming and outgoing record level fields. 

- Some returned records from API has no sense (like `annotated_as`, or `predicted_by`) since they are computed from other record fields.
- `raw_text` field for token classification becomes to `text` field, but keeping backward compatibility with old clients during some versions
- UI model now uses this new `text` field for token classification.

See #945


frontend/models/TextClassification.js
```diff
@@ -21,23 +21,24 @@ import { BaseRecord, BaseSearchQuery, BaseSearchResults } from "./Common";
...
-    predicted_as,
-    ...superData
-  }) {
+  constructor({ inputs, explanation, multi_label, ...superData }) {
     super(superData);
     this.inputs = inputs;
     this.explanation = explanation;
     this.multi_label = multi_label;
-    this.predicted_as = predicted_as;
+    // this.predicted_as = predicted_as;
```
frascuchon, line 29:
```suggestion
```

dcfidalgo (approved):
Nice! :100:

PR #1091, opened 2022-02-03
refactor(#1089): using opensearch client as default

### LINKED ISSUE #1089: The client noticed that the server is not a supported distribution of Elasticsearch
Hi All, Was trying to install today on a Macbook (Intel) and hitting this error
Version: 0.9.0

```
raise UnsupportedProductError(message)
elasticsearch.exceptions.UnsupportedProductError: The client noticed that the server is not a supported distribution of Elasticsearch
```

Here are the details of the setup process
```
# Create a directory for notebooks and another to download data 
mkdir -p rubrix/notebooks rubrix/data 

# CD into the directory
cd rubrix

# Create and activate a conda environment
conda create -n rubrix python=3.8 -y
conda activate rubrix

# Pip install the necessary libraries
pip install -U jupyterlab pandas matplotlib tqdm ipywidgets 
pip install -U "rubrix[server]" snorkel datasets

# Run the ES Docker container
docker run -d \
 --name elasticsearch-for-rubrix \
 -p 9200:9200 -p 9300:9300 \
 -e "ES_JAVA_OPTS=-Xms512m -Xmx512m" \
 -e "discovery.type=single-node" \
 docker.elastic.co/elasticsearch/elasticsearch-oss:7.10.2

# Run the Rubrix server
python -m rubrix
```

Please let me know if there's any other information needed from me to troubleshoot this issue. Appreciate the help!

frascuchon:
This PR replaces the oficial elasticsearch client  by the opensearch-py client. That is full compatible  and avoid error referred in #1089

Also refactor the github actions file for:

-  Check different supported elasticsearch versions.
- Separate test and build jobs

Closes #1089

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -47,7 +47,7 @@ server =
     # Basic dependencies
     fastapi ~= 0.63.0
     uvicorn[standard] ~= 0.13.4
-    elasticsearch >= 7.4.0,<8.0.0
+    opensearch-py
     smart-open
     brotli-asgi ~= 1.1.0
     Deprecated ~= 1.2.0
--- a/setup.cfg
+++ b/setup.cfg
@@ -47,7 +47,7 @@ server =
     # Basic dependencies
     fastapi ~= 0.63.0
     uvicorn[standard] ~= 0.13.4
-    opensearch-py
+    opensearch-py ~= 1.0.0
     smart-open
     brotli-asgi ~= 1.1.0
     Deprecated ~= 1.2.0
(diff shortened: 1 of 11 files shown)


setup.cfg
```diff
@@ -47,7 +47,7 @@ server =
     # Basic dependencies
     fastapi ~= 0.63.0
     uvicorn[standard] ~= 0.13.4
-    elasticsearch >= 7.4.0,<8.0.0
+    opensearch-py
```
dcfidalgo: are all versions compatible, should we put at least an upper limit?

dcfidalgo (approved):
:100: Just one inline comment

setup.cfg
frascuchon: Just one version of opensearch client is published, but we can set a version number
```suggestion
    opensearch-py ~= 1.0.0
```

PR #1109, opened 2022-02-07
feat(#422): introducing the rb.Dataset* classes

### LINKED ISSUE #422: [datasets] Load Rubrix dataset as HF dataset
We need to add an integration with Hugging Face Datasets library. This will make it really easy to (1) train transformers, and (2) upload dataset to the Hub, avoiding the boilerplate and potential errors. Plus it comes with really interesting features for bath processing, etc.

I see several options:

1. Add a `dataset` return type for `rb.load()`

2. Define a method `to_hf_dataset()` . Not sure what will be the input though? Maybe the dataset name, but then we might want to add a query, etc. so maybe records but it's kind of inneficient and ugly to do rb.load and then to_hf_dataset.

Any other options?

Internally this will:

1. Use Features (ClassName, Value, Sequence, etc.) to define the schema of the dataset
2. Create the Dataset.from_ 

I think we can easily support all tasks. See a dirty example here:

https://rubrix.readthedocs.io/en/master/tutorials/08-error_analysis_using_loss.html#5.-Sharing-the-dataset-in-the-Hugging-Face-Hub

I would include this in the next release

dcfidalgo:
Closes #422 

- [x] text2text
- [x] integrate with the rest of the library
- [x] provide futurewarning about the default behavior of `rb.load`
- [x] revise/complete tests
- [x] documentation

Optionally in this PR or in followup PRs:
- [ ] allow for colum2attr mappings in the `Dataset.from_*` methods (goes to a follow-up PR)


src/rubrix/client/dataset.py
```diff
@@ -0,0 +1,350 @@
...
+        ... rb.TextClassificationRecord(inputs={"text": "example"})
+        >>> dataset[0] = rb.TextClassificationRecord(inputs="replaced example")
+
+        Looping over the dataset:
+        >>> assert len(dataset) == 2
+        >>> for record in dataset:
+        ...     print(record)
+    """
+
+    def __init__(self, records: Optional[List[Record]] = None):
```
frascuchon: Maybe we can define a generic hierarchical dataset class and let pydantic apply data type checking for us:

````python
# This code works
from typing import Generic, Iterable, TypeVar

import pytest
from pydantic import ValidationError
from pydantic.generics import GenericModel

R = TypeVar("R")


class Dataset(GenericModel, Generic[R]):
    "Generic dataset"
    records: Iterable[R]


class IntegersDataset(Dataset[int]):
    """Dataset of integers"""
    pass

with pytest.raises(ValidationError):
    # pydantic will verify data type check for you
    IntegersDataset(data=["la", "lo", "les"])
````

File: src/rubrix/client/dataset.py [hunk 2]
```diff
@@ -0,0 +1,350 @@
...
+
+        if self._record_type is TextClassificationRecord:
+            return self._textclassification_to_datasets()
+        if self._record_type is TokenClassificationRecord:
+            return self._tokenclassification_to_datasets()
+        if self._record_type is Text2TextRecord:
+            return self._text2text_to_datasets()
+
+    @classmethod
+    def from_datasets(
```
frascuchon: With hiierarchical Dataset class every implementation will keep details of how to transform to/from hf datasets, instead of keep all together in the same class

src/rubrix/client/dataset.py
dcfidalgo: Yes, I thought about having an (abstract or generic) parent class, but I could not really figure out how to make it work:
1. I don't want the user to import different "types" of `Dataset`s when creating a new one, we already have the different record types for that. 
```python
# for example always
Dataset.from_datasets(...)
# instead of
TextClassificationDataset.from_datasets(...)
TokenClassificationDataset.from_datasets(...)
```
2. I would like to enable a similar "list appending" flow that we are using a lot right now in the tutorial:
```python
empty_dataset = Dataset()
for i in dataframe.itertuples():
    record = rb.TextClassificationRecord(...)
    empty_dataset.append(record)
```
Maybe a solution could be to require a `task` argument, when you want to create an empty Dataset `empty_ds = Dataset(task="TextClassification")`

Regarding pydantic, i like the idea of delegating the type checks!! Do you know if it is possible to have private attributes type checked? I don't want to expose the `_records` attribute openly, to prevent something like `Dataset.records.append(whateverthef***iwant)`.

src/rubrix/client/dataset.py
dcfidalgo: See my comment above, regarding having only one class as the user interface.

src/rubrix/client/dataset.py
dcfidalgo: > I don't want the user to import different "types" of Datasets when creating a new one, we already have the different record types for that.

I changed my mind, the implementation will be much cleaner with different Dataset types, maybe we could use `Dataset.from_datasets` as factory method, but in principle the user should use the different Dataset types directly.

frascuchon: Indeed, for private properties, the use of pydantic does not have much impact. You could use a generic class that uses some pydantic parsers for type validation. As shown in the code snippet

````python
class A(Generic[R]):
    def __init__(self, data: List[R]):
        self.__item_type__ = get_args(self.__class__.__orig_bases__[0])[0]
        self.__records__ = []
        for d in data:
            self.append_record(d)

    def append_record(self, data: R):
        self.__records__.append(parse_obj_as(self.__item_type__, data))

    @property
    def records(self) -> Iterable[R]:
        return tuple(self.__records__)

class B(A[int]):
    pass


r = B(data=[1, 23])
r.append_record("lolo") # This will raise a ValidationError

````

Comment (frascuchon) [reply src/rubrix/client/dataset.py]: Anyway, do we need a `add_record` or something like that for our `Dataset` regarding the main PR purpose?

**Comment by dcfidalgo** [comment]
Ok, this is the thing:
- There are Dataset classes for each task: `rb.DatasetForTextClassification`, `rb.DatasetForTokenClassification`, `rb.DatasetForText2Text` . Separate classes make the implementation much cleaner and more sustainable, especially when we want to extend those classes. The naming is motivated by not interfering with the record classes when using the tab completion in the notebook, that is i did not choose `TextClassificationDataset` in order to avoid always having to choose between `*Record` and `*Dataset`.
- `rb.load(as_pandas=False)` returns one of these Datasets, and issues a one-time warning, that `as_pandas` is deprecated:
```
The argument 'as_pandas' in `rb.load` will be deprecated in the future, and we will always return a `Dataset`.
To emulate the future behavior set `as_pandas=False`. To get a pandas DataFrame, call `Dataset.to_pandas()`
``` 
- there are two shortcut methods: `rb.read_pandas` and `rb.read_datasets`, that take a DataFrame/datasets.Dataset and a `task` argument. The idea is that the user should primarily use these to methods instead of `Dataset*.from_pandas` and `Dataset*.from_datasets`
- I invested a lot of time in the references, see [HERE](https://rubrix.readthedocs.io/en/feat-add_dataset_class/reference/python/python_client.html#module-rubrix.client.datasets)

so now you can do:
```python
import rubrix as rb
dataset_rb = rb.load("my_dataset")
dataset_rb.to_datasets().push_to_hub("rubrix_dataset", ...)

dataset_rb = rb.read_datasets(load_dataset("rubrix_dataset"), task="TextClassification")
```

src/rubrix/client/datasets.py
```diff
@@ -0,0 +1,780 @@
...
+
+    def _validate_record_type(self):
+        """Validates the record type.
+
+        Raises:
+            WrongRecordTypeError: When the record type in the provided
+                list does not correspond to the dataset type.
+        """
+        record_types = {type(rec): None for rec in self._records}
+        if len(record_types) > 1:
```
frascuchon, line 66: Just a suggestion but, for large records list, we could raise the error once we find an unexpected record instead of traverse the whole list

src/rubrix/client/datasets.py
```diff
@@ -0,0 +1,780 @@
...
+        records: A list of Rubrix records.
+
+    Raises:
+        WrongRecordTypeError: When the record type in the provided
+            list does not correspond to the dataset type.
+    """
+
+    _RECORD_TYPE = None
+
+    def __init__(self, records: Optional[List[Record]] = None):
```
frascuchon, line 48: As a future improvement, we should accept a lazy list of records

src/rubrix/client/datasets.py
```diff
@@ -0,0 +1,780 @@
...
+            )
+        self._records[key] = value
+
+    def __delitem__(self, key):
+        del self._records[key]
+
+    def __len__(self) -> int:
+        return len(self._records)
+
+    def append(self, record: Record):
```
frascuchon: Not sure if we really need this method instead of create a dataset from a list of records. What use case are you thinking about for `ds.append` instead of `records.append` + `Dataset(records)` ?

frascuchon (commented):
In general really nice!!! I've just left some comments

What I'm missing is the [tutorial](https://rubrix.readthedocs.io/en/stable/tutorials/08-error_analysis_using_loss.html#5.-Sharing-the-dataset-in-the-Hugging-Face-Hub) update using this new feature.

Could you tackle this in this PR ?

src/rubrix/client/datasets.py
dcfidalgo: Good point! I already have a list of improvements in mind, that I will put in an issue. This PR is already too big and took too long, so I prefer to get this merged and improve the stuff in follow-up PRs.

src/rubrix/client/datasets.py
dcfidalgo: This is one of the points in the list ...

src/rubrix/client/datasets.py
dcfidalgo: Yeah, not sure if it will be very useful. Right now a lot of times we create an empty list and populate it in a for loop, so I thought maybe one could do the same with a dataset:
```python
dataset = rb.DatasetForText2Text()
for i in some_iterable:
    dataset.append(rb.Text2TextRecord(...))
```
Anyways, it comes more or less for free, so why not put it there ...

frascuchon: Regarding the back compatibility, is easier to include a new `append` method that remove it on future versions

src/rubrix/client/datasets.py
frascuchon: Ok. From my part the only required thing would be to include changes in the tutorial

dcfidalgo:
> What I'm missing is the [tutorial](https://rubrix.readthedocs.io/en/stable/tutorials/08-error_analysis_using_loss.html#5.-Sharing-the-dataset-in-the-Hugging-Face-Hub) update using this new feature.

Yes, I looked at the tutorial and tried to change it. But in the end, I thought that it does not really showcase the new features, since the uploaded dataset in this tutorial is not really a "Rubrix dataset", but a custom one that is ready for training. Plus the tutorial is really hard to "remake" (computing the loss takes hours, correcting labels manually, replace(?) uploaded dataset). I think once we have the `prepare_for_training` feature, it offers much more potential.

Instead, I wrote a short guide that I think showcases much better the new features, like explicitly sharing Rubrix datasets via the HF Hub, [see here](https://rubrix.readthedocs.io/en/feat-add_dataset_class/guides/datasets_in_the_client.html).
@frascuchon @dvsrepo Let me know what you think.

docs/guides/datasets_in_the_client.ipynb
```diff
@@ -0,0 +1,180 @@
...
+   "cell_type": "markdown",
+   "id": "d09c1533-97fd-43f7-83ea-f3a56edd1d5e",
+   "metadata": {},
+   "source": [
+    "# Datasets in the client\n",
+    "\n",
+    "This short guide showcases some features of the `Dataset` classes in the Rubrix client.\n",
+    "The Dataset classes are basically lightweight containers for Rubrix records, that facilitate their export and import from other formats, such as a `pandas.DataFrame` or a `datasets.Dataset`.\n",
+    "\n",
+    "For each record type their is a corresponding Dataset class called `DatasetFor<RecordType>`.\n",
```
dvsrepo:
```suggestion
    "For each record type there's a corresponding Dataset class called `DatasetFor<RecordType>`.\n",
```

docs/guides/datasets_in_the_client.ipynb
```diff
@@ -0,0 +1,180 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "d09c1533-97fd-43f7-83ea-f3a56edd1d5e",
+   "metadata": {},
+   "source": [
+    "# Datasets in the client\n",
```
dvsrepo:
```suggestion
    "# Datasets\n",
```

docs/guides/datasets_in_the_client.ipynb
```diff
@@ -0,0 +1,180 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "id": "ae9d4c9e-24a1-4a59-8a17-3b6ac1a39c88",
+   "metadata": {},
+   "source": [
+    "## Working with a Dataset\n",
+    "\n",
+    "Under the hood the Dataset classes store the records in a simple Python list.\n",
+    "Therefore, working with a Dataset class is not very different than working with a simple list of records:"
```
dvsrepo:
```suggestion
    "Therefore, working with a Dataset class is not very different to working with a simple list of records:"
```

docs/guides/datasets_in_the_client.ipynb
```diff
@@ -0,0 +1,180 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "d09c1533-97fd-43f7-83ea-f3a56edd1d5e",
+   "metadata": {},
+   "source": [
+    "# Datasets in the client\n",
+    "\n",
+    "This short guide showcases some features of the `Dataset` classes in the Rubrix client.\n",
```
dvsrepo:
```suggestion
    "This guide showcases some features of the `Dataset` classes in the Rubrix client.\n",
```

docs/guides/datasets_in_the_client.ipynb
```diff
@@ -0,0 +1,180 @@
...
+    "The Dataset classes are basically lightweight containers for Rubrix records, that facilitate their export and import from other formats, such as a `pandas.DataFrame` or a `datasets.Dataset`.\n",
```
dvsrepo:
```suggestion
    "The Dataset classes are lightweight containers for Rubrix records. These classes facilitate importing from and exporting to different formats (e.g., `pandas.DataFrame`, `datasets.Dataset`) as well as sharing and versioning Rubrix datasets using the Hugging Face Hub.\n",
```

dvsrepo (changes requested):

src/rubrix/client/datasets.py
dcfidalgo: I think in the tutorials we should use the workflow described above, especially if we want to have a peek at the created dataset after the loop. It's slightly more "compact" than its alternative:
```python
dataset = rb.DatasetForText2Text()
for i in some_iterable:
    dataset.append(rb.Text2TextRecord(...))
dataset.to_pandas()

# vs

records = []
for i in some_iterable:
    dataset.append(rb.Text2TextRecord(...))
dataset = rb.DatasetForText2Text(records)
dataset.to_pandas()
```

I vote for keeping the method, but if it's a show stopper for you, I'm totally fine removing it as well.

dvsrepo:
> > What I'm missing is the [tutorial](https://rubrix.readthedocs.io/en/stable/tutorials/08-error_analysis_using_loss.html#5.-Sharing-the-dataset-in-the-Hugging-Face-Hub) update using this new feature.
> 
> Yes, I looked at the tutorial and tried to change it. But in the end, I thought that it does not really showcase the new features, since the uploaded dataset in this tutorial is not really a "Rubrix dataset", but a custom one that is ready for training. Plus the tutorial is really hard to "remake" (computing the loss takes hours, correcting labels manually, replace(?) uploaded dataset). I think once we have the `prepare_for_training` feature, it offers much more potential.
> 
> Instead, I wrote a short guide that I think showcases much better the new features, like explicitly sharing Rubrix datasets via the HF Hub, [see here](https://rubrix.readthedocs.io/en/feat-add_dataset_class/guides/datasets_in_the_client.html). @frascuchon @dvsrepo Let me know what you think.

@dcfidalgo the guide is a good addition, well done.

As for the tutorial, I'm a bit lost about which one are you referring to? Is it the loss one? If it's the fine tuning Hugging Face which was the one proposed for incrementally adding the features, let's discuss some specific changes to benefit from this feature (besides the prepare_for_training). In reality this current feature it's useful in most of the tutorials for sharing the dataset with predictions of the model so users don't have to do that step for running the tutorial. As we have one more week, let's plan this and define where to introduce this, also as a way of testing the feature.

src/rubrix/client/datasets.py
frascuchon:
````python
# vs
dataset = rb.DatasetForText2Text([
   rb.Text2TextRecord(...) 
   for r in some_iterable
])
dataset.to_pandas()
````

For me does introduce nothing, so I prefer remove and keep only needed functionality. 

Once we publish this feature with the `append` method, it will be more difficult to remove (if we decide to remove in a future).

The new Dataset class is totally operable without this method. Unless we are missing some specific use case

dcfidalgo: Yeah, the examples above are a bit simplified and there is not really the need for a for loop. Here is an [example](https://rubrix.readthedocs.io/en/stable/tutorials/02-spacy.html#Using-a-Transformer-based-pipeline) where I'm not sure if you can cramp this into a list comprehension (leaving readability aside).

Anyways, i will remove it for now, but this needs a few more changes:
- [x] update guide / docstrings
- [x] ~require a list of `records` in the `Dataset*` init~

dcfidalgo:
> As for the tutorial, I'm a bit lost about which one are you referring to? Is it the loss one? If it's the fine tuning Hugging Face which was the one proposed for incrementally adding the features, let's discuss some specific changes to benefit from this feature (besides the prepare_for_training).

Yes, it's the loss tutorial, i also had a look at the first tutorial (fine tuning HF), but was not sure where to add the new dataset feature (upload to HF Hub before or after annotation). 

>  In reality this current feature it's useful in most of the tutorials for sharing the dataset with predictions of the model so users don't have to do that step for running the tutorial. As we have one more week, let's plan this and define where to introduce this, also as a way of testing the feature.

Ok, then i will go ahead and upload the dataset more or less in the beginning after we made the first predictions. If it's ok with you, i will open another PR for this, this PR is getting really big ...

dvsrepo:
> 

Absolutely. Let's organize this before to see how is best to update the HF and then create another issue

src/rubrix/client/datasets.py
dcfidalgo: Regarding the last checkbox: We have to allow empty datasets, since `rb.load` can return empty ones. But it's a bit weird that you can create an empty `rb.DatasetFor*()`, but you cannot populate it with records.

dcfidalgo:
Ok, improved a bit further the tests and removed the `append` method. From my side, this is good to go @frascuchon

frascuchon (approved):

PR #1111, opened 2022-02-08
feat(#1061): unify records results title

### LINKED ISSUE #1061: [Unify records results ] Style
- [x] Unify style of :  records (xxx) : Use as ref. style of Records in Rules mode
- [x] Unify style of :  records (xxx) : Show it on top of records list in Explore and Annotation Mode
- [x] Explore and Annotation mode : Align records on left size (under checkbox)

See also #970

leiyre:
This PR includes show records results title on top of records list in Explore and Annotation Mode and unify styles

Closes #1061


frascuchon (changes requested):
In "Define RUle" mode, just filtering with query text will show the same message that shows when using filters, that could be a bit confusing. @Amelie-V  @leiyre  ? 

Screenshot 2022-02-08 at 13 33 45

dcfidalgo:
To me, this message should be more explicit and visually more pronounced, saying something like:
> Filters are not part of the Rule, but are applied to the following record list.

And maybe only show this when filters are set.

frascuchon:
The "Records with..." message will be applied only for "Define Rules" mode. The "Annotate" and "Explore" modes will keep intact.

frascuchon (changes requested):
The new file should be called `ResultsTitle` instead of `resultsTitle`.

And a basic test for new component should be included...

frascuchon (approved):

PR #1112, opened 2022-02-08
fix(#982): extends search area

### LINKED ISSUE #982: [Rules] Show filters in "Define Rules" mode

leiyre:
fix #982

This PR Extends search area. Search area + filters are adjusted to the "record column width"

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -135,7 +135,6 @@ jobs:
         with:
           images: recognai/rubrix
       - name: Login to DockerHub
-        if: github.event_name != 'pull_request'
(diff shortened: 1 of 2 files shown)


frascuchon (changes requested):
Cannot see filters properly 


Screenshot 2022-02-08 at 13 32 00

leiyre:
oops right, that's why it was limited to 450px. @Amelie-V how can we deal with this? maybe open the filters to the left?

Amelie-V:
@leiyre I would recommend extending the current search box 200 px more. In that way Sort modal should fit on the right side.

frascuchon (approved):

PR #1115, opened 2022-02-08
refactor(#945): add current search aggregations as metrics

frascuchon:
This PR define missing current aggregation in search as dataset metrics. For example we can find `predicted_as` or `annotated_by` as a metric for token classification datasets.

This is a part of work for better search model interaction


dcfidalgo (approved):
:100:

PR #1116, opened 2022-02-09
fix(#831): Remove sort field when only one is applied

### LINKED ISSUE #831: [Sort] List of records not refreshed when removing sort field
To reproduce:
1. Sort records by a field in the Sort "filter".
2. Remove the filter by pressing the "x".
3. You have to press the "Sort" button again to remove the sorting and update the view. For the other filters, pressing the "x" automatically refreshes the view.
bug_remove_sort
)

leiyre:
This PR allows to remove sort field from cross button when only one is applied

Closes #831

### DIFF
--- a/frontend/components/commons/header/filters/SortList.vue
+++ b/frontend/components/commons/header/filters/SortList.vue
@@ -78,7 +78,11 @@ export default {
       this.numberOfSortFields += 1;
     },
     onRemoveSortField(index) {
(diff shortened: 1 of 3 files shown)


frascuchon (changes requested):
For small PRs like this, you should include basic test specs. Please include a basic test using snapshots.

frascuchon (approved):

PR #1117, opened 2022-02-09
refactor(#945): using new search service

frascuchon:
This PR includes the implementation for `SearchRecordsService` class and changes all task services search method, that will use the new implemented service.

This will help in a future to normalize api task search and simplify development.


dcfidalgo (approved):
:100: nice!! the new search service call (`__search__.search`)  looks much neater!

PR #1125, opened 2022-02-10
Feat/advanced query dsl mode

frascuchon:
This PR includes an advanced query dsl model that uses an [enhanced lucene query dsl sintax ](https://github.com/jurismarches/luqum) to enable queries with nested fields, for example.

This "advanced mode" is enabled by a query body parameter for the sake of the compatibility. This is not enabled for weak labels rules for now, until a feature validation.

You can test this feat from UI app by setting the `advancedQueryDsl` query parameter manually

The advanced query dsl will resolve requirements mentioned in recognai/roadmap#55

### DIFF
--- a/frontend/database/modules/datasets.js
+++ b/frontend/database/modules/datasets.js
@@ -213,6 +213,10 @@ async function _updateViewSettings({ id, data }) {
 }
 
 async function _callSearchApi({ dataset, query, sort, size, from = 0 }) {
+  const { advancedQueryDsl } = $nuxt.$route.query;
+  if (advancedQueryDsl === null || advancedQueryDsl === "true") {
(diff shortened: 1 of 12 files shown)


dcfidalgo (approved):
Nice! Just one question, could you provide an example of how to activate this advanced mode? I understood that it is a switch in the address you put in the browser, right?

frascuchon:
> Nice! Just one question, could you provide an example of how to activate this advanced mode? I understood that it is a switch in the address you put in the browser, right?

Exactly ! You can just type `http://localhost:6900/ws/rubrix/<my-dataset>?advancedQueryDsl&query.... ` to activate it for from UI

PR #1133, opened 2022-02-11
fix: index records when logging

frascuchon:
Wraps properly the index not found exception that cause the logging fails.


dcfidalgo (approved):
works :100:

PR #1147, opened 2022-02-14
fix(#1140): fix/make client models more consistent

### LINKED ISSUE #1140: [Client] Improve TokenClassification and Text2Text models
I want to make two small changes to the `TokenClassificationRecord` and `Text2TextRecord`:
- automatically add the default entity score of 1.0 when not provided in the predictions (TokenClassification)
- allow unlikely but allowed text2text records, that have a mix of tuples (text, score) and only text predictions

dcfidalgo:
Closes #1140 

This PR fixes/improves a bit the `score` handling in the `TokenClassificationRecord` and `Text2TextRecord`. This enables the support of "entity scores" in the `DatasetForTokenClassification`.

Also includes some more testing.

### DIFF
--- a/src/rubrix/client/datasets.py
+++ b/src/rubrix/client/datasets.py
@@ -441,7 +441,7 @@ class DatasetForTokenClassification(DatasetBase):
             ...     "text": ["my example"],
             ...     "tokens": [["my", "example"]],
             ...     "prediction": [
-            ...         [{"label": "LABEL1", "start": 3, "end": 10, "score": 0.5}]
+            ...         [{"label": "LABEL1", "start": 3, "end": 10, "score": 1.0}]
             ...     ]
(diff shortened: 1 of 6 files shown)


src/rubrix/client/datasets.py
```diff
@@ -441,7 +441,7 @@ def from_datasets(
             ...     "text": ["my example"],
             ...     "tokens": [["my", "example"]],
             ...     "prediction": [
-            ...         [{"label": "LABEL1", "start": 3, "end": 10}]
+            ...         [{"label": "LABEL1", "start": 3, "end": 10, "score": 0.5}]
```
```suggestion
            ...         [{"label": "LABEL1", "start": 3, "end": 10, "score": 1.0}]
```

frascuchon (approved):

PR #1155, opened 2022-02-15
docs(#422): update tutorial with new dataset feature

dcfidalgo:
Adds the new `rb.DatasetForTextClassification().to_datasets()` and `rb.read_datasets` feature to our first tutorial.
Now we upload the records (after having made the predictions with the pretrained model), so in principle, the user does not have to run the predictions, but can simply load the records from the Hub and log them to Rubrix.

@dvsrepo @frascuchon


dvsrepo (approved):
👍

PR #1156, opened 2022-02-15
feat(#951): Filter uncovered records

### LINKED ISSUE #951: [Labeling Rules] Filter uncovered records

leiyre:
This PR allows filtering of records by uncovered by rules

Closes #951

### DIFF
--- a/frontend/components/commons/header/filters/FiltersList.vue
+++ b/frontend/components/commons/header/filters/FiltersList.vue
@@ -60,10 +60,11 @@
               @apply="onApply"
             />
             <FilterUncoveredByRules
-              v-else-if="filter.type === 'checkbox' && showUncoveredByRulesFilter"
+              v-else-if="showUncoveredByRulesFilter"
               class="filter"
               :filter="filter"
               :dataset="dataset"
+              @apply="onApply"
             />
           </span>
           <a
@@ -150,13 +151,6 @@ export default {
           placeholder: "Select options",
         },
         {
-          key: "status",
-          name: "Status",
-          type: "select",
-          group: "Status",
-          placeholder: "Select options",
-        },
-        {
           key: "sort",
           name: "Sort",
           type: "sort",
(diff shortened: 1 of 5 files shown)


frontend/components/commons/header/filters/FiltersList.vue
```diff
@@ -60,6 +60,10 @@
               @apply="onApply"
             />
           </span>
+          <FilterUncoveredByRules
```
frascuchon: Show this filter only if "Define Rules" modes is active, and dataset has rules already defined

frontend/components/commons/header/filters/FiltersList.vue
```diff
@@ -59,6 +59,12 @@
               :filter="filter"
               @apply="onApply"
             />
+            <FilterUncoveredByRules
+              v-else-if="filter.type === 'checkbox' && showUncoveredByRulesFilter"
```
frascuchon: Do we need the `filter.type === "checkbox"` condition?

frascuchon (changes requested):
Inline comments + fix lint messages

frontend/components/commons/header/filters/FiltersList.vue
leiyre: @frascuchon but in "Explore" and "Annotation" mode, the filter must be also visible

frascuchon: Yes, what I mean is enabled, sorry. You could see the uncovered records only if you can work with rules, so "Define Rules" mode is enabled

frontend/components/commons/header/filters/FiltersList.vue
leiyre: we can remove it

frascuchon (approved):

frascuchon (changes requested):
fix build

frascuchon (approved):

PR #1159, opened 2022-02-15
perf(settings): deprecate index template creation environment variable

frascuchon:
This PR shows a log warning when environment variable `DISABLE_ES_INDEX_TEMPLATE_CREATION` is enabled, since won't be used anymore.

**NOTE: Since changes included in this PR, tests have become unstable and were failing all the time. So I've included changes in tests to fix problems. The main change is define the mocked client as fixture and inject in all related tests.**


dcfidalgo:
This PR also includes some logging changes, is this wanted?

frascuchon:
> This PR also includes some logging changes, is this wanted?

Yes, we must setup the logging configuration prior to launch a warning, in order to show early those messages with new format. That's the reason of changes.

dcfidalgo (commented):
:+1: I think the `tests/test_something.py` should not be included, right?

frascuchon:
> 👍 I think the `tests/test_something.py` should not be included, right?

Good catch!. Removed out

dcfidalgo (approved):

PR #1162, opened 2022-02-15
refactor(metrics): compute text length as metric

frascuchon:
This PR materializes the input text length as a record metrics.

Then, you will be able to filter by text length and in a future release, sort by input text length.


dcfidalgo (approved):
Nice! :100:

PR #1177, opened 2022-02-17
fix(#1176): when converting to datasets, we convert empty dicts to Nones

### LINKED ISSUE #1176: [Datasets] datasets.Dataset with only empty dicts in one column fail when pushed to the HF Hub
This happens when we convert rubrix records with no metadata to a `datasets.Dataset`. You can create this Dataset without issues, but the push to the Hub fails.

dcfidalgo:
Closes #1176

### DIFF
--- a/src/rubrix/client/datasets.py
+++ b/src/rubrix/client/datasets.py
@@ -339,6 +339,8 @@ class DatasetForTextClassification(DatasetBase):
                 ]
             elif key == "id":
                 ds_dict[key] = [str(rec.id) for rec in self._records]
+            elif key == "metadata":
+                ds_dict[key] = [getattr(rec, key) or None for rec in self._records]
             else:
                 ds_dict[key] = [getattr(rec, key) for rec in self._records]
 
@@ -378,6 +380,8 @@ class DatasetForTextClassification(DatasetBase):
                     if row["explanation"] is not None
                     else None
                 )
+            if row.get("metadata", "no metadata in dict") is None:
+                del row["metadata"]
 
             records.append(TextClassificationRecord(**row))
 
@@ -486,6 +490,8 @@ class DatasetForTokenClassification(DatasetBase):
                 ]
             elif key == "id":
                 ds_dict[key] = [str(rec.id) for rec in self._records]
+            elif key == "metadata":
+                ds_dict[key] = [getattr(rec, key) or None for rec in self._records]
             else:
                 ds_dict[key] = [getattr(rec, key) for rec in self._records]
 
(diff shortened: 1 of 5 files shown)


src/rubrix/client/datasets.py
```diff
@@ -486,6 +490,8 @@ def entities_to_dict(
                 ]
             elif key == "id":
                 ds_dict[key] = [str(rec.id) for rec in self._records]
+            elif key == "metadata":
+                ds_dict[key] = [getattr(rec, key) or None for rec in self._records]
```
frascuchon, line 492: This line will replace the empty `metadata` to a `None` because HF does not support empty dict as value?

In that case, could we just define the `metadata` field as optional? 

What about with other fields like metrics. Should we prevent empty dict in a more general way?

What should be the behaviour with following code?:
```python
dataset = rb.DatasetForText2Text([rb.Text2TextRecord(text="mock", metrics={})])
hf_dataset = dataset.to_datasets()
hf_dataset.push_to_hub("dataset_with_empty_metrics")
```

frascuchon: As an additional comment, we could improve this code by swapping the for loops:

```python
ds_dict = defaultdict(list)
for rec in self._records:
   for key in self._RECORD_TYPE.__fields__:
      ds_dict[key].append(....)
```

src/rubrix/client/datasets.py
```diff
@@ -509,6 +515,8 @@ def entities_to_tuple(entities):
                 row["prediction"] = entities_to_tuple(row["prediction"])
             if row.get("annotation"):
                 row["annotation"] = entities_to_tuple(row["annotation"])
+            if row.get("metadata", "no metadata in dict") is None:
+                del row["metadata"]
```
frascuchon: Could we avoid these lines by defining `metadata` as `Optional` ?

tests/client/test_dataset.py
```diff
@@ -280,6 +286,12 @@ def test_to_from_datasets(self, tokenclassification_records):
         rec = rb.DatasetForTokenClassification.from_datasets(missing_optional_cols)[0]
         assert rec.text == "mock" and rec.tokens == ["mock"]
 
+    def test_datasets_empty_metadata(self):
+        dataset = rb.DatasetForTokenClassification(
+            [rb.TokenClassificationRecord(text="mock", tokens=["mock"])]
+        )
+        assert dataset.to_datasets()["metadata"] == [None]
+
```
frascuchon, line 326: I'm missing tests publishing datasets to huggingface. 

We could create private datasets, or test spaces for recognai account and publish dataset there. Not sure if we can overwrite already published.

src/rubrix/client/datasets.py
dcfidalgo: Done

tests/client/test_dataset.py
dcfidalgo: Done

frascuchon (approved):

PR #1184, opened 2022-02-18
HF datasets: spaCy, cleanlab, weak supervision

leireropl:
Uploading HF datasets for the following tutorials:

-SpaCy,
-Cleanlab,
-Weak supervision

Tutorials will be modified as well with the corresponding instructions

### DIFF
--- a/docs/tutorials/01-labeling-finetuning.ipynb
+++ b/docs/tutorials/01-labeling-finetuning.ipynb
@@ -286,7 +286,7 @@
     "dataset_rb = rb.DatasetForTextClassification(records)\n",
     "dataset_ds = dataset_rb.to_datasets()\n",
     "\n",
-    "dataset_ds.push_to_hub(\"Rubrix/sentiment-banking\")"
+    "dataset_ds.push_to_hub(\"rubrix/sentiment-banking\")"
    ]
   },
   {
--- a/docs/tutorials/01-labeling-finetuning.ipynb
+++ b/docs/tutorials/01-labeling-finetuning.ipynb
@@ -304,7 +304,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "dataset_ds = load_dataset(\"Rubrix/sentiment-banking\", split=\"train\")\n",
+    "dataset_ds = load_dataset(\"rubrix/sentiment-banking\", split=\"train\")\n",
     "dataset_rb = rb.read_datasets(dataset_ds, task=\"TextClassification\")"
    ]
   },
--- a/docs/tutorials/01-labeling-finetuning.ipynb
+++ b/docs/tutorials/01-labeling-finetuning.ipynb
@@ -294,7 +294,7 @@
    "id": "fecf89be-e58a-4e25-bb33-c346764d0c16",
    "metadata": {},
    "source": [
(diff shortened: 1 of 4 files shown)


leireropl:
I am going to do one more commit with new changes

leireropl:
Regarding the spaCy tutorial, I was wondering whether the section I placed the HF information is the most suitable. At first, I thought that the end of the tutorial would be the best place, but it doesn't make sense (people will have uploaded everything to Rubrix by then). 

Some clarifications have been made for this part, but if you think I should change it, just say it ;)

dvsrepo (changes requested):
Hi @leireropl , it looks better but I have some suggestions. The main idea is that users can skip part of the tutorial for building the dataset (which is time consuming) so we should be very clear: if you want to skip steps X and Y, you can use this dataset from the Hub.... and go directly to step W. If it's necessary we can add numbers to the sections.

- Showing the info before the user downloads the raw dataset, uses the pipeline and logs into Rubrix. The idea is to be very explicit: INFO( If you want to skip the next N steps, you can do the following:) and then show a cell with all the necessary for reading the Rubrix dataset from the hub and logging it into Rubrix.
- The Bonus part I would call it Appendix and I would explain how to upload the dataset more generally (without the rubrix/{dataset_name}. Here my recommendation is to use rb.load() instead of the records because we don't know the order in which the user is running the notebook so `records` my not be available.
- For the spaCy tutorial I would upload the imdb one for the hub and make sure the Gutenberg one includes the predictions of both model (trf and the other one).

leireropl:
Hi,

I made some changes to the files. I think I got what you meant (more or less), let me know if this is OK.

docs/tutorials/01-labeling-finetuning.ipynb
```diff
@@ -286,15 +286,15 @@
     "dataset_rb = rb.DatasetForTextClassification(records)\n",
     "dataset_ds = dataset_rb.to_datasets()\n",
     "\n",
-    "dataset_ds.push_to_hub(\"Recognai/sentiment-banking\")"
+    "dataset_ds.push_to_hub(\"Rubrix/sentiment-banking\")"
```
dcfidalgo:
```suggestion
    "dataset_ds.push_to_hub(\"rubrix/sentiment-banking\")"
```

docs/tutorials/01-labeling-finetuning.ipynb
```diff
@@ -304,7 +304,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "dataset_ds = load_dataset(\"Recognai/sentiment-banking\", split=\"train\")\n",
+    "dataset_ds = load_dataset(\"Rubrix/sentiment-banking\", split=\"train\")\n",
```
dcfidalgo:
```suggestion
    "dataset_ds = load_dataset(\"rubrix/sentiment-banking\", split=\"train\")\n",
```

docs/tutorials/01-labeling-finetuning.ipynb
frascuchon: "Rubrix" or "rubrix" @dcfidalgo ?

dcfidalgo: our organization at HF is called "rubrix", so it should be "rubrix"

PR #1187, opened 2022-02-18
fix(#1186): prevent converting None ids to str 'None's

### LINKED ISSUE #1186: [Dataset] None ids are converted to str 'None's
When transforming records to the `datasets.Dataset` format, ids that are `None`s are converted to `'None'`s (string). When reading those records again to a rubrix dataset, and logging them to rubrix, they overwrite themselves.

dcfidalgo:
Closes #1186


frascuchon (approved):
Maybe we need to revise another way to handle id types. Not sure if an string conversion is the safest way.

Issue #1144, opened 2022-02-14
[Weak Supervision] Enable weak supervision for multi_label text classification


frascuchon:
1. Weak supervision with multi_labels rules (CLIENT)
2. Define rules with multi_label classes (UI + SERVER)

## Tasks

- [x] Enable multi label selection for "Define Rules" (API and UI)
- [x] Adapt precision and correct/incorrect metrics (UII and API)
- [x] Adapt WeakLabelMatrix and LabelModel (Client)

Amelie-V:
Single/multi labels

- [x] Allow deselecting all labels while I'm creating or editing a rule (show Save disabeled)
- [x] Change are to is : "This query with these labels is already saved as rule"

PR #1190, opened 2022-02-18
feat(#1144): weak multi labels

dcfidalgo:
This PR addresses part of #1144 and implements a weak label matrix for multi-label text classification datasets.
It introduces a new `WeakMultiLabels` class that stores the weak labels as a 3 dim matrix (nr of records x nr of rules x nr of labels) with values of -1 (abstentions), 0 (no vote), and 1 (vote). The annotations are stored in a 2 dim matrix (nr of records x nr of labels). @frascuchon let's have a meeting and check that the metrics of the `WeakMultiLabels.summary` method and the ones in the UI/Server are aligned.

The `WeakLabels` and `WeakMultiLabels` classes share a common base class that facilitates the specific implementations and avoids code duplication.

The next step would be to make the label models compatible with this new weak label class.

### DIFF
--- a/src/rubrix/labeling/text_classification/weak_labels.py
+++ b/src/rubrix/labeling/text_classification/weak_labels.py
@@ -90,7 +90,7 @@ class WeakLabels:
         self._dataset = dataset
 
         self._rules = rules or load_rules(dataset)
-        if self._rules == []:
+        if not self._rules:
             raise NoRulesFoundError(
                 f"No rules were found in the given dataset '{dataset}'"
             )
@@ -134,7 +134,7 @@ class WeakLabels:
 
         if self._records[0].multi_label:
             raise MultiLabelError(
-                "Multi-label text classification is not yet supported."
+                "For multi-label text classification, use the `rb.labeling.text_classification.WeakMultiLabels` class."
             )
 
         # apply rules -> create the weak label matrix, annotation array, final label2int mapping
@@ -177,7 +177,7 @@ class WeakLabels:
         for n, record in tqdm(
             enumerate(self._records), total=len(self._records), desc="Applying rules"
         ):
-            # First: fill annotation array
+            # FIRST: fill annotation array
             try:
                 annotation = _label2int[record.annotation]
             except KeyError as error:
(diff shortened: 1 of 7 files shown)


src/rubrix/labeling/text_classification/weak_labels.py
```diff
@@ -517,18 +617,343 @@ def change_mapping(self, label2int: Dict[str, int]):
...
+    def __init__(
+        self,
+        dataset: str,
+        rules: Optional[List[Callable]] = None,
+        ids: Optional[List[Union[int, str]]] = None,
+        query: Optional[str] = None,
+    ):
+        super().__init__(dataset=dataset, rules=rules, ids=ids, query=query)
+
+        if not self._records[0].multi_label:
```
frascuchon: This check means that the WeakMultiLabels can only be generated from multi-labels datasets.

Do you think it would be interesting to allow applying it to single-label datasets? We have some use case where that would be useful.

src/rubrix/labeling/text_classification/weak_labels.py
```diff
@@ -517,18 +617,343 @@ def change_mapping(self, label2int: Dict[str, int]):
...
+            raise SingleLabelError(
+                "For single-label text classification, use the `rb.labeling.text_classification.WeakLabels` class."
+            )
+
+        # apply rules -> create the weak label matrix (3D), annotation (2D), label list
+        self._matrix, self._annotation, self._labels = self._apply_rules()
+
+    def _apply_rules(self) -> Tuple[np.ndarray, np.ndarray, List[str]]:
```
frascuchon, line 676: Could be shared some logic from `_apply_rules` method for both, single and multi-label implementations?

Also, the single-label ones, accepts a `label2int` param and the multi-label don't. Don't we need this parameter in the multi label version?

src/rubrix/labeling/text_classification/weak_labels.py
dcfidalgo: Removed check

src/rubrix/labeling/text_classification/weak_labels.py
dcfidalgo: Just a summary of our quick call:

Haven't found a good way to share logics between both implementations for now, will leave this for a future iteration. 

> Also, the single-label ones, accepts a label2int param and the multi-label don't. Don't we need this parameter in the multi label version?

- The main idea of the `label2int` parameter was to facilitate the usage of the WeakLabels with Snorkel and FlyingSquid, but both can't deal with multi-label out of the box
- with our "binary" weak multi label matrix, a label2int is not really possible. maybe one could think about making the order of the labels configurable (order of the labels on the z axis), maybe this would be roughly equivalent, but not sure how useful this is in the end.

frascuchon (approved):

PR #1197, opened 2022-02-22
feat(search): indexing keyword fields with textual info

frascuchon:
This PR configures some keyword files (like `metadata.*` or some token classification mentions metrics) to allow in search phase query as a full text search.

For example:
```bash
metadata.long_text.text: find me*
```


dcfidalgo (approved):
Nice :100:

PR #1201, opened 2022-02-23
feat(#950): include search keywords as part of record results

### LINKED ISSUE #950: [Search] Enable results highlighting for textual fields
Find a way to return highligthting info with search results. We can exploit the [elasticsearch highlighting](https://www.elastic.co/guide/en/elasticsearch/reference/current/highlighting.html) keeping in mind that, for text classification, highlighting for multiple fields is not trivial when no default string searches.

For example, given a dataset with records including  `inputs.subject` and `inputs.body` the query:
```
size weight
```
could match terms in `inputs.subject`, `inputs.body` or both.

But, if we select a field in query:
```
inputs.subject: (size weight)
```

the highlight results could bring info for both fields `subject` and `body`, resulting in highlighting information that is not totally truthful

frascuchon:
This PR includes as part of search/scan dataset record response the record keywords relevant for provided query.

These keywords will be returned only for full text search queries and are computed in approximative fashion.

Can be used for search highlighting or automatic labelling function generation for Token classification .

See #950

### DIFF
--- a/src/rubrix/client/sdk/token_classification/models.py
+++ b/src/rubrix/client/sdk/token_classification/models.py
@@ -118,6 +118,7 @@ class TokenClassificationRecord(CreationTokenClassificationRecord):
             status=self.status,
             metadata=self.metadata or {},
             metrics=self.metrics or {},
+            search_keywords=self.search_keywords or None,
         )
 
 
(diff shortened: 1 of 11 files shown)


src/rubrix/client/sdk/token_classification/models.py
```diff
@@ -118,6 +118,7 @@ def to_client(self) -> ClientTokenClassificationRecord:
             status=self.status,
             metadata=self.metadata or {},
             metrics=self.metrics or {},
```
dcfidalgo: Not related to this PR, but I think this should be None
```suggestion
            metrics=self.metrics or None,
```

dcfidalgo (dismissed):
Just one small inline comment not related to this PR.

dcfidalgo (approved):

PR #1208, opened 2022-02-28
fix(init): using api sdk wrapper is better

### LINKED ISSUE #1207: rb.init time out, replace httpx with requests, it worked
as above
image
image
image

frascuchon:
This PR changes the client initialization and use the sdk api wrapper. This normalizes the use of timeout in requests and the error handler in responses.

Closes #1207

### DIFF
--- a/src/rubrix/client/rubrix_client.py
+++ b/src/rubrix/client/rubrix_client.py
@@ -19,7 +19,6 @@ import logging
 import socket
 from typing import Any, Dict, Iterable, List, Optional, Union
 
-import httpx
 import pandas
 from tqdm.auto import tqdm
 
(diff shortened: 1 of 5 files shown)


src/rubrix/client/rubrix_client.py
```diff
@@ -104,34 +103,14 @@ def __init__(
             api_url: Address from which the API is serving.
             api_key: Authentication token.
             workspace: Active workspace for this client session.
-            timeout: Seconds to considered a connection timeout.
+            timeout: Seconds to consider a connection timeout.
```
dcfidalgo:
```suggestion
            timeout: Seconds to wait before raising a connection timeout.
```

dcfidalgo (dismissed):

dcfidalgo (approved):

PR #1209, opened 2022-02-28
feat: Implement 'prepare_for_training' for text classification datasets

dcfidalgo:
This PR implements the `prepare_for_training` method for the `DatasetForTextClassification`. I modified the first tutorial to show the usage: https://rubrix.readthedocs.io/en/feat-prepare_for_training/tutorials/01-labeling-finetuning.html
I don't think I will manage to implement the `prepare_for_training` for all tasks before the release.

@dvsrepo @frascuchon For the TokenClassification task, should we rely on spacy utilities to make the conversion from spans to ner tags, or should we provide our own utilities? To me it would seem a bit strange if we required spacy to be installed, to prepare the dataset for training with transformers ... @dvsrepo Do you know of any other library that provides some utilities for converting spans to ner tags?

### DIFF
--- a/src/rubrix/client/datasets.py
+++ b/src/rubrix/client/datasets.py
@@ -231,7 +231,7 @@ class DatasetBase:
         """Prepares the dataset for training.
 
         Args:
-            **kwargs: Specific to the tasks
+            **kwargs: Specific to the task of the dataset.
 
         Returns:
             A datasets Dataset.
@@ -264,13 +264,15 @@ def _prepend_docstring(record_type: Type[Record]):
 class DatasetForTextClassification(DatasetBase):
     """
     Examples:
-        >>> import rubrix as rb
         >>> # Import/export records:
+        >>> import rubrix as rb
         >>> dataset = rb.DatasetForTextClassification.from_pandas(my_dataframe)
         >>> dataset.to_datasets()
+        >>>
         >>> # Looping over the dataset:
         >>> for record in dataset:
         ...     print(record)
+        >>>
         >>> # Passing in a list of records:
         >>> records = [
         ...     rb.TextClassificationRecord(inputs="example"),
@@ -278,6 +280,7 @@ class DatasetForTextClassification(DatasetBase):
         ... ]
         >>> dataset = rb.DatasetForTextClassification(records)
         >>> assert len(dataset) == 2
+        >>>
         >>> # Indexing into the dataset:
         >>> dataset[0]
         ... rb.TextClassificationRecord(inputs={"text": "example"})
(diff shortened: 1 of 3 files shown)


src/rubrix/client/datasets.py
```diff
@@ -393,6 +403,48 @@ def _from_pandas(cls, dataframe: pd.DataFrame) -> "DatasetForTextClassification"
...
         )
 
+    @_requires_datasets
+    def prepare_for_training(self) -> "datasets.Dataset":
+        """Prepares the dataset for training.
+
+        This will return a "datasets.Dataset" with two columns: "text" and "label"
+        - Records without an annotation are removed.
+        - The "text" column is taken from the _inputs_ attribute of the records.
+            If there is more than one key in the dictionary, the values will be joined together with white space.
```
dvsrepo: I would discourage this behavior as users would lose the possibility to join the fields however they want (for example adding special tokens) or even just use one of the fields for training. 

My recommendation would be to create one column for each key in the dict

src/rubrix/client/datasets.py
```diff
@@ -393,6 +403,48 @@ def _from_pandas(cls, dataframe: pd.DataFrame) -> "DatasetForTextClassification"
...
+            labels.update(annotation)
+
+        class_label = datasets.ClassLabel(names=sorted(list(labels)))
+        features = datasets.Features(
+            text=datasets.Value("string"),
+            label=[class_label] if self._records[0].multi_label else class_label,
+        )
+
+        text, label = [], []
+        for rec in self._records:
```
dvsrepo: Would it be possible to only loop through the records once, or reduce the cost of computing labels and text+label?

src/rubrix/client/datasets.py
dcfidalgo: :+1: done

src/rubrix/client/datasets.py
dcfidalgo: I tried to optimize the code in a few iterations using timeit. Now, on my machine, I get an execution time of around 1s for ~500 000 records. Maybe this can be optimized further, but I think for a first version this should be sufficient.

dvsrepo: Great! 💯

dvsrepo:
> This PR implements the `prepare_for_training` method for the `DatasetForTextClassification`. I modified the first tutorial to show the usage: https://rubrix.readthedocs.io/en/feat-prepare_for_training/tutorials/01-labeling-finetuning.html I don't think I will manage to implement the `prepare_for_training` for all tasks before the release.
> 
> @dvsrepo @frascuchon For the TokenClassification task, should we rely on spacy utilities to make the conversion from spans to ner tags, or should we provide our own utilities? To me it would seem a bit strange if we required spacy to be installed, to prepare the dataset for training with transformers ... @dvsrepo Do you know of any other library that provides some utilities for converting spans to ner tags?

I agree we shouldn't rely on spaCy. Let me think of alternatives

dvsrepo (approved):

frascuchon:
> > This PR implements the `prepare_for_training` method for the `DatasetForTextClassification`. I modified the first tutorial to show the usage: https://rubrix.readthedocs.io/en/feat-prepare_for_training/tutorials/01-labeling-finetuning.html I don't think I will manage to implement the `prepare_for_training` for all tasks before the release.
> > @dvsrepo @frascuchon For the TokenClassification task, should we rely on spacy utilities to make the conversion from spans to ner tags, or should we provide our own utilities? To me it would seem a bit strange if we required spacy to be installed, to prepare the dataset for training with transformers ... @dvsrepo Do you know of any other library that provides some utilities for converting spans to ner tags?
> 
> I agree we shouldn't rely on spaCy. Let me think of alternatives

The server part computes tags at token level. You could use for this purpose. If i remember the computed tags are in the IOB format. It could be a begining

dvsrepo:
> > > This PR implements the `prepare_for_training` method for the `DatasetForTextClassification`. I modified the first tutorial to show the usage: https://rubrix.readthedocs.io/en/feat-prepare_for_training/tutorials/01-labeling-finetuning.html I don't think I will manage to implement the `prepare_for_training` for all tasks before the release.
> > > @dvsrepo @frascuchon For the TokenClassification task, should we rely on spacy utilities to make the conversion from spans to ner tags, or should we provide our own utilities? To me it would seem a bit strange if we required spacy to be installed, to prepare the dataset for training with transformers ... @dvsrepo Do you know of any other library that provides some utilities for converting spans to ner tags?
> > 
> > 
> > I agree we shouldn't rely on spaCy. Let me think of alternatives
> 
> The server part computes tags at token level. You could use for this purpose. If i remember the computed tags are in the IOB format. It could be a begining

Definitely, that's more than enough!

dcfidalgo:
> The server part computes tags at token level. You could use for this purpose. If i remember the computed tags are in the IOB format. It could be a begining

Nice, totally forgot about this. Hm, should I copy the logic to the client part? Shall we require the user to upload/load the records before calling `prepare_for_training`? Or is there an endpoint where I can send the records and get back the metrics/iob tags?

frascuchon:
Not sure about the user flow here. We can discuss tomorrow in a  call @dcfidalgo

Issue #1183, opened 2022-02-18
[NER] UI Fine tuning 


Amelie-V:
### Small interaction improvement

- [x] Improve selection (token and group) interaction. ex: When the mouse starts selecting before the beginning of a sentence. Try to find a way to match better selection "in grey" and "final selection". If the user selects a letter in the middle of a sentence, the grey selection should mark automatically the token.
- [x] Change close icon location. Show cross on hover
- [x] Remove annotation: on double click and click on the cross
- [x] Firefox: Disable grey selection when removing annotation with a double click.

- [x] Disabel shortcut pagination when dropdown is active

### Feature added

- [x] Add Clear annotations bt
- [ ] Add dropdown config.?

Amelie-V:
- [x] Scroll is blocking access to last records at page bottom, independently of numbers of records per page (Cross tasks)
- [x] Close button location labeled record:  Standard is to show the close bt on the right side. But it is not possible when the content jumps to the line. 
Do we prefer to show it always on the left side or change it regarding the record location?
>Leave it as it is ;) 
- [x] Clear annotation user flow: When a record is validated don't show "Clear annotations". Show it on *Pending* and *no Validated* status 
- [x] Apply Same width to save and validate button

Amelie-V:
- [x] Review "Save" bt size
- [x] Add disabeled state to "Clear ..." bt

leiyre:
- [x] Entities selector position
- [x] Style empty Entities in header
- [x] Single entity annotation

PR #1217, opened 2022-03-03
fix(#1183): NER adjustments

leiyre:
See #1183 

This PR fixes entities selector position, empty entities styles in header and single entity annotation


frascuchon (approved):
👍

PR #1226, opened 2022-03-04
feat(#1225): create iob tags from record spans

frascuchon:
In this PR we include client functionality for generate iob tags from record spans definitions. This will help for generate training dataset for huggingface models training

See #1225

### DIFF
--- a/src/rubrix/client/models.py
+++ b/src/rubrix/client/models.py
@@ -19,6 +19,7 @@ This module contains the data models for the interface
 
 import datetime
 import warnings
+from collections import defaultdict
 from typing import Any, Dict, List, Optional, Tuple, Union
 
 import pandas as pd
@@ -85,6 +86,7 @@ class _Validators(BaseModel):
 
     class Config:
         extra = "forbid"
+        underscore_attrs_are_private = True
 
 
 class BulkResponse(BaseModel):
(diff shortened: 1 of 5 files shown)


src/rubrix/client/models.py
```diff
@@ -246,6 +248,13 @@ class TokenClassificationRecord(_Validators):
     metrics: Optional[Dict[str, Any]] = None
     search_keywords: Optional[List[str]] = None
 
+    __chars2tokens__: Dict[int, int] = None
+    __tokens2chars__: Dict[int, Tuple[int, int]] = None
+
+    def __init__(self, **data):
+        super().__init__(**data)
+        self.__chars2tokens__, self.__tokens2chars__ = self.__build_indices_map__()
```
frascuchon: @dcfidalgo  better if we defined both maps as lazy properties. Any idea about how to do that?

frascuchon:
@dcfidalgo Any idea about how to define as read-only the `text` and `tokens` properties?

frascuchon:
Finally I didn't found a way to keep text and tokens inmutables. So the tokens/chars map will be dynamically computed every time. I've include a cache resolution to avoid extra computations.

Take a look @dcfidalgo

dcfidalgo:
@frascuchon Have a look at https://github.com/recognai/rubrix/pull/1226/commits/c0b5900ad79537146489ce920e012ff8928aa8e6 for making text and tokens immutable. Maybe we could move the "immutability" of `text` to the base `_Validators` class to have this for all record models.

dcfidalgo (approved):

PR #1228, opened 2022-03-06
feat(#981): add majority voter with multi label support

### LINKED ISSUE #981: [LabelModel] Implement a majority vote label model

dcfidalgo:
Closes #981 

This PR adds a `MajorityVoter` to our label models and supports a `WeakMultiLabels` object. In the multi-label case, this label model simply sets the probability to 1, if at least one rule voted for the respective label. Otherwise, the probability is zero.

I think a lot of the logic can be shared with the FlyingSquid label model, but I would like to do the refactoring in a separate PR. Also, Ruan is working heavily on this label model, so I would like to wait for his PR to go in. 

You can find the tutorial here: https://rubrix.readthedocs.io/en/feat-majority_vote_model/tutorials/weak-supervision-multi-label.html
I'm not super thrilled about the tutorial, mainly because of the bad results. Either the datasets are really challenging, or I'm simply bad at finding good heuristics. I doubt that a more sophisticated label model will bring much improvement, but we can use those results as a baseline. @dvsrepo feedback would be much appreciated!

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -198,6 +198,7 @@ You can join the conversation on our Github page and our Github forum.
    tutorials/08-error_analysis_using_loss
    tutorials/09-automatic_fastapi_log
    tutorials/skweak
+   tutorials/weak-supervision-multi-label
 
 .. toctree::
    :maxdepth: 4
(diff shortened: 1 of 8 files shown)


dcfidalgo:
cc @frascuchon @dvsrepo

dvsrepo:
From a quick read of the tutorial it looks great for a first iteration!!! My only comment would be to add an info box in the define rules section to explicitly say that this rules can be defined with the UI too. No need to add screenshots or anything, to avoid maintenance overhead, but I'd like to make clear you can use the UI for that. I'd probably add a link to the UI ref for the define rules module (Q: will Leire update that guide with info about multilabel weak labeling?)

frascuchon (dismissed):

dcfidalgo:
>  (Q: will Leire update that guide with info about multilabel weak labeling?)

Yes, we have to update this guide anyways, since we now also have the filters available.

PR #1231, opened 2022-03-08
feat(#1225): prepare tokenclass dataset for hf training

### LINKED ISSUE #1225: [HuggingFace Datasets] export token classification dataset for hf training

frascuchon:
This PR includes the `prepare_for_training` method implementation for token classification datasets.

This method extends generated dataset from `to_datasets` including the annotation `ner_tags` column. This allow use the same published dataset for training and/or for explore/annotate in rubrix

Closes #1225

### DIFF
--- a/src/rubrix/client/datasets.py
+++ b/src/rubrix/client/datasets.py
@@ -539,6 +539,32 @@ class DatasetForTokenClassification(DatasetBase):
         return super().from_pandas(dataframe)
 
     def prepare_for_training(self) -> "datasets.Dataset":
+        """Prepares the dataset for training.
+
+        This will return a ``datasets.Dataset`` with all columns returned by ``to_datasets`` method
+        and an additional  *ner_tags* column:
+            - Records without an annotation are removed.
+            - The *ner_tags* column corresponds to the iob tags sequences for annotations of the records
+            - The iob tags are transformed to integers.
+
+        Returns:
+            A datasets Dataset with a *ner_tags* column and all columns returned by ``to_datasets``.
+
+        Examples:
+            >>> import rubrix as rb
+            >>> rb_dataset = rb.DatasetForTokenClassification([
+            ...     rb.DatasetForTokenClassification(
+            ...         text="The text",
+            ...         tokens=["the", "text"],
+            ...         annotation=[("TAG", 0, 2)],
+            ...     )
+            ... ])
+            >>> rb_dataset.prepare_for_training().features
+            {'text': Value(dtype='string'),
+             'tokens': Sequence(Value(dtype='string')),
+             'ner_tags': ClassLabel(num_classes=2, names=['O','B-TAG'])}
+
+        """
         import datasets
 
         class_tags = ["O"]
--- a/src/rubrix/client/datasets.py
+++ b/src/rubrix/client/datasets.py
@@ -538,6 +538,7 @@ class DatasetForTokenClassification(DatasetBase):
     ) -> "DatasetForTokenClassification":
         return super().from_pandas(dataframe)
 
+    @_requires_datasets
     def prepare_for_training(self) -> "datasets.Dataset":
         """Prepares the dataset for training.
 
--- a/src/rubrix/client/datasets.py
+++ b/src/rubrix/client/datasets.py
@@ -125,7 +125,8 @@ class DatasetBase:
 
         ds_dict = self._to_datasets_dict()
         # TODO: THIS FIELD IS ONLY AT CLIENT API LEVEL. NOT SENSE HERE FOR NOW
-        del ds_dict["search_keywords"]
+        if "search_keywords" in ds_dict:
+            del ds_dict["search_keywords"]
 
         try:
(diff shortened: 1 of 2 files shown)


src/rubrix/client/datasets.py
```diff
@@ -535,6 +538,49 @@ def from_pandas(
     ) -> "DatasetForTokenClassification":
         return super().from_pandas(dataframe)
 
+    def prepare_for_training(self) -> "datasets.Dataset":
```
dcfidalgo, line 542: I would use the `@_requires_datasets` decorator to get a helpful error message when `datasets` is not installed. 
```suggestion
    @_requires_datasets
    def prepare_for_training(self) -> "datasets.Dataset":
```

src/rubrix/client/datasets.py
```diff
@@ -535,6 +538,49 @@ def from_pandas(
...
+        import datasets
```
dcfidalgo, line 581: A docstring would be nice :) I would at least describe the output format.

dcfidalgo (commented):
Nice, just a couple of inline comments.

src/rubrix/client/datasets.py
```diff
@@ -535,6 +538,76 @@ def from_pandas(
...
+            - The *ner_tags* column corresponds to the iob tags sequences for annotations of the records
+            - The iob tags are transformed to integers.
+
+        Returns:
+            A datasets Dataset with a *ner_tags* column and all columns returned by ``to_datasets``.
+
+        Examples:
+            >>> import rubrix as rb
+            >>> rb_dataset = rb.DatasetForTokenClassification([
+            ...     rb.DatasetForTokenClassification(
```
```suggestion
            ...     rb.TokenClassificationRecord(
```

src/rubrix/client/datasets.py
```diff
@@ -535,6 +538,76 @@ def from_pandas(
...
+            ...         text="The text",
+            ...         tokens=["the", "text"],
```
```suggestion
            ...         tokens=["The", "text"],
```

src/rubrix/client/datasets.py
```diff
@@ -535,6 +538,76 @@ def from_pandas(
...
+            ...         annotation=[("TAG", 0, 2)],
+            ...     )
+            ... ])
+            >>> rb_dataset.prepare_for_training().features
+            {'text': Value(dtype='string'),
+             'tokens': Sequence(Value(dtype='string')),
+             'ner_tags': ClassLabel(num_classes=2, names=['O','B-TAG'])}
```
```suggestion
            {'text': Value(dtype='string'),
             'tokens': Sequence(feature=Value(dtype='string'), length=-1),
             'prediction': Value(dtype='null'),
             'prediction_agent': Value(dtype='null'),
             'annotation': [{'end': Value(dtype='int64'),
               'label': Value(dtype='string'),
               'start': Value(dtype='int64')}],
             'annotation_agent': Value(dtype='null'),
             'id': Value(dtype='null'),
             'metadata': Value(dtype='null'),
             'status': Value(dtype='string'),
             'event_timestamp': Value(dtype='null'),
             'metrics': Value(dtype='null'),
             'ner_tags': [ClassLabel(num_classes=3, names=['O', 'B-TAG', 'I-TAG'])]}

```

dcfidalgo (commented):
I think we should unify the behavior of the `prepare_for_training` methods across the tasks:
- Should we return the full records with additional columns added (TokenClassification)?
- Should we only return the columns needed for training (TextClassification)?

frascuchon:
> I think we should unify the behavior of the `prepare_for_training` methods across the tasks:
> 
> * Should we return the full records with additional columns added (TokenClassification)?
> * Should we only return the columns needed for training (TextClassification)?

For me, the `prepare_for_training` match better as an extended datasets since (1) allow reuse dataset cache if you created the dataset before, (2) you can pub to hub and them import to rubrix for free

Let's talk in a call if you prefer.

dcfidalgo:
Ok, then for the TextClassification, the `prepare_for_training` simply adds a `label` column, right? Or do we still want to extract the `inputs` keys as columns?

Reminder for me: if we don't extract the keys, update the tutorial: https://rubrix.readthedocs.io/en/master/tutorials/01-labeling-finetuning.html#Prepare-training-and-test-datasets

frascuchon:
I'm inclined to extract the inputs as well. This way we avoid any "flatten" process that is required for training

dcfidalgo (approved):
Maybe an argument against the concept of `prepare_for_training` being an extended `to_datasets`: when we talked about this method in the beginning, we were discussing an argument that prepares the dataset for the training with different frameworks (transformers, spacy, ...). Maybe something we should keep in mind.

frascuchon:
> Maybe an argument against the concept of `prepare_for_training` being an extended `to_datasets`: when we talked about this method in the beginning, we were discussing an argument that prepares the dataset for the training with different frameworks (transformers, spacy, ...). Maybe something we should keep in mind.

Totally true! The format we're providing in this PR is compatible with the  expected format for `transformers`, that's right ? 
The format for other models should totally differs from a `datasets.Dataset`. So, it's fine as is for me. . But we definitely should keep in mind that.

PR #1233, opened 2022-03-08
feat(#1144) allow multilabel records in weak supervision

leiyre:
see #1144

### DIFF
--- a/frontend/components/text-classifier/labeling-rules/RuleLabelsDefinition.vue
+++ b/frontend/components/text-classifier/labeling-rules/RuleLabelsDefinition.vue
@@ -151,8 +151,8 @@ export default {
         return "The rule was saved";
       }
       if (this.selectedLabels && queryWithLabelsIsStored) {
-        return `This query with this ${
-          this.selectedLabels.length > 1 ? "labels" : "label"
+        return `This query with ${
+          this.selectedLabels.length > 1 ? "these labels" : "this label"
         } are already saved as rule`;
       }
     },
--- a/frontend/components/text-classifier/labeling-rules/RuleLabelsDefinition.vue
+++ b/frontend/components/text-classifier/labeling-rules/RuleLabelsDefinition.vue
@@ -150,7 +150,7 @@ export default {
       if (this.isSaved) {
         return "The rule was saved";
       }
-      if (queryWithLabelsIsStored) {
+      if (this.selectedLabels && queryWithLabelsIsStored) {
         return `This query with this ${this.selectedLabels.length > 1 ? 'labels' : 'label'} are already saved as rule`;
       }
     },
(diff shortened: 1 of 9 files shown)


frontend/components/text-classifier/labeling-rules/RuleLabelsDefinition.vue
```diff
@@ -130,23 +131,27 @@ export default {
...
         return "The rule was saved";
       }
-      if (
-        this.currentRule &&
-        this.selectedLabelsVModel.length &&
-        this.dataset.findRuleByQuery(this.currentRule.query, this.selectedLabel)
-      ) {
-        return "This query with this label are already saved as rule";
+      if (queryWithLabelsIsStored) {
+        return "This query with this labels are already saved as rule";
```
frascuchon: The message should be built depending on the number of labels the rule have defined. Should keep the singular form if the number of labels is equal to one (even for multi-label)

frascuchon (changes requested):
Precision, correct and incorrect metrics are missing. 

Screenshot 2022-03-08 at 17 13 46

The same dataset with master branch is still working.

Also, labels values in field are misaligned.

frontend/components/text-classifier/labeling-rules/RuleLabelsDefinition.vue
```diff
@@ -130,23 +131,27 @@ export default {
...
         return "The rule was saved";
       }
-      if (
-        this.currentRule &&
-        this.selectedLabelsVModel.length &&
-        this.dataset.findRuleByQuery(this.currentRule.query, this.selectedLabel)
-      ) {
-        return "This query with this label are already saved as rule";
+      if (queryWithLabelsIsStored) {
+        return `This query with this ${this.dataset.isMultiLabel ? 'labels' : 'label'} are already saved as rule`;
```
frascuchon: I think it should be better to use the number of labels in rule. Even for multi-label rules definition you could define a single label.

frascuchon:
- No rules defined but the message "This query with this labels are already saved as rule" is shown

Screenshot 2022-03-14 at 10 33 46

dvsrepo:
> * No rules defined but the message "This query with this labels are already saved as rule" is shown
> 
> Screenshot 2022-03-14 at 10 33 46

@leiyre also the message should be these labels and not this labels. Thanks!!

frascuchon:
Adding a label to an existing rule don't refresh the labes in rules summary

Screenshot 2022-03-14 at 23 06 58

Screenshot 2022-03-14 at 23 07 50

frascuchon:
1. select an already created rule from rule summary
2. The precision and correct/incorrect metrics show the rule metrics 
3. change the rule query -> related metrics are reset.
4. Precision and correct/incorrect metrics are not computed.

Screenshot 2022-03-15 at 15 00 42

frascuchon (approved):

PR #1234, opened 2022-03-08
feat(datasets): simplify load flow from hf datasets with no rb format

frascuchon:
With this PR you can convert hf datasets to rubrix datasets as follow:

```python

import datasets
amazon_reviews = datasets.load_dataset("amazon_reviews_multi", "es", split="train")
rb_ds = rb.DatasetForTextClassification.from_datasets(
    amazon_reviews, 
    inputs=["review_title","review_body"], 
    annotation="stars"
)
# Here you can just log the dataset
rb.log(rb_ds, name="amazon-reviews")
```
We can define similar approach for `metadata` or `text` for the  other tasks, or `tags` for token classification

@dcfidalgo @dvsrepo any suggestion?

TODO:
- [x] Include tests for text2text
- [x] Fix failing tests once hf bug is fixed:

```python
import datasets
ds = datasets.Dataset.from_dict({ "a": [None, [1,2], None, [3,4]]})
ds.map(lambda x: { "b": "b"}).to_pandas() # None values in a becomes empty lists
```

See this PR https://github.com/huggingface/datasets/issues/3676

### DIFF
--- a/src/rubrix/client/datasets.py
+++ b/src/rubrix/client/datasets.py
@@ -301,6 +301,8 @@ class DatasetForTextClassification(DatasetBase):
         # we implement this to have more specific type hints
         cls,
         dataset: "datasets.Dataset",
+        inputs: Optional[Union[str, List[str]]] = None,
+        annotation: Optional[str] = None,
     ) -> "DatasetForTextClassification":
         """Imports records from a `datasets.Dataset`.
 
@@ -308,6 +310,8 @@ class DatasetForTextClassification(DatasetBase):
 
         Args:
             dataset: A datasets Dataset from which to import the records.
+            inputs: A list of field names used as record inputs. Default: `None`
+            annotation: The field name used as record annotation. Default: `None`
 
         Returns:
             The imported records in a Rubrix Dataset.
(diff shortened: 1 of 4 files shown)


dvsrepo:
> With this PR you can convert hf datasets to rubrix datasets as follow:
> 
> ```python
> import datasets
> amazon_reviews = datasets.load_dataset("amazon_reviews_multi", "es", split="train")
> rb_ds = rb.DatasetForTextClassification.from_datasets(
>     amazon_reviews, 
>     inputs=["review_title","review_body"], 
>     annotation="stars"
> )
> # Here you can just log the dataset
> rb.log(rb_ds, name="amazon-reviews")
> ```
> 
> We can define similar approach for `metadata` or `text` for the other tasks, or `tags` for token classification
> 
> @dcfidalgo @dvsrepo any suggestion?

This is looking good!

For `text` it should be straightforward, for metadata there are some dataset which contain json objects in some fields. 

For token classification we need at least `tokens` and `tags` and then generate stand-off annotations with that, right?

dcfidalgo:
Nice shortcut, will definitely be helpful. Just as a side note, for token classification datasets on the Hub, often only tokens are provided, so we must reconstruct the `text` input for the `TokenClassificationRecords`.

frascuchon:
> Nice shortcut, will definitely be helpful. Just as a side note, for token classification datasets on the Hub, often only tokens are provided, so we must reconstruct the `text` input for the `TokenClassificationRecords`.

Yes, we can recompose text field in terms of tokens field by assuming white spaces between tokens

frascuchon:
> For token classification we need at least `tokens` and `tags` and then generate stand-off annotations with that, right

Agree with this. Although it would also take that logic to the creation of a token classification record

```python
record = TokenClassificationRecord(tokens=....,tags=.....)
```

frascuchon:
I've included a workaround to fix the test. The token classification model will try to build the text field from tokens if no char-level span is provided. 

But in general, operating with `None` annotation/prediction lists could raise errors related to issue: https://github.com/huggingface/datasets/issues/3676 

Anyway, we can go ahead with this feat. even if issue is not resolved since most of hf datasets will provide annotations

dcfidalgo (dismissed):
:100: Very nice!

PR #1246, opened 2022-03-10
feat: Introduce a 'text' argument for the TextClassificationRecord

dcfidalgo:
This PR introduces a `text` argument for the `TextClassificationRecord` to make the records more consistent with each other. It also helps a lot in cases, where we provide a simple string to the `inputs` argument, and when we load the records from Rubrix again, we always have to extract the `text` key from the inputs dict. In the tutorials we often do this:
```
rb.log(TextClassificationRecord(inputs="string"), "mock")
records_df = rb.load("mock")
records_df["text"] = records_df.inputs.map(lambda x: x["text"])
```
With this PR, the recommended way would be:
```
rb.log(TextClassificationRecord(text="string"), "mock")
records_df = rb.load("mock")
```
For now, we still accept strings in the `inputs` argument, but in the future, this should be deprecated.

To Do:
- [x] add tests
- [x] update documentation/tutorials


dcfidalgo:
@frascuchon Could you have a quick look at this proposal? If you are ok with it, I will tackle the ToDos.

frascuchon:
In general I feel confortable with the approach. Just some question about data loading flow:

- Logging records using the `text` property, will create records with `inputs.text` field in server, isn't it?
- Loading old datasets with this new data model with duplicate the text info in `text` and `inputs.text` columns ?

dcfidalgo:
> Logging records using the text property, will create records with inputs.text field in server, isn't it?

Yes, the server data model does not change, `text="mock"` is basically a shortcut for `inputs={"text": "mock"}` .

> Loading old datasets with this new data model with duplicate the text info in text and inputs.text columns ?

Yes, even if you create a new record with `rec = rb.TextClassificationRecord(text="mock")` it will copy it to `inputs`, so `rec.inputs == {"text": "mock"}`. Since i made text and inputs immutable, I think this is fine.

frascuchon:
Great!

docs/getting_started/concepts.rst
```diff
@@ -65,9 +63,8 @@ Another similar task to Text Classification, but yet a bit different, is Multi-l
 .. code-block:: python
 
    record = rb.TextClassificationRecord(
-        inputs={
-            "text": "I can't wait to travel to Egypts and visit the pyramids"
-        },
+        inputs="I can't wait to travel to Egypts and visit the pyramids",
```
frascuchon:
```suggestion
        text="I can't wait to travel to Egypts and visit the pyramids",
```

frascuchon (dismissed):
💯

PR #1252, opened 2022-03-11
docs: define the basic search schema

frascuchon:
I've include a bullet list for text searches in webapp.

@dcfidalgo Could you please apply the Paulo Coelho translator to generate the documentation? (i mean, could you please redact the docs from that?

We can talk next week for that, but it could be nice update the guide for weak supervision release.

### DIFF
--- a/docs/getting_started/advanced_setup_guides.rst
+++ b/docs/getting_started/advanced_setup_guides.rst
@@ -87,17 +87,13 @@ Server
 
 - ``ELASTICSEARCH``: URL of the connection endpoint of the Elasticsearch instance (Default: http://localhost:9200 ).
 
-- ``RUBRIX_ELASTICSEARCH_SSL_VERIFY``: If "False", disables SSL certificate verification when connection to the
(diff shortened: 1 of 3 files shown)


dcfidalgo:
Ok, I think this could be the first complete version: https://rubrix.readthedocs.io/en/docs-define-search-schema/reference/webapp/search_records.html
I will talk to @leireropl to make a more suitable video that showcases a few more query strings for a follow-up PR.
@frascuchon I thought about including the metrics fields in the Field glossary, what do you think? Could you check that all relevant fields are present, and maybe add some more descriptions?

frascuchon:
Very nice @dcfidalgo !!!!

Agree including metrics as part of field glossary. I will collect all of them and I'll share with you.

dcfidalgo:
@frascuchon included the metrics in the glossary. Can't select you as a reviewer since it is your PR, but my review is positive :) From my side this can be merged.

dcfidalgo (approved):

PR #1257, opened 2022-03-14
fix(#1241, #1245): show new line char in metrics plot & increase mentions in entity consistency

### LINKED ISSUE #1241: [Metrics] New line character counted as blank in token frequency metrics
When computing the **token frequency** in the metrics and plotting it, the new line chat (\n) counts as a blank space and it also is displayed in that way.

1. Run **from rubrix.metrics.token_classification import token_frequency**
2. Visualize it with a specific dataset (i.e. **token_frequency(name="med7_trf").visualize()**)
3. Look at the most frequent result.

### LINKED ISSUE #1245: [Metrics] Change default value for the argument 'mentions' in entity_consistency
When dealing with metrics (specifically with **entity_consistency**), the default value of the argument 'mentions' is set to 10. With this default value, sometimes it is not possible to obtain data.

If the default value was higher (i.e. **mentions=20**), I think we could get rid of this problem.

dcfidalgo:
This PR closes #1241 and #1245.

- New line characters are now shown explicitly in the x-axis of the metrics plots for bar plots.
- the default value for `mentions` in entity consistency is increased

### DIFF
--- a/src/rubrix/metrics/helpers.py
+++ b/src/rubrix/metrics/helpers.py
@@ -20,7 +20,10 @@ def bar(data: dict, title: str = "Bar", x_legend: str = "", y_legend: str = ""):
         return empty_visualization()
 
     keys, values = zip(*data.items())
-    keys = [key.encode("unicode-escape").decode() for key in keys]
+    keys = [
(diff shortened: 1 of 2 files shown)


frascuchon (approved):
👍

PR #1269, opened 2022-03-17
docs(#1267): move docker command to one line

### LINKED ISSUE #1267: Provide a one-line docker command for Windows users
For windows users the docker command does not work due to separators.
See https://github.com/recognai/rubrix#get-started

It would be nice to include a one-liner version too so they can copy it.  Something like: If you are using Windows copy and run this: {one-line command}

**Additional context**
This should be included in both docs and readme

dcfidalgo:
Closes #1267 

I simply moved the docker command to one line, since this works for linux/macos, too. 
This PR also contains a few minor fixes.


dvsrepo (approved):
💯

PR #1282, opened 2022-03-21
feat(#1119): Users without personal datasets

### LINKED ISSUE #1119: [Workspaces] Users without personal datasets
Users without personal datasets but that belongs to one or more workspaces which have datasets, should automatically change to one of those workspace?

Better to show all datasets from all workspaces in datasets list allowing to filter by workspace?

leiyre:
closes #1119

This PR changes workspace management, workspace is now a filtrable field in datasets list table


frascuchon:
Applied filters are not visibles as url query params. Following image shows a dataset list where recognai filter is applied from breadcrumb click and francisco is applied by clicking in filter selector. 

Screenshot 2022-03-22 at 16 59 54

The resulting url contains only the first one.

Screenshot 2022-03-22 at 17 00 01

frascuchon (changes requested):
Comments above

dcfidalgo:
@leiyre I pushed a small commit in which I updated the text when the dataset list is empty. Just to let you know.

frascuchon:
@leiyre please, resolve conflicts locally. A lot of changes are applied in user.vue and I cannot do the conflicts resolution by myself

frascuchon (approved):

PR #1286, opened 2022-03-22
ci: include es 8.0 in build process

frascuchon:
This PR includes required changes to make rubrix compatible with the new elasticsearch 8.x version.

The main changes are:

- Use `id` field as default sort key, instead of the `_id` field, which is disabled as default for this es version
- Avoid to define nested fields in mappings. This configuration was deprecated in previos versions and is not allowed in last version. This changes affect to `metadata.*` fields and how the text searches can be used over them. String metadata fields are defined as keywords, and extends the 2 subfields `text` and `exact` for textual queries. You should use them in query string as follows:
```
metadata.my_field.text:total effort
# or
metadata.my_field.exact:Total Effor~
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


dcfidalgo (approved):
:100:

PR #1298, opened 2022-03-24
updating search filter video for reference docs

leireropl:
Based on the WS tutorial. Explore mode + 3 different searches.


dcfidalgo:
Thanks @leireropl , a couple of comments:
- I would change the second search term to maybe: `text:(camp AND NOT training)`
- No need to show the second page of results, the first one is enough
- maybe we could speed up de video by x1.5, not sure if there is this option in Quick Time

leireropl:
> Thanks @leireropl , a couple of comments:
> 
> * I would change the second search term to maybe: `text:(camp AND NOT training)`
> * No need to show the second page of results, the first one is enough
> * maybe we could speed up de video by x1.5, not sure if there is this option in Quick Time

Let's see if the new video works 😄  I tried to speed up the video but I think it is not noticeable.

dcfidalgo (approved):

PR #1301, opened 2022-03-24
docs: move search ref to guides

dcfidalgo:
Moves the search reference of the web app to the guides section. From the search ref we link to this guide.
https://rubrix.readthedocs.io/en/docs-move_search_ref_to_guides/guides/queries.html


frascuchon (approved):
💯

PR #1314, opened 2022-03-25
feat(#1312): early dataset name validation

### LINKED ISSUE #1312: [Logging] Validate dataset name when rb.log
Try to validate the dataset name early and raise an well defined error if name fails validation

Closes #1200

frascuchon:
Closes #1312

### DIFF
--- a/src/rubrix/client/api.py
+++ b/src/rubrix/client/api.py
@@ -215,7 +215,7 @@ class Api:
             BulkResponse(dataset='example-dataset', processed=1, failed=0)
         """
         if not name:
-            raise InputValueError("Empty project name has been passed as argument.")
+            raise InputValueError("Empty dataset name has been passed as argument.")
 
         if not re.match(DATASET_NAME_REGEX_PATTERN, name):
             raise InputValueError(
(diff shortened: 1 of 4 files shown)


src/rubrix/client/api.py
```diff
@@ -213,6 +217,12 @@ def log(
         if not name:
             raise InputValueError("Empty project name has been passed as argument.")
```
dvsrepo: Not from this PR but shouldn't be dataset name?

src/rubrix/client/api.py
```diff
@@ -213,6 +217,12 @@ def log(
...
 
+        if not re.match(DATASET_NAME_REGEX_PATTERN, name):
+            raise InputValueError(
+                f"Provided name {name} does not match the pattern {DATASET_NAME_REGEX_PATTERN}. "
```
dvsrepo:
```suggestion
                f"Provided dataset name {name} does not match the pattern {DATASET_NAME_REGEX_PATTERN}. "
```

dvsrepo (changes requested):

src/rubrix/client/api.py
frascuchon: Sep

frascuchon:
```suggestion
            raise InputValueError("Empty dataset name has been passed as argument.")
```

frascuchon: Done

dvsrepo (approved):

PR #1345, opened 2022-03-29
docs: update define rules reference + add video to ws multi-label tutorial

dcfidalgo:
This PR updates the "Define rules" references in our docs, and adds a video to the weak supervision multi-label tutorial.

It also updates the tooltips in the UI to be more accurate in the case of multi-label tasks.
@leiyre could you have a quick look at the tooltip changes, hope i did not mess up.


frascuchon (approved):
I see the `number of labels...` meaning for correct/incorrect but not sure if can be confusing. Could we find a definition based on number of records?

dcfidalgo:
> I see the number of labels... meaning for correct/incorrect but not sure if can be confusing. Could we find a definition based on number of records?

For the multi-label case, I think it's not possible to get an accurate definition based on records. In the single-label case, label and record are equivalent, so one option could be to introduce tooltips for each case.

dcfidalgo:
@Amelie-V could you have a look at the video?

frascuchon:
> > I see the number of labels... meaning for correct/incorrect but not sure if can be confusing. Could we find a definition based on number of records?
> 
> For the multi-label case, I think it's not possible to get an accurate definition based on records. In the single-label case, label and record are equivalent, so one option could be to introduce tooltips for each case.

maybe `number of record labels`?

Amelie-V:
@dcfidalgo thank for sharing. It's good. I would recomend to continue a bit clicking on "manage rules" and showing briefly the rules list. But only if it makes sense with the tuto. Another option in to show that part on the main video (read.me)

dcfidalgo:
> maybe number of record labels?

Hm, not a big fan of that wording, but I see where you are going ... hm, not sure, I can't think of anything better ... @dvsrepo any suggestions?

frascuchon:
> > maybe number of record labels?
> 
> Hm, not a big fan of that wording, but I see where you are going ... hm, not sure, I can't think of anything better ... @dvsrepo any suggestions?

nah, it's okay like that. We can rethink tooltip if needed later.

dcfidalgo:
> [](https://github.com/recognai/rubrix/pull/1345#issuecomment-1082910631)@dcfidalgo thank for sharing. It's good. I would recomend to continue a bit clicking on "manage rules" and showing briefly the rules list. But only if it makes sense with the tuto. Another option in to show that part on the main video (read.me)

I wanted something short that simply shows that you can mark two labels at the same time. The other stuff is shown in the video of the weak supervision guide and the other weak supervision tutorial (single-label). Let's use the other video to show a more complete picture of the "Define rules" mode.

PR #1348, opened 2022-03-29
docs(#428): Add example to cookbook about fine-tuning a transformer for token class

### LINKED ISSUE #428: Add NER training example with Hugging Face transformers to the Cookbook
Similar to the following example:

https://rubrix.readthedocs.io/en/stable/guides/cookbook.html#Training

Add example to transform a Rubrix TokenClassification dataset into a Hugging Face dataset for fine-tuning a NER model

dcfidalgo:
Closes #428


docs/guides/cookbook.ipynb
```diff
@@ -319,29 +319,198 @@
...
+    "\n",
+    "# We need to properly tokenize our data and align our labels\n",
+    "tokenizer = AutoTokenizer.from_pretrained(\"google/electra-small-discriminator\")\n",
+    "data_collator = DataCollatorForTokenClassification(tokenizer=tokenizer)\n",
+    "\n",
+    "def tokenize_and_align_labels(examples):\n",
+    "    tokenized_inputs = tokenizer(examples[\"tokens\"], truncation=True, is_split_into_words=True)\n",
+    "\n",
+    "    labels = []\n",
+    "    for i, label in enumerate(examples[f\"ner_tags\"]):\n",
```
dvsrepo: Just to double-check does our `prepare_for_training` method create a `ner_tags` column?

Also, why do we need to use `f"ner_tags"` instead of just `"ner_tags"`

dvsrepo (changes requested):

docs/guides/cookbook.ipynb
dcfidalgo: yes, the prepare_for_training creates a `ner_tags` column now, and no, the f is not necessary ... :)

dcfidalgo:
```suggestion
    "    for i, label in enumerate(examples[\"ner_tags\"]):\n",
```

dvsrepo (approved):
Great job!

PR #1358, opened 2022-03-30
fix(#1265): persist pagination size after query

### LINKED ISSUE #1265: [Record navigation] Don't switch to "Records per page" default after each query
I often find myself changing the "Records per page" again and again because it is not saved between queries. When exploring short records, I want to see as many as I can at once, so I change the "Records per page" to 20. After a new search, i have to change this again.

Once the "Records per page" is changed, it should not go back to its default.

leiyre:
fix #1265
This PR prevents the pagination size from changing to the default size after each query


frascuchon (approved):
Great !!!

PR #1359, opened 2022-03-31
fix(#1244): compute capitalness based on python methods

### LINKED ISSUE #1244: [Metrics] Symbols count as UPPER tokens or characters for token/entity capitalness metrics
When using the **token capitalness** or the **entity capitalness** metrics for analyzing the capitalness on datasets, annotations, and/or predictions, symbols (*, [, (, )...) are counted as **UPPER** results (in other words, as if they were in uppercases). However, Python does not count symbols as uppercase tokens/characters, but the opposite.

1. Run some of these metrics (i.e. **from rubrix.metrics.token_classification import token_capitalness**), and the same for **token_frequency**. The most frequent results, which are probably symbols, are counted as uppercase tokens.
2. run "(".isupper( ). The result is **FALSE**.

**Expected behavior**
These tokens and entities should be counted as lowercase ones. 

**Additional context**
This problem has been encountered on med7 models.

frascuchon:
Closes #1244 

With these changes, referred symbols in issue won't be part of the metric summay (since its `capitalness` is `None`). We can discuss about include a summary class for those cases ( maybe `OTHER`), if can be relevant for the analysis.


dcfidalgo:
> We can discuss about include a summary class for those cases ( maybe OTHER), if can be relevant for the analysis.

I think for now it's fine to simply exclude them from the summary until we encounter a use case for including them in the analysis.

src/rubrix/server/tasks/token_classification/metrics.py
```diff
@@ -333,16 +333,18 @@ def density(value: int, sentence_length: int) -> float:
...
+    def capitalness(value: str) -> Optional[str]:
         """Compute capitalness for a string value"""
         value = value.strip()
-        if value.upper() == value:
+        if value.isupper():
             return "UPPER"
-        if value.lower() == value:
+        if value.islower():
             return "LOWER"
         if value[0].isupper():
```
dcfidalgo, line 341: In [Explainaboard](http://explainaboard.nlpedia.ai/leaderboard/task-ner/) the capitalness "FIRST" is actually defined as "first_caps represents only the first character of the entity is capital", which is equivalent to `str.istitle()`. I think this is more useful since "typical entities" like PER or LOC are written like this.

dcfidalgo (commented):
:100: Thanks! Just one inline comment.

src/rubrix/server/tasks/token_classification/metrics.py
frascuchon: For this case:

```python 
"Company SL".istitle()
```

The result is `False`, but it should be counted as FIRST, maybe. The docstring for `istitle` says:

> In a title-cased string, upper- and title-case characters may only follow uncased characters and lowercase characters only cased ones.

what's mean that `Peter Parker` should works,  but `The ROCK` does not. I prefer keep the custom check for first capital check.

dcfidalgo: I just checked the source code of explainaboard and we actually do the same ... ok, i will push a commit to this PR in which I update the doc string of this metric to make the definition of each "capitalness" crystal clear.

dcfidalgo (dismissed):

frascuchon:
A question here: Documentation should be updated? or changes here are not reflected in documentation (metrics guide)?

cc @dcfidalgo @leireropl

dcfidalgo:
RIght, we need to update at least the images, on it ...

frascuchon:
Everything ok here @dcfidalgo ? Can we merge this work?

PR #1363, opened 2022-03-31
fix(#1362): using active api method instead instance

### LINKED ISSUE #1362: Metrics are failing when no rb.log/rb.init is applied
**Describe the bug**
Metric summary does not work in client when no rb.log is applied before.

**To Reproduce**
```python
from rubrix.metrics.common import text_length

text_length(name="my-dataset") # --> This will fail
```
**Expected behavior**
Metric summary should works

**Screenshots**
N/A

 - OS [e.g. iOS]: macOS
 - Rubrix Version [e.g. 0.10.0]: 0.13.0
 - Docker Image (optional) [e.g. rubrix:v0.10.0]:   N/A

**Additional context**

frascuchon:
Fixing problems found after namespace cleaning with metrics.

Closes #1362

### DIFF
--- a/src/rubrix/metrics/__init__.py
+++ b/src/rubrix/metrics/__init__.py
@@ -1,0 +1,33 @@
+from .commons import records_status, text_length
+from .text_classification import f1, f1_multilabel
+from .token_classification import (
+    entity_capitalness,
+    entity_consistency,
+    entity_density,
+    entity_labels,
+)
+from .token_classification import f1 as ner_f1
+from .token_classification import (
+    mention_length,
+    token_capitalness,
+    token_frequency,
+    token_length,
+    tokens_length,
+)
+
+__ALL__ = [
+    text_length,
+    records_status,
+    f1,
+    f1_multilabel,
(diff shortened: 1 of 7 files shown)


src/rubrix/metrics/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+from .token_classification import f1 as ner_f1
+from .token_classification import (
+    mention_length,
+    token_capitalness,
+    token_frequency,
+    token_length,
+    tokens_length,
+)
+
+__ALL__ = [
```
dcfidalgo: Can we "reuse" Python's [__all__](https://docs.python.org/3/tutorial/modules.html#importing-from-a-package) for this?

dcfidalgo (dismissed):
:100: Just one inline comment

src/rubrix/metrics/__init__.py
frascuchon: sure.

dcfidalgo (approved):

PR #1366, opened 2022-03-31
docs: fix weak supervision single-label tutorial

dcfidalgo:
With this PR you can really skip the weak labeling process of the supervision tutorial (single-label) by downloading the training/test data from the HF Hub, before it was not really possible. It also fixes the results, they were a bit off.

A small cosmetical fix is also done for the find label error tutorial.

@frascuchon it also clarifies which lines exactly you have to substitute when loading the data from the Hub


dvsrepo:
> With this PR you can really skip the weak labeling process of the supervision tutorial (single-label) by downloading the training/test data from the HF Hub, before it was not really possible. It also fixes the results, they were a bit off.

We need to update also the 0.84 in the intro right?
> 
> A small cosmetical fix is also done for the find label error tutorial.
> 
> @frascuchon it also clarifies which lines exactly you have to substitute when loading the data from the Hub

dcfidalgo:
> We need to update also the 0.84 in the intro right?

good catch, fixed it.

frascuchon (approved):

PR #1371, opened 2022-04-01
fix: Computation of capitalness (metrics) when value.strip() results in empty str

### LINKED ISSUE #1244: [Metrics] Symbols count as UPPER tokens or characters for token/entity capitalness metrics

dcfidalgo:
See #1244

### DIFF
--- a/src/rubrix/server/tasks/token_classification/metrics.py
+++ b/src/rubrix/server/tasks/token_classification/metrics.py
@@ -333,7 +333,7 @@ class TokenClassificationMetrics(CommonTasksMetrics[TokenClassificationRecord]):
     @staticmethod
     def capitalness(value: str) -> Optional[str]:
         """Compute capitalness for a string value"""
-        value = value.strip()
+        value = value.strip() or " "
         if value.isupper():
             return "UPPER"
         if value.islower():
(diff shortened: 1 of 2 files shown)


src/rubrix/server/tasks/token_classification/metrics.py
```diff
@@ -333,7 +333,7 @@ def density(value: int, sentence_length: int) -> float:
     @staticmethod
     def capitalness(value: str) -> Optional[str]:
         """Compute capitalness for a string value"""
-        value = value.strip()
+        value = value.strip() or " "
```
dcfidalgo:
```suggestion
        value = value.strip()
        if not value:
            return None
```

frascuchon (approved):

PR #1380, opened 2022-04-05
fix(search): compute dataset schema properly

frascuchon:
A wrong dataset schema computation makes es query won't be built properly with nested fields.

### DIFF
--- a/environment_dev.yml
+++ b/environment_dev.yml
@@ -32,6 +32,7 @@ dependencies:
       # extra test dependencies
       - cleanlab
       - datasets>1.17.0
+      - huggingface_hub==0.4.0 # resolve problems with 0.5.0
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -1,7 +1,11 @@
 import httpx
 import pytest
 from _pytest.logging import LogCaptureFixture
-from loguru import logger
+
+try:
+    from loguru import logger
+except ModuleNotFoundError:
+    logger = None
```
dcfidalgo: For me, it would also be fine to include loguru as test dependency in the `environment_dev.yml` . Up to you!

dcfidalgo (dismissed):
:100: Just one inline comment

tests/conftest.py
frascuchon: Yes. Better including as dep. Done

dcfidalgo (approved):

PR #1386, opened 2022-04-06
fix(#1382): update image path

### LINKED ISSUE #1382: og:image not working when sharing on social media
I don't know exactly why but this seems to have stopped working.

I checked the URL here https://github.com/recognai/rubrix/pull/284 but it doesn't seem to resolve (don't know if it did before but it's been working up until the last releases).

In any case is a good opportunity to try and use the new card created by Amé.

Could you take a look @dcfidalgo?

dcfidalgo:
Closes #1382 

Seems like it works, tested it with Twitter.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -113,7 +113,9 @@ html_css_files = ["css/custom.css"]
 html_theme_options = {"logo_only": True}
 
 # Open graph meta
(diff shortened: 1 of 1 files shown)


dvsrepo (approved):
💯

PR #1391, opened 2022-04-07
feat(#1132): Introduce async version of our API

### LINKED ISSUE #1132: [Monitoring] Introduce a more general async log
Some users find `rb.monitor` very useful because it's async and they wonder if it would be possible to have a more general-purpose `rb.log` which is async too for dealing with custom cases and pipelines.

A specific use case where rb.monitor is not flexible enough (which might raise some enhancement ideas too):

User has a single pipeline/model which receives data from different "customers" and wants to log the predictions into different datasets something like `nlp(data, rubrix_dataset=f"{customer_name}_dataset")` 

In any case if feasible I think an async version of log makes total sense 

What do you think @frascuchon ?

dcfidalgo:
Closes #1132

TODO:
- [x] tests
- [x] documentation 
- [x] improve docstrings
- [x] guides
- [x] add examples
- [ ] ~~Log in batches. Server performance can be compromised otherwise~~ (NOT  IN THIS PR)
- [x] Add logging traces (success and failed logs)


dcfidalgo:
@frascuchon I did the refactoring discussed during the call. What's missing is updating the tests, documentation/docstrings, and maybe an improved behavior of the monitoring module (several retries, some automatic chunking, error messages), but maybe this can also be tackled in a separate PR.

frascuchon:
Since I didn't find a use case just using async `log_async` directly, I've included a `background` argument in the `log` function that could simplify the custom data monitoring.

```python
future= rb.log(
    name="monitor-dataset",
    records=rb_records,
    background=True, 
)

# you can always wait for the results:
future.result()
```
Anyway, the async log version remains visible in API in order to discover some use case using directly async function

I've also included a minimal example in monitoring guides using bentoML + spaCy.

Please, feel free to add your comments

@dvsrepo @dcfidalgo

docs/guides/monitoring.ipynb
```diff
@@ -216,6 +216,120 @@
...
    ]
   },
+  {
+   "cell_type": "markdown",
+   "id": "6987c362-61b3-4682-aa7b-693bed30d3ae",
+   "metadata": {},
+   "source": [
+    "## Using `rb.log` in background mode\n",
+    "\n",
+    "You can monitor your own models without latency by using the `background` param in rb.log\n",
```
dvsrepo:
```suggestion
    "You can monitor your own models without adding a response delay by using the `background` param in rb.log\n",
```

docs/guides/monitoring.ipynb
```diff
@@ -216,6 +216,120 @@
...
+    "\n",
+    "Let show an example using bentoml + spacy NER:\n"
```
dvsrepo:
```suggestion
    "Let's see an example using [BentoML](https://www.bentoml.com/) with a spaCy NER pipeline:\n"
```

dvsrepo (changes requested):

dvsrepo (dismissed):

dcfidalgo:
I removed the `Api._log` method since with the new implementation there is no need for this helper method. The example in the monitor guide looks good to me, I just added a small example in the docstring of the `rb.log` method to simply show the effect of setting `background=True`. Also improved a bit this docstring.

From my side, this is good to go.

frascuchon (approved):
YEAH!

PR #1415, opened 2022-04-18
fix(#1409): Filtering by metadata with value list

### LINKED ISSUE #1409: Filtering by metadata with value list apply a wrong query filter
**Describe the bug**
Having list values in a metadata field for a record and filtering by them return no records

**To Reproduce**
1. Add some data with a metadata field with the list as the value
2. Open the dataset in the web app, select "view metadata" and filter by the field. 
3. No records won't be found  

**Expected behavior**
At least the record should be found as query results

**Screenshots**
Screenshot 2022-04-13 at 09 34 13
Screenshot 2022-04-13 at 09 34 21
Screenshot 2022-04-13 at 09 34 59

 - OS [e.g. iOS]: NA
 - Browser [e.g. chrome, safari]: chrome
 - Rubrix Version [e.g. 0.10.0]: 0.13.2
 - ElasticSearch Version [e.g. 7.10.2]: NA
 - Docker Image (optional) [e.g. rubrix:v0.10.0]:  NA

**Additional context**
NA

leiyre:
fix #1409

This PR fixes query results when filtering list type value in metadata

### DIFF
--- a/frontend/components/commons/header/filters/SelectFilter.vue
+++ b/frontend/components/commons/header/filters/SelectFilter.vue
@@ -37,9 +37,9 @@
         <span v-if="appliedFilters.length">
           <p v-if="typeof appliedFilters === 'string'">{{ appliedFilters }}</p>
           <p
-            v-for="appliedFilter in appliedFilters"
+            v-for="(appliedFilter, index) in appliedFilters"
             v-else
-            :key="appliedFilter"
+            :key="index"
           >
             {{ appliedFilter }}
           </p>
--- a/frontend/components/commons/results/Metadata.vue
+++ b/frontend/components/commons/results/Metadata.vue
@@ -72,7 +72,12 @@ export default {
   computed: {
     normalizedMetadataItems() {
       return Object.keys(this.metadataItems).reduce(
-        (r, k) => ((r[k] = String(this.metadataItems[k])), r),
+        (r, k) => (
+          (r[k] = Array.isArray(this.metadataItems[k])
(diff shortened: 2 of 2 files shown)


frascuchon:
### Just some comments related to the UI design and the user experience. 

With changes included here, now the way that list values are shown is a bit different:

Screenshot 2022-04-18 at 21 07 01
VS
Screenshot 2022-04-18 at 21 08 03

What should be the better way to show the list of values for a metadata field in the record detail @Amelie-V @dvsrepo @leiyre? For my part, not comfortable with both of them (the old and the new one), but we should find a trade-off for this PR and plan a robust solution for another different PR.

The main cause that makes me doubt the current design is result filtering by a list of values in the global filter. The previous filter generates the following global filter status:

Screenshot 2022-04-18 at 21 07 27

3 values are active in the global metadata filter, but just one click was applied. This could introduce a misunderstanding to users if it's not clear that the selected value is really a list of values.

### The bug 
Another thing related to previous behavior is that, when we select the filter from record details, the selected filter does not remain active once the filter is applied. Because of the list of values.

Screenshot 2022-04-18 at 21 07 44

frascuchon (changes requested):
see last comment

frascuchon:
Selecting a value from the metadata list values in the global filter makes the record filter active. Is the expected behavior @Amelie-V? If not, what should be the expected result?

Screenshot 2022-04-20 at 10 21 04
Screenshot 2022-04-20 at 10 21 10

frascuchon (approved):

PR #1416, opened 2022-04-18
Refactor to `rb.DatasetFor*` by default

### LINKED ISSUE #1360: [Client] Move to 'rb.DatasetFor*' by default
- Remove `as_pandas` argument in the `rb.load` method, return a `rb.DatasetFor*` by default
- Return `rb.DatasetFor*` objects instead of records `list`s in the `labeling` module

spetrescu:
Fixes #1360 

1. Remove `as_pandas` argument in the `rb.load` method, return a `rb.DatasetFor*` by default
2. Return `rb.DatasetFor*` objects instead of records `list`s in the `labeling` module

Regarding the first requirement above, I was wondering if could you maybe have a look, to make sure that the fix is correct. If not, could you maybe suggest some changes? 

Regarding the second requirement, could you maybe provide a hint in terms of how to cast from `List[TextClassificationRecord]` to `Dataset`? (assuming that `Dataset` is indeed `rb.DatasetFor*`)


frascuchon:
Hi @spetrescu and thanks for your work.


Regarding the first line of work, you should revise where is used the `as_pandas` argument is used. Since now is removed, modules using the old fashion will fail.

Nothing to do about the second line of work, since the `rb.DatasetFor*` classes are iterable, so you use them as a list of records:

````python
for record in dataset:
   record.....
````

Anyway, it is a good practice to launch the tests with your changes applied. You should find the incompatibilities included with your code. 

````bash
pytest tests
````

Be sure your environment is already configured. See the [docs](https://rubrix.readthedocs.io/en/stable/community/developer_docs.html#development-setup) and let us know if you find some troubles

**Comment by dvsrepo** [comment]
Hi @spetrescu , thanks for this! Do you need further help with this?

**Comment by frascuchon** [comment]
Hello, @spetrescu. Did you make any progress on this?

**Comment by spetrescu** [comment]
Hi! I had some difficulties in running the tests for my changes, so this is why I haven't added anything yet. Specifically, I wasn't aware of the need of having `elasticsearch` already setup on my computer, which wasn't the case. Thus, I was trying to configure that, to be able to run the tests, however, I was unable to bypass: `Failed to establish a new connection: [Errno 61] Connection refused`.

Do you have any recommendations/resources for setting up `elasticsearch` on my machine to be able to use `rubrix`?
I am using macOS Big Sur 11.4.

**Comment by frascuchon** [comment]
Hi @spetrescu 

The simplest way to launch elasticsearch is using docker. You can check the docs [here](https://rubrix.readthedocs.io/en/stable/getting_started/advanced_setup_guides.html#setting-up-elasticsearch-via-docker), but basically, you can launch a compatible elasticsearch instance with the following command:

```bash
docker run -d --name elasticsearch-for-rubrix -p 9200:9200 -p 9300:9300 -e "ES_JAVA_OPTS=-Xms512m -Xmx512m" -e "discovery.type=single-node" docker.elastic.co/elasticsearch/elasticsearch-oss:7.10.2
```

If docker is not an option, you could try to install elasticsearch using [brew](https://brew.sh/), following the [official documentation](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/brew.html). But, unfortunately, not all es versions are available in brew repositories.

Let me know if worked.

frascuchon:
I will finish the work in #1515

Thanks, @spetrescu!

PR #1418, opened 2022-04-19
fix(#1414): Display characters between tokens words

### LINKED ISSUE #1383: When uploading text for TokenClassificationRecord if a word whose suffix matched the following word, the two words will be shown concatenated without a space in between
See below how when an example like "This is" is in the text, where the suffix of the previous word matches the following word, then the two words are not shown with a space in between. Below this, I show the original text and the tokens.

Screen Shot 2022-04-05 at 11 43 28 PM
Screen Shot 2022-04-05 at 11 44 00 PM

### LINKED ISSUE #1414: TokenClassificationRecord - if two consecutive words are identical, the white space separating the words is ignored when uploaded.
TokenClassificationRecord - if two consecutive words are identical, the white space separating the words is ignored when uploaded. As shown below, the text uploaded was "hello hello" with a space between the words. But when uploaded to Rubrix, the text appears with no space in between. I wonder if this is related to issue #1383.

Screen Shot 2022-04-14 at 10 08 19 PM
Screen Shot 2022-04-14 at 10 09 41 PM

leiyre:
fix #1414
fix #1383

This PR fixes the display of characters between duplicate words or when the suffix matches the following word in Token Classifier

### DIFF
--- a/frontend/components/token-classifier/results/RecordTokenClassification.vue
+++ b/frontend/components/token-classifier/results/RecordTokenClassification.vue
@@ -65,17 +65,10 @@ export default {
           const start = recordHasEmoji
             ? indexOf(this.record.text, token, startPosition)
             : this.record.text.indexOf(token, startPosition);
-          const nextStart = recordHasEmoji
-            ? indexOf(
-                this.record.text,
-                this.record.tokens[index + 1],
-                startPosition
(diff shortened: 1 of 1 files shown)


frascuchon (approved):
💯

PR #1427, opened 2022-04-21
refactor: change server module structure

frascuchon:
This PR includes a lot of changes, but the main idea is the new Rubrix `server` modules structure  as follows:

- `apis`: The APIs modules definition. This module contains all API models and handlers for each API version (current is v0)
- `daos`: DAO implementations for datasets and records
- `elasticsearch`: elasticseach-related operation modules: query builder, mappings, client wrapper...
- `errors`: server errors definition. All server errors should be placed here
- `security`: the security stuff modules
- `services`: The services layer modules that will be used by APIs 

As for missing work, we must define per-layer data models. For now, data models are shared by API and services layer, but each layer should have its own models. Anyway, This will be tackled in a separate PR.

If the new structure is ok for you, we can merge this PR (if CI is passing)


dcfidalgo (dismissed):
more deletes than additions are always good :smile:

PR #1433, opened 2022-04-22
feat(#1432): basic dataset settings

### LINKED ISSUE #1432: [Datasets] Allow configuring datasets with a predefined label schema

frascuchon:
This PR includes needed components to create a basic set of settings for token and text classification datasets.

The new required API endpoints are implemented as a new API version (v1). 

The client components are built in a different way. Now, the main API client contains child APIs (datasets API for now) that can be used for specific API communication. Also, the inner client wraps the `httpx` client, it's easier to use the client directly:

````python
self._client.post("/api/v1/datasets"....)
````

Closes #1432 

## Update

- A `rb.create_dataset` is provided
- The dataset settings models are found under the module `rubrix.datasets`
- Settings can only be provided on dataset creation
- No way to fetch already configured dataset settings 

A simple code snippet:

```python
import rubrix as rb
from rubrix.datasets import TextClassificationSettings

settings = TextClassificationSettings(labels_schema={ "Positive", "Negative" })
rb.create_dataset(name="a-new-dataset", settings=settings)
...
rb.log(name="a-new-dataset", records....)
```

## TODOs

- [ ] The settings documentation part
- [ ] A place in guides
- [ ] Check provided annotations in records when log records against labels schema (new API)
- [ ] Allow read settings from UI


dcfidalgo:
@frascuchon I had a quick look at the client part, and from my side, this looks goood. But maybe we could have a quick call tomorrow for a high-level overview.

dcfidalgo (dismissed):

PR #1439, opened 2022-04-26
docs(#1421): update weak-supervision guide

### LINKED ISSUE #1421: [docs] Replace majority vote example with our own implementation
We need to use our MajorityVote model here https://rubrix.readthedocs.io/en/stable/guides/weak-supervision.html#4.-Using-the-weak-labels and likely simplify this part a little bit (metrics computation, etc.)

dcfidalgo:
Closes #1421 

This PR replaces snorkels majority voter with our implementation in the weak supervision guide. I also tried to simplify the "corrected accuracy" computation, or at least make it clearer.
https://github.com/recognai/rubrix/blob/2f63534b74093a4e0988a8a1cf57f75055b26d26/docs/guides/weak-supervision.ipynb


dvsrepo (approved):
: 💯

PR #1440, opened 2022-04-27
chore: using manual version system

frascuchon:
The automatic package `setuptools-scm` system is nice for continuous development but does not fit well if we have several branches and releases, since generated versions depend on which git local info has the builder 

So, the idea here is just to define the version manually. We can control the version control and use it in the development of new features (what can be effective for compatibility features)


setup.py
```diff
@@ -1,3 +1,7 @@
 from setuptools import setup
 
-setup(use_scm_version=True)
+version = {}
+with open("src/rubrix/_version.py") as fp:
+    exec(fp.read(), version)
+
+setup(use_scm_version=False, version=version["version"])
```
dcfidalgo:
```suggestion
setup(version=version["version"])
```

src/rubrix/_version.py
```diff
@@ -0,0 +1,5 @@
+# coding: utf-8
+# file generated by setuptools_scm
+# don't change, don't track in version control
```
dcfidalgo:
```suggestion
```

docs/conf.py
```diff
@@ -29,10 +29,6 @@
 # so we just want to display the major and minor version in the tab title of the browser.
 release = ".".join(rb.__version__.split(".")[0:2])
```
dcfidalgo: I think with the changes we should now follow the convention laid out here: https://www.sphinx-doc.org/en/master/usage/configuration.html
So `release` is the full project version with bugfix/prerelease tag, and `version` is just the major and minor digits.
```suggestion
release = rb.__version__
```

docs/conf.py
```diff
@@ -29,10 +29,6 @@
...
 
-# If on the master branch, set release and version to 'master'
-if "origin/master" in os.popen("git log -n 1 --oneline --decorate").read():
-    release = "master"
-
 # The short X.Y version
 version = release
```
dcfidalgo: See comment above
```suggestion
version = ".".join(release.split(".")[0:2])
```

dcfidalgo (dismissed):
Yes, I think this makes totally sense! :+1: Just a few inline suggestions.

docs/conf.py
```diff
@@ -27,14 +27,10 @@
 # Normally the full version, including alpha/beta/rc tags.
 # But since rtd changes some files during the build process, the scm version becomes dirty,
 # so we just want to display the major and minor version in the tab title of the browser.
```
dcfidalgo:
```suggestion
```

dcfidalgo (dismissed):

dcfidalgo (approved):

PR #1444, opened 2022-04-28
feat(#983): Add multi-label support to the Snorkel label model

### LINKED ISSUE #983: [Snorkel] Add support for multi-label text classification in our Snorkel label model
The idea here is that we train x models for 1vsAll, and store the weak labels in a tensor of (nr of examples * nr of rules * nr of models).
Implementing this for FlyingSquid is also on the road map.

dcfidalgo:
Closes #983


dcfidalgo:
I applied the multi-label Snorkel model to the examples of our [tutorial](https://github.com/recognai/rubrix/blob/bb804eadd30a94666a1165d046391a69ae431c75/docs/tutorials/weak-supervision-multi-label.ipynb), but it does not seem to improve the `MajorityVoter`, it's slightly worse. I have the feeling that it is something intrinsic, and not specific to the examples of the tutorial. In multi-label tasks, you don't have abstentions due to ties that Snorkel could resolve, so I'm not sure to what extent the correlations between the rules are useful.

I don't think it's worth adding this feature in the current state, it achieves slightly worse results compared to the MajorityVoter, but is much more complex since you have to do a grid search and find the right threshold. I would simply recommend always using the MajorityVoter for multi-label classification tasks.

CC @dvsrepo @frascuchon

dvsrepo:
Yes, I agree. Let's not include it.

frascuchon:
Thanks, @dcfidalgo!

It's just a multi-label behavior or just a generic fact that the snorkel model does not improve the majority vote model? Do we know the difference between snorkel and majority vote mode for single-label problems?

dcfidalgo:
> It's just a multi-label behavior or just a generic fact that the snorkel model does not improve the majority vote model? Do we know the difference between snorkel and majority vote mode for single-label problems?

This is a specific multi-label issue. For single-label tasks, you can get ties in the majority voter, which you can resolve if you take the correlations between the rules into account (that's my high-level understanding of snorkel). In our weak label guide, we show a small improvement when comparing Snorkel vs MajorityVoter, but the exact value highly depends on your rules and your dataset, and if you take as reference the label model or the downstream model.

PR #1471, opened 2022-05-09
feat: check dataset settings integrity

### LINKED ISSUE #1432: [Datasets] Allow configuring datasets with a predefined label schema

frascuchon:
After creating the dataset settings API in #1468, this PR includes:

- logging validation to preserve label schema when logging 
- settings validation when updating settings and the dataset already contains some records


dcfidalgo (approved):
:100: Do you have a feeling for the performance impact these validation checks have?

frascuchon:
Nope. We will do general tests before including it as part of the release.

PR #1480, opened 2022-05-10
fix(#1447): change agent when validating records with annotation but default status

### LINKED ISSUE #1447: [Annotation agent] Agent is not changed when validating records with annotation but `default` status
**Describe the bug**
I have not reproduced this myself, so this should be tested. It was raised yesterday during a session with one of our production users.

The workflow:

- They created text classification records with both annotation and prediction and status = Default.
- In this way, it's easy for the annotators to just validate those records that are already labeled correctly.
- Users validate records (they actually review ~3000 per week). 
- The issue is that only records for which they change the annotation get assigned the annotation_agent. For those `validated` by accepting the annotations, the `annotation_agent` is still the `machine_process` agent.

I think the proper way would be to change the annotation_agent even if no annotation was modified as the user is confirming they agree with the annotation so the record is manually annotated.

**To Reproduce**
Not possible on my side today

leiyre:
fix #1447

This PR overwrites the agent when validating records with annotation but default status

### DIFF
--- a/frontend/components/text-classifier/results/RecordTextClassification.vue
+++ b/frontend/components/text-classifier/results/RecordTextClassification.vue
@@ -86,13 +86,8 @@ export default {
       return this.dataset.viewSettings.viewMode === "labelling-rules";
     },
     allowValidate() {
-      const isBinary =
-        this.dataset.labels.length === 2 && !this.dataset.isMultiLabel;
       return (
         this.record.status !== "Validated" &&
-        (!isBinary ||
-          (isBinary && this.record.annotation) ||
-          this.record.prediction) &&
         (this.record.annotation ||
           this.record.prediction ||
           this.dataset.isMultiLabel)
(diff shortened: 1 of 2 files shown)


frontend/components/text-classifier/results/RecordTextClassification.vue
```diff
@@ -90,7 +90,9 @@ export default {
         this.dataset.labels.length === 2 && !this.dataset.isMultiLabel;
       return (
         this.record.status !== "Validated" &&
-        !isBinary &&
+        (!isBinary ||
+          (isBinary && this.record.annotation) ||
+          this.record.prediction) &&
```
frascuchon: Could we just remove these conditions? The cases where you can validate records using the `Validate` button should be:

1. has annotation  OR
2. has prediction OR
3. is a multi_label record

Those cases are covered by lines 96, 97 and 98

leiyre: absolutely right

frascuchon (approved):
YEAH!

PR #1487, opened 2022-05-12
feat(#1134): Allow extending the weak label matrix with embeddings

### LINKED ISSUE #1134: [Weak supervision] Extend FlyingSquid with Epoxy
[Epoxy](https://github.com/HazyResearch/epoxy) should be integrated into the existing FlyingSquid implementation / documentation.

### LINKED ISSUE #1309: Utilize GPT3 embedding / classification API for more automated bulk labelling
Keyword-based bulk labeling using Rubrix Rules is still too slow if I were to label thousands of texts for classification - at maximum, I can label 20 at a time since keyword match is noisy with regard to ground-truth label. 

I have played with GPT3 embeddings, where applying UMAP on texts results in clear clusters with semantically similar texts in clusters. If one can create Rubrix Rules instead based on a selected area on a scatterplot of text embeddings, then one can label hundreds of texts at once, and can easily finish labeling 50,000 data points in a day with high quality. And that'd be a game-changer.

Adding on to that, I also played with GPT3 classification API, where it gives pretty accurate few-shot classifications. And if that is added to the pipeline, it'd expedite labeling even further. Making it possible to label thousands of texts at once.

I have the GPT3 embedding + classification pipeline done in a notebook, is there plan on Rubrix's side to look into utilizing GPT3?

dcfidalgo:
Closes #1134 and #1309


dcfidalgo:
Ok, I will merge this tour de France when the tests pass.

A quick conclusion of this journey:
- We figured out why Ruan had problems in the beginning with the ag news dataset. Since the dataset is big, he computed the embeddings in colab, but in the order of the original dataset from the HF Hub. In the tutorial we upload the records without an id, so the order in which the weak label object retrieves the records is random. So there was a mismatch between the records and the embeddings, that's why he basically obtained a random classifier. I will open an issue with a proposal to mitigate this issue in the future.
- Optimizing the threshold is important, so it is not necessarily a cheap approach to improving performance. We could do some more experimentation to find reliable and cheap optimization methods.
- In general, when one wants to extend the weak label matrix, one should aim for rules with very high precision and not necessarily a high recall, since this is what the extension of the weak label matrix is for. So one has to keep this in mind right from the start.
- Also, the right embeddings are obviously important as well.

CC @dvsrepo @frascuchon

dcfidalgo:
Let's make the QA on the master branch, this PR has been open for far too long ...

frascuchon:
Well, it looks like we're finally getting this feature...

As improvement points:

1. The embeddings should be part of the records in the weak label matrix, instead of being passed as an argument in the extension. This can remove potential mismatch embedding errors.
```python
for record in wl.records():
   record.embedding = .....
```
So, the `extend_matrix` method will accept only the thresholds argument.

2. It would be nice to provide some default method or helper to optimize the thresholds that are stopped for the extension, given the importance of this parameter.

But it should be better to tackle them (if apply) in a different PR

dcfidalgo:
I just talked about his comments with @frascuchon 

> The embeddings should be part of the records in the weak label matrix, instead of being passed as an argument in the extension. This can remove potential mismatch embedding errors.

We think it would be very helpful if one could materialize/load a `WeakLabels` object. Then you can easily share your dataset + weak labels, load it on a machine with GPU to add embeddings, etc. 

> It would be nice to provide some default method or helper to optimize the thresholds that are stopped for the extension, given the importance of this parameter.

I think more tests with different datasets and rules would be necessary to find a reliable and practical (cheap) method ... would be a good project for a student :)

PR #1490, opened 2022-05-13
feat(#1472): hide scrollbar in scrollable components

### LINKED ISSUE #1472: [UI Scrollbars] hide scrollbar in scrollable components
The scrollbar is visible in some components (metrics) in all browsers. On the other hand, in some browsers (firefox) they are visible whenever there is scroll.

In general, we can unify and hide the scrollbar in scrollable components without affecting the behavior.

leiyre:
closes #1472

This PR hides the scrollbar in all browsers without affecting the behavior

### DIFF
--- a/frontend/assets/scss/abstract/placeholders/_placeholders.scss
+++ b/frontend/assets/scss/abstract/placeholders/_placeholders.scss
@@ -221,3 +221,12 @@
     }
   }
 }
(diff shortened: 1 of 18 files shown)


frascuchon (approved):
It should be nice if we can define a base UI component where apply the common styles (like `hide-scrollbar`)

PR #1493, opened 2022-05-13
ci: pyproject.toml defines the deps and extra deps

frascuchon:
This PR tries to make accessible the [dependency graph vision](https://github.com/recognai/rubrix/network/dependencies) from Github insights

After regarding the project setup, @dcfidalgo, do we need the `setup.cfg` file? I've seen that the `pyproject.toml` covers all attributes defined in `setup.cfg`


dcfidalgo:
> After regarding the project setup, @dcfidalgo, do we need the setup.cfg file? I've seen that the pyproject.toml covers all attributes defined in setup.cfg

Nope, if we move all the setuptools stuff to the `pyproject.toml` file, the `setup.cfg` file can be removed. They still announce the configuration via the pyproject file as [experimental](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html), but from my side, we can go ahead with this. In this case, we must require a version `setuptools>=61.0.0` in the `[build-system]` section.

pyproject.toml
```diff
@@ -2,6 +2,60 @@
...
+    "keywords",
+    "authors",
+    "maintainers",
+    "license",
+    "requires-python",
+    "version",
+    "description",
+    "readme",
+]
+
```
dcfidalgo:
```suggestion
```

pyproject.toml
```diff
@@ -2,6 +2,60 @@
 requires = ["setuptools", "wheel"]
```
dcfidalgo:
```suggestion
requires = ["setuptools>=61.0.0", "wheel"]
```

pyproject.toml
```diff
@@ -2,6 +2,60 @@
...
+dynamic = [
+    "keywords",
+    "authors",
+    "maintainers",
+    "license",
+    "requires-python",
+    "version",
+    "description",
+    "readme",
+]
```
dcfidalgo: These are not really dynamic and we could move them from the `setup.cfg` file to the `pyproject.toml` file.

frascuchon: I'm not sure if they are been included in the right place. Could you take a look, and do some local tests to verify that I've pushed is fine?

dcfidalgo (approved):

Issue #1494, opened 2022-05-13
Select path for opensearch ca


zelanastasia:
I'm trying to deploy rubrix in k8s and connect with external ES instance. Connection should be secure, but i got error

`opensearchpy.exceptions.SSLError: ConnectionError([SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1091)) caused by: SSLError([SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1091))`

This is because with opensearchpy you should select path in parametr `ca_certs` with `OpenSearch` function. Reference: https://opensearch.org/docs/latest/clients/python/#sample-code

It would be great if you add `ca_cert_path` variable to rubrix. #1495

frascuchon:
Hi @zelanastasia and thanks for using Rubrix.

I've seen also your PR, thanks a lot! But, first of all, I would like to better understand why you're getting that error. 

If you have a valid certificate configured to connect to your elasticsearch instance, this extra configuration should not be needed. If I'm not wrong, you're trying to connect to an es instance using the secured `HTTPS` protocol, but the certificate is not valid for the domain you're using. It's that right?

Did you try to disable the client SSL verification? You can use the `RUBRIX_ELASTICSEARCH_SSL_VERIFY=False` configuration for that.

Anyway, I will review your PR.

Thanks!

PR #1495, opened 2022-05-13
Add ca_cert path variable

zelanastasia:
Add variable for select path to ES certs
Reference: https://opensearch.org/docs/latest/clients/python/#sample-code 

**Closes** #1494

### DIFF
--- a/src/rubrix/server/commons/settings.py
+++ b/src/rubrix/server/commons/settings.py
@@ -60,6 +60,7 @@ class ApiSettings(BaseSettings):
 
     elasticsearch: str = "http://localhost:9200"
     elasticsearch_ssl_verify: bool = True
+    elasticsearch_ca_path: str
     cors_origins: List[str] = ["*"]
 
     docs_enabled: bool = True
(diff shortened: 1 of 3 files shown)


src/rubrix/server/commons/settings.py
```diff
@@ -60,6 +60,7 @@ class ApiSettings(BaseSettings):
 
     elasticsearch: str = "http://localhost:9200"
     elasticsearch_ssl_verify: bool = True
+    elasticsearch_ca_path: str
```
frascuchon:
```suggestion
    elasticsearch_ca_path: Optional[str] = None
```

frascuchon (changes requested):
Thanks @zelanastasia 

Just a change suggestion to make the parameter optional

It would be fantastic if you add an entry in the [Environment Variables section](https://github.com/zelanastasia/rubrix/blob/main/docs/getting_started/advanced_setup_guides.rst#environment-variables) from documentation.

frascuchon:
Great, @zelanastasia !!!

Thanks for your work!! I will close the PR and move your commits to another PR created from the master branch (the development branch). See #1502

PR #1504, opened 2022-05-17
feat(#1432): get predefined labels when label schema in configured

### LINKED ISSUE #1432: [Datasets] Allow configuring datasets with a predefined label schema

leiyre:
closes #1432
This PR initializes the dataset with the labels of the predefined schema when it is configured

### DIFF
--- a/frontend/components/text-classifier/sidebar/TextClassificationRules.vue
+++ b/frontend/components/text-classifier/sidebar/TextClassificationRules.vue
@@ -124,7 +124,7 @@ export default {
     labels() {
       const labelsInRules =
         this.dataset.rules.flatMap((rule) => rule.labels) || [];
-      const availableLabels = this.dataset._labels.map((label) => {
+      const availableLabels = this.dataset.labels.map((label) => {
         return {
           label: label,
           counter: labelsInRules.filter((l) => l === label).length || 0,
--- a/frontend/models/TextClassification.js
+++ b/frontend/models/TextClassification.js
@@ -78,6 +78,7 @@ class TextClassificationDataset extends ObservationDataset {
   static fields() {
     return {
       ...super.fields(),
+      settings: this.attr({}),
       _labels: this.attr([]),
       // Search fields
       query: this.attr({}, (data) => {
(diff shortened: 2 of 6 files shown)


frontend/models/TextClassification.js
```diff
@@ -125,6 +128,21 @@ class TextClassificationDataset extends ObservationDataset {
...
+          return status === 404 || (status >= 200 && status < 300);
+        },
+      }
+    );
+    if (response.status === 404) {
+      return undefined;
+    }
+    return response.data;
+  }
+
```
frascuchon: Since the API call is perfectly parameterized, could we move this method to the parent `ObservationDataset` class?

frontend/models/TextClassification.js
```diff
@@ -297,9 +317,10 @@ class TextClassificationDataset extends ObservationDataset {
           .concat(Object.keys(aggregations.predicted_as))
       ),
     ];
-
     uniqueLabels.sort();
-    return uniqueLabels;
+    const sortedPredefinedLabels =
+      predefinedLabels && predefinedLabels.map((l) => l.id).sort();
+    return sortedPredefinedLabels || uniqueLabels;
```
frascuchon: Once we have `predefinedLabels` (the labeling schema), we could apply the map and sort and return the results before computing the labels from metadata, aggregation,...bla blabla

```javascript
const predefinedLabels = this.settings.label_schema && this.settings.label_schema.labels;
if (predefinedLabels) {
   return predefinedLabels.map((l) => l.id).sort();
}
// here we do what we did before the changes.
const { labels } = (this.metadata || {})[USER_DATA_METADATA_KEY] || {};
....
```

frascuchon (changes requested):

frontend/components/text-classifier/sidebar/TextClassificationRules.vue
```diff
@@ -124,7 +124,7 @@ export default {
     labels() {
       const labelsInRules =
         this.dataset.rules.flatMap((rule) => rule.labels) || [];
-      const availableLabels = this.dataset._labels.map((label) => {
+      const availableLabels = this.dataset.labels.map((label) => {
```
frascuchon, line 127: Is this a bug? I mean, we should use the `dataset.labels` property always (even prior to this new feature)

leiyre: Yes, in this case `dataset._labels `was wrong

frascuchon (approved):

PR #1507, opened 2022-05-19
feat(client): rubrix listener

### LINKED ISSUE #1602: Rubrix listeners
A first implementation version that allows clients to react to dataset changes.

These listeners can be the base for other integration like alerts, active learning loops, model monitoring, and other MLOPs actions.

frascuchon:
The missing component for MLOPs:

Some examples of use:

```python
import time
from typing import List

import rubrix as rb

from rubrix import listener, RBListenerContext


@listener(
    dataset="other-dataset",
    query="status:Validated AND last_updated: [now-5m TO *]",  # we set a time-windows of 5 minutes
    condition=lambda query: 80 <= query.total <= 100,
)
def next_batch_iteration(records: List[rb.TokenClassificationRecord]):
    # Do something with the records
    pass


@listener(
    dataset="sst2_listener",
    query="status:Validated AND (NOT _exists_:metadata.pending_update)",
    condition=lambda query: query.total > 0,
)
def update_external_db(
    records: List[rb.TextClassificationRecord], ctx: RBListenerContext
):
    for record in records:
        # update a mongo db or something else
        record.metadata["pending_update"] = False
    rb.log(records, name=ctx.dataset)  # update received records


@listener(
    dataset="sst2_listener",
    metrics=["F1"],
    condition=lambda query, metrics: metrics.F1.f1_macro < 0.95,
    with_records=False,
    execution_interval_in_seconds=15,
)
def some_dataset_listener(
    ctx: RBListenerContext,
):  # include a cxt: RubrixListenerContext as param instead of name, query,...args
    print(f"The model is degrading the performance: {ctx.metrics.F1}")
    # send an alert
    # or launch a model re-training
    # or buy something in amazon

some_dataset_listener.start()
update_external_db.start()

try:
    while True:
        time.sleep(1)
except KeyboardInterrupt:
    some_dataset_listener.stop()
    update_external_db.stop()
```

Waiting for your API use feedback before implementing the tests @dvsrepo @dcfidalgo

### DIFF
--- a/environment_dev.yml
+++ b/environment_dev.yml
@@ -44,4 +44,4 @@ dependencies:
       - transformers[torch]~=4.18.0
       - loguru
       # install Rubrix in editable mode
-      - -e .[server]
+      - -e .[server,listeners]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,6 +66,10 @@ server = [
     "hurry.filesize", # TODO: remove
     "psutil ~= 5.8.0",
 ]
+listeners = [
+    "schedule ~= 1.1.0",
+    "prodict ~= 0.8.0"
+]
 
 [project.urls]
 homepage = "https://www.rubrix.ml"
--- a/src/rubrix/__init__.py
+++ b/src/rubrix/__init__.py
@@ -57,6 +57,7 @@ if _TYPE_CHECKING:
         TokenClassificationSettings,
         configure_dataset,
     )
+    from rubrix.listeners import Metrics, RBListenerContext, Search, listener
     from rubrix.monitoring.model_monitor import monitor
     from rubrix.server.server import app
 
@@ -85,6 +86,7 @@ _import_structure = {
         "read_pandas",
     ],
     "monitoring.model_monitor": ["monitor"],
+    "listeners.listener": ["listener", "RBListenerContext", "Search", "Metrics"],
     "datasets": [
         "configure_dataset",
         "TextClassificationSettings",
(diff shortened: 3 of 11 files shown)


dcfidalgo:
:100: very nice, looks good to me. One question about the implementation, the `MetricsAPI` and `Searches` API are only meant to be used by the listener? If not, maybe we can move them to another PR to keep things nicely separated.

Also, not sure if we could somehow automate the `with_records` argument, that is if the decorated function has only one parameter, do not pass the records along, if it has two arguments pass them along. Or make it dependent on the argument names ... just an idea.

frascuchon:
> 💯 very nice, looks good to me. One question about the implementation, the `MetricsAPI` and `Searches` API are only meant to be used by the listener? If not, maybe we can move them to another PR to keep things nicely separated.

For now, the Metrics and Searches APIs are used only by the listener module, but the idea is to move all related operations to these APIs, but this should be applied in different PRs 

> Also, not sure if we could somehow automate the `with_records` argument, that is if the decorated function has only one parameter, do not pass the records along, if it has two arguments pass them along. Or make it dependent on the argument names ... just an idea.

Nice feat. I'll investigate a bit. I will include it in this PR if it can be tackled easily. 

#### *Update

The action function can be an arbitrary number of arguments that can be passed on `listener.start`  method. So, including this magic could be complex. 

We can start like that and test the different use cases. Maybe, in the future, this param can be removed

frascuchon:
I'll merge this PR and start testing with some use cases.

¡

PR #1515, opened 2022-05-25
refactor(client): remove as_pandas parameter

### LINKED ISSUE #1360: [Client] Move to 'rb.DatasetFor*' by default

frascuchon:
Closes #1360

### DIFF
--- a/src/rubrix/client/api.py
+++ b/src/rubrix/client/api.py
@@ -16,12 +16,12 @@ import asyncio
 import logging
 import os
 import re
+import warnings
 from asyncio import Future
 from functools import wraps
 from inspect import signature
 from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple, Union
 
-import pandas
 from tqdm.auto import tqdm
 
 from rubrix._constants import (
@@ -76,7 +76,6 @@ from rubrix.client.sdk.users.models import User
 from rubrix.utils import setup_loop_in_thread
 
 _LOGGER = logging.getLogger(__name__)
-_WARNED_ABOUT_AS_PANDAS = False
 
 
 class _RubrixLogAgent:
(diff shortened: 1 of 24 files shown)


dcfidalgo:
@frascuchon Should we include the adaptations of the docs in this PR, or do this in a follow-up PR?

frascuchon:
Yes, I forget the docs. Do you have located how many docs are affected by this change?

dcfidalgo:
Not really, but I can take care of this in this PR.

dcfidalgo:
@frascuchon I updated the docu part. I also introduced a FutureWarning instead of completely removing the parameter in order not to break code. At the moment I think there some users that have `rb.load(..., as_pandas=False)` in their codebase, so I think we should notify them first, so that they can remove this argument, before breaking their code.

dcfidalgo (dismissed):

src/rubrix/client/api.py
```diff
@@ -401,26 +400,32 @@ def load(
...
-            >>> dataframe = rb.load(name="example-dataset")
             >>> dataset = rb.load(name="example-dataset")
         """
+        if as_pandas is not None:
+            warnings.warn(
+                "The argument `as_pandas` is deprecated and will be removed in a future version. "
+                "Please adapt your code accordingly. "
+                "If you want a pandas DataFrame do `rb.load(...).to_pandas()`."
+            )
+
```
frascuchon: This method will work if `as_pandas=False` but the user flow will fail when passing `as_pandas=True`. Also, the error can be confusing for the user.

Maybe, we should include a check here to raise a `ValueError` when `as_pandas=True` 

```suggestion
        if as_pandas is False:
            warnings.warn(
                "The argument `as_pandas` is deprecated and will be removed in a future version. "
                "Please adapt your code accordingly. "
            )
        elif as_pandas is True:
           raise ValueError(
                "If you want a pandas DataFrame do `rb.load(...).to_pandas()`."
            )

```

dcfidalgo: :+1: Done, I just made the error message a bit more explicit.

dcfidalgo (approved):

dcfidalgo:
From my side, this is good to go, merge it at will.

PR #1566, opened 2022-06-16
fix: compatibility with new dataset version

dcfidalgo:
This makes Rubrix compatible with the new datasets 2.3.0 version.

It is rather a dirty fix, I'll make a follow-up PR with a proper fix while refactoring/improving a bit the `Dataset*.from_datasets` mechanics.


dcfidalgo:
See #1567 for the proper fix.

frascuchon (approved):

PR #1567, opened 2022-06-16
Feat: Improve from datasets

dcfidalgo:
This PR improves a bit the `Dataset*.from_datasets` mechanics. Instead of running x-times over the entire dataset for parsing x columns, we do the parsing of all columns in one go over the dataset.

- improves performance
- makes the ignore columns message less confusing

It also provides a "proper" fix to support the new datasets 2.3.0 version (not the push_to_hub issue, but the one when renaming a column).

### DIFF
--- a/src/rubrix/client/datasets.py
+++ b/src/rubrix/client/datasets.py
@@ -65,6 +65,8 @@ class DatasetBase:
     """
 
     _RECORD_TYPE = None
+    # record arguments that can get multiple input columns from datasets.Dataset or pandas.DataFrame
+    _ARGS_WITH_MULTIPLE_INPUT_COLUMNS = ["inputs", "metadata"]
 
     @classmethod
     def _record_init_args(cls) -> List[str]:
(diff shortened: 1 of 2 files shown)


src/rubrix/client/datasets.py
```diff
@@ -153,88 +155,146 @@ def _to_datasets_dict(self) -> Dict:
...
 
-    @classmethod
-    def _from_datasets(cls, dataset: "datasets.Dataset") -> "Dataset":
-        """Helper method to create a Rubrix Dataset from a datasets Dataset.
+    @staticmethod
+    def _join_datasets_columns_and_delete(
+        row: Dict[str, Any], columns: List[str]
+    ) -> Dict[str, Any]:
+        """Joins columns of a `datasets.Dataset` row into a dict, and deletes the single columns.
+        UPDATES TE ROW!
```
frascuchon: ?

frascuchon (dismissed):

src/rubrix/client/datasets.py
```suggestion

        Updates the ``row`` dictionary!
        
```

src/rubrix/client/datasets.py
```diff
@@ -153,88 +155,148 @@ def _to_datasets_dict(self) -> Dict:
...
-        """Helper method to create a Rubrix Dataset from a datasets Dataset.
+    @staticmethod
+    def _join_datasets_columns_and_delete(
+        row: Dict[str, Any], columns: List[str]
+    ) -> Dict[str, Any]:
+        """Joins columns of a `datasets.Dataset` row into a dict, and deletes the single columns.
 
-        Must be implemented by the child class.
+        Updates the ``row`` dictionary!
+
```
```suggestion
```

src/rubrix/client/datasets.py
dcfidalgo: Typo, wanted to explicitly state that the method updates the ``row`` dictionary.

PR #1571, opened 2022-06-17
refactor: remove `words` references in searches

frascuchon:
This PR finish work related to the issue #945  

Changes included in this PR make default query text searches change a little bit since we start using only the `text` field as the default search field, where the standard analyzer is applied. Searches with special characters, like `!` or `.` are not covered by this analyzer.

So, we should include a warning message somewhere in the documentation to advise these changes.

@dcfidalgo @dvsrepo Any recommendation about how to tackle the communication?


dcfidalgo:
I will add an admonition to the query guide, something like:
*Before version ... the default is this, after version ... the default is that*

dcfidalgo:
There was already an admonition warning about a future change, so I updated it to describe the change in more detail.
From my side, this is good to be merged.

dcfidalgo (approved):

PR #1577, opened 2022-06-22
feat: Add 'extend_matrix' to the WeakMultiLabel class

### LINKED ISSUE #1556: [WeakMultiLabels] Merge with WeakLabels using a 2 dim wl matrix
The current implementation of the weak label matrix for multi-label classifications is overcomplicated (3-dim), and the MajorityVoter does not take advantage of the format. I propose to simplify this, basically treating rules with x labels as x rules, and remove the `WeakMultiLabels` class. 

Pros: reduce code complexity, get the `extend_matrix` functionality also for multi-label cases.
Cons: if at some point we find a label model that can take advantage of the 3-dim format, we must implement a method that transforms the 2-dim format.

dcfidalgo:
This PR adds the `extend_matrix` functionality to the `MultiClassLabel`.

### DIFF
--- a/src/rubrix/labeling/text_classification/weak_labels.py
+++ b/src/rubrix/labeling/text_classification/weak_labels.py
@@ -26,6 +26,14 @@ from rubrix.client.models import TextClassificationRecord
 from rubrix.labeling.text_classification.rule import Rule, load_rules
 
 
+def _add_docstr(string: str):
+    def docstring_decorator(fn):
+        fn.__doc__ = string
+        return fn
+
+    return docstring_decorator
+
+
 class WeakLabelsBase:
     """Base class for the weak label classes.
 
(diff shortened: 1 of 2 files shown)


src/rubrix/labeling/text_classification/weak_labels.py
```diff
@@ -770,6 +746,63 @@ def change_mapping(self, label2int: Dict[str, int]):
...
+
+        self._extended_matrix = self._extend_matrix(
+            thresholds, embeddings, abstains, support, gpu
+        )
+
+        # keep the original weak label row, for which at least on rule did not abstain
+        recs_with_votes = np.argwhere(
+            (self._matrix != self._label2int[None]).sum(-1) > 0
+        ).flatten()
+        self._extended_matrix[recs_with_votes] = self._matrix[recs_with_votes]
```
frascuchon: Should we apply this post-process to the multilabel wl matrix?

src/rubrix/labeling/text_classification/weak_labels.py
```diff
@@ -1088,6 +1125,57 @@ def show_records(
...
+        ]
+
+        abstains = [
+            np.argwhere(self._matrix[:, i].sum(-1) < 0).flatten()
+            for i in range(self._matrix.shape[1])
+        ]
+
+        self._extended_matrix = self._extend_matrix(
+            thresholds, embeddings, abstains, support, gpu
+        )
```
frascuchon: Since both, single and multi-label version of this method are almost the same, could we do some refactors to create a common base method with separate internal steps? 

Something like that:

```python
def extend_matrix(
        self,
        thresholds: Union[List[float], np.ndarray],
        embeddings: Optional[np.ndarray] = None,
        gpu: bool = False,
    ):
    
    support, abstain = self._extend_matrics_preprocess()
    extended_matrix = self._extend_matrix(
            thresholds, embeddings, abstains, support, GPU
        )
    self._extended_matrix = self._extend_matrix_postprocess(extended_matrix)
```

then, every weak label child class can implement its methods.

src/rubrix/labeling/text_classification/weak_labels.py
dcfidalgo: Hm, no, for the multi-label case this does not make much sense, since you do not create conflicts between labels by extending the rules. For example:

text = "This is a sad story and it makes me angry"

**single-label case**
Imagine a sentiment analysis (negative, positive) and you have a rule `if "sad" in text: return "negative"`; If you extend the matrix you probably want to avoid that a "positive" rule extends to this example since you already have an explicit rule that votes for this example (you basically give the explicit rules more weight).

**multi-label case**
Imaging a emotion detection task (sad, angry) and you have a rule `if "sad" in text: return "sad"`; If you extend the matrix you don't mind if an "angry" rule also extends to this example;

src/rubrix/labeling/text_classification/weak_labels.py
dcfidalgo: Done

src/rubrix/labeling/text_classification/weak_labels.py
frascuchon: mkay

frascuchon (approved):

PR #1584, opened 2022-06-27
prepare_for_training supports spacy DocBin

### LINKED ISSUE #420: Add example to Cookbook to export training data for spaCy v3
Similar to the following example:

https://rubrix.readthedocs.io/en/stable/guides/cookbook.html#Training

Add example to transform a Rubrix dataset into a spaCy Docbin format and save it to disk (usable with the train spacy command for training)

ignacioct:
Closes #420.

This is the first draft of the method, as we discussed in Slack

### DIFF
--- a/src/rubrix/client/datasets.py
+++ b/src/rubrix/client/datasets.py
@@ -640,8 +640,6 @@ class DatasetForTextClassification(DatasetBase):
             ds_dict, features=datasets.Features(feature_dict)
         )
 
-    # create a Enum-type Framework class like the `SortBy` one
-
 
 class Framework(Enum):
     TRANSFORMERS = "transformers"
--- a/src/rubrix/client/datasets.py
+++ b/src/rubrix/client/datasets.py
@@ -860,6 +860,8 @@ class DatasetForTokenClassification(DatasetBase):
 
             # Creating the docbin object as in https://spacy.io/usage/training#training-data
             for record in self._records:
+                if record.annotation is None:
+                   continue
                 doc = nlp(record.text)
                 ents = []
                 for anno in record.annotation:
--- a/src/rubrix/client/datasets.py
+++ b/src/rubrix/client/datasets.py
@@ -14,6 +14,7 @@
 #  limitations under the License.
 import functools
 import logging
+from enum import Enum
 from typing import Any, Dict, List, Optional, Tuple, Type, Union
 
 import pandas as pd
@@ -639,6 +640,19 @@ class DatasetForTextClassification(DatasetBase):
             ds_dict, features=datasets.Features(feature_dict)
         )
 
+    # create a Enum-type Framework class like the `SortBy` one
+
+
+class Framework(Enum):
+    TRANSFORMERS = "transformers"
+    SPACY = "spacy"
+
+    @classmethod
+    def _missing_(cls, value):
+        raise ValueError(
+            f"{value} is not a valid {cls.__name__}, please select one of {list(cls._value2member_map_.keys())}"
+        )
+
 
 @_prepend_docstring(TokenClassificationRecord)
 class DatasetForTokenClassification(DatasetBase):
@@ -764,7 +778,7 @@ class DatasetForTokenClassification(DatasetBase):
 
     @_requires_datasets
     def prepare_for_training(
-        self, framework="transformers"
+        self, framework: Union[Framework, str] = "transformers"
     ) -> Union["datasets.Dataset", "spacy.tokens.DocBin"]:
         """Prepares the dataset for training.
 
@@ -808,7 +822,12 @@ class DatasetForTokenClassification(DatasetBase):
 
 
         """
-        if framework == "transformers":
+
+        # turn the string into a Framework instance and trigger error if str is not valid
+        if isinstance(framework, str):
+            framework = Framework(framework)
+
+        if framework is Framework.TRANSFORMERS:
 
             import datasets
 
@@ -849,7 +868,7 @@ class DatasetForTokenClassification(DatasetBase):
 
             return ds.cast(new_features)
 
-        elif framework == "spacy":
+        elif framework is Framework.SPACY:
 
             import spacy
             from spacy.tokens import DocBin
--- a/src/rubrix/client/datasets.py
+++ b/src/rubrix/client/datasets.py
@@ -869,7 +869,11 @@ class DatasetForTokenClassification(DatasetBase):
                 ents = []
                 for anno in record.annotation:
                     span = doc.char_span(anno[1], anno[2], label=anno[0])
-                    ents.append(span)
+                    # There is a misalignment between record tokenization and spaCy tokenization
+                    if span is None: 
+                       print("Something wrong with record span and spaCy tokenization")
+                    else:
+                       ents.append(span)
 
                 doc.ents = ents
                 db.add(doc)
--- a/src/rubrix/client/datasets.py
+++ b/src/rubrix/client/datasets.py
@@ -808,7 +808,7 @@ class DatasetForTokenClassification(DatasetBase):
 
 
         """
-        if framework is "transformers":
+        if framework == "transformers":
 
             import datasets
 
--- a/src/rubrix/client/datasets.py
+++ b/src/rubrix/client/datasets.py
@@ -853,11 +853,6 @@ class DatasetForTokenClassification(DatasetBase):
             import spacy
             from spacy.tokens import DocBin
 
-            has_annotations = False
-            for rec in self._records:
-                if rec.annotation is not None:
-                    has_annotations = True
-                    break
 
             nlp = spacy.blank("es")
 
--- a/src/rubrix/client/datasets.py
+++ b/src/rubrix/client/datasets.py
@@ -869,11 +869,15 @@ class DatasetForTokenClassification(DatasetBase):
                 ents = []
                 for anno in record.annotation:
                     span = doc.char_span(anno[1], anno[2], label=anno[0])
-                    # There is a misalignment between record tokenization and spaCy tokenization
-                    if span is None: 
-                       print("Something wrong with record span and spaCy tokenization")
+
+                    # There is a misalignment between record tokenization and spaCy tokenization
+                    if span is None:
+                        print(
+                            "Error between the record span and spaCy tokenization: there is a misalignment between the record and spaCy's tokenization."
+                        )
+
                     else:
-                       ents.append(span)
+                        ents.append(span)
 
                 doc.ents = ents
                 db.add(doc)
--- a/src/rubrix/client/datasets.py
+++ b/src/rubrix/client/datasets.py
@@ -764,7 +764,7 @@ class DatasetForTokenClassification(DatasetBase):
 
     @_requires_datasets
     def prepare_for_training(
-        self, framework: None
+        self, framework="transformers"
     ) -> Union["datasets.Dataset", "spacy.tokens.DocBin"]:
         """Prepares the dataset for training.
 
@@ -775,7 +775,8 @@ class DatasetForTokenClassification(DatasetBase):
             - The iob tags are transformed to integers.
 
         Args:
-            framework: A string specifying the framework of use. Default: `None`
+            framework: A string specifying the framework of use.
+                "transformers" and "spacy" are currently supported. Default: `transformers`
 
         Returns:
             A datasets Dataset with a *ner_tags* column and all columns returned by ``to_datasets``.
@@ -807,7 +808,7 @@ class DatasetForTokenClassification(DatasetBase):
 
 
         """
-        if framework is None:
+        if framework is "transformers":
 
             import datasets
 
@@ -880,6 +881,11 @@ class DatasetForTokenClassification(DatasetBase):
                 db.add(doc)
 
             return db
+
+        else:
+            print(
+                "The introduced framework does not exist or is not supported. Please, use one of the existing frameworks detailed on the documentation."
+            )
 
     def __all_labels__(self):
         all_labels = set()
--- a/src/rubrix/client/datasets.py
+++ b/src/rubrix/client/datasets.py
@@ -18,7 +18,6 @@ from typing import Any, Dict, List, Optional, Tuple, Type, Union
 
 import pandas as pd
 from pkg_resources import parse_version
-from spacy.tokens import DocBin
 
 from rubrix.client.models import (
     Record,
@@ -766,7 +765,7 @@ class DatasetForTokenClassification(DatasetBase):
     @_requires_datasets
     def prepare_for_training(
         self, framework: None
-    ) -> Union["datasets.Dataset", spacy.tokens.DocBin]:
+    ) -> Union["datasets.Dataset", "spacy.tokens.DocBin"]:
         """Prepares the dataset for training.
 
         This will return a ``datasets.Dataset`` with all columns returned by ``to_datasets`` method
@@ -851,10 +850,10 @@ class DatasetForTokenClassification(DatasetBase):
 
         elif framework == "spacy":
 
-            # Importing locally, handling errors when the modules are not found.
             import spacy
+            from spacy.tokens import DocBin
 
-            nlp = spacy.blank("en")
+            nlp = spacy.blank()
 
             db = DocBin()
 
--- a/src/rubrix/client/datasets.py
+++ b/src/rubrix/client/datasets.py
@@ -17,7 +17,6 @@ import logging
 from typing import Any, Dict, List, Optional, Tuple, Type, Union
 
 import pandas as pd
-import spacy
 from pkg_resources import parse_version
 from spacy.tokens import DocBin
 
@@ -851,6 +850,9 @@ class DatasetForTokenClassification(DatasetBase):
             return ds.cast(new_features)
 
         elif framework == "spacy":
+
+            # Importing locally, handling errors when the modules are not found.
+            import spacy
 
             nlp = spacy.blank("en")
 
--- a/src/rubrix/client/datasets.py
+++ b/src/rubrix/client/datasets.py
@@ -853,7 +853,6 @@ class DatasetForTokenClassification(DatasetBase):
             import spacy
             from spacy.tokens import DocBin
 
-
             nlp = spacy.blank("es")
 
             db = DocBin()
@@ -861,9 +860,10 @@ class DatasetForTokenClassification(DatasetBase):
             # Creating the docbin object as in https://spacy.io/usage/training#training-data
             for record in self._records:
                 if record.annotation is None:
-                   continue
+                    continue
                 doc = nlp(record.text)
                 ents = []
+
                 for anno in record.annotation:
                     span = doc.char_span(anno[1], anno[2], label=anno[0])
 
--- a/src/rubrix/client/datasets.py
+++ b/src/rubrix/client/datasets.py
@@ -853,7 +853,13 @@ class DatasetForTokenClassification(DatasetBase):
             import spacy
             from spacy.tokens import DocBin
 
-            nlp = spacy.blank()
+            has_annotations = False
+            for rec in self._records:
+                if rec.annotation is not None:
+                    has_annotations = True
+                    break
+
+            nlp = spacy.blank("es")
 
             db = DocBin()
 
--- a/src/rubrix/client/datasets.py
+++ b/src/rubrix/client/datasets.py
@@ -17,7 +17,9 @@ import logging
 from typing import Any, Dict, List, Optional, Tuple, Type, Union
 
 import pandas as pd
+import spacy
 from pkg_resources import parse_version
+from spacy.tokens import DocBin
 
 from rubrix.client.models import (
     Record,
@@ -763,7 +765,9 @@ class DatasetForTokenClassification(DatasetBase):
         return super().from_pandas(dataframe)
 
     @_requires_datasets
-    def prepare_for_training(self) -> "datasets.Dataset":
+    def prepare_for_training(
+        self, framework: None
+    ) -> Union["datasets.Dataset", spacy.tokens.DocBin]:
         """Prepares the dataset for training.
 
         This will return a ``datasets.Dataset`` with all columns returned by ``to_datasets`` method
@@ -771,6 +775,9 @@ class DatasetForTokenClassification(DatasetBase):
             - Records without an annotation are removed.
             - The *ner_tags* column corresponds to the iob tags sequences for annotations of the records
             - The iob tags are transformed to integers.
+
+        Args:
+            framework: A string specifying the framework of use. Default: `None`
 
         Returns:
             A datasets Dataset with a *ner_tags* column and all columns returned by ``to_datasets``.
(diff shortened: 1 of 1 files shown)


src/rubrix/client/datasets.py
```diff
@@ -802,44 +809,65 @@ def prepare_for_training(self) -> "datasets.Dataset":
...
+                .map(lambda example: {"ner_tags": spans2iob(example)})
+            )
+            new_features = ds.features.copy()
+            new_features["ner_tags"] = [class_tags]
+
+            return ds.cast(new_features)
+
+        elif framework == "spacy":
+
+            nlp = spacy.blank("en")
```
frascuchon: Is the language relevant for the DocBin generation? Using `en` always should be enough?

src/rubrix/client/datasets.py
```diff
@@ -802,44 +809,65 @@ def prepare_for_training(self) -> "datasets.Dataset":
...
+
+            db = DocBin()
+
+            # Creating the docbin object as in https://spacy.io/usage/training#training-data
+            for record in self._records:
+                doc = nlp(record.text)
+                ents = []
+                for anno in record.annotation:
```
frascuchon, line 885: As for the HF version, we should work only with annotated records. See `__only_annotations__` filter.

src/rubrix/client/datasets.py
```diff
@@ -802,44 +809,65 @@ def prepare_for_training(self) -> "datasets.Dataset":
...
+            new_features = ds.features.copy()
+            new_features["ner_tags"] = [class_tags]
+
+            return ds.cast(new_features)
+
+        elif framework == "spacy":
+
+            nlp = spacy.blank("en")
+
+            db = DocBin()
```
frascuchon, line 876: Can the `attrs`  init argument be left with the default value? See the [spaCy docs](https://spacy.io/api/docbin#init)

src/rubrix/client/datasets.py
```diff
@@ -17,7 +17,9 @@
 from typing import Any, Dict, List, Optional, Tuple, Type, Union
 
 import pandas as pd
+import spacy
```
frascuchon: The optional spacy modules must be imported locally, handling errors when the modules are not found.

frascuchon (commented):
Hi @ignacioct, and thanks for your work!!!

I've left some comments, but the way is good!!!

src/rubrix/client/datasets.py
ignacioct: Do we have any information for filling up the `attrs`? I've checked the attributes list but I don't know if we can extract some information from our dataset to not leave that empty.

src/rubrix/client/datasets.py
ignacioct: Apparently, if I take out import spacy, even if from spacy.tokens import DocBin is still there, I got an error with the type hinting of the function:

```
(rubrix) ➜  rubrix git:(spacy_docbin) ✗ /Users/ignacio/opt/anaconda3/envs/rubrix/bin/python /Users/ignacio/Documents/Recognai/rubrix/t
esting_docbin.py
  File "/Users/ignacio/Documents/Recognai/rubrix/src/rubrix/utils.py", line 111, in _get_module
    return importlib.import_module("." + module_name, self.__name__)
  File "/Users/ignacio/opt/anaconda3/envs/rubrix/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 961, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 843, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/Users/ignacio/Documents/Recognai/rubrix/src/rubrix/client/__init__.py", line 16, in <module>
    from .rubrix_client import RubrixClient
  File "/Users/ignacio/Documents/Recognai/rubrix/src/rubrix/client/rubrix_client.py", line 29, in <module>
    from rubrix.client.datasets import (
  File "/Users/ignacio/Documents/Recognai/rubrix/src/rubrix/client/datasets.py", line 645, in <module>
    class DatasetForTokenClassification(DatasetBase):
  File "/Users/ignacio/Documents/Recognai/rubrix/src/rubrix/client/datasets.py", line 769, in DatasetForTokenClassification
    ) -> Union["datasets.Dataset", spacy.tokens.DocBin]:
NameError: name 'spacy' is not defined

The above exception was the direct cause of the following exception:

  File "/Users/ignacio/Documents/Recognai/rubrix/testing_docbin.py", line 4, in <module>
    dataset = rb.DatasetForTokenClassification([
  File "/Users/ignacio/Documents/Recognai/rubrix/src/rubrix/utils.py", line 81, in __getattr__
    module = self._get_module(self._class_to_module[name])
  File "/Users/ignacio/Documents/Recognai/rubrix/src/rubrix/utils.py", line 113, in _get_module
    raise RuntimeError(
RuntimeError: Failed to import rubrix.client.datasets because of the following error (look up to see its traceback):
name 'spacy' is not defined
```

src/rubrix/client/datasets.py
ignacioct: With the HF version do you mean what it is now under `if framework is None`. I didn't know it was the Hugginface framework, I can name it that way and make it default. 

Then, should we only proceed with the method if there are annotations in the records? In both cases or only in HuggingFace? I think I didn't understand it properly.

src/rubrix/client/datasets.py
ignacioct: The language cannot be left blank, it requires a language code. Leaving it blank results in an error.

frascuchon: Sorry, what I try to say is what happens when the dataset contains text in other language than English. Do it work in that case or should provide a way to setup the language?

src/rubrix/client/datasets.py
frascuchon: Yes. The training data should be generated only with records containing annotation. So, the record list should be filtered in that way as the default path

src/rubrix/client/datasets.py
frascuchon: Nop. I have no idea about that. Just asking you if you have a suggestion. We can keep the defaults for now

src/rubrix/client/datasets.py
frascuchon: You should replace the line:

```python
Union["datasets.Dataset", spacy.tokens.DocBin]
```
by 
```python
Union["datasets.Dataset", "spacy.tokens.DocBin"]
```
Including the `"` should fix your problem.

src/rubrix/client/datasets.py
ignacioct: I've printed all the arguments in docbins over the same record using English and Spanish, and they seem to be the same. If we wanna include the language, we would have to pass it as metadata, which requires changing the records, right? Maybe keeping initially` en` is sufficient, as it is only generating tokens with words, but an optional piece of metadata covering the language could be useful in this case.

Screenshot 2022-06-28 at 19 06 20

Screenshot 2022-06-28 at 19 06 45

src/rubrix/client/datasets.py
ignacioct: I also have no idea about it, but from my point of view, there is nothing that we can directly relate to. Maybe open an issue to look into further?

src/rubrix/client/datasets.py
ignacioct: I have added this, taken from the HF version

```
has_annotations = False
            for rec in self._records:
                if rec.annotation is not None:
                    has_annotations = True
                    break
```

ignacioct:
@frascuchon Should I change `if framework is None:` for `if framework == "hf": ` and make HF default?

src/rubrix/client/datasets.py
```diff
@@ -802,44 +807,74 @@ def prepare_for_training(self) -> "datasets.Dataset":
...
+        elif framework == "spacy":
+
+            import spacy
+            from spacy.tokens import DocBin
+
+            has_annotations = False
+            for rec in self._records:
+                if rec.annotation is not None:
+                    has_annotations = True
+                    break
```
frascuchon: These lines of code have no effect on this path. We can remove them and try to parse only record with annotation info.

```suggestion
```

src/rubrix/client/datasets.py
frascuchon: I've left some suggestion and comments

src/rubrix/client/datasets.py
```diff
@@ -802,44 +807,74 @@ def prepare_for_training(self) -> "datasets.Dataset":
...
+
+            nlp = spacy.blank("es")
+
+            db = DocBin()
+
+            # Creating the docbin object as in https://spacy.io/usage/training#training-data
+            for record in self._records:
+                doc = nlp(record.text)
+                ents = []
+                for anno in record.annotation:
```
frascuchon:
```suggestion
            for record in self._records:
                if record.annotation is None:
                   continue
                doc = nlp(record.text)
                ents = []
                for anno in record.annotation:
```

src/rubrix/client/datasets.py
```diff
@@ -802,44 +807,74 @@ def prepare_for_training(self) -> "datasets.Dataset":
...
+                    span = doc.char_span(anno[1], anno[2], label=anno[0])
+                    ents.append(span)
```
frascuchon: Since we're building the spaCy doc without regarding the record tokenization (rubrix records contain info about text tokenization, and annotation spans are based on that tokenization), we should check if the resulting spaCy span is still valid (not`None`). If not, we should raise a warning flag, and discard it from the entities list.
```suggestion
                    span = doc.char_span(anno[1], anno[2], label=anno[0])
                    # There is a misalignment between record tokenization and spaCy tokenization
                    if span is None: 
                       print("Something wrong with record span and spaCy tokenization")
                    else:
                       ents.append(span)
```

src/rubrix/client/datasets.py
frascuchon: > @frascuchon Should I change `if framework is None:` for `if framework == "hf": ` and make HF default?

Not sure if `hf` is quite descriptive,`transformers`? `huggingface`?  Maybe @dcfidalgo or @dvsrepo can answer this.

frascuchon:
> @frascuchon Should I change `if framework is None:` for `if framework == "hf": ` and make HF default?

Not sure if `hf` is quite descriptive,`transformers`? `huggingface`?  Maybe @dcfidalgo or @dvsrepo can answer this.

dcfidalgo:
> > @frascuchon Should I change `if framework is None:` for `if framework == "hf": ` and make HF default?
> 
> Not sure if `hf` is quite descriptive,`transformers`? `huggingface`? Maybe @dcfidalgo or @dvsrepo can answer this.

I would vote for `transformers`, I think it's the most concise description.

ignacioct:
> > > @frascuchon Should I change `if framework is None:` for `if framework == "hf": ` and make HF default?
> > 
> > 
> > Not sure if `hf` is quite descriptive,`transformers`? `huggingface`? Maybe @dcfidalgo or @dvsrepo can answer this.
> 
> I would vote for `transformers`, I think it's the most concise description.

If added "transformers" as the main case.

src/rubrix/client/datasets.py
```diff
@@ -763,7 +763,9 @@ def from_pandas(
         return super().from_pandas(dataframe)
 
     @_requires_datasets
```
dcfidalgo, line 777: We should remove this decorator here, since if you choose the `spacy` framework you do not need the *datasets* package. 

I would create two helper methods:

- `_prepare_for_training_with_spacy()`
- `_prepare_for_training_with_transformers()`

and decorate the latter one with `_requires_datasets`, and create a similar `_requires_spacy` for the former.
So the `prepare_for_training` method just does the value comparison of the `framework` argument and calls the corresponding helper method.

src/rubrix/client/datasets.py
```diff
@@ -772,6 +774,10 @@ def prepare_for_training(self) -> "datasets.Dataset":
             - The *ner_tags* column corresponds to the iob tags sequences for annotations of the records
             - The iob tags are transformed to integers.
 
+        Args:
+            framework: A string specifying the framework of use.
```
```suggestion
            framework: A string specifying the framework for the training.
```

src/rubrix/client/datasets.py
```diff
@@ -763,7 +763,9 @@ def from_pandas(
         return super().from_pandas(dataframe)
 
     @_requires_datasets
-    def prepare_for_training(self) -> "datasets.Dataset":
+    def prepare_for_training(
+        self, framework="transformers"
```
dcfidalgo: Here I would use an `Enum` type and move the error message of not supported frameworks there. See the `SortBy` class for an example of this pattern: 
https://github.com/recognai/rubrix/blob/fdc851106b204683b4e144fc238267b8e3f8be48/src/rubrix/labeling/text_classification/label_errors.py#L28 

and how to use it as an argument later on:  
https://github.com/recognai/rubrix/blob/fdc851106b204683b4e144fc238267b8e3f8be48/src/rubrix/labeling/text_classification/label_errors.py#L44

src/rubrix/client/datasets.py
```diff
@@ -802,44 +808,84 @@ def prepare_for_training(self) -> "datasets.Dataset":
...
+
+            return ds.cast(new_features)
 
-        return ds.cast(new_features)
+        elif framework == "spacy":
+
+            import spacy
+            from spacy.tokens import DocBin
+
+            nlp = spacy.blank("es")
```
dcfidalgo, line 874: I think ideally the language should be an argument to the `prepare_for_training` method to get the right `Vocab` (which affects the lexemes). But since I don't know if this could be important for the training or not, I would leave it for a second iteration and choose as default the `en` language. Simply because it is the most common one.

dcfidalgo (changes requested):
Hi @ignacioct Thanks a lot for this! I left a few comments in addition to the ones by @frascuchon . Sorry, it took me a while to review this :pray:

src/rubrix/client/datasets.py
dvsrepo: Would it be difficult to add the language as an argument to the `prepare_for_training` method? I think this is an important aspect of this feature.

frascuchon: Do make sense for a more general argument to select the framework model? for example `en_core_news_sm` for spaCy or a hf model URI for hf transformers.

This could help by loading the model tokenizer and providing as result the tokenized text according to the model.

src/rubrix/client/datasets.py
```diff
@@ -802,44 +808,84 @@ def prepare_for_training(self) -> "datasets.Dataset":
 
 
         """
-        import datasets
+        if framework is "transformers":
```
dcfidalgo: One minor thing, you normally use `==` for string comparisons, since you normally test for equality (value) rather than identity (type): https://stackoverflow.com/a/1504848/16121687

src/rubrix/client/datasets.py
dcfidalgo: Regarding the HF tokenizer, not sure if I would go down this rabbit hole. There are a lot of options you can pass on to the tokenizer and its call method (truncation, padding, what to return, ...). I feel like this would need a bit more thought to cover all use cases and to have a more or less clean signature.

 Regarding the spacy language: i was reading a bit and different languages use different tokenizers
 ```python
 nlp_en, nlp_es = spacy.blank("en"), spacy.blank("es")
 list(nlp_en("don't"))  # [do, n't]
 list(nlp_es("don't"))  # [don't]
 ```
 So we definitely should include a `lang` argument in the method.

dvsrepo: I agree with @dcfidalgo on this wrt hugging face tokenizers, these should be an external step done by the users.

For spaCy it seems we do need at least the language (if we are to use blank) or the model_name. For the model name I am not sure, we need to understand what minimal components are needed/used for the docbin creation, if there are no trainable components we should be good with blank and the language code.

frascuchon: Regarding future features, could the language be a potential candidate to be included as part of record/dataset info?

Anyway, I assume we can leverage most of the use cases by including the lang as an argument. 

It will take effect for other than spaCy frameworks?

frascuchon: Regarding the options you can tune for the HF tokenizer, I think the same can be applied to the spaCy pipeline. You would like to customize your spacy tokenizer before generating the training data.

dvsrepo: Yes, you are right. 

I think that for spaCy we should pass the pipeline as arg instead of the `str` with the lang name or model name:

```python
def blabla( nlp: Language = spacy.blank("en")):

```

For Hugging Face they do not have a specialized format for training data so it's a bit different, and I don't have a clear view of how to deal with the relationship between the tokenizer and the ents defined in Rubrix

dvsrepo: The other thing is if we need to use the Language object at all. We already have the tokenized text, the clear text, and the spans, or does this conflict with the recent issue about whitespaces? Maybe better discuss this over a call.

src/rubrix/client/datasets.py
ignacioct: That´s right, it is a typo on my side, nice catch!

src/rubrix/client/datasets.py
dcfidalgo: @frascuchon and I had a quick call. The `Language` model would not be necessary for the tokenization, we have all the necessary information in the Rubrix records. However, the `DocBin` also saves information about the lemmas, for example, and this information we do not have. The question is, how important are these "extra features" for the training with the `DocBin` ...

So for now we concluded that we should leave the user with the responsibility/freedom to choose the right spacy model for the `DocBin` creation (as suggested by @dvsrepo). So the signature should look something like this:

```python
def prepare_for_training(framework="transformers", lang: Optional["spacy.Language"] = None) -> ...:
    """...
    
    Args:
        framework: ...
        lang: The spacy language model used for creating the `DocBin`. It should be the same one used for creating your Rubrix records. 
            IGNORED FOR ``framework != "spacy"``
    """
```

we also considered having two methods:
 - `prepare_for_training_with_spacy`
 - `prepare_for_training_with_transformers`
 
But this seemed a bit verbose to us...

dvsrepo: 👍

src/rubrix/client/datasets.py
ignacioct: Hmm, I think I'm not sure what you mean. Where should I move the error message if it's not in the function? In the SortBy class there is a method that raises a not found error, so it would be in two places now.

src/rubrix/client/datasets.py
ignacioct: So, if the lang is None, we use `nlp = spacy.blank("en")`? Also, are we forcing the user to use exactly the language code Spacy supports? en for English, es for Spanish, etc.

Having two methods also seems a little bit verbose for me, even though I am not a particular fan of passing a string parameter. I can proceed on this matter in both ways. 

Sorry for the late response, by the way. I hope you are all doing fine in that heat wave, especially the Madrilians. I'm going home this Friday, so I'll have my share as well 🥶

src/rubrix/client/datasets.py
dcfidalgo: Arg sorry, maybe I was a bit too quick here. What I meant is something like this

```python
# create a Enum-type Framework class like the `SortBy` one
class Framework(Enum):
    TRANSFORMERS = "transformers"
    ...
    
# then use it in the method:
def prepare_for_training(self, framework: Union[Framework, str] = "transformers"):
    # turn str into a Framework instance and trigger error if str is not valid
    if isinstance(framework, str):   
        framework = Framework(framework)
    
    if framework is Framework.TRANSFORMERS:
        ...
```

src/rubrix/client/datasets.py
dcfidalgo: The idea here is more or less following workflow:
```python
import spacy
nlp = spacy.blank("en")

my_text = "This is just a 'don't do it' example"
tokens = list(map(str, nlp(my_text)))
rec = rb.TokenClassificationRecord(text=my_text, tokens=tokens, annotation=[("do", 16, 18)])
ds = rb.DatasetForTokenClassification([rec])

docbin = ds.prepare_for_training(framework="spacy", lang=nlp)
```
Maybe, for now, I would omit a default value for `lang` to invite the user to think a bit about this argument, what do you think @frascuchon @dvsrepo

src/rubrix/client/datasets.py
ignacioct: Okay, now I get it! Thanks. I've tried it and it works fine, but as it's the first time I do something similar, maybe you can have a quick look before solving this.

src/rubrix/client/datasets.py
```diff
@@ -639,6 +640,19 @@ def prepare_for_training(self) -> "datasets.Dataset":
             ds_dict, features=datasets.Features(feature_dict)
         )
 
+    # create a Enum-type Framework class like the `SortBy` one
+
```
dcfidalgo: We can remove this line
```suggestion
```

src/rubrix/client/datasets.py
dcfidalgo: :ok_hand: perfect!

src/rubrix/client/datasets.py
frascuchon: > Maybe, for now, I would omit a default value for lang to invite the user to think a bit about this argument, what do you think @frascuchon @dvsrepo

I'm not really conscious about the complete flow, but it has the sense to me since setting a default language could raise some training troubles for text language other than English.

frascuchon:
Hi @ignacioct. I hope you're doing well and have a swimming pool near for your holidays. You really will need it.

Sorry. We're "dizzing" you a little bit with all these suggestions/API redefinitions. So I'll try to finish the `prepare_for_trainig` enhancement. Then, you can work on the cookbook examples using it, if that's fit for you

frascuchon:
@ignacioct I had to create a separate PR #1635 since I cannot push the changes on your forked repository.

I think we can go ahead once the PR is approved and merged

ignacioct:
> Hi @ignacioct. I hope you're doing well and have a swimming pool near for your holidays. You really will need it.
> 
> Sorry. We're "dizzing" you a little bit with all these suggestions/API redefinitions. So I'll try to finish the `prepare_for_trainig` enhancement. Then, you can work on the cookbook examples using it, if that's fit for you

Yeah, sounds perfect! I can start working on that once the PR is merged.

frascuchon:
> > Hi @ignacioct. I hope you're doing well and have a swimming pool near for your holidays. You really will need it.
> > Sorry. We're "dizzing" you a little bit with all these suggestions/API redefinitions. So I'll try to finish the `prepare_for_trainig` enhancement. Then, you can work on the cookbook examples using it, if that's fit for you
> 
> Yeah, sounds perfect! I can start working on that once the PR is merged.

Hi, @ignacioct !!! The PR is closed. Tell me if you need something to go ahead.

Thanks!

PR #1588, opened 2022-06-27
fix: typo corrected on cookbook

### LINKED ISSUE #1587: Typo on the Training section of the cookbook
**Describe the bug**
There is a typo in https://rubrix.readthedocs.io/en/stable/guides/cookbook.html#Training. It says `Take into account that a zeroshot_example is contain annotations.` instead of `Take into account that a zeroshot_example contains annotations.`. I´m also pretty sure that I wrote that type around a year ago. I will add a pull request to solve it.

ignacioct:
Closes #1587

Typo corrected


dcfidalgo (approved):
Thanks a lot @ignacioct !!

PR #1620, opened 2022-07-19
fix: Show predicted OK/KO when predictions exist

### LINKED ISSUE #1619: [PREDICTED OK/KO] Predicted KO when there is annotation and no prediction
**Describe the bug**
The predicted KO is displayed when there is annotation and no prediction in the text classifier explore mode

**To Reproduce**
1. Go to any single label Text Classifier dataset with annotation and without prediction
2. Go to exploration mode
3. The icon of the predicted KO is displayed

**Expected behavior**
Predicted OK/KO must be shown only in records with annotation and prediction

**Environment:**
 - OS [e.g. iOS]:
 - Rubrix Version [ 0.16.0]:

leiyre:
this PR prevents the OK/KO from being displayed when there is any prediction

Closes #1619

### DIFF
--- a/frontend/models/TextClassification.js
+++ b/frontend/models/TextClassification.js
@@ -56,7 +56,12 @@ class TextClassificationRecord extends BaseRecord {
   }
 
   get predicted() {
-    if (!this.multi_label && this.predicted_as && this.annotated_as) {
+    if (
+      !this.multi_label &&
+      this.prediction &&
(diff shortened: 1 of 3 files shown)


frascuchon (commented):
Could we include a snapshot test case with both conditions (record with and without predictions)?

frascuchon (approved):

PR #1625, opened 2022-07-20
feat: Copy Record Text

### LINKED ISSUE #1616: [Feature request] Copy text option
I'm always frustrated when I want to copy record text but I am not able to. Sometimes I want to copy text and check out it"s dependency parser on another tool but currently there is no straightforward way to do so.

A copy option in the ":" next to "View Metadata" would be great.

1. Alternative can be a copy icon next to ":" but that won"t be neat as copy is a peripheral activity
2. Inspect result item and copy text from developer tools. (Overkill)

leiyre:
closes #1616

This PR adds the "copy text" action in the extra actions menu, it also centralizes the "copy to clipboard" function and makes some small changes to the text

### DIFF
--- a/frontend/components/commons/results/RecordExtraActions.vue
+++ b/frontend/components/commons/results/RecordExtraActions.vue
@@ -17,29 +17,25 @@
 
 <template>
   <div v-click-outside="close" :key="open" class="record__extra-actions">
-    <a
-      v-if="hasMetadata || allowChangeStatus"
-      class="extra-actions__button"
-      href="#"
-      @click.prevent="open = !open"
+    <a class="extra-actions__button" href="#" @click.prevent="open = !open"
       ><svgicon name="kebab-menu" width="20" height="20" color="#4A4A4A"
     /></a>
     <div v-if="open" class="extra-actions__content">
       <div v-if="hasMetadata" @click="showMetadata()">
         <span>View metadata</span>
       </div>
-      <template v-if="allowChangeStatus">
-        <div
-          v-for="status in allowedStatusActions"
-          :key="status.key"
-          :class="record.status === 'Discarded' ? 'disabled' : null"
-          @click="onChangeRecordStatus(status.key)"
-        >
-          <span>{{
-            record.status === "Discarded" ? "Discarded" : status.name
-          }}</span>
+      <re-action-tooltip tooltip="Copied">
+        <div @click="copyToClipboard()">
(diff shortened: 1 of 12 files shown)


frascuchon:
Just a comment, regarding the changelog discussion, this should be typed as `style` instead of `feat`, isn't it @leiyre @Amelie-V ??

frontend/components/commons/results/RecordExtraActions.vue
```diff
@@ -126,6 +113,16 @@ export default {
     close() {
       this.open = false;
     },
+    copyToClipboard() {
+      if (isObject(this.record.inputs)) {
+        const text = Object.keys(this.record.inputs).map((key) => {
+          return `${key}\r\n${this.record.inputs[key]}`;
+        });
```
frascuchon: For text classification, the copied text does not keep the same layout that is shown in the record:

Screenshot 2022-07-21 at 11 29 00

In the example below, the field names should be UPPERCASED (in my opinion) and every field should include a break line (maybe two) before the previous field (see the `headline` field in the image)

Also, I think we could skip the field name for cases when just one field (text) is found in the record. But it should be discussed

cc: @Amelie-V @dvsrepo @leiyre ???

frontend/components/commons/results/RecordExtraActions.vue
```diff
@@ -126,6 +113,16 @@ export default {
...
+        return this.$copyToClipboard(text);
+      } else {
+        return this.$copyToClipboard(this.record.text);
```
frascuchon: Suggesting a code refactoring, we could move the text generation logic to each record class, keeping the `copyToClipboard` method simpler. Something like:

```javascript
copyToClipboard() {
   return this.$copyToClipboard(this.record.clipboardText); // or some other representative name 
}
```

Then each Record class should implement its own logic to generate the `clipboardText` property. For example, in the case of text classification, you can include the object parsing and all the stuff.

frascuchon (changes requested):

frontend/components/commons/results/RecordExtraActions.vue
leiyre: I agree with skip the name from the field when there is only one

frascuchon (approved):
It's ok from my side.

I think @Amelie-V should also take a look in a live dev environment before merge it

PR #1627, opened 2022-07-20
feat(Listeners): allow using query params in the condition through search parameter

### LINKED ISSUE #1622: Active learning example requires that the initial sample size and query size match exactly
TL;DR: Small problem which causes the first batch to not be displayed, but is likely easy to fix.

**Describe the bug**

In the [active learning with small-text example](https://github.com/recognai/rubrix/blob/b6239fd474694cf4fe0c5f91a45cfcdb8d9f9ec7/docs/tutorials/active_learning_with_small_text.ipynb) I used a larger size of initial samples compared to the original version. Nothing happens in this case, i.e. I don't see a first batch in the UI. 

*Details:*
In the notebook we use the same size (`NUM_SAMPLES`) for initialization...

>
num initial samples

...and for querying:
num query sizes

For this reason the listener does not fire, since the condition checks for exactly `NUM_SAMPLES`-many samples:
`condition=lambda search: search.total==NUM_SAMPLES`.

*Suggestion:*
From the active learning perspective, it should not be a requirement that the initial sample size and query size must match. Consequently, there are three cases:

1. initial sample size > query_size
2. initial sample size == query_size
3. initial sample size < query_size

The second case already works. By changing the condition to `>=` I solved my specific problem (first case) as well. Now we still have to handle the third case. Honestly, I don't know how to fix this elegantly. I could treat the first batch separately, but that would also bloat the code. *Is there an additional clause we could add to the condition?*

**To Reproduce**
Navigate to the [active learning with small-text example](https://github.com/recognai/rubrix/blob/b6239fd474694cf4fe0c5f91a45cfcdb8d9f9ec7/docs/tutorials/active_learning_with_small_text.ipynb) and increase the initial number of samples (first picture), e.g. to 100.

**Expected behavior**
The initial data should be recognized as complete and the first active learning batch (batch-0) should be created and loaded into rubrix.

**Screenshots**
none

 - Rubrix Version 0.17.0.dev0
 - ElasticSearch Version 7.10.2

**Additional context**
none

frascuchon:
In this PR the search query parameters (if any) are visible inside the condition method context. So you use the dynamic query parameters as part of your listener condition:

```python
def condition(search):
   return search.query_params["my-query-param"] == "cool" and search.total > 10
```

@chschroeder It would be nice if you can do some tests with changes in this PR, just to validate the solution. Please, let me know if it's okay for you.

Closes #1622

### DIFF
--- a/src/rubrix/listeners/listener.py
+++ b/src/rubrix/listeners/listener.py
@@ -1,3 +1,4 @@
+import copy
 import dataclasses
 import logging
 import threading
@@ -158,7 +159,9 @@ class RBDatasetListener:
             name=self.dataset, task=dataset.task, query=self.formatted_query, size=0
         )
 
-        ctx.search = Search(total=search_results.total)
+        ctx.search = Search(
+            total=search_results.total, params=copy.deepcopy(ctx.query_params)
+        )
         condition_args = [ctx.search]
         if self.metrics:
             condition_args.append(ctx.metrics)
--- a/src/rubrix/listeners/listener.py
+++ b/src/rubrix/listeners/listener.py
@@ -1,5 +1,6 @@
 import copy
 import dataclasses
+import functools
 import logging
 import threading
 import time
@@ -77,6 +78,23 @@ class RBDatasetListener:
         """True if listener is running"""
         return self.__listener_job__ is not None
 
+    def __catch_exceptions__(self, cancel_on_failure=False):
+        def catch_exceptions_decorator(job_func):
+            @functools.wraps(job_func)
+            def wrapper(*args, **kwargs):
+                try:
+                    return job_func(*args, **kwargs)
+                except:
+                    import traceback
+
+                    print(traceback.format_exc())
+                    if cancel_on_failure:
+                        self.stop()  # We stop the scheduler
+
+            return wrapper
+
+        return catch_exceptions_decorator
+
     def start(self, *action_args, **action_kwargs):
         """
         Start listen to changes in the dataset. Additionally, args and kwargs can be passed to action
@@ -88,9 +106,13 @@ class RBDatasetListener:
         if self.is_running():
             raise ValueError("Listener is already running")
 
+        job_step = self.__catch_exceptions__(cancel_on_failure=True)(
+            self.__listener_iteration_job__
+        )
+
         self.__listener_job__ = self.__scheduler__.every(
             self.interval_in_seconds
-        ).seconds.do(self.__listener_iteration_job__, *action_args, **action_kwargs)
+        ).seconds.do(job_step, *action_args, **action_kwargs)
 
         class _ScheduleThread(threading.Thread):
             _WAIT_EVENT = threading.Event()
@@ -160,7 +182,7 @@ class RBDatasetListener:
         )
 
         ctx.search = Search(
-            total=search_results.total, params=copy.deepcopy(ctx.query_params)
+            total=search_results.total, query_params=copy.deepcopy(ctx.query_params)
         )
         condition_args = [ctx.search]
         if self.metrics:
--- a/src/rubrix/listeners/models.py
+++ b/src/rubrix/listeners/models.py
@@ -13,9 +13,11 @@ class Search:
 
     Args:
         total: The total number of records affected by the listener query
+        params: The query parameters applied to the executed search
     """
 
     total: int
+    params: Optional[Dict[str, Any]] = None
 
 
 class Metrics(Prodict):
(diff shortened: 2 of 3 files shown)


chschroeder:
In your proposed solution in #1622 the batch id was exposed:
```
def al_loop_step_completed(search, ctx):
   if ctx.query_params["batch_id"] == 0:
      return search.total == INIT_SAMPLES
   return search.total == NUM_SAMPLES
```

But now I don't have the batch id anymore. Your example above you suggest I can use arbitrary search params instead if I understand that correctly? But where would I set those?

frascuchon:
Sorry @chschroeder, I was a bit laconic in my explanation.

Regarding the  listener defined in #1622, the whole configuration will be:

```python
def al_loop_step_completed(search):
   if search.params["batch_id"] == 0:
      return search.total == INIT_SAMPLES
   return search.total == NUM_SAMPLES

@listener(
    dataset=DATASET_NAME,
    query="status:Validated AND metadata.batch_id:{batch_id}",
    condition=al_loop_step_completed,
    execution_interval_in_seconds=3,
    batch_id=0
)
...
```

Instead of passing the whole listener context as a condition argument (like the proposal I wrote), we expose only the query params through the search object in a read-only way. So, the query parameter `batch_id` will be accessible inside the condition function context as `search.params["batch_id"]`. Maybe we could normalize the attribute name and call it `query_params` like the listener context.

frascuchon:
> Instead of passing the whole listener context as a condition argument (like the proposal I wrote), we expose only the query params through the search object in a read-only way. So, the query parameter batch_id will be accessible inside the condition function context as search.params["batch_id"]. Maybe we could normalize the attribute name and call it query_params like the listener context.

I've renamed the `params` to `query_params` to align with the listener context attribute. @chschroeder you should use `search.query_params["batch_id"]` instead of `search.params["batch_id"]`

The code finally looks like this:

```python
def al_loop_step_completed(search):
   if search.query_params["batch_id"] == 0:
      return search.total == INIT_SAMPLES
   return search.total == NUM_SAMPLES

@listener(
    dataset=DATASET_NAME,
    query="status:Validated AND metadata.batch_id:{batch_id}",
    condition=al_loop_step_completed,
    execution_interval_in_seconds=3,
    batch_id=0
)
...
```

chschroeder:
I tried your example from the issue of course, but I did not have the `params` attribute. Because of that (and the change in your example) I wondered if I was using it as intended. After your additional input it was clear that the problem was in front of the screen. I must have messed up installing from the branch and had an older source.

Anyway, I briefly tested the two relevant cases (initial_size < query_size and initial_size > query_size) with my existing setup and this time both of them worked. Thank you!

src/rubrix/listeners/listener.py
```diff
@@ -158,7 +159,9 @@ def __listener_iteration_job__(self, *args, **kwargs):
             name=self.dataset, task=dataset.task, query=self.formatted_query, size=0
         )
 
-        ctx.search = Search(total=search_results.total)
+        ctx.search = Search(
+            total=search_results.total, params=copy.deepcopy(ctx.query_params)
```
dcfidalgo: Shouldn't this be:
```suggestion
            total=search_results.total, query_params=copy.deepcopy(ctx.query_params)
```
?? Not sure why the tests pass...

dcfidalgo (dismissed):
Looks good from my side. Maybe I would directly expose the `RBListenerContext` in the condition callable, in this way, we would avoid having this awkward `ctx.search.query_params == ctx.query_params`, but it's a minor thing.
And I have an inline comment, but I don't understand why the tests pass, so maybe I am missing something.

src/rubrix/listeners/listener.py
frascuchon: Me either. I'll check that. Thanks !

frascuchon:
> Looks good from my side. Maybe I would directly expose the `RBListenerContext` in the condition callable, in this way, we would avoid having this awkward `ctx.search.query_params == ctx.query_params`, but it's a minor thing.

I thought about that but the change implies some other changes over the listener definition in tutorials that didn't seem convenient. 

Another thing that made me choose this approach was the fact that the condition shouldn't change the query parameters. 

Passing the whole context, we cannot resolve that situation without messy behavior (Can I change parameters of the same object in the action context but not in the condition context?! wat?). 

So I decided to expose them in a different place to reinforce the fact that those parameters don't modify the listener context.

src/rubrix/listeners/listener.py
frascuchon: Error handling related errors. Now, the listener will stop if any unhandled error occurs. Please, take a look at the changes @dcfidalgo

dcfidalgo (approved):

PR #1635, opened 2022-07-21
feat: prepare_for_training supports spacy

frascuchon:
This PR brings the works started by @ignacioct in #1584 adding support to spaCy for the `prepare_for_training` method.

Refs #420 and #1584

### DIFF
--- a/src/rubrix/client/datasets.py
+++ b/src/rubrix/client/datasets.py
@@ -14,6 +14,7 @@
 #  limitations under the License.
 import functools
 import logging
+from enum import Enum
 from typing import Any, Dict, List, Optional, Tuple, Type, Union
 
 import pandas as pd
@@ -640,6 +641,17 @@ class DatasetForTextClassification(DatasetBase):
         )
 
 
+class Framework(Enum):
+    TRANSFORMERS = "transformers"
+    SPACY = "spacy"
+
+    @classmethod
+    def _missing_(cls, value):
+        raise ValueError(
+            f"{value} is not a valid {cls.__name__}, please select one of {list(cls._value2member_map_.keys())}"
+        )
+
+
 @_prepend_docstring(TokenClassificationRecord)
 class DatasetForTokenClassification(DatasetBase):
     """
@@ -763,7 +775,9 @@ class DatasetForTokenClassification(DatasetBase):
         return super().from_pandas(dataframe)
 
     @_requires_datasets
-    def prepare_for_training(self) -> "datasets.Dataset":
+    def prepare_for_training(
+        self, framework: Union[Framework, str] = "transformers"
+    ) -> Union["datasets.Dataset", "spacy.tokens.DocBin"]:
         """Prepares the dataset for training.
 
         This will return a ``datasets.Dataset`` with all columns returned by ``to_datasets`` method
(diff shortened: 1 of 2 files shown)


src/rubrix/client/datasets.py
```diff
@@ -51,6 +52,21 @@ def check_if_datasets_installed(*args, **kwargs):
     return check_if_datasets_installed
 
 
+def _requires_spacy(func):
+    @functools.wraps(func)
+    def check_if_datasets_installed(*args, **kwargs):
```
```suggestion
    def check_if_spacy_installed(*args, **kwargs):
```

src/rubrix/client/datasets.py
```diff
@@ -51,6 +52,21 @@ def check_if_datasets_installed(*args, **kwargs):
...
+        try:
+            import spacy
+        except ModuleNotFoundError:
+            raise ModuleNotFoundError(
+                f"'spacy' must be installed to use `{func.__name__}`"
+                "You can install 'spacy' with the command: `pip install spacy`"
+            )
+        return func(*args, **kwargs)
+
+    return check_if_datasets_installed
```
```suggestion
    return check_if_spacy_installed
```

src/rubrix/client/datasets.py
```diff
@@ -802,6 +839,22 @@ def prepare_for_training(self) -> "datasets.Dataset":
...
+        # turn the string into a Framework instance and trigger error if str is not valid
+        if isinstance(framework, str):
+            framework = Framework(framework)
+
+        if framework is Framework.TRANSFORMERS:
+            return self._prepare_for_training_with_transformers()
+        # else: must be spacy for sure
+        if lang is None:
+            raise ValueError(
+                "You must provided a valid spacy language pipeline"
```
```suggestion
                "Please provide a spacy language model to prepare the dataset for training with the spacy framework."
```

src/rubrix/client/datasets.py
```diff
@@ -802,6 +839,22 @@ def prepare_for_training(self) -> "datasets.Dataset":
...
+            )  # TODO(@dcfidalgo): review text inplace
```
```suggestion
            )
```

src/rubrix/client/datasets.py
```diff
@@ -841,6 +894,40 @@ def spans2iob(example):
...
+                continue
+
+            doc = nlp(record.text)
+            entities = []
+
+            for anno in record.annotation:
+                span = doc.char_span(anno[1], anno[2], label=anno[0])
+                # There is a misalignment between record tokenization and spaCy tokenization
+                if span is None:
+                    # TODO(@dcfidalgo): Do we want to warn and continue or should we stop the training set generation?
```
dcfidalgo, line 918: I would raise an exception here since, in this case, the user should likely revise the tokenization (the spacy language model provided). Maybe in a second iteration, we could think about adding an argument that turns this exception into a warning

src/rubrix/client/datasets.py
```diff
@@ -841,6 +894,40 @@ def spans2iob(example):
...
+                    print(
+                        "Error between the record span and spaCy tokenization: "
+                        "there is a misalignment between the record and spaCy's tokenization."
+                    )
```
```suggestion
                    raise ValueError(
                        "The following annotation does not align with the tokens produced "
                        f"by the provided spacy language model: {(anno[0], record.text[anno[1]:anno[2]])}, {list(doc)}"
                    )
```

dcfidalgo (commented):
:100: Added a few inline suggestions.

src/rubrix/client/datasets.py
frascuchon: OK

PR #1643, opened 2022-07-26
Documentation improvements

maxserras:
Elasticsearch persistence how-to added to the docker compose installation guide.
Warning added that the current version might be used with "docker compose" (without  "-")
Minor typo fixes

### DIFF
--- a/docs/getting_started/advanced_setup_guides.md
+++ b/docs/getting_started/advanced_setup_guides.md
@@ -168,8 +168,8 @@ you will lose all your datasets in Rubrix!
 :::
 
 ### Persisting ElasticSearch data
-To avoid losing all the data when the docker-compose/server goes down, you can add some persistence by mounting a local
-volume in docker compose and granting the required permissions. 
+To avoid losing all the data when the docker-compose/server goes down, you can add some persistence by mounting a 
+volume in the docker compose. 
 
 To this end, **modify the elasticsearch service and create a new volume** in the docker-compse.yml file:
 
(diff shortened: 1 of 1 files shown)


docs/getting_started/advanced_setup_guides.md
```diff
@@ -157,9 +164,60 @@ Keep in mind, if you execute
...
+# At the end of the file create a volume for ElasticSearch
+volumes:
+  elasticdata:
+
+
+```
+
+Then create the elasticdata/ directory and add the required permissions:
+```bash
+mkdir elasticdata && sudo chown -R 1000:1000 elasticdata/
```
frascuchon: This line could fail for some Linux distributions, and also for macOS environments.

Could we skip the permission part and let it as a tip after in the documentation?

frascuchon (commented):
Nice work!! Thanks a lot!

Just I've included a little comment about folder permissions. Please, take a look if it is fine for you.

docs/getting_started/advanced_setup_guides.md
```diff
@@ -157,9 +164,60 @@ Keep in mind, if you execute
 ```bash
 docker-compose down
 ```
-you will loose all your datasets in Rubrix!
+you will lose all your datasets in Rubrix!
 :::
 
+### Persisting ElasticSearch data
+To avoid losing all the data when the docker-compose/server goes down, you can add some persistence by mounting a local
+volume in docker compose and granting the required permissions.
```
frascuchon: Now, we could remove references to local volume and required permissions, right?
```suggestion
To avoid losing all the data when the docker-compose/server goes down, you can add some persistence by mounting a 
volume in the docker compose. 
```

maxserras: Yeah, agree :D

frascuchon (approved):
👍

Issue #1275, opened 2022-03-18
Docker-compose and Elasticsearch disk space


mausch:
Hi, just started evaluating this project today and hit some issues.

When running Rubrix via docker-compose ( https://rubrix.readthedocs.io/en/stable/getting_started/advanced_setup_guides.html#launching-the-web-app-via-docker-compose ) which is usually used for a quick evaluation.
the docker-compose file only defines memory settings for Elasticsearch.

Problem is, Elasticsearch is by default pretty conservative about free disk space. By default it will refuse to write anything if there's less than 5% disk space or so (don't remember the exact number).

What's worse, the Rubrix client library only reports the following:

```
rubrix.client.sdk.commons.errors.GenericApiError: Rubrix server returned an error with http status: 500
Error details: [{'code': 'opensearchpy.exceptions.TransportError', 'params': {}}]
```

but when checking the server logs you see the full error:

```
opensearchpy.exceptions.TransportError: TransportError(429, 'cluster_block_exception', 'index [.rubrix.datasets-v0] blocked by: [TOO_MANY_REQUESTS/12/disk usage exceeded flood-stage watermark, index has read-only-allow-delete block];')
```

So:
1. The client library should show the *full* error message
2. The docker compose file should raise the default Elasticsearch thresholds for disk space

PR #1648, opened 2022-07-28
feat(Error Logging): send error detail in response for generic server errors

frascuchon:
This PR applies changes to send an error context from the server side for generic server errors. This change will more info to the client about what's happening.

@dvsrepo I think this PR will like you.

Refs: #1275


dvsrepo (dismissed):
🏩

PR #1659, opened 2022-08-04
Feat/search after

maxserras:
Added the "from_id" parameter for the rb.load() method, the search and stream  endpoints in FastAPI. 

Now all the results from rb.load and the search over the datasets are sorted by record identifier. 

This feature allows to iterate over large datasets (+1M samples) without the need of loading everything in one go or do hacky stuff. Also, this makes easier to integrate with dataloaders to create batches for training models without the need of loading the whole dataset.

### DIFF
--- a/src/rubrix/client/sdk/text2text/api.py
+++ b/src/rubrix/client/sdk/text2text/api.py
@@ -31,14 +31,13 @@ def data(
     limit: Optional[int] = None,
     id_from: Optional[str] = None,
 ) -> Response[Union[List[Text2TextRecord], HTTPValidationError, ErrorMessage]]:
-    url = "{}/api/datasets/{name}/Text2Text/data".format(client.base_url, name=name)
+
+    path = f"/api/datasets/{name}/Text2Text/data"
     params = build_param_dict(id_from, limit)
-    with httpx.stream(
+
+    with client.stream(
         method="POST",
-        url=url,
-        headers=client.get_headers(),
-        cookies=client.get_cookies(),
-        timeout=None,
+        path=path,
         params=params if params else None,
         json=request.dict() if request else {},
     ) as response:
--- a/src/rubrix/client/sdk/text2text/api.py
+++ b/src/rubrix/client/sdk/text2text/api.py
@@ -21,7 +21,7 @@ from rubrix.client.sdk.commons.api import build_data_response
 from rubrix.client.sdk.commons.models import ErrorMessage, HTTPValidationError, Response
 from rubrix.client.sdk.text2text.models import Text2TextQuery, Text2TextRecord
 
-from src.rubrix.client.sdk.commons.api import build_param_dict
+from rubrix.client.sdk.commons.api import build_param_dict
 
 
 def data(
--- a/src/rubrix/client/sdk/text2text/api.py
+++ b/src/rubrix/client/sdk/text2text/api.py
@@ -21,22 +21,25 @@ from rubrix.client.sdk.commons.api import build_data_response
 from rubrix.client.sdk.commons.models import ErrorMessage, HTTPValidationError, Response
 from rubrix.client.sdk.text2text.models import Text2TextQuery, Text2TextRecord
 
+from src.rubrix.client.sdk.commons.api import build_param_dict
+
 
 def data(
     client: AuthenticatedClient,
     name: str,
     request: Optional[Text2TextQuery] = None,
     limit: Optional[int] = None,
+    id_from: Optional[str] = None,
 ) -> Response[Union[List[Text2TextRecord], HTTPValidationError, ErrorMessage]]:
     url = "{}/api/datasets/{name}/Text2Text/data".format(client.base_url, name=name)
-
+    params = build_param_dict(id_from, limit)
     with httpx.stream(
         method="POST",
         url=url,
         headers=client.get_headers(),
         cookies=client.get_cookies(),
         timeout=None,
-        params={"limit": limit} if limit else None,
+        params=params if params else None,
         json=request.dict() if request else {},
     ) as response:
         return build_data_response(response=response, data_type=Text2TextRecord)
--- a/src/rubrix/client/sdk/text_classification/api.py
+++ b/src/rubrix/client/sdk/text_classification/api.py
@@ -27,7 +27,7 @@ from rubrix.client.sdk.text_classification.models import (
     TextClassificationRecord,
 )
 
-from src.rubrix.client.sdk.commons.api import build_param_dict
+from rubrix.client.sdk.commons.api import build_param_dict
 
 
 def data(
--- a/src/rubrix/client/sdk/text_classification/api.py
+++ b/src/rubrix/client/sdk/text_classification/api.py
@@ -37,16 +37,13 @@ def data(
     limit: Optional[int] = None,
     id_from: Optional[str] = None,
 ) -> Response[Union[List[TextClassificationRecord], HTTPValidationError, ErrorMessage]]:
-    url = "{}/api/datasets/{name}/TextClassification/data".format(
-        client.base_url, name=name
-    )
+
+    path = f"/api/datasets/{name}/TextClassification/data"
     params = build_param_dict(id_from, limit)
-    with httpx.stream(
+    
+    with client.stream(
         method="POST",
-        url=url,
-        headers=client.get_headers(),
-        cookies=client.get_cookies(),
-        timeout=None,
+        path=path,
         params=params if params else None,
         json=request.dict() if request else {},
     ) as response:
--- a/src/rubrix/client/sdk/text_classification/api.py
+++ b/src/rubrix/client/sdk/text_classification/api.py
@@ -12,7 +12,7 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
-from typing import List, Optional, Union
+from typing import List, Optional, Union, Dict
 
 import httpx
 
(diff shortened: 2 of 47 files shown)


src/rubrix/client/sdk/text2text/api.py
```diff
@@ -21,22 +21,25 @@
 from rubrix.client.sdk.commons.models import ErrorMessage, HTTPValidationError, Response
 from rubrix.client.sdk.text2text.models import Text2TextQuery, Text2TextRecord
 
+from src.rubrix.client.sdk.commons.api import build_param_dict
```
frascuchon:
```suggestion
from rubrix.client.sdk.commons.api import build_param_dict
```

src/rubrix/client/sdk/text_classification/api.py
```diff
@@ -27,24 +27,27 @@
     TextClassificationRecord,
 )
 
+from src.rubrix.client.sdk.commons.api import build_param_dict
```
frascuchon:
```suggestion
from rubrix.client.sdk.commons.api import build_param_dict
```

src/rubrix/client/sdk/token_classification/api.py
```diff
@@ -25,26 +25,29 @@
     TokenClassificationRecord,
 )
 
+from src.rubrix.client.sdk.commons.api import build_param_dict
```
frascuchon:
```suggestion
from rubrix.client.sdk.commons.api import build_param_dict
```

src/rubrix/server/daos/records.py
```diff
@@ -287,10 +303,19 @@ def scan_dataset(
...
+            # Scroll method does not accept read_after, thus, this case is handled as a search
+            es_query["search_after"] = [id_from]
+            results = self._es.search(index=dataset_records_index(dataset.id), query=es_query, size=limit)
+            hits = results["hits"]
+            docs = hits["hits"]
+
+        else:
+            docs = self._es.list_documents(
+                dataset_records_index(dataset.id), query=es_query,
+            )
```
frascuchon, line 308: Could we use the `search` method always for dataset loading? That means that the `rb.load` will fetch records always in batches, and can improve the client's overall performance. 

We avoid using index scrolling, what is a recommendation from the [official docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/scroll-api.html)

src/rubrix/server/apis/v0/handlers/text2text.py
```diff
@@ -130,6 +130,7 @@ def bulk_records(
 )
 def search_records(
     name: str,
+    id_from: Optional[str] = None,
```
frascuchon: Since it doesn't affect the loading of datasets with `rb.load`, I prefer to leave this parameter out of the search for now. 

The motivation is that we should solve the problem when searching with a sort provided to the API. With the proposed implementation, provided sorting has no effect and that worries me. 

It is an issue that can be resolved, but it is not easy. That's why I prefer to approach it independently.

src/rubrix/server/apis/v0/handlers/text_classification.py
```diff
@@ -146,6 +146,7 @@ async def bulk_records(
 )
 def search_records(
     name: str,
+    id_from: Optional[str] = None,
```
frascuchon: Ditto

src/rubrix/server/apis/v0/handlers/token_classification.py
```diff
@@ -141,6 +141,7 @@ async def bulk_records(
 )
 def search_records(
     name: str,
+    id_from: Optional[str] = None,
```
frascuchon: Ditto

src/rubrix/server/apis/v0/models/text2text.py
```diff
@@ -215,6 +215,7 @@ class Text2TextSearchRequest(BaseModel):
 
     query: Text2TextQuery = Field(default_factory=Text2TextQuery)
     sort: List[SortableField] = Field(default_factory=list)
+    id_from: Optional[str] = None
```
frascuchon: I think we can drop this field, since we're passing the `id_from` as id param instead of a body attibute

src/rubrix/server/apis/v0/models/text_classification.py
```diff
@@ -531,6 +531,7 @@ class TextClassificationSearchRequest(BaseModel):
 
     query: TextClassificationQuery = Field(default_factory=TextClassificationQuery)
     sort: List[SortableField] = Field(default_factory=list)
+    id_from: Optional[str] = None
```
frascuchon: Ditto

src/rubrix/server/apis/v0/models/token_classification.py
```diff
@@ -425,6 +425,7 @@ class TokenClassificationSearchRequest(BaseModel):
 
     query: TokenClassificationQuery = Field(default_factory=TokenClassificationQuery)
     sort: List[SortableField] = Field(default_factory=list)
+    id_from: Optional[str] = None
```
frascuchon: Ditto

src/rubrix/server/services/search/service.py
```diff
@@ -60,6 +60,7 @@ def search(
         size: int = 100,
         exclude_metrics: bool = True,
         metrics: Optional[List[BaseMetric]] = None,
+        id_from: Optional[str] = None
```
frascuchon: As I said before, we can wait to apply changes to the search

src/rubrix/server/services/text2text.py
```diff
@@ -86,6 +86,7 @@ def search(
         sort_by: List[SortableField],
         record_from: int = 0,
         size: int = 100,
+        id_from: Optional[str] = None,
```
frascuchon: Ditto

src/rubrix/server/services/text_classification.py
```diff
@@ -97,6 +97,7 @@ def search(
         sort_by: List[SortableField],
         record_from: int = 0,
         size: int = 100,
+        id_from: Optional[str] = None,
```
frascuchon: Ditto

frascuchon (commented):
It looks great to me. Just a few comments and some reflections about changes on the search endpoint.

src/rubrix/server/elasticseach/client_wrapper.py
```diff
@@ -102,7 +102,8 @@ def list_documents(
         A sequence of documents resulting from applying the query on the index
 
         """
-        return es_scan(self.__client__, query=query or {}, index=index)
+        # Preserve order is used to ensure that the samples are sorted by identifier
+        return es_scan(self.__client__, query=query or {}, index=index, preserve_order=True)
```
frascuchon: From elasticsearch client documentation:

> preserve_order – don’t set the search_type to scan - this will cause the scroll to paginate with preserving the order. Note that this can be an extremely expensive operation and can easily lead to unpredictable results, use with caution.

I'm a bit unquiet about the performance impact caused by activating the `preserve_order` flag. If I understand well, this must be set up to fetch the proper last id for the next batch in `rb.load`, right? But, when no limit is applied, we will return the whole dataset, so no need for order preservation, right?

So, for those cases where a limit is provided, I suggest using the search method instead, sorted by the `id` field. Anyway, the better solution is to implement an alternative helper to `es_scan` using automatically the `search_after` attribute. Let me think about a solution that could fit in this PR

frascuchon: I've been doing some tests and I think this should generally work. We avoid using scroll and use the `search_after` mechanism to iterate the whole dataset:

```suggestion
        size = size or 1000
        query = query.copy() or {}
        query["sort"] = [{"_id": {"order": "asc"}}] # Force sorting by id
        response = self.__client__.search(index=index, body=query, size=size)
        while response["hits"]["hits"]:
            for hit in response["hits"]["hits"]:
                yield hit

            last_id = hit["_id"]
            query["search_after"] = [last_id]
            response = self.__client__.search(index=index, body=query, size=size)
```

src/rubrix/server/daos/records.py
maxserras: Agree :D

src/rubrix/server/daos/records.py
```diff
@@ -181,6 +181,7 @@ def search_records(
         record_from: int = 0,
         exclude_fields: List[str] = None,
         highligth_results: bool = True,
+        id_from: Optional[str] = None,
```
frascuchon: We should revert this change, and keep search_records as before

src/rubrix/server/daos/records.py
```diff
@@ -208,16 +213,21 @@ def search_records(
         aggregation_requests = (
             {**(search.aggregations or {})} if compute_aggregations else {}
         )
-
-        sort_config = self.__normalize_sort_config__(records_index, sort=search.sort)
-
```
frascuchon: ditto

src/rubrix/server/daos/records.py
```diff
@@ -208,16 +213,21 @@ def search_records(
...
         es_query = {
             "_source": {"excludes": exclude_fields or []},
             "from": record_from,
             "query": search.query or {"match_all": {}},
-            "sort": sort_config,
```
frascuchon: ditto

src/rubrix/server/daos/records.py
```diff
@@ -208,16 +213,21 @@ def search_records(
...
         }
+
+        if not id_from:
+            sort_config = self.__normalize_sort_config__(records_index, sort=search.sort)
+        else:
+            sort_config = [{"id": {"order": "asc"}}]
+            search_after = [id_from]
+            es_query["search_after"] = search_after
+        es_query["sort"] = sort_config
+
```
frascuchon: ditto

src/rubrix/server/elasticseach/client_wrapper.py
```diff
@@ -102,7 +102,17 @@ def list_documents(
         A sequence of documents resulting from applying the query on the index
 
         """
-        return es_scan(self.__client__, query=query or {}, index=index)
+        size = size or 1000
```
frascuchon, line 107: `size` must be defined as an argument. Otherwise, it fails on run:


```python
def list_documents(
        self, index: str, query: Dict[str, Any] = None, size: Optional[int] = None
    ) -> Iterable[Dict[str, Any]]:
```

frascuchon (approved):

bilykigor:
Screen Shot 2022-09-08 at 1 55 35 PM

Seems id_from not working in 0.17.0

frascuchon:
Hi @bilykigor, and thanks for reaching out.

Are you using v0.17.0 also for the server?

PR #1666, opened 2022-08-08
fix: Update progress bar when refreshing after adding new records

### LINKED ISSUE #1590: [Progress Metrics] Not updated when pressing "Refresh"
**Describe the bug**
When you add records to a dataset and press the "Refresh" button, the new records will appear, and also the "Records (x)" in the upper right corner will get updated. However, the Progress Metrics will not update and the total will not reflect the total number of records in the dataset. The Progress bar gets updated only when you start annotating.

**To Reproduce**
1. create a dataset
2. start annotating
3. add some more records
4. press the Refresh button
5. the total in the Progress metrics stays the same

**Expected behavior**
The "Refresh" button should also update the Progress metrics. Otherwise, for an active learning loop this is can get a bit confusing. 

**Screenshots**

 - OS [e.g. iOS]: Ubuntu
 - Browser [e.g. chrome, safari]: firefox
 - Rubrix Version [e.g. 0.10.0]: master
 - ElasticSearch Version [e.g. 7.10.2]: 7.10.2

**Additional context**

leiyre:
closes #1590

this PR forces "Refresh" button to update progress metrics

### DIFF
--- a/frontend/database/modules/datasets.js
+++ b/frontend/database/modules/datasets.js
@@ -282,15 +282,9 @@ async function _refreshDatasetAggregations({ dataset }) {
     size: 0,
   });
 
-  const globalResults = await _querySearch({
-    dataset,
-    query: {},
-    size: 0,
-  });
-
   return await _updateTaskDataset({
     dataset,
-    data: { globalResults, results: { aggregations } },
+    data: { results: { aggregations } },
   });
 }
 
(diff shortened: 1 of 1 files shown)


frascuchon (changes requested):
Weird behavior validating token classification records. After validate some records, the refresh button reset the counter to the initial state.  See the gif

compute-wrong-metrics

frascuchon (approved):

PR #1671, opened 2022-08-09
feat(#1579): validate token classification annotations in client

### LINKED ISSUE #1579: [TokenClassificationRecord] Add validation for the alignment of tokens and predictions/annotations to the client
[Copied from Slack] What I have in mind is creating the list of my TokenClassificationRecord and filtering the ones out which have an issue (what I am getting is an rubrix.api.errors::ValidationError. Cant log my dataset as it has some misalignments due to token missmatch (e.g. subword was labeled as the entity text and the list of tokens doest not reflect that).

I would like to be able to drop the ones that would cause a validation error before log by checking the alignment when creating the records in the client.

**Additional context**

dcfidalgo:
Closes #1579 

Validates the prediction/annotation spans in the client when creating the token classification records. For this, we introduce a new `SpanUtils` class that also takes care of transforming the spans into tags.


dcfidalgo:
> Comments after the call with @frascuchon :
> 
>  - from_tags should support BILOU
>  - add a private attribute to Record classes that hold the SpanUtils instance (this avoids the overhead of computing the char to token mappings every time)
>   - remove/deprecate token_span and char_id2token_id methods, and private chars2token and tokens2chars attributes
>   - create a utils module and with it a span_utils.py, move current utils.py to the new utils folder

- [x] support BILOU in `from_tags` 
- [x] add private `_span_utils` attribute in corresponding token classification record classes
- [x] remove/deprecate old helper methods
- [x] create utils module  

still missing:
 - [x] add/adapt unit tests

dcfidalgo:
Ok, this should be ready to review. @frascuchon I left one inline comment, that I think should be corrected before merging.

src/rubrix/server/apis/v0/models/metrics/token_classification.py
```diff
@@ -417,22 +419,26 @@ def build_tokens_metrics(
                 idx=token_idx,
                 value=token_value,
                 char_start=char_start,
-                char_end=char_end,
+                char_end=char_end - 1,
```
dcfidalgo, line 422: I think the `char_end` definition should be changed to `text[char_start:char_end] == "token"`. Right now it is `text[char_start:char_end+1] == "token"`, that is why I added a `- 1`.

frascuchon: The `char_end` character should be aligned to the way annotations are defined, so, text[char_start:char_end] == "token", but we will tackle this in a separate PR.

frascuchon:
Closing.

Merged from #1709

PR #1721, opened 2022-09-15
feat(Datasets): delete records by query

### LINKED ISSUE #1714: [Dataset records] delete dataset records
Sometimes we could clean a rubrix dataset. The current workflow should be load desired records and log in a new dataset and then delete the old dataset.

It would be nice a "delete by query" action to clean dataset records. What do you think @dvsrepo @dcfidalgo ?

## [UPDATE]

- Records deletion will be applied by the python client
- The new method/function will have the same signature as `rb.load`:
```python 
import rubrix as rb
...
rb.delete_records(
   name="my-dataset", 
   ids=[...],  # A list of record ids. If provided, `query` param will be ignored
   query="a query", # The query matching records
   discard_only=False, # If True, matched records won't be deleted. Instead of that, the matched records will be marked as `Discarded`
)
```
- Only super-users or dataset creators can delete records. Other users only can mark as `Discarded`, even if `discard_only` is set to `True`

frascuchon:
Closes #1714

### DIFF
--- a/src/rubrix/client/api.py
+++ b/src/rubrix/client/api.py
@@ -421,7 +421,7 @@ class Api:
             ids:
                 If provided, deletes dataset records with given ids.
             discard_only:
-                If `True`, matched records won't be delete. Instead of that, they will be marked
+                If `True`, matched records won't be deleted. Instead, they will be marked
                 as `Discarded`
             discard_when_forbidden:
                 Only super-user or dataset creator can delete records from a dataset.
--- a/src/rubrix/client/api.py
+++ b/src/rubrix/client/api.py
@@ -20,7 +20,7 @@ import warnings
 from asyncio import Future
 from functools import wraps
 from inspect import signature
-from typing import Any, Callable, Dict, Iterable, List, Optional, Union
+from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple, Union
 
 from tqdm.auto import tqdm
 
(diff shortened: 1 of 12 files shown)


frascuchon:
@dvsrepo Could you please review the docstring?

src/rubrix/client/api.py
```diff
@@ -401,6 +401,57 @@ async def log_async(
...
+                as `Discarded`
+            discard_when_forbidden:
+                Only super-user or dataset creator can delete records from a dataset.
+                So, running "hard" deletion for other users will raise an `ForbiddenApiError` error.
+                If this parameter is `True`, the client API will automatically try to mark as ``Discarded``
+                records instead. Default, `True`
+
+        Returns:
+        --------
+            The total of matched records and real number of processed errors. These numbers could not
```
dvsrepo, line 434:
```suggestion
            The total number of matched records and real number of processed errors. These numbers could not
```

src/rubrix/client/api.py
```diff
@@ -401,6 +401,57 @@ async def log_async(
...
+        -----------
+            name:
+                The dataset name.
+            query:
+                An ElasticSearch query with the
+                `query string syntax <https://rubrix.readthedocs.io/en/stable/guides/queries.html>`_
+            ids:
+                If provided, deletes dataset records with given ids.
+            discard_only:
+                If `True`, matched records won't be delete. Instead of that, they will be marked
```
dvsrepo:
```suggestion
                If `True`, matched records won't be deleted. Instead, they will be marked
```

dvsrepo (changes requested):

dvsrepo (approved):

PR #1736, opened 2022-09-28
fix: search tag reset prior annotation

### LINKED ISSUE #1711: Search-tag on labels, resets prior annotation on text classification hand labeling with multi_label=True
It appears that the search function on labels in case of hand labeling - text classification with multiple labels clears all prior annotations on close. This creates a major bug, because it is not apparent immediately that the prior annotation labels have been reset since they are out of visible scope. The problems is even more pronounced if you are working with a large number of labels. 
Steps to reproduce:
Create a ```DatasetForTextClassification``` with an array of records created using 
```python

records = []
for idx, row in df.iterrows():
    records.append(make_record(row))
dataset_rb = rb.DatasetForTextClassification(records)

def make_record(row):
  record = rb.TextClassificationRecord(
          text = row["text"],
          multi_label = True
  )
  return row
```
Assign a large amount of labels to the dataset
```python

  settings = rb.TextClassificationSettings(label_schema=get_lots_of_labels())

  # apply settings to new or already existing dataset
  rb.configure_dataset("my_dataset_name", settings=settings)

  # logging to the newly created dataset triggers the validation checks
  rb.log(dataset_rb, "my_dataset_name")
```

Switch to the web app and try hand labeling, use the search on the labels (not the record) for toggling select, try a few search string and clear out search string after making selections, only the most recent labels maintain state, all prior label toggles get reset.

Appears to be a state management issue.

keithCuniah:
Close #1711

Also add some none regression unit test with Jest

### DIFF
--- a/frontend/components/text-classifier/results/classifier-annotation/classifierExplorationArea.spec.js
+++ b/frontend/components/text-classifier/results/classifier-annotation/classifierExplorationArea.spec.js
@@ -1,0 +1,81 @@
+import { shallowMount } from "@vue/test-utils";
+import ClassifierAnnotationArea from "./ClassifierAnnotationArea";
+import ClassifierAnnotationButton from "../ClassifierAnnotationButton";
+import { TextClassificationRecord } from "@/models/TextClassification";
+
+let wrapper = null;
+const options = {
+  propsData: {
+    record: new TextClassificationRecord({
+      inputs: { text: "My text", multi_label: false },
+    }),
+    dataset: {
+      task: "TextClassification",
+      isMultiLabel: true,
+      viewSettings: {
+        annotationEnabled: false,
+      },
+      labels: ["label1"], // WARNING, if labels is empty then the ClassifierAnnotationButton will no be render => test 2/3/4/5 will failed
+    },
+  },
+};
+const spyUpdateLabels = jest.spyOn(
+  ClassifierAnnotationArea.methods,
+  "updateLabels"
+);
+beforeEach(() => {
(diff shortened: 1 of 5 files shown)


frontend/components/text-classifier/results/classifier-annotation/classifierExplorationArea.spec.js
```diff
@@ -0,0 +1,78 @@
...
+
+const testIfEmittedValuesFromValidateEventIsEqualToUserSelectedAnnotations = async (
+  selectedAnnotations
+) => {
+  const annotationButtons = wrapper.findComponent(ClassifierAnnotationButton);
+  const emittedValuesFromAnnotationButtons = selectedAnnotations;
+  annotationButtons.vm.$emit("change", emittedValuesFromAnnotationButtons);
+  await wrapper.vm.$nextTick();
+  expect(spyUpdateLabelsMethod).toHaveBeenCalled();
+  if (selectedAnnotations.lenth) {
```
leiyre:
```suggestion
  if (selectedAnnotations.length) {
```

leiyre (changes requested):

frontend/components/text-classifier/results/classifier-annotation/classifierExplorationArea.spec.js
keithCuniah: Done

frascuchon (dismissed):

frascuchon (approved):

leiyre (approved):

Issue #482, opened 2021-10-18
Accept flat input text for token classification in RubrixLogHTTPMiddleware ner mapper


frascuchon:
The default `RubrixLogHTTPMiddleware` [record mapper for token classification](https://github.com/recognai/rubrix/blob/master/src/rubrix/client/asgi.py#L35) expect a structured including a `text` field for inputs. This could make prediction model inputs a bit cumbersome. Default mapper could accepts also flat strings as inputs:

```python
def token_classification_mapper(inputs, outputs):
    if isinstance(inputs, str):
        text = inputs
     ...

Ankush-Chander:
Hi @frascuchon 
I intend to get some clarity on this issue if it"s not already undertaken.

Thanks.

frascuchon:
Hi @Ankush-Chander and sorry for the late response. 

Sure! You can go ahead with this issue. 

The idea is to improve [HTTP monitor middleware for token classification](https://github.com/recognai/rubrix/blob/master/src/rubrix/monitoring/asgi.py#L49) by accepting string text as inputs, instead of a dict containing the `text` key.

PR #1738, opened 2022-09-30
docs: using official token class. mapper since is compatible now

frascuchon:
Once the issue #482  is resolved, we can remove the custom mapper for spacy on this tutorial and use the one provided by the monitoring module.


dvsrepo (approved):
💯

PR #1713, opened 2022-09-13
feat(Datasets): restrict dataset deletion only to creators and super-users

frascuchon:
This PR includes restrictions on dataset deletion. Only datasets the creator or super-users (with full workspaces access) can delete datasets

@dvsrepo just review the error message, pls

PR #1740, opened 2022-09-30
fix: handle separately 401 and 403 api errors in UI

frascuchon:
Since the changes included in #1713, users without permission trying to delete datasets from UI will hit into a re-login operation.

This PR handles separate authentication errors (401) and forbidden errors (403).

We should find a better way to handle error codes and info in general.

Refs: #1713

### DIFF
--- a/frontend/plugins/vuex-orm-axios.js
+++ b/frontend/plugins/vuex-orm-axios.js
@@ -42,7 +42,7 @@ export default ({ $axios, app }) => {
 
   $axios.onError((error) => {
     const code = parseInt(error.response && error.response.status);
-    if (error instanceof ExpiredAuthSessionError || [401, 403].includes(code)) {
+    if (error instanceof ExpiredAuthSessionError || 401 === code) {
       app.$auth.logout();
     }
 
@@ -70,13 +70,13 @@ export default ({ $axios, app }) => {
         break;
       case 404:
         Notification.dispatch("notify", {
-          message: "Warning: " + error.response.data.detail,
+          message: "Warning: " + messageDetail.params.detail,
           type: "warning",
         });
         break;
       default:
         Notification.dispatch("notify", {
-          message: error,
+          message: "Error: " + messageDetail.params.detail,
           type: "error",
         });
     }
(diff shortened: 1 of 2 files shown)


frontend/plugins/vuex-orm-axios.js
```diff
@@ -70,13 +70,13 @@ export default ({ $axios, app }) => {
         break;
       case 404:
         Notification.dispatch("notify", {
-          message: "Warning: " + error.response.data.detail,
+          message: "Warning: " + messageDetail.params.detail,
```
keithCuniah: Some code smeller recommends to concatenate strings with backtick like so : (backtick)Warning : ${messageDetail.params.detail}(backtick)

frontend/plugins/vuex-orm-axios.js
```diff
@@ -70,13 +70,13 @@ export default ({ $axios, app }) => {
...
           type: "warning",
         });
         break;
       default:
         Notification.dispatch("notify", {
-          message: error,
+          message: "Error: " + messageDetail.params.detail,
```
keithCuniah: Some code smeller recommends to concatenate strings with backtick like so : (backtick)Error: ${messageDetail.params.detail}(backtick)

frascuchon: We could apply your suggestion here, isn't it?
```suggestion
          message: `Error:  ${messageDetail.params.detail}`,
```

frontend/plugins/vuex-orm-axios.js
frascuchon: like that ? 

```suggestion
          message: `Warning: ${messageDetail.params.detail}`,
```

keithCuniah: yes like your correction

frontend/plugins/vuex-orm-axios.js
keithCuniah: yes, but don't forget the brackets  around the variable : `Error:  ${messageDetail.params.detail}`,

PR #1744, opened 2022-10-02
docs: raise small-text version to 1.1.0 and adapt tutorial

### LINKED ISSUE #1693: [Active Learning] Small text tutorial hits errors when running on local
The small text tutorial should be review since some weird errors are hit when running on a local machine with some customizations. (Setting a different initial batch size, no CUDA installed...)

chschroeder:
Hi,

with the new small-text v1.1.0 release the requirement that all classes must be in the training data (#1693) is now gone and should not cause any more problems.

- Moreover, I changed the code to use the new convenience methods to construct a `TransformersDataset`.
- The small-text imports have (finally) been cleaned up and you can import all public methods from the top level.
- For dataset construction, passing the target_labels is now recommended (and I changed that as it returns a warning otherwise).
- Since the new TREC dataset (despite your configured `revision` parameter) always fell back to the old version (which resulted in an error since the attributes don't mach) therefore I changed this to use the `version` parameter instead. This seems to work well and it is probably reasonable to pin the example to a dataset version. 

- Proposal: I changed the transformer model name to use a larger model if a GPU (i.e. if CUDA) is available. Otherwise, with only the tiny model, people might undererstimate the strength of this combination.

I did manually test it only until the labeling starts. Sorry, short on time today :).

This PR is just intended to make this as easy as possible for you. Don't feel obligated to use this and change/revert anything that does not fit.

### DIFF
--- a/docs/tutorials/active_learning_with_small_text.ipynb
+++ b/docs/tutorials/active_learning_with_small_text.ipynb
@@ -77,7 +77,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "%pip install \"rubrix[listeners]\" datasets \"small-text>=1.1.0\" \"transformers[torch]\""
+    "%pip install \"rubrix[listeners]\" \"datasets~=2.5.0\" \"small-text>=1.1.0\" \"transformers[torch]\""
    ]
   },
   {
--- a/docs/tutorials/active_learning_with_small_text.ipynb
+++ b/docs/tutorials/active_learning_with_small_text.ipynb
@@ -77,7 +77,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "%pip install \"rubrix[listeners]\" datasets \"small-text\" \"transformers[torch]\""
+    "%pip install \"rubrix[listeners]\" datasets \"small-text>=1.1.0\" \"transformers[torch]\""
    ]
   },
   {
@@ -111,7 +111,7 @@
    "source": [
     "import datasets\n",
     "\n",
-    "trec = datasets.load_dataset('trec', revision=\"bc790b9ce61d4c2b1ea9622cd65da40182725a61\")"
+    "trec = datasets.load_dataset('trec', version=datasets.Version(\"2.0.0\"))\n"
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


frascuchon (approved):
Wow!! 

Thanks a lot @chschroeder

docs/tutorials/active_learning_with_small_text.ipynb
```diff
@@ -77,7 +77,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "%pip install \"rubrix[listeners]\" datasets \"small-text\" \"transformers[torch]\""
+    "%pip install \"rubrix[listeners]\" datasets \"small-text>=1.1.0\" \"transformers[torch]\""
```
frascuchon: Using older `datasers` versions, the downloaded dataset has different feature names. I think that is why the settled revision finds different dataset versions for both of us.

```suggestion
    "%pip install \"rubrix[listeners]\" \"datasets~=2.5.0\" \"small-text>=1.1.0\" \"transformers[torch]\""
```

docs/tutorials/active_learning_with_small_text.ipynb
```diff
@@ -140,52 +140,47 @@
...
    "outputs": [],
    "source": [
+    "import torch\n",
     "from transformers import AutoTokenizer\n",
     "\n",
-    "# Choose transformer model\n",
-    "TRANSFORMER_MODEL = \"prajjwal1/bert-tiny\"\n",
-    "\n",
+    "# Choose transformer model: In non-gpu environments we use a tiny model to reduce the runtime\n",
+    "if torch.cuda.is_available():\n",
```
frascuchon: Great catch for running different environments!! But, I think a `not` was missed here :-)
```suggestion
    "if not torch.cuda.is_available():\n",
```

frascuchon (changes requested):
I've been testing the tutorial locally and I found just 2 minimal problems. I've included suggestions to simplify the revision.

The rest looks really great and works like a charm

docs/tutorials/active_learning_with_small_text.ipynb
chschroeder: Yes, good catch :D! This one would have been noticed if I would have done some labeling.

chschroeder:
Thanks a lot for your review @frascuchon! Two very valid points, I completely agree.

docs/tutorials/active_learning_with_small_text.ipynb
chschroeder: I noticed the problem with the feature names, but using the `revision` argument still loaded the previous dataset version for me. And this resulted in a situation where I had the old dataset with your code for the new attribute name which again lead to that attribute error (in exactly the opposite way as before). This occurred even after clearing the cache, the data was definitely re-downloaded. I can't rule out an error on my side here but this is the reason why I switched to `version`.

frascuchon (approved):
💯

PR #1768, opened 2022-10-14
feat: Similarity vector search client and backend

### LINKED ISSUE #1757: Similarity search with vectors
In order to enable similarity and vector search features (bulk labeling, similarity-based data curation, etc.) the first step is to enable the ability to store a vector associated with a record.

The main idea is to enable users to embed a record (using their embedding method of choice) and then add the vector as a field associated with the record. In this way, it can already be used with ES/Opensearch for KNN as well as for other use cases.

There are some things to take into account:

- Compat between Opensearch and Elasticsearch
- Work needs to be done at the Server and the Client level

We already started working on this: https://github.com/recognai/rubrix/pull/214 but stopped because Elasticsearch's support was limited at the time.

There are several discussions around this topic:
https://github.com/recognai/rubrix/discussions/177
https://github.com/recognai/rubrix/discussions/1079

## Tasks
- [x] #2026
- [x] #2027 
- [x] #2028 
- [x] #2029

ufukhurriyetoglu:
I was able to decrease the failed test count to six and the errors to one error. Currently debugging the failing or error containing tests to have a clean test run.

Steps planned for the vector search task:

- Upgrade ElasticSearch version and use official python client library [ongoing]
- Opensearch server and python client upgrade, and relevant changes [planned]
- Modularize the above steps, to avoid code repetition
- Add vector search attribute to record class
- Elasticsearch knn integration 
- Opensearch knn integration 
- Final refactoring and tests for new cases 

# TODO

- [x] Elasticsearch 8.5.x integration
- [x] Integrate `k` parameter and pagination size https://github.com/argilla-io/argilla/pull/1973)
- [ ] Apply vector setup optimization for elasticsearch/opensearch
- [x] Compute the` num_candidates` argument for elasticsearch (opensearch has a similar setup but at index mapping definition). https://github.com/argilla-io/argilla/pull/1973
- [ ] Add documentation guides (or maybe a mini tutorial for data deduplication) cc @dvsrepo @davidberenstein1957 
- [x] Review python client docstrings for new record `embeddings` client property.
- [x] ~~Enhance python client usability (allow passing numpy arrays or pandas.Series as vector values)~~ -> Next Iteration

Closes #1757


tests/server/datasets/test_api.py
```diff
@@ -195,6 +195,8 @@ def test_open_and_close_dataset(mocked_client):
 
     assert mocked_client.put(f"/api/datasets/{dataset}:close").status_code == 200
 
+    print("Done")
```
ufukhurriyetoglu: I will remove these stale print statements.

tests/server/datasets/test_api.py
```diff
@@ -203,6 +205,7 @@ def test_open_and_close_dataset(mocked_client):
             "params": {"name": dataset},
         }
     }
+    print("Done 2")
```
ufukhurriyetoglu: will remove this one

tests/server/text_classification/test_api.py
```diff
@@ -377,7 +377,14 @@ def test_some_sort_by(mocked_client):
...
     }
-
+    assert (
+        response.json()["detail"]["code"]
+        == expected_response_property_name_2_value["detail"]["code"]
+    )
+    assert (
+        response.json()["detail"]["params"]["message"]
+        == expected_response_property_name_2_value["detail"]["params"]["message"]
+    )
```
ufukhurriyetoglu, line 499: @frascuchon the remaining test failures seem to be error checks, will examine them properly. this is an example of them. The status code is 500, however the message ["params"]["message"] is same. we may need to update the respective test data, wdyt ?

frascuchon: I've seen that you already changed the [error handler wrapper](https://github.com/recognai/rubrix/blob/a179d5e1028660fab7a60052df1679c80da5a2f7/src/argilla/server/daos/backend/elasticsearch.py#L87). Are these errors already fixed?

frascuchon:
Regarding the build pipeline, I see that client version >=8.0 is not compatible with server instance <8.0 (see [here](https://github.com/recognai/rubrix/actions/runs/3250216738/jobs/5333596326#step:11:54))

We should implement a clever mechanism to allow:

- Configure rubrix with the `elasticsearch-py` client (from 7.4 to 8.x)
- Configure rubrix with the `opensearch-py` client (from 1.0 to 2.x) See[ compatibility notes](https://github.com/opensearch-project/opensearch-py/blob/main/COMPATIBILITY.md#compatibility-with-opensearch) 

Also, regarding the latest elasticsearch docs referring to the [dense vector configuration](https://www.elastic.co/guide/en/elasticsearch/reference/current/knn-search-api.html), the separate knn-search is now deprecated in favor of a knn option in the common [`_search`](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-search.html#search-api-knn).

This new endpoint was included from version 8.x and is still in technical revision, so both (knn option and knn search) could be modified. 

Since the dense vector is also available in the scripting score searches, and this is available for older versions, maybe be a wider way to implement cross-version compatibility by using this kind of search (see [here](https://www.elastic.co/guide/en/elasticsearch/reference/8.0/query-dsl-script-score-query.html#vector-functions)). At least if you have already found problems with your performance experiments.

tests/server/text_classification/test_api.py
ufukhurriyetoglu: I replaced the OpenSearch error methods with ElasticSearch methods, however the OpenSearchException don't have a match so I converted it to Exception temporarily. After these changes the errors caused by library incompatibility fixed. But still 500 == 404 assertion error happens.

src/argilla/client/models.py
```diff
@@ -176,7 +176,48 @@ class TextClassificationRecord(_Validators):
...
+        ...                         "embeddingVectors": [
+        ...                                {
+        ...                                     "vectorizerName": "bert_base_uncased"
+        ...                                     "embeddingVector": [1.3, 2.1, 3.4, 5.1]
+        ...                                 }
+        ...                          ]
+        ...                      }
+        ...                  ]
+        ...             }
+        ...         ]
```
ufukhurriyetoglu: @frascuchon to speed up the end to end vector search functionality implementation, I designed this data mapping as a kickstart. will proceed with the rest. in final PR reviews we can discuss and adjust this better.

docker-compose.yaml
```diff
@@ -1,51 +1,37 @@
...
-    image: docker.elastic.co/elasticsearch/elasticsearch:7.11.1
-    container_name: elasticsearch
+    image: docker.elastic.co/elasticsearch/elasticsearch:8.2.0
+    ports:
+      - "9200:9200"
+      - "9300:9300"
     environment:
       - node.name=elasticsearch
-      - cluster.name=es-local
+      - cluster.name=es-rubrix-local
```
dvsrepo: @ufukhurriyetoglu If we need to change this, we should use argilla instead of rubrix

src/argilla/server/daos/backend/elasticsearch.py
```diff
@@ -1129,6 +1146,31 @@ def _create_index_alias(self, index: str, alias: str):
...
+        vector_search_index_name_2_index_mapping = {}
+        for embedding_name, embedding_dimension in embeddings:
+            vector_search_index_name = "_".join(
+                id, embedding_name
+            )  ## aims to make the vector search index name unique
+            index_mapping = {
+                "type": "dense_vector",
+                "dims": embedding_dimension,
+                "index": True,
+                "similarity": "cosine",  ## can similarity property also be part of config @frascuchon ? relates vector search similarity metric
```
frascuchon: > `## can similarity property also be part of config @frascuchon? relates vector search similarity metric`

Totally agree. Make sense for a global similarity function for all dataset embeddings computation or one per embedding field should be a better fit? Does it depend on the embedding nature?

src/argilla/server/daos/backend/elasticsearch.py
```diff
@@ -1129,6 +1146,31 @@ def _create_index_alias(self, index: str, alias: str):
...
+            }
+            vector_search_index_name_2_index_mapping[
+                vector_search_index_name
+            ] = index_mapping
+        # raise NotImplementedError(
+        #    "Here, we should configure embeddings mappings by using the provided "
+        #    "per-field embeddings configuration. "
+        # )
+        return vector_search_index_name_2_index_mapping
```
frascuchon: Here, we can create the index mapping updates directly. I will push the changes

src/argilla/server/daos/backend/search/query_builder.py
```diff
@@ -180,6 +195,16 @@ def _to_es_query(cls, query: BackendRecordsQuery) -> Dict[str, Any]:
...
+        # TODO(@ufukhurriyetoglu): Here, we will use the query embeddings settings to build the knn search
+        if query.embedding_vector:
+            # do your magic ;-) !
+            knn_query = knn.knn_query(
+                field=query.embedding_name,
+                query_vector=query.embedding_vector,
+                k=3,
+                num_candidates=50,
+            )
+
```
frascuchon: Given the extensive use of the `map_2_es_query` method, and the fact that the `knn` config could be a separate configuration of than query. I will include the `knn query` generation isolated prior to searching the results. I will push the changes for that

src/argilla/server/daos/records.py
```diff
@@ -102,13 +102,21 @@ def add_records(
...
+        # TODO(@ufukhurriyetoglu): At this point, we can setup the embeddings setting at index level
         self._es.create_dataset_index(
-            dataset.id,
+            id=dataset.id,
             task=dataset.task,
             metadata_values=metadata_values,
+            # Here, we can pass the embeddings configuration to create. We can do index level validations
+            # inside the es.created_dataset_index method (like embeddings fields are
+            # already created and things like that)
+            embeddings_cfg=...,
```
frascuchon: This won't be needed if we finally decide to configure the dataset embedding by the dataset settings. (Personally, I prefer this approach. Can also simplify interactions with the client)

docker-compose.yaml
ufukhurriyetoglu: @dvsrepo will update that before reopening this PR. this is temporary.

ufukhurriyetoglu: corrected.

src/argilla/server/daos/models/records.py
```diff
@@ -107,6 +125,7 @@ def prepare_record_for_db(cls, values):
 
         values = cls.update_annotation(values, "prediction")
         values = cls.update_annotation(values, "annotation")
+        values = cls.update_annotation(values, "embedding")
```
frascuchon: Not sure if this line is needed

tests/server/datasets/test_api.py
frascuchon: 👍

docker-compose.yaml
```diff
@@ -1,51 +1,37 @@
...
+    ports:
+      - "9200:9200"
+      - "9300:9300"
     environment:
       - node.name=elasticsearch
-      - cluster.name=es-local
+      - cluster.name=es-argilla-local
       - discovery.type=single-node
-      - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
+      - “ES_JAVA_OPTS=-Xms512m -Xmx512m”
```
frascuchon, line 31: Be careful with quotation marks. It should be the standard ones `"` instead of `“` `”`

src/argilla/server/daos/backend/elasticsearch.py
ufukhurriyetoglu: there are several distance calculation functions, and not all of them supported by each opensearch, elasticsearch versions. cosine distance generally works best and we can set it as default, by also taking care of the elasticsearch/opensearch version

src/argilla/server/daos/backend/search/query_builder.py
ufukhurriyetoglu: yes, makes sense

src/argilla/server/daos/records.py
ufukhurriyetoglu: yeah me too

src/argilla/server/daos/models/records.py
ufukhurriyetoglu: actually not required, because annotation and embedding share the same data class, however embedding is not relevant to annotation. I will remove this

src/argilla/server/daos/backend/search/model.py
```diff
@@ -74,6 +74,13 @@ class BaseRecordsQuery(BaseQuery):
 
     has_annotation: Optional[bool] = None
     has_prediction: Optional[bool] = None
+    # TODO(@ufukhurriyetogl): In order to allow embedding-based searches, we can include here
+    #  (the base records query data model class) an embeddings related query params that
+    #  we can use later for the records search.
+    # The input vector:
+    embedding_vector: Optional[List[float]] = None
+    # The name of the embedding to uses (if several):
+    embedding_name: Optional[str] = None
```
frascuchon, line 83: Here, which name should be provided to match the stored one? Regarding the data model discussed yesterday, given the embedding config for a record:

```python
embeddings = {
        "text": {
            "embedding_vectors": { 
                "bert_uncased": [...],
                "bert_case": [...]
            }
        },
        "inputs": {
            "property_names": ["body", "subject"],
            "embedding_vectors": { 
                "bert_uncased": [...],
                "bert_case": [...]
            }
        }
}
```
If a user wants to filter using the `bert_case` for the input text, must compound the final embedding vector like the server does, which is not desirable.

Also, for several embedding based on different fields from the input, this data model does not match (since we have only one `inputs` key)


I think it should be better if we let the users to naming on their own embeddings, so those names can be used easily in query time. For example:

```python
embeddings = {
   "my_text_bert_cased": { 
       "vector": [...],
       # this field is not needed for now, and we could avoid it for the moment. 
       # Once we introduce automatic embedding generation we can refine how 
       # to choice the record field used for the embeding
       "record_properties": ["inputs.body", "inputs.subject"] 
   }, 

   "another_vector_uncased": {
      "vector": [...],
      "record_properties": [ { "inputs": ["body", "subject"]}, "text"] # Another definition way 
      "model": "bert_uncased" # Once we have automatic embedding generation, we can provide the model name used for that in this way
   }
}
```

So, under this configuration, we can easily query data using embedding likes follows:

````python
embedding_name = "another_vector_uncased"
embedding_vector = [...]
````

What do you think @ufukhurriyetoglu ?

File: docker-compose.yaml (hunk 11 above)
Comment (ufukhurriyetoglu) [reply docker-compose.yaml:31]: sure

File: src/argilla/server/daos/backend/search/model.py (hunk 12 above)
Comment (ufukhurriyetoglu) [reply src/argilla/server/daos/backend/search/model.py:83]: totally agree with you

File: src/argilla/client/sdk/commons/models.py [hunk 13]
```diff
@@ -45,6 +45,7 @@ class BaseRecord(GenericModel, Generic[T]):
     status: Optional[TaskStatus] = None
     prediction: Optional[T] = None
     annotation: Optional[T] = None
+    embeddings: Optional[Dict[str, Any]] = None
```
frascuchon, line 48: Could we simplify the client embedding configuration and pass only the embedding name and the vector array as value?

Also, to be more user-friendly, could we accept NumPy arrays or pandas series for the vector value?


```python
EmbeddingVector = Union[List[float], <the-numpy-array-typing>, <the-pandas-series-typing>]
embeddings: Optional[Dict[str,EmbeddingVector]]
```

This change will add some extra logic to the SDK mapping but could simplify the user integration and engage the use of the library.

src/argilla/client/sdk/token_classification/models.py
```diff
@@ -140,3 +142,5 @@ class TokenClassificationQuery(BaseModel):
     score: Optional[ScoreRange] = Field(default=None)
     status: List[TaskStatus] = Field(default_factory=list)
     predicted: Optional[PredictionStatus] = Field(default=None, nullable=True)
+    embedding_name: str = Field(default=None)
+    embedding_vector: List[float] = Field(default_factory=list)
```
frascuchon, line 146: Here, the same comment as before.

src/argilla/client/sdk/commons/models.py
ufukhurriyetoglu: yes, I agree with both changes.

src/argilla/client/sdk/token_classification/models.py
ufukhurriyetoglu: yes, will do

tests/server/token_classification/test_api.py
```diff
@@ -224,6 +225,130 @@ def test_create_records_for_token_classification(
...
+        "predicted_mentions": {"TEST": {"This": 3}},
+        "status": {"Default": 3},
+        "words": {},
+    }
+
+    assert "This" in results.aggregations.predicted_mentions[entity_label]
+    assert "This" in results.aggregations.mentions[entity_label]
+    expected_record_ids = [1, 2, 0]
+    for index, record in enumerate(results.records):
+        # assert metrics_validator(record)
```
ufukhurriyetoglu: @frascuchon if we don't comment out this line, the test is failing. Can you take a look ?

tests/server/datasets/test_api.py
ufukhurriyetoglu: done

tests/server/datasets/test_api.py
ufukhurriyetoglu: done

opensearch.compose.yaml
```diff
@@ -0,0 +1,27 @@
...
+      - bootstrap.memory_lock=true # along with the memlock settings below, disables swapping
+      - plugins.security.disabled=true
+      - "OPENSEARCH_JAVA_OPTS=-Xms512m -Xmx512m" # minimum and maximum Java heap size, recommend setting both to 50% of system RAM
+      - network.host=0.0.0.0 # required if not using the demo security configuration
+    ulimits:
+      memlock:
+        soft: -1
+        hard: -1
+    ports:
+      - 19200:9200
```
ufukhurriyetoglu: @frascuchon maybe this was the cause for opensearch github action not work, after correcting to "9200:9200" locally started to work in my laptop

src/argilla/client/models.py
```diff
@@ -503,7 +572,26 @@ class TextGenerationRecord(_Validators):
         >>> import argilla as rg
         >>> record = rg.Text2TextRecord(
         ...     text="My name is Sarah and I love my dog.",
-        ...     prediction=["Je m'appelle Sarah et j'aime mon chien."]
+        ...     prediction=["Je m'appelle Sarah et j'aime mon chien."],
```
ufukhurriyetoglu, line 540: all these docstrings needs to be updated, note to @ufukhurriyetoglu

tests/server/token_classification/test_api.py
ufukhurriyetoglu: @frascuchon this metrics_validator issue still there

src/argilla/client/models.py
ufukhurriyetoglu: done

opensearch.compose.yaml
frascuchon: Not related, for the GitHub ci we're using this [action step ](https://github.com/ankane/setup-opensearch). But I cannot do it to work

PR #1771, opened 2022-10-17
Including some info about the upgrade and migration

frascuchon:


dvsrepo (changes requested):
I've reviewed this. 

The only change to make is to include the link in the readme's menu to the md in `docs`, what would be the best link until we merge this into main?

frascuchon:
> I've reviewed this.
> 
> The only change to make is to include the link in the readme's menu to the md in `docs`, what would be the best link until we merge this into main?

The README doc is included in the pypi Argilla page. Could we share from here?

frascuchon:
I've already setup the relative link. 

Take a look [here](https://github.com/recognai/rubrix/tree/docs/add-a-migration-guide-in-readme#) @dvsrepo

dvsrepo (approved):

dvsrepo:
This could be merged @frascuchon

Issue #1805, opened 2022-10-27
Link key features


chschroeder:
Hey, I have a small but maybe useful suggestion: In the README.md file, of all things your "Key Features" do not link to the respective concepts in the docs. Maybe there is a reason against this, then consider this withdrawn. Otherwise, this is easy to do and much nicer to navigate, especially for new users.

I can provide a PR with suggestions if you like.

**Is your feature request related to a problem? Please describe.**

For example, this...
![Auswahl_397](https://user-images.githubusercontent.com/6340318/198347715-fded9303-23bc-4a02-93ed-f7e7424994d4.png)

**Describe the solution you'd like**

...could look like this:
![Auswahl_396](https://user-images.githubusercontent.com/6340318/198347681-e153ad41-6f23-4fb9-a2f5-c2aa1cbff365.png)

I was not sure which documentation page to link in the middle, otherwise each line would have had at least one link.

dvsrepo:
@chschroeder This is such a great suggestion! We would be glad to receive a PR!

PR #1809, opened 2022-10-28
docs: Link key features (#1805)

chschroeder:
I have linked all concepts that I could relate to some part of the docs. Moreover, I changed the captialization of "Weak Supervision" to "weak supervision" as it seemed more consistent with the rest of the README.md (e.g., active learning, data collection, ...).

Out of scope but related:
Do you plan to keep multiple versions of the documents online? If so it might be better to change all documentation links (not only the new ones) to a fixed version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ https://user-images.githubusercontent.com/1107111/197567844-4370487d-fe44-441e-9
 ### Monitoring
 
 - Close the gap between production data and data collection activities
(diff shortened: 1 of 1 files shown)


dvsrepo:
Thanks so much @chschroeder !

This looks great on my side, I think for now the base URL can be the one you used (`latest`) and then we can see if we use (`stable`) instead. 

@ufukhurriyetoglu on my side this is ready to be merged

davidberenstein1957 (approved):
LFTM @chschroeder thanks for the great contribution:)

PR #1880, opened 2022-11-10
feat: Using new `top_k_mentions` metrics instead of `entity_consistency`

### LINKED ISSUE #1834: [Token Classification metrics] Review entity consistency and mentions histograms
The first version of `entity_consistency` adds a threshold param and forces it to be greater than 2 (see https://github.com/recognai/rubrix/blob/7e806ad8b9d3b2abf54b14c87c335022c60633ff/src/rubrix/metrics/token_classification/metrics.py#L378). I'd like to get rid of this logic as I think counterintuitive.

Related to this, we briefly discussed this but having a list of the top N mentions (and their types) in the metrics module would be really useful. I say this because this would be solved by threshold=1 but maybe the name `entity_consistency` is not the best naming. One might argue that is not the best in any case, but I could live with both options:

1. Changing the name `entity_consistency` to something more meaningful and keeping the threshold parameter (without enforcing >=2, but still have the default=2)
2.  Have a separate metric top_k_mentions for the mentions and keep entity_consistency as is (although I still think forcing >=2 is not the best option).
3. Keep the name `entity_consistency` and keep the threshold parameter (without enforcing >=2, but ). This is probably the fastest way to do this (without enforcing >=2, but still have the default=2)

What do you think @frascuchon @dcfidalgo ?

frascuchon:
This PR includes changes to use the old `entity_consistency` metrics as a `top_k_mentions` distribution.

This new function also allows values equal to 1 for the `threshold` parameter and a labeling post-filter, that can help to fine-grain the results.

Closes #1834

### DIFF
--- a/docs/_source/guides/steps/4_monitoring.ipynb
+++ b/docs/_source/guides/steps/4_monitoring.ipynb
@@ -4140,9 +4140,9 @@
     }
    ],
    "source": [
-    "from argilla.metrics.token_classification import entity_consistency\n",
+    "from argilla.metrics.token_classification import top_k_mentions\n",
     "\n",
-    "entity_consistency(name=\"spacy_sm_wnut17\", mentions=5000, threshold=2).visualize()\n"
+    "top_k_mentions(name=\"spacy_sm_wnut17\", k=5000, threshold=2).visualize()\n"
    ]
   },
   {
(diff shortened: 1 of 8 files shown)


dvsrepo:
Looks great to me! Could we do a search for mentions to the old function name in the docs and replace accordingly?

frascuchon:
Done!

dvsrepo:
💯

dvsrepo (approved):

ufukhurriyetoglu (commented):
missed the notification. however went over the changes and they are ok 👍

PR #1884, opened 2022-11-11
feat: Add, delete and edit rules from Python client

### LINKED ISSUE #1855: [Weak supervision] Add Rules to Web UI via Python
### Discussed in https://github.com/recognai/rubrix/discussions/1554

<div type='discussions-op-text'>

<sup>Originally posted by **hanshupe** June 11, 2022</sup>
Is there a way to add Rules via Python script, so that they show up in the Web UI. And after editing them there, can they be imported in Python again?</div>

frascuchon:
Closes #1855


src/argilla/labeling/text_classification/rule.py
```diff
@@ -143,6 +147,18 @@ def __call__(
...
+def add_rules(dataset: str, rules: List[LabelingRule]):
+    """Adds the rules to a given dataset
+
+    Args:
+        dataset: Name of the dataset.
+
+    Returns:
+        api response
+    """
+    return api.active_api().add_dataset_labeling_rules(dataset, rules)
```
frascuchon, line 186: Transform Rule to LabelingRule

src/argilla/client/api.py
```diff
@@ -606,6 +606,11 @@ def compute_metric(
 
         return MetricResults(**metric_.dict(), results=response.parsed)
 
+    def add_dataset_labeling_rules(self, dataset: str, rules: List[LabelingRule]):
+        text_classification_api.add_dataset_labeling_rules(
+            self._client, name=dataset, rules=rules
+        )
```
frascuchon:
```suggestion
        for rule in rules:
            text_classification_api.add_dataset_labeling_rule(
                self._client,
                name=dataset,
                rule=rule,
            )
```

src/argilla/client/sdk/text_classification/api.py
```diff
@@ -58,6 +58,22 @@ def data(
...
+
+    response = httpx.post(
+        url=url,
+        headers=client.get_headers(),
+        cookies=client.get_cookies(),
+        timeout=client.get_timeout(),
+    )
+    return build_list_response(response)
+
+
```
frascuchon: The API endpoint is as POST HTTP method to the path `/api/datasets/{name}/TextClassification/labeling/rules`.

You should create a method to create a single rule and use it in a loop from the `api.add_dataset_labeling_rules` method.

```suggestion
def add_dataset_labeling_rule(
    client: AuthenticatedClient,
    name: str,
    rule: LabelingRule,
) -> Response[
    Union[LabelingRule, HTTPValidationError, ErrorMessage]
]:
    url = "{}/api/datasets/{name}/TextClassification/labeling/rules".format(
        client.base_url,
        name=name,
        json=rule.dict(),
    )

    response = httpx.post(
        url=url,
        headers=client.get_headers(),
        cookies=client.get_cookies(),
        timeout=client.get_timeout(),
    )
    return build_typed_response(response, rule.__class__)
```

src/argilla/labeling/text_classification/rule.py
```diff
@@ -143,6 +147,18 @@ def __call__(
             return self._label
 
 
+def add_rules(dataset: str, rules: List[LabelingRule]):
```
frascuchon: The `LabelingRule` class is an internal data model used to communicate with the API. You should use the `Rule` data model and do the mappings to the `LabelingRule`, like function `load_rules` does.

```suggestion
def add_rules(dataset: str, rules: List[Rule]):
```

src/argilla/client/sdk/text_classification/api.py
```diff
@@ -58,6 +58,67 @@ def data(
...
+    name: str,
+    rule: LabelingRule,
+) -> Response[Union[LabelingRule, HTTPValidationError, ErrorMessage]]:
+    url = "{}/api/datasets/{name}/TextClassification/labeling/rules".format(
+        client.base_url, name=name
+    )
+
+    response = httpx.post(
+        url=url,
+        json={"query": rule.query, "label": rule.label},
```
frascuchon: This won't work for text-classification with multi-label. You should pass also the `rule.labels` field.

src/argilla/client/sdk/text_classification/api.py
```diff
@@ -58,6 +58,67 @@ def data(
...
+
+    return build_typed_response(response, LabelingRule)
+
+
+def update_dataset_labeling_rule(
+    client: AuthenticatedClient,
+    name: str,
+    rule: LabelingRule,
+) -> Response[Union[HTTPValidationError, ErrorMessage]]:
+    url = "{}/api/datasets/TextClassification/{name}/labeling/rules/".format(
```
frascuchon: The update endpoint works at rule level. So, you should pass the query as part of the path.

```"{}/api/datasets/TextClassification/{name}/labeling/rules/{query}"....```

File: src/argilla/client/sdk/text_classification/api.py [hunk 7]
```diff
@@ -58,6 +58,67 @@ def data(
...
+        client.base_url, name=name
+    )
+
+    response = httpx.patch(
+        url,
+        json={"query": rule.query, "label": rule.label},
```
frascuchon: Here, only the label/labels can be updated

src/argilla/client/sdk/text_classification/api.py
```diff
@@ -58,6 +58,67 @@ def data(
...
+
+    return build_typed_response(response, rule.__class__)
+
+
+def delete_dataset_labeling_rule(
+    client: AuthenticatedClient,
+    name: str,
+    rule: LabelingRule,
+) -> Response[Union[LabelingRule, HTTPValidationError, ErrorMessage]]:
+    url = "{}/api/datasets/TextClassification/{name}/labeling/rules?{rule}".format(
```
frascuchon: The way to build the path is the same that for the update method: passing the query in the path. The query is the rule key. Argilla does not support two rules with the same query.

dvsrepo:
Hi @ufukhurriyetoglu great work. For updating the docs, I suggest the following:

1. Update https://docs.argilla.io/en/latest/guides/techniques/weak_supervision.html to reflect that Rules created from python can (and should) be added to the dataset. This will make everything much more comfortable for users. Also show you can update and delete rules.
2.  Update tutorials https://docs.argilla.io/en/latest/tutorials/notebooks/labelling-textclassification-snorkel-weaksupervision.html and https://docs.argilla.io/en/latest/tutorials/notebooks/labelling-textclassification-sklearn-weaksupervision.html . In this case, I would add the rules we use to the dataset. In this way the tutorials become much more reproducible.

In both cases 1 and 2, test that they are working as expected, and take the opportunity to simplify text if you see the opportunity.

src/argilla/client/sdk/text_classification/api.py
ufukhurriyetoglu: added the "labels"

src/argilla/client/sdk/text_classification/api.py
ufukhurriyetoglu: fixed

src/argilla/client/sdk/text_classification/api.py
ufukhurriyetoglu: fixed

src/argilla/client/sdk/text_classification/api.py
ufukhurriyetoglu: fixed

frascuchon:
Not sure if everything is pushed, but some new tests are failing. Could you please take a look https://github.com/argilla-io/argilla/actions/runs/3479400438/jobs/5817898907?

docker-compose.yaml
```diff
@@ -1,22 +1,13 @@
...
-    restart: unless-stopped
-    ports:
-      - "6900:80"
-    environment:
-      ARGILLA_ELASTICSEARCH: http://elasticsearch:9200
-      # ARGILLA_LOCAL_AUTH_USERS_DB_FILE: /config/.users.yaml
-      # volumes:
-      #- ${PWD}/.users.yaml:/config/.users.yaml
-    networks:
-      - argilla
```
ufukhurriyetoglu: maybe this change causes the tests failures @frascuchon

src/argilla/labeling/text_classification/rule.py
```diff
@@ -74,6 +74,31 @@ def author(self):
...
         return self._author
 
+    def update_label(self, label):
+        """Updates the label"""
+        self._label = label
+
+    def add_to_dataset(self, dataset: str):
+        """Add to rule to the given dataset"""
+        labeling_rule = LabelingRule(
+            query=self.query, label=self.label, labels=[], author=self.author or "None"
```
frascuchon: As far as I remember, the author will be filled with the user request if no provided. We should remove de "None" defaults

src/argilla/labeling/text_classification/rule.py
```diff
@@ -74,6 +74,31 @@ def author(self):
...
+        """Removes the rule from the given dataset"""
+        labeling_rule = LabelingRule(
+            query=self.query, label=self.label, labels=[], author=self.author or "None"
+        )
+        api.active_api().delete_dataset_labeling_rules(dataset, rules=[labeling_rule])
+
+    def update_at_dataset(self, dataset: str):
+        """Updates the rule at the given dataset"""
+        labeling_rule = LabelingRule(
+            query=self.query, label=self.label, labels=[], author=self.author or "None"
```
frascuchon: Does work for multi label rules? I think we can have errors since rule.label can be a list

tests/labeling/text_classification/test_rule.py
```diff
@@ -86,6 +93,37 @@ def test_name(name, expected):
     assert rule.name == expected
 
 
+def test_atomic_crud_operations(monkeypatch, mocked_client, log_dataset):
+    rule = Rule(query="inputs.text:(NOT positive)", label="negative")
+    with pytest.raises(RuleNotAppliedError):
+        rule(TextClassificationRecord(text="test"))
+
+    monkeypatch.setattr(httpx, "post", mocked_client.post)
+    monkeypatch.setattr(httpx, "stream", mocked_client.stream)
```
frascuchon, line 104: I think you should also mock the "patch" and "delete" httpx functions since are used by the update/delete functionalities.

src/argilla/labeling/text_classification/rule.py
ufukhurriyetoglu: yes, updating according to our discussion "rule.labels = [self.label] or self.label" with sanity check

src/argilla/labeling/text_classification/rule.py
```diff
@@ -74,6 +74,40 @@ def author(self):
...
+        )
+
+    def update_label(self, label):
+        """Updates the label"""
+        self._label = label
+
+    def add_to_dataset(self, dataset: str):
+        """Add to rule to the given dataset"""
+        api.active_api().add_dataset_labeling_rules(
+            dataset, rules=[self.convert_to_labeling_rule()]
```
frascuchon: Here, we have coupled components between `Rule` and `Api`. We've already had it with the `apply` method. We can go forward as is, but we will assume the technical debt. 

If in the future we allow defining contextualized clients, as the example shows:

```python
with rb.init("<an-url>", "<an-api-key") as client:
   client.log..
   client.load...
```
We need to redesign all this stuff.

Just to keep in mind.

src/argilla/labeling/text_classification/rule.py
```diff
@@ -74,6 +74,40 @@ def author(self):
...
+        if isinstance(self._label, str):
+            labels = [self._label]
+        else:
+            labels = self._label
+
+        return LabelingRule(
+            query=self.query, labels=labels, author=self.author or "None"
+        )
+
+    def update_label(self, label):
```
frascuchon: could we just define it as a public attribute or define it as a setter?

```python
@property
def label(self):
   return self._label
@label.setter
def update_label(self, value):
   self._label = value
```

src/argilla/labeling/text_classification/rule.py
```diff
@@ -74,6 +74,40 @@ def author(self):
...
+
+        api.active_api().delete_dataset_labeling_rules(
+            dataset, rules=[self.convert_to_labeling_rule()]
+        )
+
+    def update_at_dataset(self, dataset: str):
+        """Updates the rule at the given dataset"""
+        api.active_api().update_dataset_labeling_rules(
+            dataset, rules=[self.convert_to_labeling_rule()]
+        )
```
frascuchon: So, this method will allow also rules individually, isn't it?

```python
rule = Rule(...)

rule.add_to_dataset(ds)

rule.label = ...
rule.update_at_dataset(ds)

````

For me, it is a bit weird here without a dataset container class.

We can create a rule and update it to a dataset where does not exists.

```python
rule = Rule(...)
rule.update_at_dataset(ds)
```

But not sure if could be helpful in the flow.  @dvsrepo what do you think?

tests/labeling/text_classification/test_rule.py
ufukhurriyetoglu: yes, that fixed the tests

src/argilla/labeling/text_classification/rule.py
ufukhurriyetoglu: done

docker-compose.yaml
ufukhurriyetoglu: this was not relevant to test failures

src/argilla/labeling/text_classification/rule.py
ufukhurriyetoglu: sure

ufukhurriyetoglu: done

src/argilla/labeling/text_classification/rule.py
ufukhurriyetoglu: @frascuchon yes it will work for indivual rules. if the rule is already there then it will be updated

src/argilla/labeling/text_classification/rule.py
ufukhurriyetoglu: done

davidberenstein1957 (commented):
@ufukhurriyetoglu could you also take a look at the docs under the reference/webapp. They currently only contain information about weak labelling for text classification and I feel token classification should be added.

https://docs.argilla.io/en/latest/reference/webapp/features.html#weak-labelling

docs/_source/tutorials/notebooks/labeling_rules.csv
```diff
@@ -0,0 +1,6 @@
+,query,label
```
davidberenstein1957: could you add this to a new separate datasets folder in _static/datasets for the docs? Then we can collect all datasets in this folder.

src/argilla/labeling/text_classification/rule.py
ufukhurriyetoglu: created an issue by referencing this comment,

docs/_source/tutorials/notebooks/labeling_rules.csv
ufukhurriyetoglu: do you mean _static/datasets/{tutorial_name}/ folder path convention ?

ufukhurriyetoglu: added

ufukhurriyetoglu:
> 

@davidberenstein1957 for the webapp/features I checked the post and it seems related to web UI features. This PR is related to python client and UI does not change. As I understand python side is auto generated, we can check if the added modules reflected.

dvsrepo (changes requested):
- [x] Please review my changes, especially for typos.
- [x] Remove output cell from https://docs.argilla.io/en/feat-1855-weak-supervision-rules-python-client/tutorials/notebooks/labelling-textclassification-sklearn-weaksupervision.html#Define-rules

ufukhurriyetoglu (approved):
great! thanks for the review and corrections @dvsrepo

ufukhurriyetoglu (approved):

dvsrepo (approved):
Hi @ufukhurriyetoglu, typically after changes a new review should be requested. This looks good on my side.

@frascuchon do you need to take another look? Otherwise please merge

src/argilla/labeling/text_classification/rule.py
```diff
@@ -143,6 +173,47 @@ def __call__(
...
 
 
+def add_rules(dataset: str, rules: List[Rule]):
+    """Adds the rules to a given dataset
+
+    Args:
+        dataset: Name of the dataset.
+
+    Returns:
+        api response
```
frascuchon:
```suggestion
    Returns:
```

src/argilla/labeling/text_classification/rule.py
```diff
@@ -143,6 +173,47 @@ def __call__(
...
+
+def delete_rules(dataset: str, rules: List[Rule]):
+    """Deletes the rules from the given dataset
+
+    Args:
+        dataset: Name of the dataset
+
+    Returns:
+        api response
+
```
frascuchon:
```suggestion
    Returns:
```

src/argilla/labeling/text_classification/rule.py
```diff
@@ -143,6 +173,47 @@ def __call__(
...
+def update_rules(dataset: str, rules: List[Rule]):
+    """Updates the rules of the given dataset
+
+    Args:
+        dataset: Name of the dataset
+
+    Returns:
+        api response
+
```
frascuchon:
```suggestion
    Returns:
```

src/argilla/labeling/text_classification/rule.py
```diff
@@ -143,6 +173,47 @@ def __call__(
...
+    """
+    rules = [rule.convert_to_labeling_rule() for rule in rules]
+    return api.active_api().delete_dataset_labeling_rules(dataset, rules)
+
+
+def update_rules(dataset: str, rules: List[Rule]):
+    """Updates the rules of the given dataset
+
+    Args:
+        dataset: Name of the dataset
```
frascuchon, line 206: add `rules` in args description

src/argilla/labeling/text_classification/rule.py
```diff
@@ -143,6 +173,47 @@ def __call__(
...
+    """
+    rules = [rule.convert_to_labeling_rule() for rule in rules]
+    return api.active_api().add_dataset_labeling_rules(dataset, rules)
+
+
+def delete_rules(dataset: str, rules: List[Rule]):
+    """Deletes the rules from the given dataset
+
+    Args:
+        dataset: Name of the dataset
```
frascuchon, line 193: add rules in args description

src/argilla/labeling/text_classification/rule.py
```diff
@@ -143,6 +173,47 @@ def __call__(
             return self._label
 
 
+def add_rules(dataset: str, rules: List[Rule]):
+    """Adds the rules to a given dataset
+
+    Args:
+        dataset: Name of the dataset.
```
frascuchon, line 180: add rules in args description

src/argilla/labeling/text_classification/rule.py
```diff
@@ -143,6 +173,47 @@ def __call__(
...
+
+    Args:
+        dataset: Name of the dataset
+
+    Returns:
+        api response
+
+    """
+    rules = [rule.convert_to_labeling_rule() for rule in rules]
+    return api.active_api().delete_dataset_labeling_rules(dataset, rules)
```
frascuchon: the `api` methods return nothing. You can remove the `return` here

src/argilla/labeling/text_classification/rule.py
```diff
@@ -143,6 +173,47 @@ def __call__(
...
+
+    Args:
+        dataset: Name of the dataset
+
+    Returns:
+        api response
+
+    """
+    rules = [rule.convert_to_labeling_rule() for rule in rules]
+    return api.active_api().update_dataset_labeling_rules(dataset, rules)
```
frascuchon: the `api` methods return nothing. You can remove the `return` here

src/argilla/labeling/text_classification/rule.py
```diff
@@ -74,6 +78,34 @@ def author(self):
         """Who authored the rule."""
         return self._author
 
+    def convert_to_labeling_rule(self):
```
frascuchon: this method must be private 

rename to `__convert_to_labeling_rule` and its references

frascuchon (commented):
Just some comments to improve the python reference documentation

See [here](https://docs.argilla.io/en/feat-1855-weak-supervision-rules-python-client/reference/python/python_labeling.html#python-labeling)

src/argilla/labeling/text_classification/rule.py
ufukhurriyetoglu: done

src/argilla/labeling/text_classification/rule.py
ufukhurriyetoglu: fixed

src/argilla/labeling/text_classification/rule.py
ufukhurriyetoglu: fixed

src/argilla/labeling/text_classification/rule.py
ufukhurriyetoglu: added

src/argilla/labeling/text_classification/rule.py
ufukhurriyetoglu: added

src/argilla/labeling/text_classification/rule.py
ufukhurriyetoglu: added

PR #1895, opened 2022-11-14
fix: Avoid manipulating DOM

### LINKED ISSUE #1765: Client (components): components should interact with virtual DOM and not with DOM directly
**Describe the bug**
With VueJS we should interact with virtual DOM and not with DOM directly. 
Means that if we need to add or remove class, we need to bind to the template and not add the class from document object

**To Reproduce**

**Expected behavior**
bind the class "--fixed"

**Screenshots**

 - OS [e.g. iOS]:
 - Rubrix Version [e.g. 0.10.0]:

**Additional context**

leiyre:
This PR removes DOM manipulation and include the logic (toggle class "--fixed") inside component virtual DOM

Closes #1765


frontend/components/base/BaseModal.vue
```diff
@@ -16,10 +16,13 @@
   -->
 
 <template>
-  <div v-if="modalVisible" v-click-outside="onClickOutside" class="modal-mask">
+  <div v-if="modalVisible" :class="modalMaskClass">
```
keithCuniah: with modalMaskClass, you are binding with 2 classes:
 - one with a logical rendering "prevent-scroll"
 - one wich will always rendered "modal-mask"

=> instead, bind only with the class you had a logical rendering

leiyre: yeah, makes sense

keithCuniah (approved):

PR #1905, opened 2022-11-15
perf: improve log from pipeline monitors

### LINKED ISSUE #1727: Optimize pipelines monitoring
When monitoring pipelines (like spacy ner or transformers), records are sent in the background to the Argilla instance every time they are received.

We should include a way to reduce the number of communications with the server, sending data in batches.

frascuchon:
Closes #1727


ufukhurriyetoglu (approved):
there are some code coverage notifications, @frascuchon

PR #1908, opened 2022-11-15
fix(Monitoring): Serializable log middleware

frascuchon:
Also, review and format the class name


src/argilla/monitoring/asgi.py
```diff
@@ -113,18 +113,36 @@ def __init__(
...
-        self, request: Request, call_next: RequestResponseEndpoint
+        self,
+        request: Request,
+        call_next: RequestResponseEndpoint,
     ) -> Response:
+        if not self._initialized:
+            self.init()
+
         if self._endpoint != request.url.path:  # Filtering endpoint path
             return await call_next(request)
```
davidberenstein1957, line 152: The Middleware is serializable but doesn´t allow for requests without a "Content-type" header. If it isn´t present `None` is returned. So the `if "application/json" not in content_type:`will result in an error. 

Additionally, I think we would also like to be able to log GET requests to endpoints that are processed correctly by the FastAPI. 

In both cases, I would say that the error handling must be fore explicit in terms of only letting the logging fail and not the entire requests.

src/argilla/monitoring/asgi.py
```diff
@@ -101,7 +101,7 @@ async def cached_receive() -> Message:
         return self._receive
 
 
-class argillaLogHTTPMiddleware(BaseHTTPMiddleware):
+class ArgillaLogHTTPMiddleware(BaseHTTPMiddleware):
     """An standard starlette middleware that enables argilla logs for http prediction requests"""
 
     def __init__(
```
davidberenstein1957, line 105: I think these variables can be a bit more aligned with the pipeline monitors, wherever possible. I am explicitly talking about `log_interval`and `sample_rate`here.
https://github.com/argilla-io/argilla/pull/1905/files

src/argilla/monitoring/asgi.py
```diff
@@ -67,16 +67,52 @@ def token_classification_mapper(inputs, outputs):
 
 
 def text_classification_mapper(inputs, outputs):
```
davidberenstein1957: @frascuchon I updated the logic for the default mapper

src/argilla/monitoring/asgi.py
```diff
@@ -113,30 +150,60 @@ def __init__(
...
-            scope=request.scope, receive=request.receive, send=request._send
+            scope=request.scope,
+            receive=request.receive,
+            send=request._send,
         )
-        # Must read body before call_next
-        inputs = await cached_request.json()
+
+        # Must read body or query params before call_next
+        if cached_request.method == "POST":
```
davidberenstein1957: I updated the logic to be able to deal with data from PUT and GET requests, however, I think we might need to combine the dict? or is this an edge case?

src/argilla/monitoring/asgi.py
```diff
@@ -113,30 +150,60 @@ def __init__(
...
+            inputs = await cached_request.json()
+            url = str(request.url)
+        elif cached_request.method == "GET":
+            inputs = cached_request.query_params._dict
+            url = str(request.url.remove_query_params(inputs.keys()))
+            request.url.include_query_params()
```
davidberenstein1957: I also added the URL without query params, because the prediction agent has a max length of 64, maybe this can be also set more specific bij selecting the first 64 characters?

src/argilla/monitoring/asgi.py
```diff
@@ -190,14 +258,6 @@ def _log_to_argilla(
         url: str,
         **tags
     ):
-        records = [
```
davidberenstein1957, line 193: This might have been some old logic regarding the usage of the queue? when calling Queue.get() we always get a single item/request from the queue right? To me, it makes more sense to leave potential "batching" behaviour for a potential /predict_batch endpoint to the `_records_mapper`, which can be customized by the users themselves.

src/argilla/monitoring/asgi.py
```diff
@@ -190,14 +256,16 @@ def _log_to_argilla(
...
-            record
-            for _inputs, _outputs in zip(inputs, outputs)
-            for record in [self._records_mapper(_inputs, _outputs)]
-            if record
-        ]
-
-        if records:
+        records = self._records_mapper(inputs, outputs)
```
davidberenstein1957: maybe it makes sense to create a try-except for the records mapper that provides some information about the input-output and that the _records_mapper is causing the error and can be adapted manually.

src/argilla/monitoring/asgi.py
davidberenstein1957: I was also thinking about giving preference of being able to set the prediction agent explicitly like `agent` in `rg.monitor`.

src/argilla/monitoring/asgi.py
```diff
@@ -67,16 +67,52 @@ def token_classification_mapper(inputs, outputs):
...
+                (label, score)
+                for label, score in zip(
+                    outputs.get("labels", []), outputs.get("scores", [])
+                )
+            ],
+            event_timestamp=datetime.datetime.now(),
+        )
+    elif isinstance(inputs, dict):
+        text = inputs.get("text")
+        if text:
```
frascuchon: Here, I would include an extra condition: `len(inputs) == 1`
```suggestion
        if text and len(inputs) == 1:
```

This avoids conflicts when we have the text and other fields in the inputs

src/argilla/monitoring/asgi.py
```diff
@@ -67,16 +67,52 @@ def token_classification_mapper(inputs, outputs):
...
+        return TextClassificationRecord(
+            text="",
+            prediction=[
+                (label, score)
+                for label, score in zip(
+                    outputs.get("labels", []), outputs.get("scores", [])
+                )
+            ],
+            event_timestamp=datetime.datetime.now(),
+        )
```
frascuchon: Does make sense this case? I mean, do we need to log empty texts into argilla? Maybe some kind of user message should be enough. 

What do you think?

src/argilla/monitoring/asgi.py
frascuchon: Yes. This is a very early implementation and some of the behavior found in `rg.monitor` is not placed here. 

+1 for including the `agent` field. But maybe also the dataset `tags` and `metadata` too.

src/argilla/monitoring/asgi.py
frascuchon: The body expected to be found in the request for this middleware is a json-based text. We're more comfortable adding this condition.

We can bypass it (`ignore_content_type` or some similar setting) and be more resilient in `cache_request.json()` statement.

src/argilla/monitoring/asgi.py
```diff
@@ -113,30 +149,60 @@ def __init__(
...
+            receive=request.receive,
+            send=request._send,
         )
-        # Must read body before call_next
-        inputs = await cached_request.json()
+
+        # Must read body or query params before call_next
+        if cached_request.method == "POST":
+            inputs = await cached_request.json()
+            url = str(request.url)
```
frascuchon: This info could be interesting if we have a Fastapi cluster and store it as metadata. Now, the middleware is sending as part of the global dataset tags, meaning that new values will override old ones.

src/argilla/monitoring/asgi.py
frascuchon: I think is fine to keep only the URL with the path, and exclude the query params.  Even more, if we're storing the query params as record inputs.

src/argilla/monitoring/asgi.py
frascuchon: This is exactly what we've already implemented in #1905 

I think that we can reuse the [`DatasetRecordsConsumer`](https://github.com/argilla-io/argilla/pull/1905/files#diff-ee0c8bf2e7a1a161f5d457be5f06188d723f597e3dc39a92f46cff2da6fba5fe) inside this middleware, and handle all logs through it.

src/argilla/monitoring/asgi.py
frascuchon: +1

src/argilla/monitoring/asgi.py
frascuchon: see the comment below

frascuchon: The `PUT` method is still missing, isn't it?

In that case, I think we could do something like this:

```suggestion
        if cached_request.method in ["POST", "PUT"]:
```

frascuchon:
Done @davidberenstein1957 

Now, the middleware is using a `BaseMonitor` object, so the logic to send data to Argilla remains aligned with the `rg.monitor` behavior.

The code has changed a little bit. The main thing here is to provide the records properly in the `_prepare_argilla_data` method (old `log_to_argilla`).

Let me know if you have questions.

frascuchon:
Cool @davidberenstein1957 !!! Great work.

Only a comment. It would be nice if we add a minimal test for the `GET` method part.

question: Do we need to change the record_mapper for the rest of the tasks?

davidberenstein1957:
@frascuchon 

I set the record_mapper to be a required variable to make the usage a bit more explicit and avoid potential questions. So, people always need to define a record mapper, especially since using it with ASGI is going to be usecase dependent. For the tests and as examples, I left the current record_mappers within the asgi file.  

I will add a `GET` test.

src/argilla/monitoring/asgi.py
```diff
@@ -183,21 +207,30 @@ async def _extract_response_content(
...
+                records = [records]
+        except Exception as ex:
+            records = []
+            _logger.error(
+                "Cannot log to argilla. Error in records mapper.", exc_info=ex
+            )
+
+        for record in records:
+            if self._monitor.agent is not None:
+                record.prediction_agent = self._monitor.agent
```
frascuchon: In case that record mapper already set the agent:
```suggestion
            if self._monitor.agent is not None and not record.prediction_agent:
                record.prediction_agent = self._monitor.agent
```

PR #1915, opened 2022-11-17
View Docs link in frontend header `users.vue`

bengsoon:
Hi I just noticed that the "View Docs" link in the header of the Argilla's frontend interface currently leads to the old Rubrix documentation. 

Proposal to change `https://docs.rubrix.ml/en/stable/` to `https://docs.argilla.io/en/latest/`


leiyre (approved):
Thanks @bengsoon! 💯​

PR #1919, opened 2022-11-17
fix: dataset regex pattern, allow upper case

### LINKED ISSUE #1918: Upper case letters are not allowed in dataset name
**Describe the bug**
Dataset's name is invalid, because of upper letters. I assume it was not intended.

**To Reproduce**

```python
record = rg.TextClassificationRecord(text="Some text")
rg.log(record, "DS_NAME")
```

> ```powershell
> argilla.client.api.InputValueError: Provided dataset name ALtempdataset does not match
> the pattern ^(?!-|_)[a-z0-9-_]+$. Please, use a valid name for your dataset
> ```

jamnicki:
Fixes #1918


frascuchon:
Hi @jamnicki and thanks for the contribution

We cannot support uppercased dataset names without decoupling them from the elasticsearch index creation. Elasticsearch introduces this limitation in the index naming and we're using the dataset name as part of the elasticsearch index. 

We need first to decouple that relation and then relax the name pattern validation for the dataset name.

PR #1924, opened 2022-11-18
feat: Allow sort records by event_timestamp or last_updated fields

### LINKED ISSUE #1835: [Sort] Allow sort value by event_timestamp or last_updated fields
API is ready to sort at least by `event_timestamp` . So you can proceed for this field. Also , we can tackle the `last_updated` field to retrieve records sorted by recently updated or created.

leiyre:
closes #1835

This PR includes the fields event_timestamp and last_updated in the sort filter allowing to sort records by these values

### DIFF
--- a/frontend/components/commons/header/filters/FiltersList.vue
+++ b/frontend/components/commons/header/filters/FiltersList.vue
@@ -299,11 +299,11 @@ export default {
       this.$emit("applySortBy", sortList);
       this.close();
     },
-    sortByDateFilter(id, name) {
+    sortByDateFilter(id, name, group = "Sort") {
       return {
-        id: id,
+        id,
         key: id,
-        group: "Sort",
+        group: group,
         name: name,
         disabled: this.recordPropertyHasValue(id),
       };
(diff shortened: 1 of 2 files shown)


frontend/components/commons/header/filters/FiltersList.vue
```diff
@@ -235,6 +235,22 @@ export default {
...
+          disabled: this.recordPropertyHasValue("last_updated"),
+        },
+        {
+          id: "event_timestamp",
+          key: "event_timestamp",
+          group: "Sort",
+          name: "Event timestamp",
+          disabled: this.recordPropertyHasValue("event_timestamp"),
+        },
+      ].filter(({ disabled }) => !disabled);
```
keithCuniah, line 241: code cleanning : there is a duplicated code to create the array => use a method with a property to difene wich object to return (with last_update or event_timestamp

frontend/components/commons/header/filters/FiltersList.vue
```diff
@@ -290,6 +299,20 @@ export default {
...
       this.close();
     },
+    sortByDateFilter(id, name) {
+      return {
+        id: id,
+        key: id,
+        group: "Sort",
+        name: name,
+        disabled: this.recordPropertyHasValue(id),
+      };
```
keithCuniah, line 309: 1/ shortcut: in es6, there is a shortcut in case the key and the value from object are the same => you can write return {id, key: id, group: "Sort", name, disabled: this.recordPropertyHasValue(id),}
2/ put `group= "Sort"` as an optional parameter in the function argument

keithCuniah (approved):

PR #1930, opened 2022-11-21
feat: #1929 add warning about using wrong hostnames

### LINKED ISSUE #1929: feat: provide an userwarning about using underscores in hostnames
HTTPX cannot handle this during SSL verification and it is bad practice https://stackoverflow.com/questions/10959757/the-use-of-the-underscore-in-host-names

A warning, when connecting to a wrong hostname.

N.A.

**Additional context**
N.A.

davidberenstein1957:
closes #1929

### DIFF
--- a/src/argilla/client/sdk/client.py
+++ b/src/argilla/client/sdk/client.py
@@ -50,6 +50,12 @@ class _Client:
         self.base_url = self.base_url.strip()
         if self.base_url.endswith("/"):
             self.base_url = self.base_url[:-1]
+        if "_" in self.base_url:
+            self.__httpx__ = None
+            raise ValueError(
+                'Avoid using hostnames with underscores "_". For reference see:'
+                " https://stackoverflow.com/questions/10959757/the-use-of-the-underscore-in-host-names"
+            )
 
 
 @dataclasses.dataclass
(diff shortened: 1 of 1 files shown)


src/argilla/client/sdk/client.py
```diff
@@ -50,6 +50,12 @@ def __post_init__(self):
         self.base_url = self.base_url.strip()
         if self.base_url.endswith("/"):
             self.base_url = self.base_url[:-1]
+        if "_" in self.base_url:
+            self.__httpx__ = None
+            raise ValueError(
+                'Avoid using hostnames with underscores "_". For reference see:'
+                " https://stackoverflow.com/questions/10959757/the-use-of-the-underscore-in-host-names"
+            )
```
frascuchon: Doing some tests, I've seen that we can customize the SSL verification context. Maybe for those cases, we can just disable the hostname verification:

```python
import ssl
from urllib.parse import urlparse

argilla_url =  urlparse("https://test_httpx.argilla.io")

ssl_context = ssl.create_default_context()
if "_" in argilla_url.hostname:
   ssl_context.check_hostname = False

httpx = Client(...,verify=ssl_context)
```
Also, not sure if this check will be applied to regular HTTP requests (I suppose the SSL verify is used ONLY for secure requests (HTTPS)

```python
argilla_url.scheme == "https"
```

davidberenstein1957: I know it is possible, but I'm not sure if we want to allow for this, even with a warning. Assume a random user deploys Argilla and we automatically disable SSL hidden away in our source code and they might miss the logging and get a security leak.

frascuchon: Yes, I understand and agree with that.

So let's raise the `ValueError` ONLY for secure connections. If users are using a non-secure (HTTP) connection, we should let them connect even if the hostname (and maybe only the hostname) includes an underscore.

PR #1965, opened 2022-11-25
feat: Rules management for token classification weak labelling

### LINKED ISSUE #1749: Token classification weak labelling
As in the text classification task, the **"Weak labeling"** mode in token classification must allow tag entities by defining a query and entity label (the rule).

Given a rule, the **Weak labeling** mode for token classification will tag entities based on the matched tokens/words in the search results returned by the API.

The way the entity will be tagged from the matched token will be determined by a **labeling function** provided as an attribute of the rule. For now, just one single labeling function will be supported, the `exact_match`, where all matched tokens/words will be tagged as the provided rule.

For example, given a labeling rule with the query `Par*`, the label `PLACE`, and the matched record `Paris is the city of light`, the labeling function will tag the token `Paris` as a `PLACE`.

An important behavior of this feature is to provide a visualization of the tagged entities in the visible records from the UI.

frascuchon:
PR for adding weak labelling feature in TokenClassification

I did some refactoring:
 - create a component for the tooltip functionnality
 - create a generic component for the RulesMetricToken Component which will be reused in the TextClassification

Closes #1749


frascuchon:
Sorry, @keithCuniah. I would like to relaunch the build since it was stopped but I needed to recreate a new PR.

Please, close this PR and create a new one using the same branch, so you will be the final commit author once the PR is merged. 

Also, please review the failing js tests 
https://github.com/argilla-io/argilla/actions/runs/3551291314/jobs/5965346489#step:6:605

PR #1975, opened 2022-11-28
fix: Compute predicted properly for token classification [NEEDS_DATA_UPGRADE]

### LINKED ISSUE #1955: Predicted filter not working for token classification
**Describe the bug**
When querying for `ko` on a token classification set, the client returns many records for which all tokens have been correctly tagged. 
 
**To Reproduce**
1. Compile a dataset for a tokenclassification task, including annotations and predictions and log it to argilla. Make sure to  also include a score for the predictions.  
2. Load all records with query="predicted:ko" 
3. Verify that this will include records for which all tokens have been correctly predicted. 
4. Verify that the ko records only include those records without any prediction or annotation.

**Expected behavior**
I would expect to see only records with at least one wrongly predicted token.

**Screenshots**
image

 - OS [e.g. iOS]: 
 - Argilla Version [e.g. 1.0.0]: 1.0.1

**Additional context**

frascuchon:
This PR fixes the way predicted ok/ko info is computed for token classification records.

To apply this fix to already created datasets, you must first re-log records. Otherwise, stored info won't be updated.

Closes #1955

### DIFF
--- a/src/argilla/server/services/tasks/token_classification/model.py
+++ b/src/argilla/server/services/tasks/token_classification/model.py
@@ -145,11 +145,21 @@ class ServiceTokenClassificationRecord(
     @property
     def predicted(self) -> Optional[PredictionStatus]:
         if self.annotation and self.prediction:
-            return (
-                PredictionStatus.OK
-                if self.annotation.entities == self.prediction.entities
-                else PredictionStatus.KO
-            )
+
+            annotated_entities = self.annotation.entities
+            predicted_entities = self.prediction.entities
+            if len(annotated_entities) != len(predicted_entities):
+                return PredictionStatus.KO
+
+            for ann, pred in zip(annotated_entities, predicted_entities):
+                if (
+                    ann.start != pred.start
+                    or ann.end != pred.end
+                    or ann.label != pred.label
+                ):
+                    return PredictionStatus.KO
+
+            return PredictionStatus.OK
         return None
 
     @property
--- a/tests/server/token_classification/test_model.py
+++ b/tests/server/token_classification/test_model.py
@@ -335,3 +335,48 @@ def test_whitespace_in_tokens():
     record = ServiceTokenClassificationRecord.parse_obj(record)
     assert record
     assert record.tokens == ["every", "four", "(", "4", ")", " "]
+
+
+def test_predicted_ok_ko_computation():
+
+    text = "A text with some empty spaces that could bring not cleanly annotated spans"
+    record = ServiceTokenClassificationRecord(
+        text=text,
+        tokens=text.split(),
+        prediction=TokenClassificationAnnotation(
+            agent="pred.test",
+            entities=[
+                EntitySpan(
+                    start=0,
+                    end=6,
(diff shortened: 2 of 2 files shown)


frascuchon:
@dvsrepo @davidberenstein1957 

This issue introduces a need to recreate or update already stored dataset records. Do you see a new feature here for this kind of data update?

davidberenstein1957:
@frascuchon I am not entirely sure what you mean, but are you referring to some kind of upsert method? This afternoon we got a mention about this in Slack, and I thought Dani had a fix for that but maybe it is limited to metadata changes?

dvsrepo:
> @frascuchon I am not entirely sure what you mean, but are you referring to some kind of upsert method? This afternoon we got a mention about this in Slack, and I thought Dani had a fix for that but maybe it is limited to metadata changes?

@davidberenstein1957 what I have shared this morning works for every field in the record which is not computed during indexing time (`predicted`, `predicted_as`, `metrics` and a few others). 

I think @frascuchon is referring to a feature for recomputing these computed fields with a new version (for example this fix). In my view, this is not worth the effort, I'd be happy with just notifying the user about the need to reindex and also include this in the release notes.

frascuchon:
Not really an upstert. This change is done in the record data model class. Users must re-log stored records to propagate this and store the correct predicted ok/ko computation. Otherwise, the `predicted:ko` still brings unexpected records.

A way to do this is :

```python
import argilla as rg

ds = rg.load("dataset") 
rg.log(ds, "dataset") 
# Even if we do nothing, the records will be updated with the proper predicted ok/ko computation
```

davidberenstein1957:
Clear, I understand the difference. However, I am not sure if it ought to be a feature or a specific part in the docs.

frascuchon:
Where is the better place to put this kind of required task?

cc @davidberenstein1957 @dvsrepo

davidberenstein1957:
@frascuchon I believe streamlit has some kind of logging mechanism that is triggered when there is a newer release available with relevant updates/bug fixes. I am not sure how often this occurs but it might be relevant.

davidberenstein1957:
Alternatively, we might publish our release notes in our docs along with a section about breaking changes, bugfixes etc?

frascuchon:
As a solution for this PR:

TODOs:
- [ ] Find a section place for the dataset update/reindex @davidberenstein1957 
- [ ] Refer to the docs in the issue changelog entry


cc @dvsrepo @davidberenstein1957

frascuchon:
Hi @davidberenstein1957.


The section in docs including the data upgrade should contain a snippet as follows:
```python
import argilla as rg

dataset = "my-outdated-dataset"
ds = rg.load(dataset) 
rg.log(ds, dataset) 
```

Could you tackle this directly on this PR, please?

davidberenstein1957:
Thanks, does Elastic actually re-index if you log the same record? Or do they need to be logged to a new index.

https://discuss.elastic.co/t/is-there-any-best-way-to-update-existing-index-without-re-indexing-and-creating-alias-of-index/207622/2

PR #1998, opened 2022-12-02
feat: Similarity vector search (#1768)

### LINKED ISSUE #1757: Similarity search with vectors

frascuchon:
This is the main PR for the new feature #1757 

# TODO (Backend)
- [x] Apply vector setup optimization for elasticsearch/opensearch 
version and opensearch >= 2.2)
- [x] Adapt client data model -> https://github.com/argilla-io/argilla/pull/2008
- [x] Accept vector search from python client -> https://github.com/argilla-io/argilla/pull/2010
- [x] Name normalization for client and backend (`embedding` to `vector`)  -> https://github.com/argilla-io/argilla/pull/2009


docker-compose.yaml
```diff
@@ -19,37 +20,36 @@ services:
...
     networks:
       - argilla
-
 networks:
   argilla:
     driver: bridge
-
 volumes:
-  elasticdata:
+  elasticdata85:
```
frascuchon: Leaving the default volume requires an internal elasticsearch migration (from 7.10 to 8.5 cannot be done automatically). 

Changing to a new volume will hide all existing data.

We need to decide on the final option and document it

frascuchon: cc @dvsrepo @davidberenstein1957

dvsrepo:
[EDITED]
Ok, I see now this is a limit from Elasticsearch. Not sure why in some other cases it shows a clearer message (shows this limit when doing the ES mapping). In any case:

* we need to add this limitation in the docs (I left warning blocks for this)
* For Cohere and OpenAI I'll recommend the smaller models to be sure it works.

Question: Do you know where to find the official docs for these limits in ES? Is the same for Opensearch? Changes from one version to another?

@frascuchon:

Testing this feature with OpenAI embeddings, which are very large (>1200 dimensions) fails with a generic server error and no further info. I have tested with truncating to 1000 dimensions and it works.

I have pushed the dataset to  the Hub for testing: https://huggingface.co/datasets/argilla/banking77_openai_embeddings

The error stack trace can be found here: https://github.com/argilla-io/argilla/blob/8178542fccb57d908c2ec5edc91cf21f0cc653f5/docs/_source/guides/features/similarity-search.ipynb

I'm using the elasticsearch container defined in the `docker-compose` file of that branch

docker-compose.opensearch.yaml
```diff
@@ -0,0 +1,30 @@
...
+      memlock:
+        soft: -1
+        hard: -1
+    ports:
+      - "9200:9200"
+    networks:
+      - argilla
+    volumes:
+      - opensearch-data:/usr/share/opensearch/data
+
```
frascuchon, line 37: Add opensearch-dashboards setup:

```yaml
opensearch-dashboards:
    image: opensearchproject/opensearch-dashboards:latest # Make sure the version of opensearch-dashboards matches the version of opensearch installed on other nodes
    container_name: opensearch-dashboards
    ports:
      - 5601:5601 # Map host port 5601 to container port 5601
    expose:
      - "5601" # Expose port 5601 for web access to OpenSearch Dashboards
    environment:
      OPENSEARCH_HOSTS: '["https://opensearch-node1:9200","https://opensearch-node2:9200"]' # Define the OpenSearch nodes that OpenSearch Dashboards will query
    networks:
      - opensearch-net
```

frascuchon:
> [EDITED] Ok, I see now this is a limit from Elasticsearch. Not sure why in some other cases it shows a clearer message (shows this limit when doing the ES mapping). In any case:
> 
> * we need to add this limitation in the docs (I left warning blocks for this)
> * For Cohere and OpenAI I'll recommend the smaller models to be sure it works.
> 
> Question: Do you know where to find the official docs for these limits in ES? Is the same for Opensearch? Changes from one version to another?


Sorry @dvsrepo, I've seen this message now.

For elasticsearch, the official docs talk about a max dimension of **1024** for indexed vectors (our case) See https://www.elastic.co/guide/en/elasticsearch/reference/current/dense-vector.html#dense-vector-params

For Opensearch, the official docs mentions similar limit for lucene engines (what we're using). See [this section](https://opensearch.org/docs/latest/search-plugins/knn/approximate-knn/#get-started-with-approximate-k-nn):

> The knn_vector data type supports a vector of floats that can have a dimension of up to 10000 for the nmslib and faiss engines, as set by the dimension mapping parameter. The maximum dimension for the Lucene library is 1024.

But, maybe the error message could not be the same

frascuchon:
> > [EDITED] Ok, I see now this is a limit from Elasticsearch. Not sure why in some other cases it shows a clearer message (shows this limit when doing the ES mapping). In any case:
> > 
> > * we need to add this limitation in the docs (I left warning blocks for this)
> > * For Cohere and OpenAI I'll recommend the smaller models to be sure it works.
> > 
> > Question: Do you know where to find the official docs for these limits in ES? Is the same for Opensearch? Changes from one version to another?
> 
> Sorry @dvsrepo, I've seen this message now.
> 
> For elasticsearch, the official docs talk about a max dimension of **1024** for indexed vectors (our case) See https://www.elastic.co/guide/en/elasticsearch/reference/current/dense-vector.html#dense-vector-params
> 
> For Opensearch, the official docs mentions similar limit for lucene engines (what we're using). See [this section](https://opensearch.org/docs/latest/search-plugins/knn/approximate-knn/#get-started-with-approximate-k-nn):
> 
> > The knn_vector data type supports a vector of floats that can have a dimension of up to 10000 for the nmslib and faiss engines, as set by the dimension mapping parameter. The maximum dimension for the Lucene library is 1024.
> 
> But, maybe the error message could not be the same

The latest [doc section](https://docs.argilla.io/en/feature-similarity-vector-search/guides/features/semantic-search.html#Semantic-search) does not include the dimension limit reference. We could include warnings directly here. 

@dvsrepo could you share with us where to add those warnings?

dvsrepo:
> > > [EDITED] Ok, I see now this is a limit from Elasticsearch. Not sure why in some other cases it shows a clearer message (shows this limit when doing the ES mapping). In any case:
> > > 
> > > * we need to add this limitation in the docs (I left warning blocks for this)
> > > * For Cohere and OpenAI I'll recommend the smaller models to be sure it works.
> > > 
> > > Question: Do you know where to find the official docs for these limits in ES? Is the same for Opensearch? Changes from one version to another?
> > 
> > 
> > Sorry @dvsrepo, I've seen this message now.
> > For elasticsearch, the official docs talk about a max dimension of **1024** for indexed vectors (our case) See https://www.elastic.co/guide/en/elasticsearch/reference/current/dense-vector.html#dense-vector-params
> > For Opensearch, the official docs mentions similar limit for lucene engines (what we're using). See [this section](https://opensearch.org/docs/latest/search-plugins/knn/approximate-knn/#get-started-with-approximate-k-nn):
> > > The knn_vector data type supports a vector of floats that can have a dimension of up to 10000 for the nmslib and faiss engines, as set by the dimension mapping parameter. The maximum dimension for the Lucene library is 1024.
> > 
> > 
> > But, maybe the error message could not be the same
> 
> The latest [doc section](https://docs.argilla.io/en/feature-similarity-vector-search/guides/features/semantic-search.html#Semantic-search) does not include the dimension limit reference. We could include warnings directly here.
> 
> @dvsrepo could you share with us where to add those warnings?

Added to OpenAI and cohere sections but maybe it's a good idea to add somewhere earlier too? https://docs.argilla.io/en/feature-similarity-vector-search/guides/features/semantic-search.html#OpenAI-Embeddings

Please feel free to add it somewhere else you see fit, better to be repetitive 😅

Issue #1928, opened 2022-11-21
Show scroll bar into dropdown


Amelie-V:
**Is your feature request related to a problem? Please describe.**
Adding two new items in the sort dropdown, we detect the lack of evidence we need to scroll to view the whole list.
(view screenshot)

**Describe the solution you'd like**
Show scroll bar in all dropdowns from the filter bar: With/without checkbox
Don't active scroll when there are until 6 items (to avoid behavior viewed in video)

**Describe alternatives you've considered**

**Additional context**
6 items are showing before the scroll, active scrollbar from seven items

Amelie-V:
<img width="433" alt="Captura de pantalla 2022-11-21 a las 12 05 42" src="https://user-images.githubusercontent.com/84135211/203035869-da79a8f7-15ab-4f5f-aeaf-12545e479341.png">

Amelie-V:
https://user-images.githubusercontent.com/84135211/203036464-a3ffb7a5-b6f8-469a-87eb-51e8baf61fa6.mov

Amelie-V:
@leiyre, please give to @keithCuniah  UI material in figma

frascuchon:
Closed by #2000

PR #2000, opened 2022-12-02
feat(Client): remove mixins to hide scroll bar in drop down

keithCuniah:
refs #1928

### DIFF
--- a/frontend/components/commons/header/filters/SelectOptions.vue
+++ b/frontend/components/commons/header/filters/SelectOptions.vue
@@ -109,7 +109,7 @@ export default {
   hyphens: auto;
 }
 ul {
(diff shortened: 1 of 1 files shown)


frascuchon:
@leiyre, this should be typed as `style` or it's fine as `feat`?

leiyre:
@frascuchon I think better style

leiyre (approved):

leiyre:
@keith we need also to change a little the max-height of the scroll to prevent the behavior that Amelie said in the issue:

> Don't active scroll when there are until 6 items (to avoid behavior viewed in video)

leiyre (approved):

frascuchon (approved):

PR #2001, opened 2022-12-02
Refactor/convert embedding name convention to vector 

ufukhurriyetoglu:
Draft PR for renaming the embeddings to vectors and updating the data models as requested in Notion spec


src/argilla/server/daos/backend/search/model.py
```diff
@@ -74,13 +74,11 @@ class BaseRecordsQuery(BaseQuery):
...
-    # TODO(@ufukhurriyetogl): In order to allow embedding-based searches, we can include here
-    #  (the base records query data model class) an embeddings related query params that
+    # TODO(@ufukhurriyetoglu): In order to allow vector searches, we can include here
+    #  (the base records query data model class) a vectors related query params that
     #  we can use later for the records search.
     # The input vector:
-    embedding_vector: Optional[List[float]] = None
-    # The name of the embedding to uses (if several):
-    embedding_name: Optional[str] = None
+    # query: Optional[VectorQuery]
```
ufukhurriyetoglu, line 81: not sure if I need to use the VectorQuery data model here, wdyt @frascuchon

src/argilla/server/daos/models/records.py
```diff
@@ -66,11 +64,11 @@ class BaseRecordInDB(GenericModel, Generic[AnnotationDB]):
...
-    embeddings: Optional[Dict[str, BaseEmbeddingVectorDB]] = Field(
+    vectors: Optional[Dict[str, List[float]]] = Field(
         None,
-        description="Provide the embedding info as a list of key - value dictionary."
-        "The dictionary contains the dimension and dimension sized embedding float list"
-        "Using this way you can skip passing the agent inside of the embedding",
+        description="Provide the vector information as a list of key - value dictionary."
+        "The dictionary contains the dimension and dimension sized float list"
+        "Using this way you can skip passing the agent inside of the vector",
     )
```
ufukhurriyetoglu, line 72: I could not figure out how to convert the client model to backend model, 

in client we have

vectors = {"my_bert": [....]}

and in backend we defined VectorDB, which will in turn provide

vectors: {"my_bert": {"value": [...]}}

ufukhurriyetoglu: @frascuchon can you check if we need such client to backend transformation and point out how to do ?

src/argilla/server/daos/backend/search/model.py
frascuchon: We should have a `VectorQuery` modeling the vector query parameters. But not should be the same class for the client and server. The client and server should have separate data models even if they share the same schema.

So, create here the new class and add it as a property in the query search model.

src/argilla/server/daos/models/records.py
frascuchon: We've already those kind of transformations for predictions and/or annotations. Client and backed has different schemas. So, the SDK classes apply the transformations: 

See [here](https://github.com/argilla-io/argilla/blob/2ff3b959d69b8212b18b50ea66f02465fa66d63b/src/argilla/client/sdk/text_classification/models.py#L56) and [here](https://github.com/argilla-io/argilla/blob/2ff3b959d69b8212b18b50ea66f02465fa66d63b/src/argilla/client/sdk/text_classification/models.py#L99) for two examples

frascuchon (changes requested):

frascuchon:
Won't APPLY

PR #2010, opened 2022-12-07
feat: Enable vector search in python cli

frascuchon:
Refs #1998


davidberenstein1957:
@frascuchon I think there might be a versioning problem here.

```
try:
    import ujson as json
except ModuleNotFoundError:
    import json
```

PR #2057, opened 2022-12-16
ci: optimize docker images

rhoitjadhav:
# Description
Improve the build process for the argilla-server docker image.

**Checklist**


scripts/start_argilla_server.sh
```diff
@@ -0,0 +1,15 @@
...
+
+if [[ "$status_code" -ne 200 ]]; then
+  echo "ElasticSearch connection error. Returned status code: $status_code"
+  exit 1
+else
+  echo "ElasticSearch connected"
+fi
+
+# Run argilla-server
+uvicorn argilla:app --host "0.0.0.0" --port 6900
```
frascuchon, line 15: Since we're fixing the port here, could you also add the `EXPOSE` statement in dockerfile?

Also, we need to update the provided [docker-compose.yml](https://github.com/argilla-io/argilla/blob/ci/optimize-docker-images/docker-compose.yaml#L8) and update the mapped ports for argilla. Please, review it and do the proper changes.

frascuchon (changes requested):

scripts/start_argilla_server.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/bash
+set -e
+
+# Check elasticsearch response status
+status_code=$(curl -s -o /dev/null -w "%{http_code}" "$ARGILLA_ELASTICSEARCH")
```
frascuchon, line 5: Do we still need this step? The `wait-for-it.sh` script should finish once the elasticsearch is up and running, right? So, the start script will be run once the elasticsearch is ready.

scripts/start_argilla_server.sh
rhoitjadhav: Made the changes

scripts/start_argilla_server.sh
rhoitjadhav: I think this is an extra check if the elasticsearch is returning the expected output or not. But please correct if any

frascuchon (approved):

PR #2082, opened 2022-12-20
build: Some argilla-server docker image improvements

frascuchon:
# Description

- Handle wait-for-it call inside the start command
- Using a slim version of the base image
- Upgrade the python version to 3.9
- Build fails if pip install fails
- Pre-download the wait-for-it.sh shell script

**Type of change**

CI/BUILD

**Checklist**

### DIFF
--- a/scripts/start_argilla_server.sh
+++ b/scripts/start_argilla_server.sh
@@ -4,5 +4,11 @@ set -e
 # Check elasticsearch response status
 /wait-for-it.sh -t 60 -s `echo "${ARGILLA_ELASTICSEARCH}" | awk -F'^http[s]?://' '{print $2}'` -- echo "ElasticSearch connected"
 
-# Run argilla-server
+# Run argilla-server (See https://www.uvicorn.org/settings/#settings)
+#
+# From uvicorn docs:
+#   You can also configure Uvicorn using environment variables
+#   with the prefix UVICORN_. For example, in case you want to
+#   run the app on port 5000, just set the environment variable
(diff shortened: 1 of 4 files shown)


scripts/start_argilla_server.sh
```diff
@@ -2,14 +2,7 @@
...
-  echo "ElasticSearch connection error. Returned status code: $status_code"
-  exit 1
-else
-  echo "ElasticSearch connected"
-fi
+/wait-for-it.sh -t 60 -s `echo "${ARGILLA_ELASTICSEARCH}" | awk -F'^http[s]?://' '{print $2}'` -- echo "ElasticSearch connected"
 
 # Run argilla-server
-uvicorn argilla:app --host "0.0.0.0" --port 6900
+uvicorn argilla:app --host "0.0.0.0"
```
rhoitjadhav, line 14: Should we pass UVICORN_PORT as a parameter here?

frascuchon: The UVICORN_PORT is automatically read by the uvicorn command. So, we don't need it. Maybe a comment referring to uvicorn docs.  Could you add the comment, please? https://www.uvicorn.org/settings/#settings

From docs:

> You can also configure Uvicorn using environment variables with the prefix UVICORN_. For example, in case you want to run the app on port 5000, just set the environment variable UVICORN_PORT to 5000.

rhoitjadhav: Got it thanks

PR #2091, opened 2022-12-22
perf: Optimize word cloud computation

frascuchon:
# Description

Sometimes, for huge datasets, the word cloud metric can introduce some problems loading the dataset. This PR introduces a better way to compute it and avoid those situations.

Also:
- Remove stopwords-iso deps
- Create the multi-lingual stop analyzer using standard filters (All but English. See https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-stop-tokenfilter.html#analysis-stop-tokenfilter)
- Remove refs to unused field `words`

Refs #1757 

**Checklist**

### DIFF
--- a/src/argilla/server/daos/backend/mappings/helpers.py
+++ b/src/argilla/server/daos/backend/mappings/helpers.py
@@ -14,13 +14,12 @@
 
 from typing import Any, Dict, List
 
+from argilla.server.daos.backend.mappings.stopwords import english
 from argilla.server.settings import settings
 
 EXTENDED_ANALYZER_REF = "extended_analyzer"
 
 MULTILINGUAL_STOP_ANALYZER_REF = "multilingual_stop_analyzer"
-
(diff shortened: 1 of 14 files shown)


src/argilla/server/daos/backend/mappings/helpers.py
```diff
@@ -14,14 +14,13 @@
...
 from argilla.server.settings import settings
 
 EXTENDED_ANALYZER_REF = "extended_analyzer"
 
 MULTILINGUAL_STOP_ANALYZER_REF = "multilingual_stop_analyzer"
 
-DEFAULT_SUPPORTED_LANGUAGES = ["es", "en", "fr", "de"]  # TODO: env var configuration
-
 
 class mappings:
```
rhoitjadhav, line 25: Class name should be upper case?

frascuchon: Yes, we will tackle it in another place. This class contains only static methods, so maybe can be converted into a module itself.

rhoitjadhav: Got it

PR #2245, opened 2023-01-24
feat: better log error handling

### LINKED ISSUE #2005: too verbose failure logging during failed `rg.log()` (with too large chunk_size)
**Describe the bug**
The terminal gets bombarded with the complete records after failed logging attempts. When logging a lot of records or larger records, this is problematic because it freezes the IDE and doesn´t really stop.

Also the `rg.log` generally failed a lot with larger chunk_sizes, however, instead of failing so easily, I would initially expect this to just influence speed and only fail after a relatively large timeout (60s?).

**To Reproduce**
1. Go to Any arbitrary large dataset with larger records for example `news-fakenews`within our pre-release.
2. upload data via `rg.log()`with a chunk size larger than 100
3. it starts failing after a while

**Expected behavior**
I would expect to get a message `# records with IDs x,y,z have failed` to process. Or a response item that contains the failed / processed record info.

**Screenshots**
N.A.

 - OS [e.g. iOS]: iOS
 - Browser [e.g. chrome, safari]: Firefox
 - Argilla Version [e.g. 1.0.0]: 1.1.1.
 - ElasticSearch Version [e.g. 7.10.2]:7.10.2
 - Docker Image (optional) [e.g. argilla:v1.0.0]: 1.1.1

**Additional context**
Not the finicky failure behaviour of the logging with larger chunk sizes.

frascuchon:
# Description

This PR adds code to better error parsing for bulk index errors, and other HTTP errors in general:

- For bulk errors, return the reason and the causes of the errors for each failed record in the bulk.
- Reduce args info when an error is raised from a `rg.log`
- Parse HTTP status codes and responses other than those defined by the Argilla server.

Closes #2005 

**Type of change**

Please delete options that are not relevant.

- [x] Improvement

**Checklist**

### DIFF
--- a/src/argilla/client/client.py
+++ b/src/argilla/client/client.py
@@ -92,7 +92,7 @@ class _ArgillaLogAgent:
             dataset = kwargs["name"]
             records = kwargs["records"]
             _LOGGER.error(
-                f"\nCannot log in dataset '{dataset}'\n records '{str(records)[:100]}...'\n"
+                f"\nCannot log data in dataset '{dataset}'\n"
                 f"Error: {type(ex).__name__}\n"
                 f"Details: {ex}"
             )
--- a/src/argilla/client/client.py
+++ b/src/argilla/client/client.py
@@ -90,7 +90,6 @@ class _ArgillaLogAgent:
             return await api.log_async(*args, **kwargs)
         except Exception as ex:
             dataset = kwargs["name"]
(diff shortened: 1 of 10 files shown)


davidberenstein1957 (approved):
LFTM besides some fine-tuning handled [here](https://github.com/argilla-io/argilla/issues/2260).

PR #2291, opened 2023-02-03
feat(Client): Add a new component for multi-label text classification bulk annotation

### LINKED ISSUE #2267: Add a new component for multi-label text classification bulk annotation

keithCuniah:
# Description

Components : 
- [x] TextClassificationBulkAnnotation 
- [x] BulkAnnotation 
- [x] BulkAnnotationForm

Note : 
- I removed the `remove` attribute from the form since we can guess which annotation will be remove only with the `selected` attribute, I will update the doc
- refactoring for later : it could be a nice to have a builder design pattern to manage forms 

What is left : 
- [x] use Vuelidate plugin to manage disable enable submit and cancel buttons (I didn't use vuelidate, since only one computed was enough)
- [x] disable state of the dropdown button
- [x] implement the search in the BulkAnnotation component ( **bug to correct : need to keep the previous selection after search** )
- [x] Style
- [x] unit test
- [x] update documentation

Closes #2267 

**Type of change**

- [ ] Documentation update

**How Has This Been Tested**

- [x] TextClassification  multilabel

**Checklist**


leiyre:
One comment, in the case of a label already applied to some records, the icon of the cross and the label have different behavior. 
Let's imagine that we have POSITIVE (2), in this case the cross must be visible, when we click on the cross we have to delete the annotation, when we click on the label we have to apply the label to all the selected records. The number of records applied should change, if we click on the cross, we remove the record number and if we apply the label we need to show the number of all selected records.

If we have 20 records selected in a page:

X` POSITIVE (2)`: If I click the cross existing labels must be deleted. Label looks like =>  `POSITIVE`
X `POSITIVE (2)`: If I click the label existing labels must be applied to all records. Label looks like =>  X `POSITIVE (20)`

frontend/components/commons/header/global-actions/text-classification-bulk-annotation/textClassificationBulkAnnotation.spec.js
```diff
@@ -0,0 +1,151 @@
+import { shallowMount } from "@vue/test-utils";
```
leiyre: the whole text-classifier-bulk-annotation folder is better to be outside of commons, in the textClassifier folder

frascuchon:
I've been doing some testing in dev and I've seen some weird behavior for single-label:

**1. Discard a validated record**
Screenshot 2023-02-07 at 13 23 17

Screenshot 2023-02-07 at 13 23 22

**2. Enabling by clicking the selected label let the record empty, and no pending status is shown:**
Screenshot 2023-02-07 at 13 23 30

cc @keithCuniah @leiyre

leiyre:
@frascuchon your example for me is the correct behavior, we emit a reset in case of single label with empty labels (it's not new behavior and we can discuss). I see some things that don't fit, there is supposed to be no pending status in the single Text Classifier Classifier,  so for me the "clear" button doesn't make sense for this case. I will comment in specs.

leiyre (approved):

Issue #1843, opened 2022-03-31
Use Rich for logging, stack traces, and progress bars


dcfidalgo:
I propose to substitute tqdm and loguru with the rich library:
- use rich to format the stack traces
- use rich for the progress bars
- use rich to format our logging
- (optional) add a specific logging handler for notebooks
- (optional) hide certain stack traces in the notebook

tomaarsen:
Some sample outputs when Rich is used:
### Loading and logging `gutenberg_spacy-ner`
#### Rich:
![image](https://user-images.githubusercontent.com/37621491/218731397-1d90099d-440d-46ff-ac30-d214cadaf5c9.png)
#### Current (with Loguru & tqdm):
![image](https://user-images.githubusercontent.com/37621491/218732160-6db8a097-5366-4877-b304-3936e490c1ce.png)


### Providing an incorrect value for initialising a Record
#### Rich:
![image](https://user-images.githubusercontent.com/37621491/218731844-ffff5802-f1ca-466a-9165-7804916e0bb3.png)
#### Current:
![image](https://user-images.githubusercontent.com/37621491/218732279-f810d3bb-d3b6-4ca7-817b-ac2d4b0659f9.png)

I'd love to hear your thoughts on the before and after.

davidberenstein1957:
For me, rich is more readable in both cases.

dvsrepo:
Same for me!

tomaarsen:
Alright, wonderful. I'll prepare a PR.

PR #2329, opened 2023-02-09
Enhancement: Distinguish between error message and context in validation of spans

tomaarsen:
# Description

I expanded slightly on the error message provided when providing spans that do not match the tokenization. 
Consider the following example script:
```python
import argilla as rg

record = rg.TokenClassificationRecord(
    text = "This is my text",
    tokens=["This", "is", "my", "text"],
    prediction=[("ORG", 0, 6), ("PER", 8, 14)],
)
```
The (truncated) output on the `develop` branch:
```
ValueError: Following entity spans are not aligned with provided tokenization
Spans:
- [This i] defined in This is my ...
- [my tex] defined in ...s is my text
Tokens:
['This', 'is', 'my', 'text']
```
The distinction between `defined in` and `This is` is unclear. I've worked on this.

The (truncated) output after this PR:
```
ValueError: Following entity spans are not aligned with provided tokenization
Spans:
- [This i] defined in 'This is my ...'
- [my tex] defined in '...s is my text'
Tokens:
['This', 'is', 'my', 'text']
```
Note the additional `'`. Note that the changes rely on `repr`, so if the snippet contains `'` itself, it uses `"` instead, e.g.:
```python
import argilla as rg

record = rg.TokenClassificationRecord(
    text = "This is Tom's text",
    tokens=["This", "is", "Tom", "'s", "text"],
    prediction=[("ORG", 0, 6), ("PER", 8, 16)],
)
```
```
ValueError: Following entity spans are not aligned with provided tokenization
Spans:
- [This i] defined in 'This is Tom...'
- [Tom's te] defined in "...s is Tom's text"
Tokens:
['This', 'is', 'Tom', "'s", 'text']
```

**Type of change**

**How Has This Been Tested**

Modified the relevant tests, ensured they worked.

**Checklist**

- Tom Aarsen

### DIFF
--- a/docs/_source/conf.py
+++ b/docs/_source/conf.py
@@ -149,7 +149,7 @@ if docs_version == "latest":
     branch = "main"
 else:
     branch = docs_version.replace("-", "/")
-    
+
 nbsphinx_prolog = (
     """
 .. raw:: html
--- a/src/argilla/utils/span_utils.py
+++ b/src/argilla/utils/span_utils.py
@@ -100,7 +100,7 @@ class SpanUtils:
                 context += self.text[max(char_start - 5, 0) : char_end + 5]
                 if char_end + 5 < len(self.text):
                     context += "..."
-                message = f"- [{span_str}] defined in {repr(context)}"
+                message = f"{span} - [{span_str}] defined in {repr(context)}"
 
                 misaligned_spans_errors.append(message)
 
--- a/src/argilla/utils/span_utils.py
+++ b/src/argilla/utils/span_utils.py
@@ -93,12 +93,14 @@ class SpanUtils:
                 self._start_to_token_idx.get(char_start),
                 self._end_to_token_idx.get(char_end),
             ):
-                message = f"- [{self.text[char_start:char_end]}] defined in "
+                span_str = self.text[char_start:char_end]
+                context = ""
                 if char_start - 5 > 0:
-                    message += "..."
-                message += self.text[max(char_start - 5, 0) : char_end + 5]
+                    context = "..."
+                context += self.text[max(char_start - 5, 0) : char_end + 5]
                 if char_end + 5 < len(self.text):
-                    message += "..."
+                    context += "..."
+                message = f"- [{span_str}] defined in {repr(context)}"
 
                 misaligned_spans_errors.append(message)
 
--- a/src/argilla/utils/span_utils.py
+++ b/src/argilla/utils/span_utils.py
@@ -94,14 +94,7 @@ class SpanUtils:
                 self._end_to_token_idx.get(char_end),
             ):
                 span_str = self.text[char_start:char_end]
-                context = ""
-                if char_start - 5 > 0:
-                    context = "..."
-                context += self.text[max(char_start - 5, 0) : char_end + 5]
-                if char_end + 5 < len(self.text):
-                    context += "..."
-                message = f"{span} - [{span_str}] defined in {repr(context)}"
-
+                message = f"{span} - {repr(span_str)}"
                 misaligned_spans_errors.append(message)
 
         if not_valid_spans_errors or misaligned_spans_errors:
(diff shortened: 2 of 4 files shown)


davidberenstein1957:
Hi @tomaarsen 

Looks much more readable! Perhaps it makes sense to directly link them to the provided input. Also, I think the substring part (`this is tom...`) can be a bit longer for referencing.
```bash
("ORG", 0, 6) - 'This i' defined in 'This is Tom...'
```

tomaarsen:
Even better. I also considered output along the lines of this:
```
("ORG", 0, 6): 'This is Tom...'
                ^^^^^^
```
But I suspect that we would be best off implementing #1843 instead and using colors. 
That said, we may want to have a "color toggle" of sorts for users directly logging away into files, in which case this solution would work well.

Thoughts?

I'm down to update this PR according to your suggestion, by the way.

davidberenstein1957:
Yes, I agree that something like that would even be better, but might require a bit more custom code/logic, so I would, for now, go for a quick win, and based on the initial `rich` improvements continue fine-tuning from there.

So, if it doesn't take up too much time, the last proposed design seems best to me.

tomaarsen:
I've updated the error to match the proposed message from https://github.com/argilla-io/argilla/pull/2329#issuecomment-1425614519. I think we should skip the proposed error from https://github.com/argilla-io/argilla/pull/2329#issuecomment-1425619096.

```python
import argilla as rg

record = rg.TokenClassificationRecord(
    text = "This is Tom's text",
    tokens=["This", "is", "Tom", "'s", "text"],
    prediction=[("ORG", 0, 6), ("PER", 8, 16)],
)
```
```
ValueError: Following entity spans are not aligned with provided tokenization
Spans:
('ORG', 0, 6) - [This i] defined in 'This is Tom...'
('PER', 8, 16) - [Tom's te] defined in "...s is Tom's text"
Tokens:
['This', 'is', 'Tom', "'s", 'text']
```

I think this PR is ready. The next step could come when #1843 is implemented.

- Tom Aarsen

davidberenstein1957:
Hi @tomaarsen ,

Sorry, readability is still a bit off for me. I think the "defined in" makes it a bit unclear
I think I prefer the second option.
```bash
Spans:
("ORG", 0, 6, "This i") 
("PER", 8, 16, "Tom's te") 
Tokens:
['This', 'is', 'Tom', "'s", 'text']
```

```bash
Spans:
("ORG", 0, 6) - "This i"
("PER", 8, 16) - "Tom's te" 
Tokens:
['This', 'is', 'Tom', "'s", 'text']
```

```bash
Spans:
("ORG", 0, 6, "This i") - "This is Tom..."
("PER", 8, 16, "Tom's te") - "...s is Tom's text"
Tokens:
['This', 'is', 'Tom', "'s", 'text']
```

tomaarsen:
No worries, I'm open to iterating here. 
I think options 1 and 3 from your proposal here are unintuitive as you strictly provide `("ORG", 0, 6)` as a part of the prediction, and these proposals arbitrarily include e.g. `"This i"` in the tuple. I like option 2, but we can also make it more word-y:
```
These spans do not correspond with token boundaries:
("ORG", 0, 6) corresponds with "This i" in "This is Tom..."
("PER", 8, 16) corresponds with "Tom's te" in "...s is Tom's text"
Tokens:
['This', 'is', 'Tom', "'s", 'text']
```
I'm curious to hear your thoughts.

davidberenstein1957:
I wanted to avoid wordiness because we are dealing with a lot of text/words. If you feel option 2 is too unintuitive, I would prefer your final proposal.

tomaarsen:
No, I think option 2 is fairly clear, especially as we must consider that the snippet follows `"Following entity spans are not aligned with provided tokenization"`. I'll get on it :)

tomaarsen:
Updated the output:
```python
import argilla as rg

record = rg.TokenClassificationRecord(
    text = "This is Tom's text",
    tokens=["This", "is", "Tom", "'s", "text"],
    prediction=[("ORG", 0, 6), ("PER", 8, 16)],
)
```
```
ValueError: Following entity spans are not aligned with provided tokenization
Spans:
('ORG', 0, 6) - 'This i'
('PER', 8, 16) - "Tom's te"
Tokens:
['This', 'is', 'Tom', "'s", 'text']
```
I'm open to any additional feedback.

davidberenstein1957 (approved):

Issue #1852, opened 2022-03-02
Improve label organization and interaction in Token Classification


dvsrepo:
## Description

Creating a new UI mode that helps with annotation sessions. 

This mode must be included for all defined tasks and should keep the user interaction as simple as possible and hard oriented to annotation.

### From issue #1851 
**Describe the bug**
With more than 10 labels the button hot-links stop.

**To Reproduce**
Go to token / text-classification and assign >10 label options.

**Expected behavior**
I would expect them to continue on function keys or initial qwerty keys.

**Screenshots**
![image](https://user-images.githubusercontent.com/25269220/196913426-5c46cf7a-8472-4454-b6f4-b51310d9dc71.png)

cceyda:
Some ideas I would like to add to this:
- Would also be good to get more variation in colors. I know it is pretty hard to choose a distinctive enough [color pallet](https://seaborn.pydata.org/tutorial/color_palettes.html) but still important.

- allowing custom key(and color) assignment. maybe something like:
```
rg.log(..., tags={"ORG":"organization","LOC":"location",...} ,config={"shortcuts":{"ORG":"1","LOC":"a",...}})
```
- allow labels on the top bar to be clickable (so you can use the drop down list or the top bar to pick the label)(not so sure how useful this is)

- A slightly more complex suggestion: group labels hierarchically. Sometimes there can be a logical grouping of labels.
purely for  example let's say:
<img width="1349" alt="image" src="https://user-images.githubusercontent.com/15624271/218690189-67cd9620-b471-4a4c-b9d9-eac69d4b48e6.png">
can be divided like this (excuse the paint edit):

![image](https://user-images.githubusercontent.com/15624271/218692053-59222e99-8c49-4b9b-94df-0367bdd8dd07.png)

then for keyboard shortcuts you would first press the parent tag key, then the sub. 
example: [1][3] would be food, [2][5]organization

nataliaElv:
@dvsrepo Would you say this issue has been addressed properly in the `SpanQuestion` included in v1.26.0?

cceyda:
`SpanQuestion` looks nice 🥳  I'll try it out this week migrating the old TokenClassification dataset.

cceyda:
Here are my first impressions & feedback about the new `SpanQuestion`.
I have migrated my old `TokenClassification` dataset to a Feedback dataset:

@nataliaElv 
[...]

Issue #2003, opened 2022-12-05
[Dataset settings] New section in UI


frascuchon:
We need to change the front-end part to create new labels by using the labeling schema already defined as part of the dataset settings.

Including these changes, we need to think about how to restrict the use of label creation to users.

cc @dvsrepo 

# Updated

This view cannot be visible for ALL users. For example, for now, the "Create label" button is only shown when no settings resource is found (from an API call), or deleting a dataset is only possible for dataset creators.

The condition will be (until argilla has no user role assignment):

Show dataset settings view only for users that have created the dataset (cc @dvsrepo)

Also, we should unify the current labels schema to the settings resource (cc @keithCuniah @leiyre )

# Tasks
- [x] #2210  
- [x] #2368
- [x] #2369
- [x] #2370
- [x] #2371
- [x] #2372

Amelie-V:
We will create a proper area for dataset settings. The two main sets accessible to users from the UI are:

- Label creation
- Delete dataset

These two features already exist on the UI.

It consists of relocating them into a dedicated page.

frascuchon:
Refs [#1854](https://github.com/argilla-io/argilla/issues/1854)

frascuchon:
@keithCuniah  see https://github.com/argilla-io/argilla/blob/develop/frontend/models/Dataset.js#L22 references for the action migration

PR #2339, opened 2023-02-14
Enhancement: extend shortcuts to include alphabet for token classification

cceyda:
# Description

This is a stop-gap solution to issue https://github.com/argilla-io/argilla/issues/1852
A more sophisticated solution can be found but in the mean time this solves the inconvenience when there are tags more than 10. 

**Type of change**

**How Has This Been Tested**

- [ ] Test A
- [ ] Test B

**Checklist**

### DIFF
--- a/frontend/components/token-classifier/header/EntitiesHeader.vue
+++ b/frontend/components/token-classifier/header/EntitiesHeader.vue
@@ -53,7 +53,7 @@ export default {
   }),
   computed: {
     visibleEntities() {
-      const characters = "1234567890qwertyuiopasdfghjklzxcvbnm".split("");
+      const characters = "1234567890QWERTYUIOPASDFGHJKLZXCVBNM".split("");
       let entities = [...this.dataset.entities]
         .sort((a, b) => a.text.localeCompare(b.text))
         .map((ent, index) => ({
--- a/frontend/components/token-classifier/header/EntitiesHeader.vue
+++ b/frontend/components/token-classifier/header/EntitiesHeader.vue
@@ -53,7 +53,7 @@ export default {
   }),
   computed: {
     visibleEntities() {
-      const characters = "1234567890".split("");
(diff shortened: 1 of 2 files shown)


tomaarsen:
I understand that this is designed as a simple temporary solution, but I want to point out that this will only work for the "default" QWERTY keyboard layout. Users on QWERTZ, AZERTY or others would likely find this unintuitive.

cceyda:
okay realized shortcuts are uppercased in [code](https://github.com/argilla-io/argilla/blob/91a77ad85244cd885fafd40c38251ecdf2e37147/frontend/components/token-classifier/results/EntitiesSelector.vue#L126) so changed that. 

Also if anyone is looking for a quick custom shortcutkeys fix:
(chrome) using developer tools search for "1234567890" in js files, there should be 2.
then override, reload should work!

dvsrepo:
thanks so much @cceyda for the PR. I would vote to merge this and tackle the keyboard layout configuration mentioned by @tomaarsen as part of the new Dataset Setting feat #2003 (maybe not for the first version but for following releases, that would be configurable there). what do you think @leiyre @Amelie-V ?

Could you review this PR @leiyre ?

leiyre (approved):
👌

dvsrepo:
@leiyre could this be merged already?

keithCuniah:
Hello everyone.
Just for information, as mention by @dvsrepo, this correction has also impact in dataset settings feature. See #[2444](https://github.com/argilla-io/argilla/pull/2444). So I moved the solution to the the front ORM.

Second things, I was looking for a way to get the user keyboard layout dynamically (which could solve the problem pointed by @tomaarsen), and there is a keyboard API in javascript for that but it's experimental (see this [link](https://developer.mozilla.org/en-US/docs/Web/API/Keyboard)).  
So when this API will be stable, it would be a nice solution to have dynamical shorcuts depending on the users keyboard.

PR #2349, opened 2023-02-15
fix: Allow bulk annotation when records have Default status

leiyre:
# Description

Allow bulk annotation when records have Default status

**Type of change**

**Checklist**

### DIFF
--- a/frontend/components/text-classifier/header/TextClassificationHeader.vue
+++ b/frontend/components/text-classifier/header/TextClassificationHeader.vue
@@ -102,7 +102,7 @@ export default {
         const labelsToSend = this.labelsFactoryBySingleOrMultiLabel(
           labels,
           labelsToRemove,
-          record.currentAnnotation?.labels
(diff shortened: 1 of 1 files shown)


frontend/components/text-classifier/header/TextClassificationHeader.vue
```diff
@@ -102,7 +102,7 @@ export default {
         const labelsToSend = this.labelsFactoryBySingleOrMultiLabel(
           labels,
           labelsToRemove,
-          record.currentAnnotation?.labels
+          record.currentAnnotation?.labels || []
```
keithCuniah: 1/ Clean code : don't do too much thing in one line : 
 It's better to do the test before in a specific variable and pass this variable as argument

2/ it's a very little comment, but maybe you want to use nullish coallescing instead of || => record.currentAnnotation?.labels ?? []

keithCuniah (approved):

PR #2350, opened 2023-02-15
Use `rich` for logging, tracebacks, printing, progressbars

### LINKED ISSUE #1843: Use Rich for logging, stack traces, and progress bars

tomaarsen:
Closes #1843

Hello!

## Pull Request overview
* Use [`rich`](https://github.com/Textualize/rich) for logging, tracebacks, printing and progressbars.
* Add `rich` as a dependency.
* Remove `loguru` as a dependency and remove all mentions of it in the codebase.
* Simplify logging configuration according to the logging documentation.
* Update logging tests.

## Before & After
[`rich`](https://github.com/Textualize/rich) is a large Python library for very colorful formatting in the terminal. Most importantly (in my opinion), it improves the readability of logs and tracebacks. Let's go over some before-and-afters:
<details><summary>Printing, Logging & Progressbars</summary>

### Before:
image

### After:
image

Note that for the logs, the repeated information on the left side is removed. Beyond that, the file location from which the log originates is moved to the right side. Beyond that, the progressbar has been updated, ahd the URL in the printed output has been highlighted automatically.

</details>

<details><summary>Tracebacks</summary>

### Before:
image

### After:
image

---

### Before:
image

### After:
image

---

### Before
image

### After
image

</details>

## Notes
Note that there are some changes in the logging configuration. Most of all, it has been simplified according to the note from [here](https://docs.python.org/3/library/logging.html#logging.Logger.propagate). In my changes, I only attach our handler to the root logger and let propagation take care of the rest.

Beyond that, I've set `rich` to 13.0.1 as newer versions experience a StopIteration error like discussed [here](https://github.com/Textualize/rich/issues/2800#issuecomment-1428764064).

I've replaced `tqdm` with `rich` Progressbar when logging. However, I've kept the `tqdm` progressbar for the [Weak Labeling](https://github.com/argilla-io/argilla/blob/develop/src/argilla/labeling/text_classification/weak_labels.py) for now. 

One difference between the old situation and now is that all of the logs are displayed during `pytest` under "live log call" (so, including expected errors), while earlier only warnings were shown.

## What to review?
Please do the following when reviewing:
1. Ensuring that `rich` is correctly set to be installed whenever someone installs `argilla`. I always set my dependencies explicitly in setup.py like [here](https://github.com/nltk/nltk/blob/develop/setup.py#L115) or [here](https://github.com/huggingface/setfit/blob/78851287535305ef32f789c7a87004628172b5b6/setup.py#L47-L48), but the one for `argilla` is [empty](https://github.com/argilla-io/argilla/blob/develop/setup.py), and `pyproject.toml` is used instead. I'd like for someone to look this over.
2. Fetch this branch and run some arbitrary code. Load some data, log some data, crash some programs, and get an idea of the changes. Especially changes to loggers and tracebacks can be a bit personal, so I'd like to get people on board with this. Otherwise we can scrap it or find a compromise. After all, this is also a design PR.
3. Please have a look at my discussion points below.

## Discussion
`rich` is quite configurable, so there's some changes that we can make still.
1. The `RichHandler` logging handler can be modified to e.g. include rich tracebacks in their logs as discussed [here](https://rich.readthedocs.io/en/latest/logging.html#handle-exceptions). Are we interested in this?
2. The `rich` traceback handler can be set up to include local variables in its traceback:
    <details><summary>Click to see a rich traceback with local variables</summary>

    image
    </details>
    Are we interested in this? I think this is a bit overkill in my opinion.
3. We can suppress frames from certain Python modules to exclude them from the rich tracebacks. Are we interested in this?
4. The default rich traceback shows a maximum of 100 frames, which is a *lot*. Are we interested in reducing this to only show the first and last X?
5. The progress bar doesn't automatically stretch to fill the full available width, while `tqdm` does. If we want, we can set `expand=True` and it'll also expand to the entire width. Are we interested in this?
6. The progress "bar" does not need to be a bar, we can also use e.g. a spinner animation. See some more info [here](https://rich.readthedocs.io/en/latest/progress.html#columns). Are we interested in this?

---

**Type of change**

**How Has This Been Tested**

I've updated the tests according to my changes.

**Checklist**

- Tom Aarsen

### DIFF
--- a/environment_dev.yml
+++ b/environment_dev.yml
@@ -46,6 +46,6 @@ dependencies:
       - spacy==3.1.0
       - https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-3.1.0/en_core_web_sm-3.1.0.tar.gz
       - transformers[torch]~=4.18.0
-      - loguru
+      - rich==13.0.1
       # install Argilla in editable mode
       - -e .[server,listeners]
--- a/environment_dev.yml
+++ b/environment_dev.yml
@@ -43,8 +43,8 @@ dependencies:
       - pgmpy
       - plotly>=4.1.0
       - snorkel>=0.9.7
-      - spacy==3.1.0
-      - https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-3.1.0/en_core_web_sm-3.1.0.tar.gz
+      - spacy==3.5.0
+      - https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-3.5.0/en_core_web_sm-3.5.0.tar.gz
       - transformers[torch]~=4.18.0
       - rich==13.0.1
       # install Argilla in editable mode
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,11 +38,13 @@ dependencies = [
     "wrapt >= 1.13,< 1.15",
     # weaksupervision
     "numpy < 1.24.0",
+    # for progressbars
     "tqdm >= 4.27.0",
     # monitor background consumers
     "backoff",
-    "monotonic"
-
+    "monotonic",
+    # for logging, tracebacks, printing, progressbars
+    "rich <= 13.0.1"
 ]
 dynamic = ["version"]
 
--- a/src/argilla/__init__.py
+++ b/src/argilla/__init__.py
@@ -21,12 +21,18 @@ as well as in the `_import_structure` dictionary.
 import sys as _sys
 from typing import TYPE_CHECKING as _TYPE_CHECKING
 
-from rich.traceback import install as _install_rich
-
 from argilla.logging import configure_logging as _configure_logging
 
 from . import _version
 from .utils import LazyargillaModule as _LazyargillaModule
+
+try:
+    from rich.traceback import install as _install_rich
+
+    # Rely on `rich` for tracebacks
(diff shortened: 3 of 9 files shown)


tomaarsen:
It seems to have some issues installing `rich`. In particular, I think it's struggling to read the `__version__` of `argilla` without triggering errors that `rich` hasn't been installed yet. This issue persists regardless of whether `rich==13.0.1` is in the `- pip:` section of the `environment_dev.yml` or not.

It'll have to be investigated some more.

tomaarsen:
In order to allow rich to be installed, I had to:
* Allow `argilla.__version__` to be accessed without having `rich` installed by hiding the imports under `try-except ModuleNotFound`.
* Update `spacy` from `3.1.0` to `3.5.0` due to a dependency conflict between `spacy==3.1.0` and `rich==13.0.1` over the `typing-extensions` version.

I'll solve the merge conflicts.

This is ready for review.

frascuchon (approved):
So nice work! 

Thanks @tomaarsen

PR #2352, opened 2023-02-16
ci: Setup black line-length in toml file

frascuchon:
Upgrade the black line-length parameter in the project toml  file.


tomaarsen (approved):
I love it!

PR #2363, opened 2023-02-17
Add new endpoint to delete users

jfcalvo:
Add new endpoint to delete users.
- The endpoint returns a 404 error when the provided `user_id` is not found.
- I have found a problem with `UUID` type not being correctly serialized to JSON for some reason. We need to find an alternative solution to this (more info in the code).

### DIFF
--- a/src/argilla/server/apis/v0/handlers/users.py
+++ b/src/argilla/server/apis/v0/handlers/users.py
@@ -13,8 +13,8 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-from uuid import UUID
 from typing import List
+from uuid import UUID
 
 from fastapi import APIRouter, Depends, Request, Security
 from pydantic import parse_obj_as
--- a/src/argilla/server/apis/v0/handlers/users.py
+++ b/src/argilla/server/apis/v0/handlers/users.py
@@ -13,12 +13,15 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+from uuid import UUID
+
 from fastapi import APIRouter, Depends, Request, Security
 from sqlalchemy.orm import Session
 
 from argilla.server.commons import telemetry
 from argilla.server.contexts import accounts
 from argilla.server.database import get_db
+from argilla.server.errors import EntityNotFoundError
 from argilla.server.security import auth
 from argilla.server.security.model import User, UserCreate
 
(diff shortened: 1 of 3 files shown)


src/argilla/server/apis/v0/handlers/users.py
```diff
@@ -54,3 +57,20 @@ def create_user(
...
+def delete_user(
+    *, db: Session = Depends(get_db), user_id: UUID, current_user: User = Security(auth.get_user, scopes=[])
+):
+    user = accounts.get_user_by_id(db, user_id)
+    if not user:
+        # TODO: Forcing here user_id to be an string.
+        # Not casting it is causing a `Object of type UUID is not JSON serializable`.
+        # Possible solution redefining JSONEncoder.default here:
+        # https://github.com/jazzband/django-push-notifications/issues/586
+        raise EntityNotFoundError(name=str(user_id), type=User)
```
frascuchon, line 81: I think adding `json_encoders`  to out error class models should work.

Take a look here: https://docs.pydantic.dev/usage/exporting_models/#json_encoders

src/argilla/server/apis/v0/handlers/users.py
```diff
@@ -54,3 +57,20 @@ def create_user(
...
+
+    accounts.delete_user(db, user)
+
+    return User.from_orm(user)
```
frascuchon, line 85: Here, do you have any problem with UUID?

src/argilla/server/contexts/accounts.py
```diff
@@ -43,6 +49,13 @@ def create_user(db: Session, user_create: UserCreate):
     return user
 
 
+def delete_user(db: Session, user: User):
+    db.delete(user)
```
frascuchon, line 57: Removing this user from the user_workspaces relationship table will be tackled in a next step, right?

src/argilla/server/apis/v0/handlers/users.py
jfcalvo: I will take a look.

src/argilla/server/apis/v0/handlers/users.py
jfcalvo: Nope, on Pydantic schemas UUID it's correctly serialized. I'm using it with workspaces without any problem.

src/argilla/server/contexts/accounts.py
jfcalvo: That's correct.

src/argilla/server/apis/v0/handlers/users.py
frascuchon: We will resolve this once we review the error classes

frascuchon (approved):

Issue #1854, opened 2022-03-02
Dataset settings


dvsrepo:
## Description

Allow configure dataset level settings for several purposes:

- Edit the labels (add, edit, remove) ( refs #1887 )
- Define metadata fields schema. This could include hidden fields, for example
- Customize the inner elaticsearch index settings/mappings (for advanced users)
- Configure the users that can operate over the dataset (annotators, administrators, reviewers,...). Here, a previous design for user roles should be defined.
- Some task could define their own settings (the `multi_label` parameter could be a setting for the text classification task)

### From issue #1853 
Edit the tags and workspace for the dataset, but only was able to do so via deletion and re-creation.

**Describe the solution you'd like**
The possibility to have a button next o the delete and copy that allows for editting.

frascuchon:
This suggestion could match with some config that could be included as part of the dataset settings.

https://github.com/argilla-io/argilla/issues/1761#issuecomment-1316455118

PR #2365, opened 2023-02-18
chore: Replace old recognai emails with argilla ones

jfcalvo:
# Description

Replace old `recogn.ai` emails with `argilla.io` ones.

This will improve [argilla Pypi package page](https://pypi.org/project/argilla/) showing the correct contact emails.

We still have a reference to `recogn.ai` on this JS model file: https://github.com/argilla-io/argilla/blob/develop/frontend/models/Dataset.js#L22

Please @frascuchon and @keithCuniah can you confirm if that JS code can be safely changed and how?

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -60,7 +60,7 @@ representative at an online or offline event.
 
 Instances of abusive, harassing, or otherwise unacceptable behavior may be
 reported to the community leaders responsible for enforcement at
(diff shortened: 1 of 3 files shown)


frascuchon (approved):
great! THanks, @jfcalvo

frascuchon:
@jfcalvo The js reference you mentioned cannot be replaced on this PR. This will be tackled with the #1854  implementation y using the specific create settings/label schema endpoint.

PR #2366, opened 2023-02-18
Add new endpoint to list users

jfcalvo:
This PR adds a new API endpoint to list all the users present in the database.

It doesn't includes pagination. We need to discuss if it's something that we need right now, maybe it can be implemented later.

### DIFF
--- a/src/argilla/server/apis/v0/handlers/users.py
+++ b/src/argilla/server/apis/v0/handlers/users.py
@@ -13,7 +13,10 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+from typing import List
+
 from fastapi import APIRouter, Depends, Request, Security
+from pydantic import parse_obj_as
 from sqlalchemy.orm import Session
 
 from argilla.server.commons import telemetry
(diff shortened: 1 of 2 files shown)


src/argilla/server/apis/v0/handlers/users.py
```diff
@@ -47,6 +50,13 @@ async def whoami(request: Request, current_user: User = Security(auth.get_user,
     return current_user
 
 
+@router.get("/users", response_model=List[User], response_model_exclude_none=True)
+def list_users(*, db: Session = Depends(get_db), current_user: User = Security(auth.get_user, scopes=[])):
+    users = accounts.list_users(db)
+
+    return parse_obj_as(List[User], users)
```
frascuchon, line 57: We can go with this response but in the next step, we should change the response mode, even if the endpoint has no support for pagination, just to simplify upgrades once we include support for it.

frascuchon (approved):
Good work.

We will include endpoint tests in future sessions

src/argilla/server/apis/v0/handlers/users.py
jfcalvo: Sounds great.

PR #2380, opened 2023-02-21
Refactor: Add `require_version` function and `requires_version` decorator

### LINKED ISSUE #2367: Decide if we add sklearn as dep or include a error message when computing train_test_split
We currently use sklearn in two places: 1) prepare for training train test splits (new) and 2) metrics.

The way we handled it when we only had 2) was to have it as optional dep, and show the following message to users:

```python
try:
  import sklearn  # noqa: F401
        except ModuleNotFoundError:
            raise ModuleNotFoundError(
                "'sklearn' must be installed to compute the metrics! "
                "You can install 'sklearn' with the command: `pip install scikit-learn`"
            )
```

I remember @frascuchon commenting on this previously but I missed it. 

We could add sklearn as core dep but I have some concerns about opening the door to some installation problems and making the distribution of argilla less lightweight.

Thoughts @frascuchon , @davidberenstein1957 , @tomaarsen ?

For context: if you don't have scikit-learn installed prepare_for_training with train/test size params fails:

```python

  File "/Users/dani/argilla/projects/idealista-ner/scripts/preprocess.py", line 32, in <module>
    typer.run(main)

  File "/Users/dani/argilla/projects/idealista-ner/scripts/preprocess.py", line 26, in main
    ds = ds.prepare_for_training(framework="spacy", train_size=train_size, lang=nlp)

  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/argilla/client/datasets.py", line 492, in prepare_for_training
    from sklearn.model_selection import train_test_split

ModuleNotFoundError: No module named 'sklearn'
```

tomaarsen:
Closes #2367

Hello!

## Pull Request overview
* Implementation for `require_version` based on the transformer module.
* Implementation for `requires_version`, a decorator that wraps `require_version`
  * `requires_datasets`, `requires_spacy` and `requires_sklearn` as shorthands for `requires_version` for specific modules.
* Removed `try-except ModuleNotFoundError` throughout the project for e.g. `cleanlab`, `snorkel`, `flyingsquid`, `pgmpy`, `starlette` and replaced them with `requires_version` decorators.
* Added tests to accompany the additions
* Modified existing tests

## Details
I have added `require_version` which ought to be used like:
```python
...
if train_size and test_size:
    require_version("scikit-learn")
    from sklearn.model_selection import train_test_split
    ...
```

And also `requires_version`:
```python
...
@requires_version("cleanlab")
def find_label_errors(
    records: Union[List[TextClassificationRecord], DatasetForTextClassification],
    ...
```
and some shorthands, e.g.:
```python
...

@requires_datasets
def to_datasets() -> "datasets.Dataset":
    """Exports your records to a `datasets.Dataset`.
    ...
```

When `require_version` is called, or when a function/method wrapped with `requires_...` is called, [`importlib.metadata`](https://docs.python.org/3/library/importlib.metadata.html) is used to see if the package is installed, and at which version. This is the recommended approach for collecting version information, i.e. recommended over the less efficient `pkg_resources`.

### Advanced usage
We can set version requirements using this approach:
```python
require_version("datasets>1.17.0")
# or
@requires_version("datasets>1.17.0")
def ...
```
And not just one, we can set multiple:
```python
require_version("datasets>1.17.0,<2.0.0")
# or
@requires_version("datasets>1.17.0,<2.0.0")
def ...
```
We can also specify Python versions for certain functions/methods/sections of code:
```python
require_version("python>3.7")
# or
@requires_version("python>3.7")
def ...
```

### Missing & outdated dependencies
Consider the following example:
```python
from argilla.utils.dependency import requires_datasets

@requires_datasets
def foo():
    pass

foo()
```
When executed without `datasets` installed, the following exception is thrown:
```
ModuleNotFoundError: 'datasets' must be installed to use `foo`! You can install 'datasets' with this command: `pip install datasets>1.17.0`
```
Alternatively, if I install `datasets==1.16.0` and run it again, I get this exception:
```
ImportError: datasets>1.17.0 must be installed to use `foo`, but found datasets==1.16.0. You can install a supported version of 'datasets' with this command: `pip install -U datasets>1.17.0`
```

### Notes
I had to update various tests that contained the following snippet:
```python
monkeypatch.setitem(sys.modules, "datasets", None)
with pytest.raises(ModuleNotFoundError):
    ...
```
The reasoning is that although `import datasets` would fail in this case, the `require_version` function is not fooled by this monkeypatch as it reads directly from files in `sys.meta_path`. As a result, `require_version` still recognises that the module is installed. That way, it becomes a bit harder to pretend that a module is not installed. An alternative that I went with is clearing out `sys.meta_path`, which means that nothing can be imported after the monkeypatch.

One other tiny note: the following snippet also stopped working.
```
monkeypatch.setattr(FlyingSquid, "_predict", mock_predict)
```
This is because `FlyingSquid` is now decorated by `requires_version`, and this seems to have broken the `monkeypatch.setattr`. As a result, I now perform the `monkeypatch.setattr` on the class instance instead of the class.

I want to reiterate that everything from this section only affects the tests: the behaviour in practice works well.

Lastly, the tests result in a 97% coverage on the new file.

---

**Type of change**

**How Has This Been Tested**

The introduction of various tests under `tests/utils/test_dependency.py` and through updating tests throughout the project.
I also ran some test cases manually.

**Checklist**

---

I'm open to feedback, as always.

cc: @davidberenstein1957 @dvsrepo due to the discussion in #2367

- Tom Aarsen

### DIFF
--- a/src/argilla/client/datasets.py
+++ b/src/argilla/client/datasets.py
@@ -28,7 +28,7 @@ from argilla.client.models import (
     TokenClassificationRecord,
 )
 from argilla.client.sdk.datasets.models import TaskType
-from argilla.utils.dependency import require_version, requires_datasets, requires_spacy
+from argilla.utils.dependency import require_version, requires_version
 from argilla.utils.span_utils import SpanUtils
 
 if TYPE_CHECKING:
@@ -119,7 +119,7 @@ class DatasetBase:
     def __str__(self):
         return repr(self)
 
-    @requires_datasets
+    @requires_version("datasets>1.17.0")
     def to_datasets(self) -> "datasets.Dataset":
         """Exports your records to a `datasets.Dataset`.
 
@@ -462,7 +462,7 @@ class DatasetBase:
         else:
             raise NotImplementedError(f"Framework {framework} is not supported. Choose from:" f" {list(Framework)}")
 
-    @requires_spacy
+    @requires_version("spacy")
     def _prepare_for_training_with_spacy(
         self, **kwargs
     ) -> Union["spacy.token.DocBin", Tuple["spacy.token.DocBin", "spacy.token.DocBin"]]:
@@ -477,7 +477,7 @@ class DatasetBase:
 
         raise NotImplementedError
 
-    @requires_datasets
+    @requires_version("datasets>1.17.0")
     def _prepare_for_training_with_transformers(self, **kwargs) -> "datasets.Dataset":
         """Prepares the dataset for training using the "transformers" framework.
 
@@ -490,7 +490,7 @@ class DatasetBase:
 
         raise NotImplementedError
 
-    @requires_datasets
+    @requires_version("datasets>1.17.0")
     def _prepare_for_training_with_spark_nlp(self, **kwargs) -> "datasets.Dataset":
         """Prepares the dataset for training using the "spark-nlp" framework.
 
@@ -559,7 +559,7 @@ class DatasetForTextClassification(DatasetBase):
         super().__init__(records=records)
 
     @classmethod
-    @requires_datasets
+    @requires_version("datasets>1.17.0")
     def from_datasets(
         cls,
         dataset: "datasets.Dataset",
@@ -710,7 +710,7 @@ class DatasetForTextClassification(DatasetBase):
     def _from_pandas(cls, dataframe: pd.DataFrame) -> "DatasetForTextClassification":
         return cls([TextClassificationRecord(**row) for row in dataframe.to_dict("records")])
 
-    @requires_datasets
+    @requires_version("datasets>1.17.0")
     def _prepare_for_training_with_transformers(
         self,
         train_size: Optional[float] = None,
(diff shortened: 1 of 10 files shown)


davidberenstein1957 (commented):
Hi, @tomaarsen. It looks great, but had one discussion point.

Maybe it would make sense to simplify the workflow of `requires_specific_package` + `requires_decorator` and use the `require_version` or `@requires_version` whenever needed.

tomaarsen:
To be concrete, do you propose the removal of `@requires_datasets`, `@requires_spacy` and `@requires_sklearn`, and replace them with `@requires_version("datasets>1.17.0")`, `@requires_version("spacy")`, `@requires_version("scikit-learn")` instead? That does simplify things for us.
Please let me know if that's indeed what you mean, and if so, then I'll make the changes.

davidberenstein1957:
Yes, that is exactly what I mean. Thanks for picking this up.

tomaarsen:
I've implemented your feedback 🎉

dvsrepo:
Love this @tomaarsen !

dvsrepo (approved):
💯

PR #2392, opened 2023-02-22
CI: Skip rather than failing in 2 common scenarios

tomaarsen:
Hello!

## Pull Request overview
* Skip CI jobs rather than letting them fail in 2 common scenarios:
  1. If a PR is created from a fork.
  2. If a PR does not contain code changes.

### Scenario 1
If a PR is created from a fork, then certain secrets will be inaccessible, e.g. `secrets.AR_DOCKER_USERNAME` and `secrets.AR_DOCKER_PASSWORD`. This causes the "Docker Deploy" job to fail, understandably. It would be better if these jobs would be skipped instead.
See [here](https://github.com/argilla-io/argilla/actions/runs/4240437103/jobs/7369703791) for an example case.

### Scenario 2
There are two subsequent jobs: One to build the Python package, and one to deploy it as a Docker image. The first one will stop halfway through if there are no code changes found in the PR, but the second job will still try to deploy the build artifact. Upon trying to load it, it will crash.
See [here](https://github.com/argilla-io/argilla/actions/runs/4229327600/jobs/7345607762) for an example case.

### Scenario 1
Sadly, secrets are not accessible on the `build.if` and `deploy_docker.if` level, so we can't use e.g. `if: secrets.AR_DOCKER_USERNAME != ''`. Instead, we are required to use an environment variable, e.g. `IS_DEPLOYABLE` to track whether the secrets are accessible, and then add `if env.IS_DEPLOYABLE == 'true'` to all steps.

### Scenario 2
This was as simple as setting `code_outputs` as an output from the `build` job, which can then be read as input for the `deploy_docker` job. This job is now skipped if there are no code changes in the PR/commit.

---

**Type of change**

**How Has This Been Tested**

Through this PR. Note: I have only been able to test scenario 1 (PR from a fork). I haven't been able to test scenario 2 nor the "normal" scenarios where these jobs do actually execute (i.e. first-party PR with code changes). These are annoying to test easily. I think we're best off merging this and reverting if any scenarios work unexpectedly, rather than trying to test all of the scenarios in a separate repo.

**Checklist**

---

With this PR I hope to have reached the point where CI failures imply actual issues and CI passes imply that a PR is "likely ready".

- Tom Aarsen

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -211,9 +211,9 @@ jobs:
     # Only build the package if we can deploy it as a docker image
     env:
       IS_DEPLOYABLE: ${{ secrets.AR_DOCKER_USERNAME != '' }}
-    if: env.IS_DEPLOYABLE == 'true'
     outputs:
       code_changes: ${{ steps.filter.outputs.code_changes }}
+
     steps:
       - name: Checkout Code 🛎
         uses: actions/checkout@v2
@@ -233,7 +233,7 @@ jobs:
               - '.github/workflows/package.yml'
       - name: Cache pip 👜
         uses: actions/cache@v2
-        if: steps.filter.outputs.code_changes == 'true'
+        if: steps.filter.outputs.code_changes == 'true' && env.IS_DEPLOYABLE == 'true'
         env:
           # Increase this value to reset cache if pyproject.toml has not changed
           CACHE_NUMBER: 0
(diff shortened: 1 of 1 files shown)


frascuchon (commented):
Nice !!!

The only thing is the condition duplication on each job action step. Do you think we could define a job-level condition (at least with the IS_DEPLOYABLE check) to skip the whole job?

Take a look at this https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution

tomaarsen:
I had a look at those docs earlier, but I was under the impression that the job-level `if` is unable to access the `env` values that are set in that same job. Let me give it a try, though. 
Oh, and I agree that spamming the `if` on like 8 subsequent steps is horrible, I'd much rather avoid it.

.github/workflows/package.yml
```diff
@@ -262,6 +267,9 @@ jobs:
       - build
       - test-elastic
       - test-opensearch
+    env:
+      IS_DEPLOYABLE: ${{ secrets.AR_DOCKER_USERNAME != '' }}
+    if: needs.build.outputs.code_changes == 'true' && env.IS_DEPLOYABLE == 'true'
```
tomaarsen: If it works, then we can also remove the `env` from here: if `IS_DEPLOYABLE` would ever be false, then `build` would have been skipped, and the `docker_deploy` job would never have started as `build` is one of its requirements.

tomaarsen:
I believe it did not work. The `Build the python package` job was ran, despite `IS_DEPLOYABLE` being false: https://github.com/argilla-io/argilla/actions/runs/4242551009/jobs/7374203626. 
The reason the CI was green is because the scenario 2 fix does still work: this PR has no code changes, so the deployment jobs were skipped.

I believe I'll have to revert my commit to get the PR ready again.

.github/workflows/package.yml
tomaarsen: Sadly, this doesn't apply as the job-level conditional using `env.IS_DEPLOYABLE` did not skip the `build` job despite `IS_DEPLOYABLE` being false.

frascuchon (approved):
Okay, no worries. We can take a look in the future.

Thanks @tomaarsen

PR #2395, opened 2023-02-22
Change tests configuration to allow testing with new endpoints using database

jfcalvo:
In this PR we are making changes on how tests are configured to be used with the database:
* Moved test seeds to a new `test_seeds` function at `argilla.server.seeds` package.
* One the new `test_seeds` function we are defining a default "argilla" user. In the future `test_seeds` should only have proper seeds necessary for the application to work normally and users and other stuff should be added as test fixtures.
* We didn't find a proper way to use transactions and rollback changes in every unit test and allow endpoints access the same data so we are truncating all the database tables after every unit test execution.
* We added a test for one of the new workspaces endpoint that interact with the database so we can check that everything is working as expected.
* Now our GitHub actions run (alembic) database migrations in the tests section.

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -97,6 +97,7 @@ jobs:
         if: steps.filter.outputs.python_code == 'true'
         run: |
           pip install -e ".[server,listeners]"
+          alembic upgrade head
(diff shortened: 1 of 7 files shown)


frascuchon (approved):
💯

PR #2401, opened 2023-02-23
Fix wrong method name when user is not found on authenticate_user function

jfcalvo:
A small fix for a wrong method name used when user is not found on `authenticate_user`.

Waiting for changes on test settings to be merged so we can write some new tests for this one.


frascuchon (approved):
cooL!

PR #2404, opened 2023-02-23
Better workspace/owner initialization

frascuchon:
Better data sync between `workspace` and `owner` attributes.

Also, force the workspace is present on a dataset instance creation

Note: Tests will be fixed in a future PR. This incremental PR is just to simplify the Reviewer's work.

### DIFF
--- a/src/argilla/server/daos/models/datasets.py
+++ b/src/argilla/server/daos/models/datasets.py
@@ -15,7 +15,7 @@
 from datetime import datetime
 from typing import Any, Dict, Optional, TypeVar, Union
 
-from pydantic import BaseModel, Field, validator
+from pydantic import BaseModel, Field, root_validator
 
 from argilla._constants import DATASET_NAME_REGEX_PATTERN
 from argilla.server.commons.models import TaskType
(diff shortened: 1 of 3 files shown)


src/argilla/server/daos/models/datasets.py
```diff
@@ -36,24 +36,33 @@ class BaseDatasetDB(BaseModel):
...
 
-    @validator("workspace", pre=True, always=True)
-    def set_workspace_defaults(cls, value, values):
-        if value:
-            return value
-        else:
-            return values.get("owner")
+    @root_validator(pre=True)
+    def set_defaults(cls, values):
+        workspace = values.get("workspace")
```
jfcalvo:
```suggestion
```

src/argilla/server/daos/models/datasets.py
```diff
@@ -36,24 +36,33 @@ class BaseDatasetDB(BaseModel):
...
+        workspace = workspace or values.get("owner")
```
jfcalvo:
```suggestion
        workspace = values.get("workspace") or values.get("owner")
```

tests/server/commons/test_records_dao.py
```diff
@@ -26,6 +26,7 @@ def test_raise_proper_error():
         dao.search_records(
             dataset=BaseDatasetDB(
                 name="mock-notfound",
+                workspace="ee",
```
jfcalvo: If `ee` is a random string maybe we should use more contextual strings like:

```suggestion
                workspace="workspace-name",
```

Correct me if I'm wrong.

jfcalvo (changes requested):

tests/server/commons/test_records_dao.py
frascuchon: +1

jfcalvo (approved):

PR #2413, opened 2023-02-24
Adding id and timestamps to user

frascuchon:
The pydantic `User` class contains id, and timestamp (`inserted_at` and `updated_at` attributes).

Also, old user components (service, dao,...) are removed, and all affected tests removed/adapted.


src/argilla/server/security/model.py
```diff
@@ -82,6 +83,7 @@ def get(self, key: str, default: Any) -> Any:
 class User(BaseModel):
     """Base user model"""
 
+    id: UUID = Field(default_factory=uuid.uuid4)
```
jfcalvo: We should not need defaults for these fields because `User` model should be used to represent existent users from database.

For creating users we should use `UserCreate` model and maybe we should move all the validations to that model.

src/argilla/server/security/model.py
```diff
@@ -90,8 +92,8 @@ class User(BaseModel):
     workspaces: Optional[List[str]] = None
     # TODO: Adding inserted_at and updated_at is causing multiple errors so we are disabling them by now.
     # Once that we refactor all the old code we should enable this again.
```
jfcalvo:
```suggestion
```

tests/server/security/test_model.py
```diff
@@ -11,42 +11,68 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
+import uuid
+from datetime import datetime
 
 import pytest
 from argilla.server.errors import EntityNotFoundError
-from argilla.server.security.model import User
+from argilla.server.security.model import User, UserCreate
```
jfcalvo, line 19: You are not using `UserCreate` right?

```suggestion
from argilla.server.security.model import User
```

jfcalvo (changes requested):

tests/server/security/test_model.py
frascuchon: I'm using it in tests related to the username, since is the model which validate this behavior

jfcalvo (approved):

PR #2415, opened 2023-02-24
Add factories and additional tests for new endpoints

jfcalvo:
This PR add tests for new endpoints and also:
- Adding [`factory-boy`](https://factoryboy.readthedocs.io/en/stable/) as development dependency.
- Adding factories: `UserFactory`, `WorkspaceFactory`, and `UserWorkspaceFactory`.
- Fixing relationship warnings for `UserWorkspace` SQLAlchemy model.

Additionally some of the tests are marked as pending for two different reasons:
- SQLite is by default [not enforcing referencial integrity with foreign keys](https://docs.sqlalchemy.org/en/20/dialects/sqlite.html#foreign-key-support), we have different options here:
  1. Enforce foreign key support executing `PRAGMA` statement on connection.
  2. Add additional checks to the endpoints where necessary.
- Pydantic is by default accepting empty strings as valid values for fields that are not optional. This is causing for example that create an user with an empty string as `first_name` is totally valid.
  - We need to check if it's necessary to add a constraint to our models or an alternative solution (it is a dangerous default behavior).

We have discussed these points and we will solve them as follows:
1. We will enforce SQLite foreign keys integrity using a `PRAGMA` statement in every connection.
    - The changes are included in this PR.
    - Even with this I have added some additional checks in some endpoints to give more information to the user when necessary.
2.  For Pydantic models accepting empty strings for non-optional string fields:
     - We will create a new PR with a solution for this, adding more constraints of changing this Pydantic default behavior.
3. We will change `users_workspaces` table name and models to `workspaces_users`, ' WorkspaceUser`, etc. :
     - The changes will be done in a different PR.
     - This has sense because the most important entity is workspace.


tests/server/api/v0/test_users.py
```diff
@@ -0,0 +1,135 @@
...
+    user = {"first_name": "first-name", "username": "username", "password": "12345678"}
+
+    response = client.post("/api/users", json=user)
+
+    assert response.status_code == 401
+    assert db.query(User).count() == 1
+
+
+# TODO: We need to review all the models because of this.
+# More info here: https://github.com/pydantic/pydantic/issues/1626
```
frascuchon, line 84: Could we just add ad min_lenght to those required string attributes? ([`constr`](https://docs.pydantic.dev/usage/types/#arguments-to-constr))

jfcalvo: We had an slack call to discuss this and we agree on review Pydantic models and add some constraint if necessary or investigate a global default config change for Pydantic (if it's possible and the effects are ok).

These changes are gonna be in a different PR.

src/argilla/server/apis/v0/handlers/workspaces.py
```diff
@@ -87,6 +87,14 @@ def create_workspace_user(
     user_id: UUID,
     current_user: User = Security(auth.get_user, scopes=[]),
 ):
+    workspace = accounts.get_workspace_by_id(db, workspace_id)
```
jfcalvo, line 90: Apart from enabling foreign keys I have added some additional checks here. I think can be of value to the user in the case a workspace or user is not there for some reason.

src/argilla/server/contexts/accounts.py
```diff
@@ -57,7 +50,7 @@ def delete_user_workspace(db: Session, user_workspace: UserWorkspace):
 
 
 def get_workspace_by_id(db: Session, workspace_id: UUID):
-    return db.query(Workspace).get(workspace_id)
+    return db.get(Workspace, workspace_id)
```
jfcalvo, line 53: These `db.query(...).get()` calls was causing a warning and asking to replace them with new `session.get`.

src/argilla/server/models.py
```diff
@@ -56,7 +56,9 @@ class Workspace(Base):
     inserted_at: Mapped[datetime] = mapped_column(default=datetime.utcnow)
     updated_at: Mapped[datetime] = mapped_column(default=default_inserted_at, onupdate=datetime.utcnow)
 
-    users: Mapped[List["User"]] = relationship(secondary="users_workspaces", back_populates="workspaces")
+    users: Mapped[List["User"]] = relationship(
+        secondary="users_workspaces", back_populates="workspaces", order_by="asc(UserWorkspace.inserted_at)"
```
jfcalvo: I have added here a default order for `users` and the same one later for `workspaces`.

tests/conftest.py
```diff
@@ -35,14 +35,9 @@
...
 
-@pytest.fixture(scope="session")
-def api_key_header():
-    return {API_KEY_HEADER_NAME: DEFAULT_API_KEY}
-
-
 @pytest.fixture(scope="session")
 def client():
-    with TestClient(app) as c:
+    with TestClient(app, raise_server_exceptions=False) as c:
```
jfcalvo: Setting `raise_server_exceptions` to `False` is making the tests working the same that normal behavior. The exceptions raised by the server will now return a 500 response for example (instead of returning that exception and not returning a response).

src/argilla/server/apis/v0/handlers/workspaces.py
frascuchon: 👍

src/argilla/server/contexts/accounts.py
frascuchon: Also, it seems to be more efficient

src/argilla/server/database.py
```diff
@@ -12,12 +12,24 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-from sqlalchemy import create_engine
+from sqlite3 import Connection as SQLite3Connection
```
frascuchon, line 15: Is `sqlite3` installed as default? or is an optional dep from sqlalchemy. If second, we should validate that the database module still works if sqlite is not installed (using another db engine)

src/argilla/server/models.py
frascuchon: Could we use the model field instead of evaluating a string? I don't like using strings refs since names refactoring can be hell.

```python
from sqlalchemy import asc

...
order_by=asc(UserWorkspace.inserted_at)
...

```

src/argilla/server/contexts/accounts.py
jfcalvo: More efficient like on number of characters in the code or a more efficient db query?

tests/conftest.py
frascuchon: Be careful, this will by-pass other errors than 500 (Remember 401 exception in tests)

tests/server/api/v0/test_users.py
```diff
@@ -0,0 +1,135 @@
...
+    user = {"first_name": "first-name", "username": "username", "password": "12345678"}
+
+    response = client.post("/api/users", headers=admin_auth_header, json=user)
+
+    assert response.status_code == 200
+    assert db.query(User).count() == 2
+
+    response_body = response.json()
+    assert response_body["username"] == "username"
+    assert db.query(User).filter_by(username="username").first()
```
frascuchon: The api-key generation is important and, for now, the only way to generate api keys is with user creation

```suggestion
    user_db =  assert db.query(User).filter_by(username="username").first()
    assert user_db
    assert user_db.api_key == response_body["api_key"] 
```

src/argilla/server/database.py
jfcalvo: It's part of the python standard lib. Looks like SQLAlchemy is internally using it to construct the db connection.

tests/server/api/v0/test_workspaces.py
```diff
@@ -12,19 +12,193 @@
...
+    pass
+
+
+def test_list_workspace_users(client: TestClient, db: Session, admin_auth_header: dict):
+    workspace_a = WorkspaceFactory.create()
+    UserWorkspaceFactory.create(user_id=UserFactory.create(username="username-a").id, workspace_id=workspace_a.id)
+    UserWorkspaceFactory.create(user_id=UserFactory.create(username="username-b").id, workspace_id=workspace_a.id)
+    UserWorkspaceFactory.create(user_id=UserFactory.create(username="username-c").id, workspace_id=workspace_a.id)
+
+    WorkspaceFactory.create(users=[UserFactory.build(), UserFactory.build()])
```
frascuchon, line 96: What's the difference between `create` and `build` for the `UserFactory` class method? I've seen other pieces of code where `.create` is used.

tests/server/api/v0/test_users.py
jfcalvo: Sure, I can add something to check that the `api_key` has been created.

frascuchon (commented):
Looks nice to me. Just a few comments to be validated before approve it.

tests/server/api/v0/test_workspaces.py
jfcalvo: * `create` is inserting the record into the database.
* `build` is creating an instance of the ORM model without inserting it into the database.

tests/server/api/v0/test_users.py
jfcalvo: Apart from that we should discuss (or I can investigate) possible ways to check that the responses are the expected ones. i.e. using json schemas

tests/conftest.py
jfcalvo: Yes, you're right. Let's be cautious and remove it by now.

src/argilla/server/models.py
jfcalvo: Sure, let me try.

tests/server/api/v0/test_workspaces.py
frascuchon: Okay, Thanks!

frascuchon (approved):

tests/server/api/v0/test_users.py
frascuchon: +1

src/argilla/server/contexts/accounts.py
frascuchon: I can imagine some cache improvement mechanisms when using a get-by id instead of a more generic query. But not sure if really are done this kind of optimization.

PR #2417, opened 2023-02-24
added fields to retrieve in load

### LINKED ISSUE #2398: rg.load takes too long because of the vector field
`rg.load` takes too long because of the `vector` field, even when I don't need it.

Allow passing a param or through the query
use the elasticsearch `_source` query field to exclude `vector` when not needed.

```
"_source": {        
        "exclude": ["vector"]
    }
```

wait

**Additional context**
Low priority

orasik:
# Description

Add the fields to retrieve when loading the data from argilla. 
`rg.load` takes too long because of the vector field, even when I don't need it.

Closes #2398

**Type of change**

- [X] Documentation update

**How Has This Been Tested**

- [X] Using the following code: 
```
import argilla as rg
labelled_ds = rg.load("my_data_set", limit=2, fields=["text"])
# will load id, inputs, and text
labelled_ds = rg.load("my_data_set", limit=2)
# will load everything
```

**Checklist**


src/argilla/client/client.py
```diff
@@ -710,6 +716,7 @@ def _load_records_new_fashion(
             # Query
             query_text=query,
             ids=ids,
+            fields=fields,
```
frascuchon: The `projection`  is already providing this functionality.

```python
if fields:
    projection = {"id", "inputs"}
    projection.update(fields)
else: 
    projection = {"*"} 

self.datasets.scan(....projection=projection)
```
```suggestion
fields_to_load = ["*"]

```

src/argilla/client/apis/datasets.py
```diff
@@ -181,8 +183,13 @@ def scan(
         if limit == 0:
             limit = None
 
+        fields_to_load = ["*"]
+        if fields:
+            fields.extend(["id", "inputs"])
+            fields_to_load = list(set(fields))
+
```
frascuchon: This logic should be moved to the `load` method.

```suggestion
```

src/argilla/client/apis/datasets.py
```diff
@@ -181,8 +183,13 @@ def scan(
...
         request = {
-            "fields": list(projection) if projection else ["id"],
+            "fields": fields_to_load,
```
frascuchon:
```suggestion
            "fields": list(projection) if projection else ["id"],
```

frascuchon (changes requested):
Hi @orasik  and thanks for your PR.

I've suggested some changes since we can implement this logic using the existing attribute `projection`

Please look into the changes and let me know if you have any doubts.

src/argilla/client/apis/datasets.py
orasik:
```
else ["id"]
```
should be `["id", "inputs"]`.

This is bug as the object verification expecting at least id and input. See my comment here https://github.com/argilla-io/argilla/issues/2398#issuecomment-1444482828

orasik:
Thank you for the review @frascuchon 
I have made the required changes, can you please have a look and let me know if you have more feedback?

src/argilla/client/apis/datasets.py
```diff
@@ -182,7 +182,7 @@ def scan(
             limit = None
 
         request = {
-            "fields": list(projection) if projection else ["id"],
+            "fields": projection,
```
frascuchon: here, we must be sure that `fields` attribute is a list. Also, since is an optional argument, I would keep the original setup.

```suggestion
            "fields": list(projection) if projection else ["id"]
```

src/argilla/client/client.py
```diff
@@ -701,10 +707,12 @@ def _load_records_new_fashion(
                 vector=vector_search,
             )
             return dataset_class(results.records)
+        if fields:
+            fields.extend(["id", "inputs"])
 
         records = self.datasets.scan(
             name=name,
-            projection={"*"},
+            projection=list(set(fields)) if fields else ["*"],
```
frascuchon: Parsing projection inside the `scan` method allows us to pass sets instead of lists.
```suggestion
            projection=set(fields) if fields else {"*"},
```

frascuchon (commented):
Hi, @orasik, and sorry for the late response.

It's better now. I've just added some suggestions not to break the current `scan` method signature. 

Please, take a look and tell me what you think.

Thanks!!

src/argilla/client/apis/datasets.py
orasik: @frascuchon sure, but ` else ["id"]` is a bug. The validation requirements expect at least id and inputs.

frascuchon: Sorry, I missed your answer.

This method will be used in a wider way to fetch part of records from a dataset, without parsing into the Record structure. So, is not really a bug, is the default behavior: retrieve at least the record ids.

The error must be handled from the rb.load functions, where records will be parsed into Record instances, so is there where a minimum value of fields must be assured

For text classification, should be `id` and `inputs`
For token classification: `id`, `text` and `tokens`
For text2text, `id` and `text`

src/argilla/client/apis/datasets.py
```diff
@@ -193,7 +193,7 @@ def scan(
         query = self._parse_query(query=query)
 
         request = {
-            "fields": list(projection) if projection else ["id"],
+            "fields": list(projection) if projection else ["id"]
```
frascuchon:
```suggestion
            "fields": list(projection) if projection else ["id"],
```

frascuchon (approved):

PR #2429, opened 2023-02-27
fix: error when loading record with empty string query

### LINKED ISSUE #2303: passing an empty string "" as query param returns no records instead of all records
**Describe the bug**
For the [keywords](https://github.com/argilla-io/argilla/pull/2290/commits/9d192390f4f77a61cc41fbeb54e94d7d8dc440a3) this was fixed, but not yet done structurally. 

**To Reproduce**
N.A.

**Expected behavior**
N.A.

**Screenshots**
N.A.

N.A.

**Additional context**
N.A.

### LINKED ISSUE #2400: passing empty string `""` to `query` param results in 0 returned records
**Describe the bug**
passing empty string `""` to `query` param results in 0 returned records

**To Reproduce**
rg.load("datasets", query="") 
returns
0 records

**Expected behavior**
I would expect it to return all records

**Screenshots**
N.A.

 - OS [e.g. iOS]: N.A. 
 - Browser [e.g. chrome, safari]: N.A.
 - Argilla Version [e.g. 1.0.0]: 1.3.0
 - ElasticSearch Version [e.g. 7.10.2]: N.A.
 - Docker Image (optional) [e.g. argilla:v1.0.0]: N.A.

**Additional context**
N.A.

davidberenstein1957:
# Description
empty query strings were not returning any records. These are currently set to None and therefore return records again.

Closes #2400 
Closes #2303 
**Type of change**

**How Has This Been Tested**

- [x] [Test A](https://github.com/argilla-io/argilla/blob/34da84b5b0ac2c71d17459519915630dda7393a2/tests/client/test_api.py#L179)

**Checklist**

N.A.

### DIFF
--- a/docs/_source/getting_started/installation/deployments/docker.md
+++ b/docs/_source/getting_started/installation/deployments/docker.md
@@ -9,12 +9,13 @@ First, you need to create a network to make both standalone containers visibles 
 Just run the folowing command:
 ```bash
 docker network create argilla-net
+```
 
 Setting up Elasticsearch (ES) via docker is straightforward.
 Simply run the following command:
 
 ```bash
-docker run -d --name elasticsearch-for-argilla --network argilla-net  -p 9200:9200 -p 9300:9300 -e "ES_JAVA_OPTS=-Xms512m -Xmx512m" -e "discovery.type=single-node" docker.elastic.co/elasticsearch/elasticsearch:8.5.3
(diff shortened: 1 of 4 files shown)


frascuchon (commented):
Good catch David,

Anyway, I think this should be resolved in the server part. Since we can have queries giving different results depending on if they are run from the python client or from UI.

What do you think?

davidberenstein1957:
﻿I agree that it would be better to fix this in the server. I did not directly think about making this distinction. As of now I don't think the python client executed the query when having an empty field but it is cleaner to change this Iwill have a look tomorrow. On 27 Feb 2023, at 22:31, Francisco Aranda ***@***.***> wrote:﻿
@frascuchon commented on this pull request.

Good catch David,
Anyway, I think this should be resolved in the server part. Since we can have queries giving different results depending on if they are run from the python client or from UI.
What do you think?

frascuchon (approved):

PR #2432, opened 2023-02-28
Improve field constraints for some Pydantic schemas

jfcalvo:
We found that Pydantic schema string fields allow empty strings by default, being the field optional or not.

More info about that here: https://github.com/pydantic/pydantic/issues/1626

With this behavior we can finish with empty strings used in fields like `user.first_name`, `user.last_name` and others. Something that can have a bad impact in our application.

I have explored different solutions, like defining a common validation or a custom field type. At the end I think the most simple solution is to be explicit and define the constraint that we want for the fields. In this case `min_length=1` when necessary.

Additional changes:
* Added new max length constraint for `user.password` field.
* Notice how I have changed `constr` for `str = Field(...)`. If I'm not wrong looks like they are similar but `Field` is used to generate a validation response when errors are found.

### DIFF
--- a/src/argilla/server/security/model.py
+++ b/src/argilla/server/security/model.py
@@ -51,7 +51,7 @@ class Workspace(BaseModel):
 
 
 class WorkspaceCreate(BaseModel):
-    name: str = Field(regex=_WORKSPACE_NAME_REGEX, min_length=1)
+    name: constr(regex=_WORKSPACE_NAME_REGEX, min_length=1)
 
 
 class UserGetter(GetterDict):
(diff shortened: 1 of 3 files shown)


tomaarsen:
@jfcalvo 
I'm not quite aware of the differences between `constr` and `Field`, but it seems like `constr` allows some other parameters that may be interesting: https://docs.pydantic.dev/usage/types/#arguments-to-constr

I think the validation errors can be enabled for `constr` using `strict=True`, and in particular the `strip_whitespace=True` may be interesting, depending on whether you would like to allow `" "` as a valid username, etc.

This PR is already certainly a step up from allowing empty strings, though.

jfcalvo:
> I think the validation errors can be enabled for `constr` using `strict=True`, and in particular the `strip_whitespace=True` may be interesting, depending on whether you would like to allow `" "` as a valid username, etc.

I think `strip_whitespace=True` can be useful for some fields like `first_name` and `last_name`. For `username` we have a regular expression that will do the trick of not allowing empty strings.

I will take a look and add use `constr` where necessary.

Thanks for the feedback @tomaarsen

jfcalvo:
@tomaarsen I have been taking a look to [Pydantic documentation about `strict` parameter](https://docs.pydantic.dev/usage/types/#strict-types).

Looks like when using `strict=True` inside a `constr` then only string values are allowed:

Not using `strict=True` on `first_name` field for `UserCreate` will allow something like the following:

```python
# Here integer 12 will be transformed to the string "12"
UserCreate(first_name=12)
```

Using `strict=True` on `first_name` filed will not allow values that are not strings:

```python
# Here a ValidationError will be raised not allowing a integer as value for first_name
UserCreate(first_name=12)
first_name
  str type expected (type=type_error.str)
```

tomaarsen:
Thanks for clearing that up!

frascuchon (approved):

PR #2434, opened 2023-02-28
refactor: Improve efficiency of  `.scan` (and `.load`) if `limit` is set

tomaarsen:
Hello!

## Pull Request overview
* Refactor `Datasets.scan` to never request more samples than required.
* Implements a `batch_size` variable which is now set to `self.DEFAULT_SCAN_SIZE`. This parameter can easily be exposed in the future for `batch_size` support, as discussed previously in Slack etc.

## Details
The core of the changes is that the URL is now built on-the-fly at every request, rather than just once. For each request, the limit for that request is computed using `min(limit, batch_size)`, and `limit` is decremented, allowing it to represent a "remaining samples to fetch". When `limit` reaches 0, i.e. 0 more samples to fetch, we return from `scan`.

I've also added a check to ensure that the `limit` passed to `scan` can't negative.

## Tests
For tests, I've created a `test_scan_efficient_limiting` function which verifies the new and improved behaviour. It contains two monkeypatches:
1. `DEFAULT_SCAN_SIZE` is set to 10. Because our dataset in this test only has 100 samples, we want to ensure that we can't just sample the entire dataset with one request.
2. The `http_client.post` method is monkeypatched to allow us to track the calls to the server.

We test the following scenarios:
* `limit=23` -> 3 requests: for 10, 10 and 3 samples.
* `limit=20` -> 2 requests: for 10, 10 samples.
* `limit=6` -> 1 request: for 6 samples.

There's also a test to cover the new ValueError if `limit` < 0.

## Effects
Consider the following script:
```python
import argilla as rg
import time

def test_scan_records(fields):
    client = rg.active_client()

    data = client.datasets.scan(
        name="banking77-topics-setfit",
        projection=fields,
        limit=1,  # <- Note, just one sample
    )
    start_t = time.time()
    list(data)
    print(f"load time for 1 sample with fields={fields}: {time.time() - start_t:.4f}s")

test_scan_records(set())
test_scan_records({"text"})
test_scan_records({"tokens"})
```

On this PR, this outputs:
```
load time for 1 sample with fields=set(): 0.0774s
load time for 1 sample with fields={'text'}: 0.0646s
load time for 1 sample with fields={'tokens'}: 0.0669s
```
On the `develop` branch, this outputs:
```
load time for 1 sample with fields=set(): 0.1355s
load time for 1 sample with fields={'text'}: 0.1347s
load time for 1 sample with fields={'tokens'}: 0.1173s
```
This can be repeated for `rg.load(..., limit=1)`, as that relies on `.scan` under the hood.

Note that this is the most extreme example of performance gains. The performance increases in almost all scenarios if `limit` is set, but the effects are generally not this big. Going from fetching 250 samples 8 times to fetching 250 samples 7 times and 173 once doesn't have as big of an impact.

---

**Type of change**

**How Has This Been Tested**

See above.

**Checklist**

---

- Tom Aarsen

### DIFF
--- a/src/argilla/client/apis/datasets.py
+++ b/src/argilla/client/apis/datasets.py
@@ -12,6 +12,7 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+import math
 import warnings
 from dataclasses import dataclass
 from datetime import datetime
@@ -163,23 +164,23 @@ class Datasets(AbstractApi):
             name: the dataset
             query: the search query
             projection: a subset of record fields to retrieve. If not provided,
-            limit: The number of records to retrieve
+                only id's will be returned
+            limit: The number of records to retrieve.
             id_from: If provided, starts gathering the records starting from that Record.
                 As the Records returned with the load method are sorted by ID, ´id_from´
                 can be used to load using batches.
-            only id's will be returned
 
         Returns:
-
             An iterable of raw object containing per-record info
-
         """
 
-        url = f"{self._API_PREFIX}/{name}/records/:search?limit={self.DEFAULT_SCAN_SIZE}"
-        query = self._parse_query(query=query)
+        if limit and limit < 0:
+            raise ValueError("The scan limit must be non-negative.")
 
-        if limit == 0:
-            limit = None
+        batch_size = self.DEFAULT_SCAN_SIZE
+        limit = limit if limit else math.inf
+        url = f"{self._API_PREFIX}/{name}/records/:search?limit={{limit}}"
+        query = self._parse_query(query=query)
 
         request = {
             "fields": list(projection) if projection else ["id"],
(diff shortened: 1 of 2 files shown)


tomaarsen:
I've taken it a little further than we originally discussed, i.e. I've also improved the performance if `limit > DEFAULT_SCAN_SIZE`. However, no new parameters (e.g. `batch_size`) are introduced yet, as I imagine there should be a discussion on how to introduce that throughout Argilla as a whole.

frascuchon (approved):
Good work, @tomaarsen !!!

davidberenstein1957:
Hi, @frascuchon @tomaarsen, currently `rg.log` uses `chunk_size`. I am in favour of deprecating this variable and moving over to the more intuitive `batch_size` when reviewing this discussion. If you don't agree, we should also use `chunk_size` within this PR.

tomaarsen:
I'm in favour of `batch_size` over `chunk_size`. I think (in a future PR?) we should expose `batch_size` in both functions.

dvsrepo:
I'm in favor of `batch_size` too

davidberenstein1957:
Yes, for me it can be the PR where/when/if we actually start using the `batch_size` within  `rg.load`

tomaarsen:
I think we're best off keeping that separate from this PR for now, but we should work on that soon. Do you want to raise an issue, @davidberenstein1957? (i.e. one issue with two parts: rename `chunk_size` to `batch_size` for `rg.log` and introduce `batch_size` for `rg.load`)

frascuchon:
Yes, I prefer to keep PRs as simple as possible. We can tackle the `batch_size` in 2 separate PRs:

- Deprecate the current `chunk_size` in favour of `batch_size` for `rg.load`
- Add a `batch_size` argument to `rg.load` to load data (adding also this new argument to `datasets.scan` method

UPDATE.

It's the same thing that wrote @tomaarsen one hour ago 😄

tomaarsen:
Great minds think alike ;)

Issue #2454, opened 2023-03-02
Introduce a `batch_size` argument for `rg.load`


tomaarsen:
**Is your feature request related to a problem? Please describe.**
I'd like to be able to specify a `batch_size` argument in `rg.load` to have more influence over potential timeouts.

**Describe the solution you'd like**
I'd like to be able to do:
```python
data = rg.load(name, batch_size=64)
```

**Describe alternatives you've considered**
Sticking with the current behaviour, i.e. using a batch size of 250 at all times.

**Additional context**
See https://github.com/argilla-io/argilla/pull/2434#issuecomment-1448303805

PR #2455, opened 2023-03-02
Deprecate `chunk_size` in favor of `batch_size` for `rg.log`

### LINKED ISSUE #2453: Deprecate the current `chunk_size` in favour of `batch_size` for `rg.log`
I'd like to be able to log using the more intuitive `batch_size` parameter for batching my records. I believe that `batch_size` better aligns with the remainder of Argilla.

I'd like to be able to perform
```python
rg.log(records, name, batch_size=64)
```

Sticking with the current `chunk_size`. 

**Additional context**
See https://github.com/argilla-io/argilla/pull/2434#issuecomment-1448092478

tomaarsen:
Closes #2453

## Pull Request overview
* Deprecate `chunk_size` in favor of `batch_size` in `rg.log`:
  * Move `chunk_size` to the end of all related signatures.
  * Set `chunk_size` default to None and update the typing accordingly.
  * Introduce `batch_size` in the old position in the signature.
  * Update docstrings accordingly.
  * Introduce a `FutureWarning` if `chunk_size` is used.
* Introduce test showing that `rg.log(..., chunk_size=100)` indeed throws a warning.
* Update a warning to no longer include a newline and a lot of spaces (see first comment of this PR)

## Details
Note that this deprecation is non-breaking: Code that uses `chunk_size` will still work, as `batch_size = chunk_size` after a FutureWarning is given, if `chunk_size` is not `None`.

## Discussion
* Should I use a FutureWarning? Or a DeprecationWarning? Or a PendingDeprecationWarning? The last two make sense, but they are [ignored by default](https://docs.python.org/3/library/warnings.html#default-warning-filter), I'm afraid. 
* Is the deprecation message in the format that we like?

---

**Type of change**

**How Has This Been Tested**

I introduced a test, and ran all tests.

**Checklist**

---

- Tom Aarsen

### DIFF
--- a/src/argilla/client/client.py
+++ b/src/argilla/client/client.py
@@ -115,7 +115,7 @@ class Argilla:
     """
 
     # Larger sizes will trigger a warning
-    _MAX_CHUNK_SIZE = 5000
+    _MAX_BATCH_SIZE = 5000
 
     def __init__(
         self,
@@ -256,9 +256,10 @@ class Argilla:
         name: str,
         tags: Optional[Dict[str, str]] = None,
         metadata: Optional[Dict[str, Any]] = None,
-        chunk_size: int = 500,
+        batch_size: int = 500,
         verbose: bool = True,
         background: bool = False,
+        chunk_size: Optional[int] = None,
     ) -> Union[BulkResponse, Future]:
         """Logs Records to argilla.
 
@@ -269,11 +270,12 @@ class Argilla:
             name: The dataset name.
             tags: A dictionary of tags related to the dataset.
             metadata: A dictionary of extra info for the dataset.
-            chunk_size: The chunk size for a data bulk.
+            batch_size: The batch size for a data bulk.
             verbose: If True, shows a progress bar and prints out a quick summary at the end.
             background: If True, we will NOT wait for the logging process to finish and return
                 an ``asyncio.Future`` object. You probably want to set ``verbose`` to False
                 in that case.
+            chunk_size: DEPRECATED! Use `batch_size` instead.
 
         Returns:
             Summary of the response from the REST API.
@@ -286,8 +288,9 @@ class Argilla:
             name=name,
             tags=tags,
             metadata=metadata,
-            chunk_size=chunk_size,
+            batch_size=batch_size,
             verbose=verbose,
+            chunk_size=chunk_size,
         )
         if background:
             return future
@@ -303,8 +306,9 @@ class Argilla:
         name: str,
         tags: Optional[Dict[str, str]] = None,
         metadata: Optional[Dict[str, Any]] = None,
-        chunk_size: int = 500,
+        batch_size: int = 500,
         verbose: bool = True,
+        chunk_size: Optional[int] = None,
     ) -> BulkResponse:
         """Logs Records to argilla with asyncio.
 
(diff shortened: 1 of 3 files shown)


src/argilla/client/client.py
```diff
@@ -115,7 +115,7 @@ class Argilla:
     """
 
     # Larger sizes will trigger a warning
-    _MAX_CHUNK_SIZE = 5000
+    _MAX_BATCH_SIZE = 5000
```
davidberenstein1957, line 118: @tomaarsen, FYI I believe we also need to take this into account for https://github.com/argilla-io/argilla/issues/2454

src/argilla/client/client.py
```diff
@@ -367,23 +379,23 @@ async def log_async(
...
                     ),
                 )
 
                 processed += response.parsed.processed
                 failed += response.parsed.failed
 
-                progress_bar.update(task, advance=len(chunk))
+                progress_bar.update(task, advance=len(batch))
 
         # TODO: improve logging policy in library
```
davidberenstein1957, line 400: hahaha nice to see this todo here

tests/client/test_api.py
```diff
@@ -228,6 +228,14 @@ def test_log_passing_empty_records_list(mocked_client):
         api.log(records=[], name="ds")
 
 
+def test_log_deprecated_chunk_size(mocked_client):
```
davidberenstein1957, line 231: maybe we can add an additional test w.r.t. to too large `batch_size`?

davidberenstein1957 (commented):
FutureWarning seems fine to me. I found some other `chunk-size` references within the `scan_data_response` methods for `src/argilla/server/apis/v0/handlers`. IMO these can also be renamed, but @frascuchon not sure if you agree.

@frascuchon these methods

src/argilla/client/client.py
tomaarsen: Sounds good, shall I introduce a warning like this for `rg.load`?
https://github.com/argilla-io/argilla/blob/f5834a5408051bf1fa60d42aede6b325dc07fdbd/src/argilla/client/client.py#L338-L343

davidberenstein1957:
Yes, but I think this can be handled within the PR for the other issue. However, because the batch_size was set by default, I wanted to mention before we forget.

tomaarsen:
I'll include it in the other PR that I'm preparing :)
Edit: I will not. See https://github.com/argilla-io/argilla/pull/2455#discussion_r1122883970

src/argilla/client/client.py
tomaarsen: I will not be doing this. For #2454, the server limits the maximum batch size to 500, so there is no need to have a warning when using a batch size of over 5000.

tests/client/test_api.py
tomaarsen: Done

frascuchon:
> FutureWarning seems fine to me. I found some other `chunk-size` references within the `scan_data_response` methods for `src/argilla/server/apis/v0/handlers`. IMO these can also be renamed, but @frascuchon not sure if you agree.
> 
> @frascuchon these methods

No worries about that. These endpoints are not used anymore and will be removed in v1.5.0

frascuchon (approved):

PR #2459, opened 2023-03-02
Add migration to create users on database from YAML config file

jfcalvo:
This PR adds a new users migration task that will allow to move users defined at a YAML file to database.

After checking that this is working as expected and it's what we really want we can made the necessary changes to our documentation (maybe in this current PR).

## Details about the migration process:

- It's executed inside a database transaction so an error in one specific user should rollback all the changes.
- I have added some Pydantic basic schemas to validate the content from the YAML file. Please check if some additional checks or validations can be added.
- The code is capturing the exception when the file is not found. Knowing that this is a task to be manually executed, maybe we should not capture the exception and let it crash.

## How to execute the migration process:

```sh
$ ARGILLA_LOCAL_AUTH_USERS_DB_FILE=users.yml python -m argilla.tasks.users_migrator
```


src/argilla/tasks/users_migrator.py
```diff
@@ -0,0 +1,90 @@
...
+        print(f"Migrating User with username {user['username']!r}")
+
+        user_create = self._build_user_create(user)
+
+        session.add(self._build_user(session, user_create))
+        session.flush()
+
+    def _build_user_create(self, user: dict):
+        return UserCreate(
+            first_name=user["full_name"],
```
frascuchon: I realized that this field is optional in the current model. Maybe we can just set as default the username if the `full_name was not provided in the file?

jfcalvo: Ouch, you're right. It's optional. We can use the username or use an empty string (we can modify the "local" Pydantic schema).

Note for myself: add this scenario to our tests.

frascuchon (approved):
This is working like a charm!

PR #2460, opened 2023-03-02
Expose `batch_size` parameter for `rg.load`

### LINKED ISSUE #2454: Introduce a `batch_size` argument for `rg.load`

tomaarsen:
Closes #2454 

Hello!

## Pull Request overview
* Extend my previous work to expose a `batch_size` parameter to `rg.load` (and `rg.scan`, `_load_records_new_fashion`).
  * Update the docstrings for the public methods.
  * The actual implementation is very simple due to my earlier work in #2434:
    * `batch_size = self.DEFAULT_SCAN_SIZE` -> `batch_size = batch_size or self.DEFAULT_SCAN_SIZE`
* Add an (untested!) warning if `batch_size` is supplied with an old server version.
  * (We don't have compatibility tests for old server versions)
* I've simplified `test_scan_efficient_limiting` with `batch_size` rather than using a mocker to update `DEFAULT_SCAN_SIZE`.
  * I've also extended this test to test `rg.load` in addition to `active_api().datasets.scan`.

## Note
Unlike previously discussed with @davidberenstein1957, I have *not* added a warning like so:
https://github.com/argilla-io/argilla/blob/f5834a5408051bf1fa60d42aede6b325dc07fdbd/src/argilla/client/client.py#L338-L343
The server limits the maximum batch size to 500 for loading, so there is no need to have a warning when using a batch size of over 5000.

## Discussion
1. Should I include in the docstring that the default batch size is 250? That would be "hardcoded" into the docs, so if we ever change the default (`self.DEFAULT_SCAN_SIZE`), then we would have to remember to update the docs too.
2. Alternatively, should we deprecate `self.DEFAULT_SCAN_SIZE` and enforce that `batch_size` must be set for `datasets.scan`, with a default size of 250 everywhere? Then people can see the default batch size in the signature?

I think we should do option 2. Let me know how you feel.

---

**Type of change**

**How Has This Been Tested**

Modified and updated a test, ran all tests.

**Checklist**

---

- Tom Aarsen

### DIFF
--- a/src/argilla/client/api.py
+++ b/src/argilla/client/api.py
@@ -206,6 +206,7 @@ def load(
     ids: Optional[List[Union[str, int]]] = None,
     limit: Optional[int] = None,
     id_from: Optional[str] = None,
+    batch_size: Optional[int] = None,
     as_pandas=None,
 ) -> Dataset:
     """Loads a argilla dataset.
@@ -220,6 +221,8 @@ def load(
         id_from: If provided, starts gathering the records starting from that Record.
             As the Records returned with the load method are sorted by ID, ´id_from´
             can be used to load using batches.
+        batch_size: If provided, load `batch_size` samples per request. A lower batch
+            size may help avoid timeouts.
         as_pandas: DEPRECATED! To get a pandas DataFrame do
             ``rg.load('my_dataset').to_pandas()``.
 
@@ -251,6 +254,7 @@ def load(
         ids=ids,
         limit=limit,
         id_from=id_from,
+        batch_size=batch_size,
         as_pandas=as_pandas,
     )
 
--- a/src/argilla/client/api.py
+++ b/src/argilla/client/api.py
@@ -206,7 +206,7 @@ def load(
     ids: Optional[List[Union[str, int]]] = None,
     limit: Optional[int] = None,
     id_from: Optional[str] = None,
-    batch_size: Optional[int] = None,
+    batch_size: int = 250,
     as_pandas=None,
 ) -> Dataset:
     """Loads a argilla dataset.
--- a/src/argilla/client/client.py
+++ b/src/argilla/client/client.py
@@ -439,7 +439,7 @@ class Argilla:
         ids: Optional[List[Union[str, int]]] = None,
         limit: Optional[int] = None,
         id_from: Optional[str] = None,
-        batch_size: Optional[int] = None,
+        batch_size: int = 250,
         as_pandas=None,
     ) -> Dataset:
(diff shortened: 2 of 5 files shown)


src/argilla/client/api.py
```diff
@@ -206,6 +206,7 @@ def load(
     ids: Optional[List[Union[str, int]]] = None,
     limit: Optional[int] = None,
     id_from: Optional[str] = None,
+    batch_size: Optional[int] = None,
```
frascuchon: Maybe it makes sense to have an explicit default value here. 
`batch_size: int = 250`

src/argilla/client/client.py
```diff
@@ -495,6 +497,13 @@ def load(
                 f"`pip install argilla=={err.api_version}`",
                 category=UserWarning,
             )
+            if batch_size is not None:
```
frascuchon, line 500: Nice catch!

frascuchon (approved):

src/argilla/client/api.py
tomaarsen: Will do, I'll use 250 throughout 👍

davidberenstein1957:
@frascuchon is the 500 max size a ES limit or an Argilla limit? 
If Argilla limit:
What happens if we increase it? If the python client defaults to 250, I would once again argue that it might be worth increasing/removing the server-side limit and letting this be configurable by the user. Most clients will just use the defaults and clients that want to scale, want to be able to configure this and potentially assign extra resources.

src/argilla/client/api.py
tomaarsen: Done

frascuchon:
@davidberenstein1957 It's an argilla endpoint limit. For elasticsearch will be set normally to 10000. Yes, we can increase this limit once the batch_size is exposed in `rg.load`. @tomaarsen could you include the change in this PR? The value to change is here https://github.com/argilla-io/argilla/blob/develop/src/argilla/server/apis/v0/handlers/records_search.py#L60

We could start  by aligning it with the current pagination value for search (1000)

tomaarsen:
@frascuchon Done, it's at 1000 now

frascuchon (approved):

PR #2461, opened 2023-03-02
Allow reading datasets that can contain empty info about dataset creator

frascuchon:
Some datasets can contain no info about the creator.

This change allows reading those datasets. This behavior will change with change in migration DB and the `created_by` will be a required field

### DIFF
--- a/tests/server/datasets/test_model.py
+++ b/tests/server/datasets/test_model.py
@@ -68,7 +68,7 @@ def test_dataset_creation_fails_on_no_workspace_and_owner():
         BaseDatasetDB(task=TaskType.text_classification, name="tedb", workspace=None, owner=None)
 
 
-def test_accept_create_dataset_with_creator():
+def test_accept_create_dataset_without_created_by():
     ds = Dataset(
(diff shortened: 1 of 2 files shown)


tests/server/datasets/test_model.py
```diff
@@ -64,3 +66,18 @@ def test_dataset_creation_sync(dataset, expected_workspace):
 def test_dataset_creation_fails_on_no_workspace_and_owner():
     with pytest.raises(ValueError, match="Missing workspace"):
         BaseDatasetDB(task=TaskType.text_classification, name="tedb", workspace=None, owner=None)
+
+
+def test_accept_create_dataset_with_creator():
```
jfcalvo:
```suggestion
def test_accept_create_dataset_without_created_by():
```

Or

```suggestion
def test_accept_create_dataset_without_creator():
```

frascuchon: Oh yes, thanks

jfcalvo (approved):

PR #2467, opened 2023-03-02
Add basic role support to users

jfcalvo:
- Add a new string `role` column to `users` table. The column is non nullable and indexed.
- Add `UserRole` class as enumerable. The class inherits from `Enum` and `str` (to avoid some JSON encoding problems, maybe we can create a custom JSON encoder to avoid possible problems).
- Add `role` as enumerable field to `User` SQLAlchemy model with `annotator` as default.
- Add `role` as optional for `UserCreate` Pydantic schema.

Changes has been tested locally and are working with PostgreSQL too.


frascuchon (approved):
Nice!

We will manage the current `superuser`  logic by using this new `role` info

PR #2468, opened 2023-03-02
Add new ARGILLA_HOME_PATH environment variable

jfcalvo:
This PR adds `ARGILLA_HOME_PATH` new environment variable with a default value equal to `~/.argilla`.

After that `ARGILLA_HOME_PATH` is used to define the default value for `ARGILLA_DATABASE_URL`.

I have added these changes to `_const.py` file but maybe we should create a `settings.py` file or something similar.

### DIFF
--- a/src/argilla/_constants.py
+++ b/src/argilla/_constants.py
@@ -13,8 +13,16 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-DEFAULT_MAX_KEYWORD_LENGTH = 128
+import os
+from pathlib import Path
 
+ARGILLA_HOME_PATH = os.getenv("ARGILLA_HOME_PATH", os.path.join(Path.home(), ".argilla"))
+ARGILLA_DATABASE_URL = os.getenv("ARGILLA_DATABASE_URL", f"sqlite:///{os.path.join(ARGILLA_HOME_PATH, 'argilla.db')}?check_same_thread=False")
+
+# Create argilla home path directory if it doesn't exist
+Path(ARGILLA_HOME_PATH).mkdir(parents=True, exist_ok=True)
+
+DEFAULT_MAX_KEYWORD_LENGTH = 128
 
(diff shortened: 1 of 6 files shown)


frascuchon:
We are using this https://docs.pydantic.dev/usage/settings for managing app settings from environment vars.

I think (need to validate) this is also compatible with the idea of an environment file (json, ini or whatever) containing all the application settings (See https://docs.pydantic.dev/usage/settings/#adding-sources)

So, maybe we can add these new env vars to the common app settings that we have already defined in https://github.com/argilla-io/argilla/blob/develop/src/argilla/server/settings.py#L28

src/argilla/_constants.py
```diff
@@ -13,8 +13,18 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-DEFAULT_MAX_KEYWORD_LENGTH = 128
+import os
+from pathlib import Path
+
+ARGILLA_HOME_PATH = os.getenv("ARGILLA_HOME_PATH", os.path.join(Path.home(), ".argilla"))
+ARGILLA_DATABASE_URL = os.getenv(
```
frascuchon: Read my general comment in PR

jfcalvo:
> We are using this https://docs.pydantic.dev/usage/settings for managing app settings from environment vars.
> 
> I think (need to validate) this is also compatible with the idea of an environment file (json, ini or whatever) containing all the application settings (See https://docs.pydantic.dev/usage/settings/#adding-sources)
> 
> So, maybe we can add these new env vars to the common app settings that we have already defined in https://github.com/argilla-io/argilla/blob/develop/src/argilla/server/settings.py#L28

I noticed the `settings.py` file but because the class was named `ApiSettings` I thought it was related with the API alone. Maybe we need to change it to `AppSettings` or something similar because this is not only related with the API.

frascuchon (approved):
Yeah!

PR #2469, opened 2023-03-02
Add user workspace validations

frascuchon:
This PR replicates workspace/user validations using new DB models.

This PR shows where to define the validations, and how to use them in handlers. Once this is well defined/validated, we will  remove the old validation implemented in pydantic `User` class (`check_workspace`...) in a separate PR.

### DIFF
--- a/src/argilla/server/contexts/validations.py
+++ b/src/argilla/server/contexts/validations.py
@@ -1,0 +1,36 @@
+#  Copyright 2021-present, the Recognai S.L. team.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+from argilla.server.contexts import accounts
+from argilla.server.database import SessionLocal
+from argilla.server.errors import EntityNotFoundError, ForbiddenOperationError
+from argilla.server.models import User, Workspace
+
+
+def validate_workspaces_user_access(user: User, *workspaces: Workspace):
+    for workspace in workspaces:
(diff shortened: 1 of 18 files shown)


src/argilla/server/contexts/validations.py
```diff
@@ -0,0 +1,36 @@
...
+        if workspace in user.workspaces:
+            continue
+        elif accounts.is_superuser(user):
+            continue
+        else:
+            raise ForbiddenOperationError(f"Cannot access workspace {workspace}")
+    return True
+
+
+def validate_and_get_workspace_by_name(db: SessionLocal, workspace_name: str) -> Workspace:
```
jfcalvo: @frascuchon we can create a function on `accounts` context with the following code:

```python
def get_workspace_by_name_or_raise(db: Session, workspace_name: str):
    workspace = get_workspace_by_name(db, workspace_name)

    if not workspace:
        raise EntityNotFoundError(name=str(workspace.id), type=Workspace)

    return workspace
```

frascuchon: Done for workspace_by_name. See the code [here](https://github.com/argilla-io/argilla/blob/a5b3294a3423d0724ab88749335669e252d35393/src/argilla/server/contexts/accounts.py#L63)

src/argilla/server/contexts/validations.py
```diff
@@ -0,0 +1,25 @@
...
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+from argilla.server.contexts import accounts
+from argilla.server.models import User, Workspace
+
+
+def can_user_access_to_workspace(user: User, workspace: Workspace):
```
jfcalvo: I think we need more granularity with the permissions. Maybe something like `is_authorized(user, "create", workspace)` or something similar, "access" is a little bit ambiguous.

In any case we can have a discussion about this.

frascuchon: I have done some changes here. We have no functional pattern matching in python so, this kind of function definition could grow easily when we need to authorize several resources for a user. Anyway, let's discuss this from here on Monday.

jfcalvo:
Closing. A different approach has been defined on https://github.com/argilla-io/argilla/pull/2491

PR #2486, opened 2023-03-06
Run alembic migrations for server startup and users migrator

jfcalvo:
This PR adds a new `run_db_migrations` function to `argilla.server.database` package which execute alembic migrations like `$ alembic upgrade head` command.

This new function is used on the following places:
- On `argilla.tasks.users_migrator` execution
- On `argilla.server` startup.

Once that we create a  new task to add users on demand we should run this function too.


jfcalvo:
We have finally decided that database migrations will be on the documentation and it's something that users should execute in the own argilla installations.

PR #2488, opened 2023-03-06
Create task CLI for Argilla users creation

frascuchon:
Adding a task to create users from command line.

```bash 
# To create an user
 python -m argilla.tasks.create_user --username 3233edrito2df --role admin --password 12345678932

# To show help
python -m argilla.tasks.create_user --help
```

TODO:
- [x] Add tests

### DIFF
--- a/src/argilla/tasks/create_user.py
+++ b/src/argilla/tasks/create_user.py
@@ -19,9 +19,7 @@ import click
 from argilla.server.contexts import accounts
 from argilla.server.database import SessionLocal
 from argilla.server.models import User, UserRole
-from argilla.server.security.model import (
-    UserCreate,
-)
+from argilla.server.security.model import UserCreate
 
 
 def _show_created_user(user: User):
--- a/src/argilla/tasks/create_user.py
+++ b/src/argilla/tasks/create_user.py
@@ -15,16 +15,11 @@
 from typing import Optional
 
 import click
-from pydantic import BaseModel, constr
 
 from argilla.server.contexts import accounts
-from argilla.server.contexts.accounts import CRYPT_CONTEXT
 from argilla.server.database import SessionLocal
 from argilla.server.models import User, UserRole
 from argilla.server.security.model import (
-    USER_PASSWORD_MAX_LENGTH,
-    USER_PASSWORD_MIN_LENGTH,
-    USER_USERNAME_REGEX,
     UserCreate,
 )
 
--- a/src/argilla/tasks/create_user.py
+++ b/src/argilla/tasks/create_user.py
@@ -17,6 +17,7 @@ from typing import Optional
 import click
 from pydantic import BaseModel, constr
 
+from argilla.server.contexts import accounts
 from argilla.server.contexts.accounts import CRYPT_CONTEXT
 from argilla.server.database import SessionLocal
 from argilla.server.models import User, UserRole
@@ -24,39 +25,38 @@ from argilla.server.security.model import (
     USER_PASSWORD_MAX_LENGTH,
     USER_PASSWORD_MIN_LENGTH,
     USER_USERNAME_REGEX,
+    UserCreate,
 )
 
 
-class UserCreate(BaseModel):
-    username: constr(regex=USER_USERNAME_REGEX, min_length=1)
-    role: Optional[UserRole]
-    password: constr(min_length=USER_PASSWORD_MIN_LENGTH, max_length=USER_PASSWORD_MAX_LENGTH)
-    api_key: Optional[str]
+def _show_created_user(user: User):
+    message = f"\nUsername: {user.username!r}" f"\nRole: {user.role.value!r}" f"\nFirst name: {user.first_name!r}"
+
+    if user.last_name:
+        message += f"\nLast name: {user.last_name!r}"
+
+    message += f"\nAPI-Key: {user.api_key!r}"
+
+    return message
 
 
 @click.command()
+@click.option("--first-name", prompt=True)
+@click.option("--last-name")
 @click.option("--username", prompt=True)
+@click.option(
+    "--role", prompt=True, type=click.Choice(UserRole.__members__, case_sensitive=False), default=UserRole.annotator
+)
 @click.option("--password", prompt=True, hide_input=True, confirmation_prompt=True)
-@click.option("--role", type=click.Choice(UserRole, case_sensitive=False))
-@click.option("--api-key")
-def create_user(username: str, password: str, role: Optional[UserRole], api_key: Optional[str]):
-    with SessionLocal() as session, session.begin():
-        user_create = UserCreate(username=username, password=password, role=role)
-
-        user = User(
-            first_name="",
-            username=user_create.username,
-            role=user_create.role,
-            password_hash=CRYPT_CONTEXT.hash(user_create.password),
-            api_key=api_key,
+def create_user(username: str, password: str, role: Optional[UserRole], first_name: str, last_name: Optional[str]):
+    with SessionLocal() as session:
+        user_create = UserCreate(
+            username=username, password=password, role=role, first_name=first_name, last_name=last_name
         )
-
-        session.add(user)
-        session.flush()
-        session.refresh(user)
+        user = accounts.create_user(session, user_create)
 
         print(f"User created successfully!")
-        print(f"\nUsername: {user.username!r}" f"\nRole: {user.role.value!r}" f"\nAPI-Key: {user.api_key!r}")
+        print(_show_created_user(user))
         print("\nPlease, save user credentials")
 
 
--- a/src/argilla/tasks/create_user.py
+++ b/src/argilla/tasks/create_user.py
@@ -1,0 +1,64 @@
+#  Copyright 2021-present, the Recognai S.L. team.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+from typing import Optional
+
+import click
+from pydantic import BaseModel, constr
+
+from argilla.server.contexts.accounts import CRYPT_CONTEXT
+from argilla.server.database import SessionLocal
+from argilla.server.models import User, UserRole
+from argilla.server.security.model import (
+    USER_PASSWORD_MAX_LENGTH,
+    USER_PASSWORD_MIN_LENGTH,
+    USER_USERNAME_REGEX,
+)
+
+
+class UserCreate(BaseModel):
+    username: constr(regex=USER_USERNAME_REGEX, min_length=1)
+    role: Optional[UserRole]
+    password: constr(min_length=USER_PASSWORD_MIN_LENGTH, max_length=USER_PASSWORD_MAX_LENGTH)
+    api_key: Optional[str]
+
+
+@click.command()
+@click.option("--username", prompt=True)
+@click.option("--password", prompt=True, hide_input=True, confirmation_prompt=True)
+@click.option("--role", type=click.Choice(UserRole, case_sensitive=False))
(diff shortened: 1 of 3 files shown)


tests/tasks/test_create_user.py
```diff
@@ -0,0 +1,146 @@
...
+    runner = CliRunner()
+    cli_params = f"--username {username} --password {password} --role admin --first-name First --last-name {last_name}"
+    runner.invoke(create_user, cli_params)
+
+    user = accounts.get_user_by_username(db, username)
+    assert user
+    assert user.last_name == last_name
+
+
+def test_create_user_with_wrong_username(db):
```
jfcalvo: We are using `invalid` instead of `wrong` in the rest of the tests.

```suggestion
def test_create_user_with_invalid_username(db: Session):
```

tests/tasks/test_create_user.py
```diff
@@ -0,0 +1,146 @@
...
+
+def test_create_user_passing_all_params(db):
+    username = "test-user"
+    first_name = "First"
+    role = "admin"
+    password = "12345678"
+
+    runner = CliRunner()
+    cli_params = f"--username {username} --password {password} --role {role} --first-name {first_name}"
+    runner.invoke(create_user, cli_params)
```
jfcalvo: If we are not using the `CliRunner` instance then I guess we can use it right away:

```suggestion
    cli_params = f"--username {username} --password {password} --role {role.value} --first-name {first_name}"
    CliRunner().invoke(create_user, cli_params)
```

tests/tasks/test_create_user.py
```diff
@@ -0,0 +1,146 @@
...
+@pytest.fixture
+def db():
+    with SessionLocal() as db, db.begin():
+        yield db
+
+        db.query(User).delete()
+        db.commit()
+
+
+def test_create_user_passing_all_params(db):
```
jfcalvo:
```suggestion
def test_create_user_passing_all_params(db: Session):
```

tests/tasks/test_create_user.py
```diff
@@ -0,0 +1,146 @@
...
+    username = "test-user"
+    first_name = "First"
+    role = "admin"
+    password = "12345678"
+
+    runner = CliRunner()
+    cli_params = f"--username {username} --password {password} --role {role} --first-name {first_name}"
+    runner.invoke(create_user, cli_params)
+
+    user = accounts.get_user_by_username(db, username)
```
jfcalvo: The reason why I try to not use `accounts` in the tests (if possible) is to isolate what we are testing in the test from the rest of the application logic.

```suggestion
    user = db.query(User).filter_by(username="username").first()
```

tests/tasks/test_create_user.py
```diff
@@ -0,0 +1,146 @@
...
+        yield db
+
+        db.query(User).delete()
+        db.commit()
+
+
+def test_create_user_passing_all_params(db):
+    username = "test-user"
+    first_name = "First"
+    role = "admin"
```
jfcalvo: `UserRole.admin` will interpolate correctly to string in the f string:

```suggestion
    role = UserRole.admin
```

tests/tasks/test_create_user.py
```diff
@@ -0,0 +1,146 @@
...
+    password = "12345678"
+
+    runner = CliRunner()
+    cli_params = f"--username {username} --password {password} --role {role} --first-name {first_name}"
+    runner.invoke(create_user, cli_params)
+
+    user = accounts.get_user_by_username(db, username)
+    assert user
+    assert user.role.value == role
```
jfcalvo:
```suggestion
    assert user.role == UserRole.admin
```

tests/tasks/test_create_user.py
```diff
@@ -0,0 +1,146 @@
...
+    assert accounts.get_user_by_username(db, username) is None
+    assert str(result.exception) == (
+        "1 validation error for UserCreate\n"
+        "username\n"
+        '  string does not match regex "^(?!-|_)[a-z0-9-_]+$" '
+        "(type=value_error.str.regex; pattern=^(?!-|_)[a-z0-9-_]+$)"
+    )
+
+
+def test_create_user_with_wrong_password(db):
```
jfcalvo:
```suggestion
def test_create_user_with_invalid_password(db: Session):
```

tests/tasks/test_create_user.py
```diff
@@ -0,0 +1,146 @@
...
+    cli_params = f"--username {username} --password {password} --role admin --first-name First --last-name {last_name}"
+    runner.invoke(create_user, cli_params)
+
+    user = accounts.get_user_by_username(db, username)
+    assert user
+    assert user.last_name == last_name
+
+
+def test_create_user_with_wrong_username(db):
+    username = "Wrong-username "
```
jfcalvo:
```suggestion
    username = "Invalid-Username"
```

tests/tasks/test_create_user.py
```diff
@@ -0,0 +1,146 @@
...
+def test_create_user_with_input_username(db):
+    username = "test-user"
+    password = "12345678"
+
+    runner = CliRunner()
+    cli_params = f"--password {password} --first-name First"
+    runner.invoke(create_user, cli_params, input=f"{username}\n")
+
+    user = accounts.get_user_by_username(db, username)
+    assert user
```
jfcalvo:
```suggestion
    assert accounts.get_user_by_username(db, username)
```

src/argilla/tasks/create_user.py
```diff
@@ -0,0 +1,59 @@
...
+from typing import Optional
+
+import click
+
+from argilla.server.contexts import accounts
+from argilla.server.database import SessionLocal
+from argilla.server.models import User, UserRole
+from argilla.server.security.model import (
+    UserCreate,
+)
```
jfcalvo:
```suggestion
from argilla.server.security.model import UserCreate
```

src/argilla/tasks/create_user.py
```diff
@@ -0,0 +1,59 @@
...
+
+@click.command()
+@click.option("--first-name", prompt=True)
+@click.option("--last-name")
+@click.option("--username", prompt=True)
+@click.option(
+    "--role", prompt=True, type=click.Choice(UserRole.__members__, case_sensitive=False), default=UserRole.annotator
+)
+@click.option("--password", prompt=True, hide_input=True, confirmation_prompt=True)
+def create_user(username: str, password: str, role: Optional[UserRole], first_name: str, last_name: Optional[str]):
```
jfcalvo: Should we order here the params in the same way that the click options?

jfcalvo (changes requested):

src/argilla/tasks/create_user.py
frascuchon: Yes, it should be better

tests/tasks/test_create_user.py
frascuchon: good point, yes

tests/tasks/test_create_user.py
```diff
@@ -0,0 +1,138 @@
...
+
+@pytest.fixture
+def db():
+    with SessionLocal() as db, db.begin():
+        yield db
+
+        db.query(User).delete()
+        db.commit()
+
+
```
jfcalvo: This is not necessary we're already "truncating" `users` table on our original `db` fixture:

```suggestion
```

src/argilla/tasks/create_user.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+@click.command("create-user")
+@click.option("--first-name", prompt=True)
+@click.option("--username", prompt=True)
+@click.option(
+    "--role", prompt=True, type=click.Choice(UserRole.__members__, case_sensitive=False), default=UserRole.annotator
+)
+@click.option("--password", prompt=True, hide_input=True, confirmation_prompt=True)
+@click.option("--last-name")
```
jfcalvo: Possible changes/improvements:
- Remove the name of the command from `@click.command` because it takes the name of the function (but it's totally ok if you want to still use it).
- Follow the same order of the columns that we have in the database model (I have changed them to follow that order in the suggestion).
- Use `@click.password_option()` for the password option. It's exactly the same that you are using (but it's ok to be explicit if you want).
- Add `show_default=True` to ` --role` option (so it shows the default value if you pass `--help` to the task).

```suggestion
@click.command()
@click.option("--first-name", prompt=True)
@click.option("--last-name")
@click.option("--username", prompt=True)
@click.option(
    "--role",
    prompt=True,
    type=click.Choice(UserRole.__members__, case_sensitive=False),
    default=UserRole.annotator,
    show_default=True,
)
@click.password_option()
```

src/argilla/tasks/create_user.py
```diff
@@ -0,0 +1,57 @@
...
+def create_user(first_name: str, username: str, role: UserRole, password: str, last_name: Optional[str]):
```
jfcalvo, line 48: If you think changing the order of params to follow the database model it's fine you can apply this suggestion too:

```suggestion
def create_user(first_name: str, last_name: Optional[str], username: str, role: UserRole, password: str):
```

src/argilla/tasks/create_user.py
```diff
@@ -0,0 +1,57 @@
...
+    with SessionLocal() as session:
+        user_create = UserCreate(
+            username=username, password=password, role=role, first_name=first_name, last_name=last_name
+        )
+        user = accounts.create_user(session, user_create)
```
jfcalvo, line 55: You are not using `user_create` so maybe you can use it directly as parameter:

```suggestion
        user = accounts.create_user(
            session,
            UserCreate(
                first_name=first_name,
                last_name=last_name,
                username=username,
                role=role,
                password=password,
            )
        )

```

tests/tasks/test_create_user.py
```diff
@@ -0,0 +1,138 @@
...
+def test_create_user_passing_all_params(db: Session):
+    username = "test-user"
+    first_name = "First"
+    role = UserRole.admin
+    password = "12345678"
+
+    cli_params = f"--username {username} --password {password} --role {role.value} --first-name {first_name}"
```
jfcalvo: You can use `role` instead of `role.value` and improve order:

```suggestion
    cli_params = f"--first-name {first_name} --username {username} --role {role} --password {password}"
```

jfcalvo (changes requested):

src/argilla/tasks/create_user.py
frascuchon: DOne

src/argilla/tasks/create_user.py
```diff
@@ -0,0 +1,57 @@
...
+
+        print(f"User created successfully!")
```
jfcalvo: You can use `click.echo` instead (the same for the rest of prints):

```suggestion
        click.echo(f"User created successfully!")
```

src/argilla/tasks/create_user.py
jfcalvo, line 48: You should add a comment after the function declaration so it's showed when `--help` is used:

```suggestion
def create_user(first_name: str, username: str, role: UserRole, password: str, last_name: Optional[str]):
    """Creates a new user into the database with the specified parameters blah blah blah.........."""
```

src/argilla/tasks/create_user.py
frascuchon: I prefer keep 2 step on this. If some error is raised within the pydantic model, the trace will be more accurate to where the errors is raised.

src/argilla/tasks/create_user.py
frascuchon: +1

src/argilla/tasks/create_user.py
frascuchon: +1

src/argilla/tasks/create_user.py
frascuchon: > Possible changes/improvements:
> 
> * Remove the name of the command from `@click.command` because it takes the name of the function (but it's totally ok if you want to still use it).

Totally, I was doing some tests and I forgot remove the name.

> * Follow the same order of the columns that we have in the database model (I have changed them to follow that order in the suggestion).

I prefer keep the order only for prompted and required args. Since last_name is not requested in prompt and is an optional param, I would like keep at the end of function signature.

> * Use `@click.password_option()` for the password option. It's exactly the same that you are using (but it's ok to be explicit if you want).

+1 

> * Add `show_default=True` to ` --role` option (so it shows the default value if you pass `--help` to the task).

+1

jfcalvo (approved):

PR #2491, opened 2023-03-07
Add basic authorization policies

jfcalvo:
This is a first proposal of the new authorization policies after visiting them for the first time on @frascuchon PR: https://github.com/argilla-io/argilla/pull/2469

The general idea is to have a `policies` package where we can create different policies classes including the authorization logic for every model in our application.

The `policies` package will include also two method to facilitate check authorization policies on API handlers.
- `authorized`: that will raise a `ForbiddenOperationError` exception if the policy does not check.
- `is_authorized`: that will return `True` or `False` depending if the policy checks or not.

Things to have into account:
- The user that we get as `current_user` on API handlers is not an SQLAlchemy ORM `User` model (we need to take a look to this @frascuchon).
- The creation of default user on server startup it's breaking some of my tests so I have disabled it (we can discuss if with the new user creation task this can be removed).
- There is an alternative version of authorization policies that we can implement using closures. It has a lot of flexibility but adds a little bit of complexity to the code.

### DIFF
--- a/src/argilla/server/policies.py
+++ b/src/argilla/server/policies.py
@@ -18,61 +18,52 @@ from argilla.server.models import User, UserRole, Workspace, WorkspaceUser
 
 class WorkspaceUserPolicy:
     @classmethod
-    def list(cls, actor: User, workspace_user: WorkspaceUser):
-        return True
+    def list(cls, actor: User, workspace_user: WorkspaceUser) -> bool:
+        return actor.role == UserRole.admin
 
     @classmethod
-    def create(cls, actor: User, workspace_user: WorkspaceUser):
-        return True
+    def create(cls, actor: User, workspace_user: WorkspaceUser) -> bool:
+        return actor.role == UserRole.admin
 
     @classmethod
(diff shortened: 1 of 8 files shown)


src/argilla/server/policies.py
```diff
@@ -0,0 +1,69 @@
...
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+from argilla.server.errors import ForbiddenOperationError
+from argilla.server.models import User, UserRole, Workspace, WorkspaceUser
+
+
+class WorkspaceUserPolicy:
+    @classmethod
+    def list(cls, actor: User, workspace_user: WorkspaceUser) -> bool:
```
frascuchon: Since we're using the list policy without params, it's a bit weird having the resource in the definition. Would it be easy to remove it from the method signature?

jfcalvo: After talking with @frascuchon we have decide to take some time to try dynamic arguments aka `kargs` for policy actions.

frascuchon (approved):

PR #2498, opened 2023-03-07
Remove creation of default user from server startup

jfcalvo:
This PR removes the creation of the default user from the server startup process.

The idea is to move this functionality to a new task that will allow users to create the default `argilla` user from command line.


frascuchon (approved):
Good!

PR #2499, opened 2023-03-07
refactor:  send username when no workspace setup

frascuchon:
1. Client will always send the workspace (even the username for "Private" workspaces)
2. The workspace is not required for this current version, since old clients don't send workspace when matches with username


src/argilla/client/client.py
```diff
@@ -152,8 +152,7 @@ def __init__(
         )
 
         self._user: User = users_api.whoami(client=self._client)
-        if workspace is not None:
-            self.set_workspace(workspace)
+        self.set_workspace(workspace or self.user.username)
```
```suggestion
        self.set_workspace(workspace or self._user.username)
```

PR #2502, opened 2023-03-08
Add task to create default user on database

jfcalvo:
This PR adds a new task to create user with defaults credentials on database.

```sh
# Get help
$ python -m argilla.tasks.users.create_default --help

# Create the default user
$ python -m argilla.tasks.users.create_default

# Create the default user without output
$ python -m argilla.tasks.users.create_default --quiet
```

I have refactored a little bit some `contexts.accounts` functions related with passwords, maybe it could be applied later to some other PRs on the go.


frascuchon (approved):
Cool!

PR #2516, opened 2023-03-10
Support api-key and password as parameter when using users.create_default task

jfcalvo:
This PR includes the following changes:
* Adds `--api-key` and `--password` parameters to `argilla.tasks.users.create_default`.
* Now `argilla.tasks.users.create_default` checks if there is a user using default username before creating it and showing a message in such case (but not raising an error).

### DIFF
--- a/src/argilla/tasks/users/create_default.py
+++ b/src/argilla/tasks/users/create_default.py
@@ -14,10 +14,11 @@
 
 import click
 
-from argilla._constants import DEFAULT_API_KEY, DEFAULT_PASSWORD, DEFAULT_USERNAME
+from argilla._constants import DEFAULT_USERNAME
 from argilla.server.contexts import accounts
 from argilla.server.database import SessionLocal
 from argilla.server.models import User, UserRole, Workspace
+from argilla.server.settings import settings
 
 
 @click.command()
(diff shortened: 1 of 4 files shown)


src/argilla/tasks/users/create_default.py
```diff
@@ -30,17 +31,17 @@ def create_default(quiet: bool):
                 first_name="",
                 username=DEFAULT_USERNAME,
                 role=UserRole.admin,
-                api_key=DEFAULT_API_KEY,
-                password_hash=accounts.hash_password(DEFAULT_PASSWORD),
+                api_key=settings.api_key,
```
frascuchon: If these settings are used only for the argilla task, I would prefer using a click.option instead, and calling the tasks:

```bash
python -m argilla.tasks.users.create_default --password $MY_PASS --api-key $MY_API_KEY
```

For me, it's a bit weird having app env vars used only on this task.

jfcalvo: I agree but then `users.create_default` is useless because we already have `users.create` (and it will have the possibility of setting the api key, that is necessary for the team user of the QuickStart docker image).

frascuchon: I Agree. Also, in docs we can just say: 

"Create the default user by typing:

```bash
argilla users create --username argilla --password 1234 --api-key argilla.apiKey --role admin
```

# EDIT
Indeed is not possible since the password constraints. And I would like to keep this constraints  for the common user creation command

frascuchon (approved):

PR #2523, opened 2023-03-13
Return all workspaces for admin users

frascuchon:
The current client will check if a user can work on a specific workspace by using workspaces info returning in `/api/me`  Returning all workspaces from the `/api/me` for admin users keeps the backward compatibility with the client flow.


src/argilla/server/apis/v0/handlers/users.py
```diff
@@ -52,7 +56,13 @@ async def whoami(request: Request, current_user: models.User = Security(auth.get
 
     await telemetry.track_login(request, username=current_user.username)
 
-    return User.from_orm(current_user)
+    user = User.from_orm(current_user)
+
+    if current_user.is_admin:
```
jfcalvo, line 67: I think we should add a TODO comment here explaining why this logic is necessary and the conditions when it can be removed.

src/argilla/server/apis/v0/handlers/users.py
```diff
@@ -52,7 +56,13 @@ async def whoami(request: Request, current_user: models.User = Security(auth.get
...
+        workspaces = accounts.list_workspaces(db)
+        user.workspaces.extend([workspace.name for workspace in workspaces])
```
jfcalvo: Correct me if I'm wrong but I believe you can finish with duplications in the workspaces. For example when the user has one workspace you are adding that same workspace at the end.

jfcalvo (changes requested):

src/argilla/server/apis/v0/handlers/users.py
jfcalvo: If you fix this please add a test for this scenario.

src/argilla/server/apis/v0/handlers/users.py
frascuchon: Done

src/argilla/server/apis/v0/handlers/users.py
```diff
@@ -20,7 +20,7 @@
 from pydantic import parse_obj_as
 from sqlalchemy.orm import Session
 
-from argilla.server import models
+from argilla.server import database, models
```
jfcalvo: You don't need to import `database` we are importing `get_db` later:

```suggestion
from argilla.server import models
```

src/argilla/server/apis/v0/handlers/users.py
```diff
@@ -33,7 +33,11 @@
 
 
 @router.get("/me", response_model=User, response_model_exclude_none=True, operation_id="whoami")
-async def whoami(request: Request, current_user: models.User = Security(auth.get_current_user)):
+async def whoami(
+    request: Request,
+    db: Session = Depends(database.get_db),
```
jfcalvo:
```suggestion
    db: Session = Depends(get_db),
```

tests/server/api/v0/test_users.py
```diff
@@ -37,6 +43,43 @@ def test_me_without_authentication(client: TestClient):
...
+    response_body = response.json()
+
+    assert response_body["id"] == str(admin.id)
+
+    user_workspace_names = response_body["workspaces"]
+    user_workspace_names.sort()
+
+    assert user_workspace_names == [ws1.name, ws2.name]
+
+
```
jfcalvo: I think you should create one for admin, another for non-admin (annotator) and that's it:

```suggestion
def test_me_as_admin(client: TestClient, db: Session):
    admin = AdminFactory.create(
        workspaces=[
            WorkspaceFactory.build(name="workspace-a"),
            WorkspaceFactory.build(name="workspace-b")
        ]
    )
    WorkspaceFactory.create(name="workspace-c")

    response = client.get("/api/me", headers={API_KEY_HEADER_NAME: admin.api_key})

    assert response.status_code == 200

    response_body = response.json()
    assert response_body["id"] == str(admin.id)
    assert response_body["workspaces"] == ["workspace-a", "workspace-b", "workspace-c"]


def test_me_as_annotator(client: TestClient, db: Session):
    annotator = AnnotatorFactory.create(
        workspaces=[
            WorkspaceFactory.build(name="workspace-a"),
            WorkspaceFactory.build(name="workspace-b")
        ]
    )
    WorkspaceFactory.create(name="workspace-c")

    response = client.get("/api/me", headers={API_KEY_HEADER_NAME: annotator.api_key})

    assert response.status_code == 200

    response_body = response.json()
    assert response_body["id"] == str(annotator.id)
    assert response_body["workspaces"] == ["workspace-a", "workspace-b"]


```

jfcalvo (changes requested):

jfcalvo (approved):

PR #2539, opened 2023-03-14
fix: create equal sized batches within batch size for logging #2341

### LINKED ISSUE #2341: create equally sized batches before/during `rg.log`
I just wanted to upload 35K records and logging is not as robust as I would like.

Ideally, I would like to divide records based on the allowed batch size, while equally dividing the text lengths over these batches. IMO, this could already alleviate some of the problems relatively easily. And this might also help with assessing and analyzing payload limitations(each payload is roughly equal in this case).

N.A.

**Additional context**
N.A.

davidberenstein1957:
# Description

dives logging data in equal batches in two ways
1. len(record) ==700 AND batch_size == 500 => batch_size = 350
2. if size of record [2,2,1,1] batches 0=>[[2,1][2,1]]

Closes #2341 
**Type of change**

**How Has This Been Tested**

Not sure how to test this

**Checklist**
N.A.

### DIFF
--- a/src/argilla/client/client.py
+++ b/src/argilla/client/client.py
@@ -393,13 +393,18 @@ class Argilla:
         # divide records in chunks of chunk size based on their byte size
         batch_size = int(batch_size / (len(records) / batch_size)) + 1
         n_batches = math.ceil(len(records) / batch_size)
-        rec_size = [sys.getsizeof(rec.json()) for rec in records]
-        records = [rec for _, rec in sorted(zip(rec_size, records), key=lambda pair: pair[0])]
-        records_new = dict.fromkeys(range(n_batches + 1), [])
-        while records:
-            for i in records_new:
-                records_new[i].append(records.pop(0))
-        records = [item for sublist in records_new.values() for item in sublist]
+        
+        if n_batches > 1:
+            rec_size = [sys.getsizeof(rec.json()) for rec in records]
+            records = [rec for _, rec in sorted(zip(rec_size, records), key=lambda pair: pair[0])]
+            
+            records_new = dict.fromkeys(range(n_batches), [])
+            for idx, record in enumerate(records):
+                records_new[idx % n_batches].append(record)
+            
+            records = []
+            for batch_list in records_new.values():
+                records.extend(batch_list)
 
         # log records
         processed, failed = 0, 0
--- a/src/argilla/client/client.py
+++ b/src/argilla/client/client.py
@@ -393,15 +393,15 @@ class Argilla:
         # divide records in chunks of chunk size based on their byte size
         batch_size = int(batch_size / (len(records) / batch_size)) + 1
         n_batches = math.ceil(len(records) / batch_size)
-        
+
         if n_batches > 1:
             rec_size = [sys.getsizeof(rec.json()) for rec in records]
             records = [rec for _, rec in sorted(zip(rec_size, records), key=lambda pair: pair[0])]
-            
+
             records_new = dict.fromkeys(range(n_batches), [])
             for idx, record in enumerate(records):
                 records_new[idx % n_batches].append(record)
-            
+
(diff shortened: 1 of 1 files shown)


tomaarsen:
@davidberenstein1957 
I've previously attempted similar efforts to try and figure out how "large" some arbitrary Python objects are, and I came to the conclusion that `sys.getsizeof` is a really poor way to measure the size of an object, as it only works consistently with builtin objects. See this example:
```python
>>> import torch
>>> import sys
>>> vector = torch.rand(10000) 
>>> vector
tensor([0.3660, 0.8881, 0.5079,  ..., 0.2980, 0.9991, 0.7953])
>>> sys.getsizeof(vector)
72
>>> sys.getsizeof("Just an arbitrary string, for comparison purposes")
98
```

Personally, I ended up using `len(pickle.dumps(obj))`, but I used it to measure the memory usage of torch Tensors and large lists in my efforts to minimize the memory usage of various NER models. That said, it seems unnecessarily expensive to call `len(pickle.dumps(obj))` on all records.

In conclusion, I think this PR's approach won't work well, but I don't have an easy and cheap alternative for you, I'm afraid.

- Tom Aarsen

davidberenstein1957:
That is why I was using the `json` of the records. I assumed this would correspond to the payload being sent. Anyhow, In this case, I did not care about an exact approach,  but one that would equally divide the records based on a comparable way of measuring. Not sure what the overhead is in terms of time though.

davidberenstein1957:
@frascuchon Do you see a quick fix here? Otherwise, we can leave it for now.

frascuchon:
> @frascuchon Do you see a quick fix here? Otherwise, we can leave it for now.

I agree with @davidberenstein1957 that measuring the json representation is a good approach for the solution. Not sure the impact of this extra computation for each record, but we can do some empirical tests.

Regarding the record size, could we use just the length of generated json string and distribute data based on that sort? 

cc @davidberenstein1957 @tomaarsen

tomaarsen:
Considering that we convert to stdlib Python objects via the json call, the `sys.getsizeof` may still work fine, it just needs some quick checks to verify that it does indeed correctly understand that a long list corresponds with a lot of bytes. Otherwise, the string approach may work too.

src/argilla/client/client.py
```diff
@@ -391,6 +390,18 @@ async def log_async(
...
+        # divide records in chunks of chunk size based on their byte size
+        batch_size = int(batch_size / (len(records) / batch_size)) + 1
+        n_batches = math.ceil(len(records) / batch_size)
+        rec_size = [sys.getsizeof(rec.json()) for rec in records]
+        records = [rec for _, rec in sorted(zip(rec_size, records), key=lambda pair: pair[0])]
+        records_new = dict.fromkeys(range(n_batches + 1), [])
+        while records:
+            for i in records_new:
+                records_new[i].append(records.pop(0))
+        records = [item for sublist in records_new.values() for item in sublist]
```
frascuchon: Using the modulo operator (%) we can iterate safety the records list to build the distributed batches dict.

```suggestion
        batch_size = int(batch_size / (len(records) / batch_size)) + 1
        n_batches = math.ceil(len(records) / batch_size)
        
        if n_batches > 1:
            rec_size = [sys.getsizeof(rec.json()) for rec in records]
            records = [rec for _, rec in sorted(zip(rec_size, records), key=lambda pair: pair[0])]
            
            records_new = dict.fromkeys(range(n_batches), [])
            for idx, record in enumerate(records):
                records_new[idx % n_batches].append(record)
            
            records = []
            for batch_list in records_new.values():
                records.extend(batch_list)
```

davidberenstein1957: Hi @frascuchon good suggestion. I reasoned wrongly about using the modulo operator, but gladto see it is working.

PR #2540, opened 2023-03-14
fix: added flag check for re-logging projection block #2534

### LINKED ISSUE #2534: add `projection` flag to records obtained through `fields`-projecten param
Re-uploading these records could potentially overwrite previous info.

on a record level, have a projection flag, which should be checked when logging data.

only providing a warning message

**Additional context**
N.A.

davidberenstein1957:
# Description

avoids allowing users to potentially overwrite their original data after having obtained a projection of their data.

Closes #2534

**Type of change**

**How Has This Been Tested**

See changes.

**Checklist**

N.A.

### DIFF
--- a/src/argilla/client/client.py
+++ b/src/argilla/client/client.py
@@ -50,7 +50,6 @@ from argilla.client.sdk.client import AuthenticatedClient
 from argilla.client.sdk.commons.api import async_bulk
 from argilla.client.sdk.commons.errors import (
     AlreadyExistsApiError,
-    ApiCompatibilityError,
     InputValueError,
     NotFoundApiError,
 )
@@ -69,7 +68,6 @@ from argilla.client.sdk.text_classification.models import (
     LabelingRule,
     LabelingRuleMetricsSummary,
     TextClassificationBulkData,
-    TextClassificationQuery,
 )
 from argilla.client.sdk.text_classification.models import (
     TextClassificationRecord as SdkTextClassificationRecord,
@@ -77,7 +75,6 @@ from argilla.client.sdk.text_classification.models import (
 from argilla.client.sdk.token_classification.models import (
     CreationTokenClassificationRecord,
     TokenClassificationBulkData,
-    TokenClassificationQuery,
 )
 from argilla.client.sdk.token_classification.models import (
     TokenClassificationRecord as SdkTokenClassificationRecord,
@@ -391,6 +388,14 @@ class Argilla:
         else:
             raise InputValueError(f"Unknown record type {record_type}. Available values are {Record.__args__}")
 
+        # TODO: remove after https://github.com/argilla-io/argilla/issues/2535
+        if any([rec.projection for rec in records]):
+            raise InputValueError(
+                "Re-logging field projected records will remove previous info.",
+                "If desidered, set record.projection to None before re-logging.",
+                "If not, look at https://github.com/argilla-io/argilla/issues/2535 for resolving this.",
+            )
+
         processed, failed = 0, 0
         with Progress() as progress_bar:
             task = progress_bar.add_task("Logging...", total=len(records), visible=verbose)
@@ -679,4 +684,7 @@ class Argilla:
             ids=ids,
         )
         records = [sdk_record_class.parse_obj(r).to_client() for r in records]
+        if fields is not None:
+            for rec in records:
+                rec.projection = True
         return dataset_class(records)
--- a/src/argilla/client/models.py
+++ b/src/argilla/client/models.py
@@ -343,7 +343,7 @@ class TokenClassificationRecord(_Validators):
     search_keywords: Optional[List[str]] = None
     _span_utils: SpanUtils = PrivateAttr()
 
-    projection: Optional[bool] = None  # TODO: remove after https://github.com/argilla-io/argilla/issues/2535
+    projection: Optional[bool] = Field(default=None, exclude=True)  # TODO: remove after https://github.com/argilla-io/argilla/issues/2535
 
     def __init__(
         self,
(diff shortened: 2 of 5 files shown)


src/argilla/client/models.py
```diff
@@ -341,6 +343,8 @@ class TokenClassificationRecord(_Validators):
     search_keywords: Optional[List[str]] = None
     _span_utils: SpanUtils = PrivateAttr()
 
+    projection: Optional[bool] = None  # TODO: remove after https://github.com/argilla-io/argilla/issues/2535
```
frascuchon: Some tests are failing since this new field is serialized with the rest of the fields. Since is an internal temporary value depending on how data where loaded, we can exclude it for serialization. See [here](https://docs.pydantic.dev/usage/exporting_models/#model-and-field-level-include-and-exclude)

```suggestion
    projection: Optional[bool] = Field(default=None, exclude=True)  # TODO: remove after https://github.com/argilla-io/argilla/issues/2535
```

davidberenstein1957:
Thanks @frascuchon

src/argilla/client/models.py
```diff
@@ -341,6 +343,10 @@ class TokenClassificationRecord(_Validators):
     search_keywords: Optional[List[str]] = None
     _span_utils: SpanUtils = PrivateAttr()
 
+    projection: Optional[bool] = Field(
+        default=None, exclude=True
+    )  # TODO: remove after https://github.com/argilla-io/argilla/issues/2535
+
```
frascuchon, line 351: This should be applied to the rest of the models, right @davidberenstein1957?

tests/client/test_dataset.py
```diff
@@ -802,6 +804,7 @@ def test_to_from_datasets(self, text2text_records):
             "status",
             "event_timestamp",
             "metrics",
+            "projection",
```
frascuchon, line 807: Do we want this field for exportable datasets? In that case, why we are using the `Field(exclude=True)` set up?

For me, it makes sense to remove this field when exporting to HF datasets

src/argilla/client/models.py
davidberenstein1957: Correct

tests/client/test_dataset.py
davidberenstein1957: I would exclude it by default. Not sure what you think.

davidberenstein1957:
@frascuchon I found this, what was the reasoning behind this?
https://github.com/argilla-io/argilla/blob/a559b6bee951a9cef7bc6d2a1a55ac798f745c64/src/argilla/client/client.py#L668
 
Maybe we want a minimum required set per dataset task? Or a deep-copy due to the mutability of lists.

src/argilla/client/models.py
```diff
@@ -238,6 +238,8 @@ class TextClassificationRecord(_Validators):
     metrics: Optional[Dict[str, Any]] = None
     search_keywords: Optional[List[str]] = None
 
+    projection: Optional[bool] = None  # TODO: remove after https://github.com/argilla-io/argilla/issues/2535
```
frascuchon:
```suggestion
    projection: Optional[bool] = None  Field(
        default=None, exclude=True
    )  # TODO: remove after https://github.com/argilla-io/argilla/issues/2535
```

src/argilla/client/models.py
```diff
@@ -546,6 +552,8 @@ class Text2TextRecord(_Validators):
...
```
frascuchon:
```suggestion
    projection: Optional[bool] = None  Field(
        default=None, exclude=True
    )  # TODO: remove after https://github.com/argilla-io/argilla/issues/2535
```

src/argilla/client/models.py
frascuchon: 👍

frascuchon:
> @frascuchon I found this, what was the reasoning behind this?
> 
> https://github.com/argilla-io/argilla/blob/a559b6bee951a9cef7bc6d2a1a55ac798f745c64/src/argilla/client/client.py#L668
> 
> Maybe we want a minimum required set per dataset task? Or a deep-copy due to the mutability of lists.

Yes. Indeed, here are all the minimum required fields per task but all in one. I've pushed some changes to have a more readable code. Also, I've introduced some tests related to `fields` projection.

tests/client/test_dataset.py
frascuchon: Regarding the flow, maybe it makes sense to have this field in HF structure for those cases where projection was applied. Just to prevent reloading data from a projected HF snapshot.

So, we could keep it, but maybe only when `projection=True`

src/argilla/client/client.py
```diff
@@ -664,12 +677,11 @@ def _load_records_internal(
 
             return dataset_class(results.records)
 
-        if fields:
-            fields.extend(["id", "text", "tokens", "inputs"])
+        fields_set = required_fields.union(set(fields)) if fields else set("*")
```
```suggestion
        projection_fields = required_fields.union(set(fields)) if fields else set("*")
```

src/argilla/client/client.py
```diff
@@ -664,12 +677,11 @@ def _load_records_internal(
...
 
         records = self.datasets.scan(
             name=name,
-            projection=set(fields or "*"),
+            projection=fields_set,
```
```suggestion
            projection=projection_fields,
```

src/argilla/client/client.py
```diff
@@ -635,13 +640,21 @@ def _load_records_internal(
         task = dataset.task
 
         task_config = {
-            TaskType.text_classification: (SdkTextClassificationRecord, DatasetForTextClassification),
-            TaskType.token_classification: (SdkTokenClassificationRecord, DatasetForTokenClassification),
-            TaskType.text2text: (SdkText2TextRecord, DatasetForText2Text),
+            TaskType.text_classification: (
+                SdkTextClassificationRecord,
+                DatasetForTextClassification,
+                {"id", "inputs", "multi_label"},
```
frascuchon: I think the `event_timestamp` field should be returned always. Otherwise, a new value will be created when a record is initialized from a projection, right @davidberenstein1957?

PR #2542, opened 2023-03-14
Explicit setup to add alembic.ini in package distribution

frascuchon:
It's what it is

@jfcalvo I see also that the `script.mako` file is not included by default. Do we need it?


jfcalvo:
> @jfcalvo I see also that the script.mako file is not included by default. Do we need it?

Not needed, that file is the template to generate new alembic migrations so is something that we only need on development.

jfcalvo (approved):

PR #2545, opened 2023-03-15
Fix Pydantic User schema full_name attribute to don't use last_name when is None

jfcalvo:
`last_name` column on `users` database is optional and was wrongly rendered on `User` Pydantic schema for `full_name` dict property.

This PR only returns the user `first_name` when `last_name` is not available (when is `None`).

### DIFF
--- a/src/argilla/server/security/model.py
+++ b/src/argilla/server/security/model.py
@@ -62,7 +62,7 @@ class UserCreate(BaseModel):
 class UserGetter(GetterDict):
     def get(self, key: str, default: Any = None) -> Any:
         if key == "full_name":
-            return " ".join(filter(None, [self._obj.first_name, self._obj.last_name]))
+            return f"{self._obj.first_name} {self._obj.last_name}" if self._obj.last_name else self._obj.first_name
         elif key == "workspaces":
             return [workspace.name for workspace in self._obj.workspaces]
         else:
--- a/src/argilla/server/security/model.py
+++ b/src/argilla/server/security/model.py
@@ -62,7 +62,7 @@ class UserCreate(BaseModel):
 class UserGetter(GetterDict):
     def get(self, key: str, default: Any = None) -> Any:
         if key == "full_name":
-            return f"{self._obj.first_name} {self._obj.last_name}"
+            return " ".join(filter(None, [self._obj.first_name, self._obj.last_name]))
         elif key == "workspaces":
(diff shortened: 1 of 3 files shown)


src/argilla/server/security/model.py
```diff
@@ -62,7 +62,7 @@ class UserCreate(BaseModel):
 class UserGetter(GetterDict):
     def get(self, key: str, default: Any = None) -> Any:
         if key == "full_name":
-            return f"{self._obj.first_name} {self._obj.last_name}"
+            return " ".join(filter(None, [self._obj.first_name, self._obj.last_name]))
```
frascuchon: I see a lot of logic just to prevent a default value. Wdyt @jfcalvo?

```suggestion
            return f"{self._obj.first_name} {self._obj.last_name or ''}".strip()
```

jfcalvo: I think using `strip` is a bit misleading. If we want to really simplify then I would go with something like this:

```suggestion
            return f"{self._obj.first_name} {self._obj.last_name}" if self._obj.last_name else self._obj.first_name
```

tests/server/security/test_model.py
```diff
@@ -11,38 +11,41 @@
...
+
+
+def test_user_full_name(db: Session):
+    user = UserFactory.create(first_name="first-name", last_name="last-name")
+
+    assert User.from_orm(user).full_name == "first-name last-name"
+
+
+def test_user_full_name_without_last_name(db: Session):
+    user = UserFactory.create(first_name="first-name", last_name=None)
```
frascuchon: We could extend this using some extra values for `last_name` matching the tests: "" and "      "

```suggestion
@pytest.mark.parameterize("last_name", [None, "", " ", "     "])
def test_user_full_name_without_last_name(db: Session, last_name):
    user = UserFactory.create(first_name="first-name", last_name=last_name)
```

jfcalvo: We are only covered for `None` value and `""`, when `last_name` is `" "` then `full_name` should be `first-name  ` (with two additional spaces at the end), and that's ok because it's the value that we have on `last_name`.

jfcalvo: We will check `None` and `""`.

jfcalvo: Feedback applied @frascuchon. Please review again.

frascuchon (approved):

PR #2546, opened 2023-03-15
Add first name and last name to user schema

jfcalvo:
**Note:** This PR is based on changes added by https://github.com/argilla-io/argilla/pull/2545 so please review https://github.com/argilla-io/argilla/pull/2545 first.

In this PR I'm adding `first_name` and `last_name` to `User` Pydantic schema. I have not found a proper way to deprecate `full_name` field (I have only found this https://github.com/pydantic/pydantic/issues/2255). @frascuchon if you know of some proper way to deprecate this field and automagically show that on swagger docs please tell me.

### DIFF
--- a/src/argilla/server/security/model.py
+++ b/src/argilla/server/security/model.py
@@ -73,14 +73,14 @@ class User(BaseModel):
     """Base user model"""
 
     id: UUID
+    first_name: str
+    last_name: Optional[str]
+    full_name: Optional[str]
     username: str = Field()
     role: UserRole
-    full_name: Optional[str] = None
-
     superuser: Optional[bool]
     workspaces: Optional[List[str]]
     api_key: str
-
     inserted_at: datetime
     updated_at: datetime
 
--- a/src/argilla/server/security/model.py
+++ b/src/argilla/server/security/model.py
@@ -75,7 +75,7 @@ class User(BaseModel):
     id: UUID
     first_name: str
     last_name: Optional[str]
-    full_name: Optional[str]
+    full_name: Optional[str] = Field(description="Deprecated. Use `first_name` and `last_name` instead")
     username: str = Field()
     role: UserRole
     superuser: Optional[bool]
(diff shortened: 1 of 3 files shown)


src/argilla/server/security/model.py
```diff
@@ -73,14 +73,14 @@ class User(BaseModel):
     """Base user model"""
 
     id: UUID
+    first_name: str
+    last_name: Optional[str]
+    full_name: Optional[str]
```
frascuchon: You could add a description with a deprecation message. This will be shown in swagger documentation directly.
```suggestion
    full_name: Optional[str] = Field(None, description="Deprecated: You should user `first_name` and `last_name` fields instead,....")
```

src/argilla/server/security/model.py
```diff
@@ -62,7 +62,7 @@ class UserCreate(BaseModel):
 class UserGetter(GetterDict):
     def get(self, key: str, default: Any = None) -> Any:
         if key == "full_name":
-            return f"{self._obj.first_name} {self._obj.last_name}"
+            return " ".join(filter(None, [self._obj.first_name, self._obj.last_name]))
```
frascuchon: I feel too much logic here just to prevent default value. Wdyt @jfcalvo ?

```suggestion
            return f"{self._obj.first_name} {self._obj.last_name or ''}"
```

src/argilla/server/security/model.py
jfcalvo: Cool!

src/argilla/server/security/model.py
jfcalvo: I will apply this feedback in the base PR.

src/argilla/server/security/model.py
jfcalvo: Done.

src/argilla/server/security/model.py
frascuchon: 👍

frascuchon (approved):

PR #2547, opened 2023-03-15
Review annotator policies for dataset settings

frascuchon:
For an annotator, dataset settings are visible in read-only mode.

This PR adds task dataset settings-related policies and uses them in dataset settings-related actions.

Also, new tests are added to test this functionality.

### DIFF
--- a/src/argilla/server/policies.py
+++ b/src/argilla/server/policies.py
@@ -103,7 +103,7 @@ class DatasetPolicy:
         return lambda actor: actor.is_admin or is_get_allowed(actor) and cls.create(actor)
 
 
-class TaskDatasetSettingsPolicy:
+class DatasetSettingsPolicy:
     @classmethod
     def list(cls, dataset: Dataset) -> PolicyAction:
         return DatasetPolicy.get(dataset)
--- a/src/argilla/server/policies.py
+++ b/src/argilla/server/policies.py
@@ -103,6 +103,20 @@ class DatasetPolicy:
         return lambda actor: actor.is_admin or is_get_allowed(actor) and cls.create(actor)
 
 
+class TaskDatasetSettingsPolicy:
+    @classmethod
+    def list(cls, dataset: Dataset) -> PolicyAction:
+        return DatasetPolicy.get(dataset)
+
+    @classmethod
+    def save(cls, dataset: Dataset) -> PolicyAction:
+        return lambda actor: actor.is_admin
+
+    @classmethod
+    def delete(cls, dataset: Dataset) -> PolicyAction:
+        return lambda actor: actor.is_admin
+
+
 def authorize(actor: User, policy_action: PolicyAction) -> None:
     if not is_authorized(actor, policy_action):
         raise ForbiddenOperationError()
--- a/src/argilla/server/services/datasets.py
+++ b/src/argilla/server/services/datasets.py
@@ -32,7 +32,7 @@ from argilla.server.errors import (
 from argilla.server.models import User, Workspace
 from argilla.server.policies import (
     DatasetPolicy,
-    TaskDatasetSettingsPolicy,
+    DatasetSettingsPolicy,
     is_authorized,
 )
 from argilla.server.schemas.datasets import CreateDatasetRequest, Dataset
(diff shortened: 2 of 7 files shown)


src/argilla/server/policies.py
```diff
@@ -103,6 +103,20 @@ def copy(cls, dataset: Dataset) -> PolicyAction:
         return lambda actor: actor.is_admin or is_get_allowed(actor) and cls.create(actor)
 
 
+class TaskDatasetSettingsPolicy:
```
jfcalvo: Why `TaskDatasetSettingsPolicy` instead of `DatasetSettingsPolicy`?

src/argilla/server/services/datasets.py
```diff
@@ -226,6 +226,11 @@ async def get_settings(
         settings = self.__dao__.load_settings(dataset=dataset, as_class=class_type)
         if not settings:
             raise EntityNotFoundError(name=dataset.name, type=class_type)
+
+        if not is_authorized(user, TaskDatasetSettingsPolicy.list(dataset)):
+            raise ForbiddenOperationError(
+                "You don't have the necessary permissions to list settings for this dataset. "
```
jfcalvo:
```suggestion
                "You don't have the necessary permissions to list settings for this dataset."
```

src/argilla/server/services/datasets.py
```diff
@@ -234,8 +239,18 @@ def raw_dataset_update(self, dataset):
     async def save_settings(
         self, user: User, dataset: ServiceDataset, settings: ServiceDatasetSettings
     ) -> ServiceDatasetSettings:
+        if not is_authorized(user, TaskDatasetSettingsPolicy.save(dataset)):
+            raise ForbiddenOperationError(
+                "You don't have the necessary permissions to save settings for this dataset. "
```
jfcalvo:
```suggestion
                "You don't have the necessary permissions to save settings for this dataset."
```

src/argilla/server/services/datasets.py
```diff
@@ -234,8 +239,18 @@ def raw_dataset_update(self, dataset):
...
+            )
+
         self.__dao__.save_settings(dataset=dataset, settings=settings)
         return settings
 
     async def delete_settings(self, user: User, dataset: ServiceDataset) -> None:
+        if not is_authorized(user, TaskDatasetSettingsPolicy.delete(dataset)):
+            raise ForbiddenOperationError(
+                "You don't have the necessary permissions to delete settings for this dataset. "
```
jfcalvo:
```suggestion
                "You don't have the necessary permissions to delete settings for this dataset."
```

tests/server/text_classification/test_api_settings.py
```diff
@@ -126,5 +131,97 @@ def log_some_data(mocked_client, name):
...
+    annotator = AnnotatorFactory.create(workspaces=[WorkspaceFactory.build(name=workspace_name)])
+
+    test_client.delete(f"/api/datasets/{dataset_name}?workspace={workspace_name}", headers=argilla_auth_header)
+
+    response = test_client.post(
+        "/api/datasets",
+        json={"name": dataset_name, "task": task, "workspace": workspace_name},
+        headers=argilla_auth_header,
+    )
+    assert response.status_code == 200
```
jfcalvo, line 149: Not sure if there is an easy solution for this but you are testing not only the save settings endpoint but the dataset creation as well.

The ideal scenario is to only test the specific endpoint and create the dataset manually.

The same happen with the rest of tests.

jfcalvo (changes requested):

tests/server/text_classification/test_api_settings.py
frascuchon: Please, fill free to add the test setup using the elasticsearch client to create the document directly.

These tests are testing user flows more than simple endpoint actions. It's faster and gives us the confidence to check the flow interactions

src/argilla/server/policies.py
frascuchon: At the moment, `DatasetSettingsPolicy` name fits better.  I will change it

tests/server/text_classification/test_api_settings.py
jfcalvo: > These tests are testing user flows more than simple endpoint actions. It's faster and gives us the confidence to check the flow interactions

Sure and that's fine, but in that case I would move those tests to some special folder like `test_api_flows` or something similar (and left regular files for unit testing). We can talk about this if you want.

jfcalvo (approved):

PR #2549, opened 2023-03-15
Change policies for dataset update

frascuchon:
This PR changes the update dataset policy to accept updating a dataset (tags and metadata) for all users linked to the dataset workspace.

This may change in the future.

### DIFF
--- a/src/argilla/server/services/datasets.py
+++ b/src/argilla/server/services/datasets.py
@@ -120,9 +120,6 @@ class DatasetsService:
         tags: Dict[str, str],
         metadata: Dict[str, Any],
     ) -> Dataset:
-        if not (tags or metadata):
-            return dataset
-
         found = self.find_by_name(user=user, name=dataset.name, task=dataset.task, workspace=dataset.workspace)
 
         if not is_authorized(user, DatasetPolicy.update(found)):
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -59,6 +59,24 @@ def admin(db):
     db.add(user)
     db.commit()
     db.refresh(user)
+
+    return user
+
+
+@pytest.fixture(scope="function")
+def annotator(db):
```
jfcalvo, line 68: We already have `AnnotatorFactory`, instead of using this fixture you can do:

```python
annotator = AnnotatorFactory.create()
```

And use `annotator` where necessary.

src/argilla/server/services/datasets.py
```diff
@@ -120,6 +120,9 @@ def update(
         tags: Dict[str, str],
         metadata: Dict[str, Any],
     ) -> Dataset:
+        if not (tags or metadata):
+            return dataset
```
jfcalvo: I don't understand this part. 😄

jfcalvo: We discussed it and we believe this can be removed. @frascuchon

jfcalvo (approved):

PR #2550, opened 2023-03-15
Map SQLAlchemy IntegrityError as Argilla EntityAlreadyExistsError

jfcalvo:
This is a draft pull request with a possible solution to the handlers that need to manage situations where is possible to have an integrity error coming from database and was returning a complete error trace with too many internal details.

The suggested solution is to capture `sqlalchemy.exc.IntegrityError` on the place where this error can happen and raise a new Argilla internal `EntityAlreadyExistsError`.

There is of course alternative solutions to this problem like:
* Change `GenericServerError` to not include complex error messages. (This should include some improvements over the logger system to allow users capture the error messages, even if those are not included in the response).
* Refactor our API error system, right now the responses are too generic and doesn't match sometimes with the required responses, and improve our logging system (including request ids, etc.).

### DIFF
--- a/src/argilla/server/apis/v0/handlers/users.py
+++ b/src/argilla/server/apis/v0/handlers/users.py
@@ -16,6 +16,7 @@
 from typing import List
 from uuid import UUID
 
+import sqlalchemy.exc
 from fastapi import APIRouter, Depends, Request, Security
 from pydantic import parse_obj_as
 from sqlalchemy.orm import Session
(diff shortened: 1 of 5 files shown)


jfcalvo:
Instead of capturing the exception and raising an Argilla error we will check that the entity exists before hand.

jfcalvo:
Instead of capturing the exception and raising an Argilla error we will check that the entity exists before hand.

PR #2560, opened 2023-03-16
Add argilla_data volume to docker compose

jfcalvo:
In this PR we are defining an argilla data volume to be used on our `docker-compose.yaml`.

Things to have into account:
- <del>I'm defining the volume as `data` because docker adds the `argilla_` prefix once it's created being `argilla_data` at the end.</del>
- Environment variables defined on the docker compose cannot being used on the right side of the volumes section. This can potentially provoke errors if the users sets the `ARGILLA_HOME_PATH` value.

I have successfully tested to set a custom volume pointing to a host folder with:

```dockerfile
argilla:
    volumes:
      - ${HOME}/.argilla:/var/lib/argilla
     
networks:
  argilla:
    driver: bridge
volumes:
  # argilladata:
  elasticdata:
```

And the database file is correctly created on my home folder. Restarting the docker compose is working too.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -20,7 +20,9 @@ services:
     networks:
       - argilla
     volumes:
-      - data:/var/lib/argilla
+      # ARGILLA_HOME_PATH is used to define where Argilla will save it's application data.
+      # If you change ARGILLA_HOME_PATH value please copy that same value to argilladata volume too.
+      - argilladata:/var/lib/argilla
   elasticsearch:
     image: docker.elastic.co/elasticsearch/elasticsearch:8.5.3
     environment:
@@ -55,5 +57,5 @@ networks:
   argilla:
     driver: bridge
 volumes:
-  data:
+  argilladata:
   elasticdata:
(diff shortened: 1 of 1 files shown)


docker-compose.yaml
```diff
@@ -18,7 +19,8 @@ services:
       #- ${PWD}/.users.yaml:/config/.users.yaml
     networks:
       - argilla
-
+    volumes:
+      - data:/var/lib/argilla
```
frascuchon: The prefix given by docker-compose depends on where docker-compose.yaml is placed. Normally, the parent folder name will be used. If we download this file and place it in folder `local` the volume name will be `local_data`. I prefer adding the service name to the created volume: `argilladata` is fine

frascuchon: Also, we can add a comment noticing the volume could be modified when the `ARGILLA_HOME_PATH` is modified.

frascuchon: Also, we can add a comment noticing the volume could be modified when the `ARGILLA_HOME_PATH` is modified.
```suggestion
     # If the ARGILLA_HOME_PATH is customized, this value should be change....
      - data:/var/lib/argilla
```

jfcalvo: Done

jfcalvo: Done.

frascuchon (approved):

Issue #2340, opened 2023-02-14
create/align database ID creation with Python client to skip processed records during failing log


davidberenstein1957:
**Is your feature request related to a problem? Please describe.**
I was just passing 35K records, and it failed at record X.

**Describe the solution you'd like**
Ideally, I would like to have records with IDs and add a rg.log `skip_if_present` flag or something to avoid duplicating records and processing time.

**Describe alternatives you've considered**
N.A.

**Additional context**
Is partially related to https://github.com/argilla-io/argilla/issues/2272

davidberenstein1957:
```
import uuid
uuid.uuid4()
```

PR #2576, opened 2023-03-20
feat: Allow partial record update

### LINKED ISSUE #2535: allow for updating only certain `fields` of a record ID via `rg.log` projection

for now we can do this only for the metadata info and the status, but we could extend it to other fields
- [] add batch support
- [] add other field support
- [] remove logic surround [projection flag](https://github.com/argilla-io/argilla/issues/2534)

**Additional context**
/docs#/records/partial_update_dataset_record_api_datasets__name__records__record_id__patch

frascuchon:
# Description

Changes in this PR affect directly the way the clients log data. If the record info contains the `id`, the record will be partially updated. This means that logging data can be done by passing a subset of per-record attributes, instead of the whole record. 

Closes #2535

Refs #2534

**Type of change**

**How Has This Been Tested**

Added tests covering this new feature

**Checklist**


davidberenstein1957:
@frascuchon I agree, #2534 can be ignored when setting this up correctly. However, in order to update records, I expect to have some record-specific state, as to which fields ought to be upserted. Similar to the projection state used in #2534. 

So, perhaps we can copy some of the logic used in #2534 and add an 'as_upsert' and a 'updated_fields'-attributes to the pydantic models. The `as_upsert` should default to `False` when creating a record, but should be set to `True` when a `record.id` is passed (might be an issue when someone uses manual record ids) or when data is loaded with `rg.load` (more flexible but requires a bit more implementation efforts from our side). Similarly, the `updated_fields` should default to `None` or an `empty list []`, and attribute names should be added when records are updated during item assignment. 

- When records have `as_upsert==True` and `updated_fields==True`, we can then log the `id` and `**updated_fields` as an upsert. 
- When records have `as_upsert==True` and `updated_fields==False`, we will not log anything.
- When`as_upsert==False`, we will just do normal logging.
- In case of failed upsert logs (due to IDs already present), we can reference that in a logging message and mention that someone might want to overrule this behaviour by setting `rg.log(upsert=True)`.

frascuchon:
> @frascuchon I agree, #2534 can be ignored when setting this up correctly. However, in order to update records, I expect to have some record-specific state, as to which fields ought to be upserted. Similar to the projection state used in #2534.
> 
> So, perhaps we can copy some of the logic used in #2534 and add an 'as_upsert' and a 'updated_fields'-attributes to the pydantic models. The `as_upsert` should default to `False` when creating a record, but should be set to `True` when a `record.id` is passed (might be an issue when someone uses manual record ids) or when data is loaded with `rg.load` (more flexible but requires a bit more implementation efforts from our side). Similarly, the `updated_fields` should default to `None` or an `empty list []`, and attribute names should be added when records are updated during item assignment.
> 
> * When records have `as_upsert==True` and `updated_fields==True`, we can then log the `id` and `**updated_fields` as an upsert.
> * When records have `as_upsert==True` and `updated_fields==False`, we will not log anything.
> * When`as_upsert==False`, we will just do normal logging.
> * In case of failed upsert logs (due to IDs already present), we can reference that in a logging message and mention that someone might want to overrule this behaviour by setting `rg.log(upsert=True)`.

Got it. This PR works only at the API backend level by defining a new endpoint ONLY for record updates. If provided ids were not found, the batch update won't do anything. The existing ones will work as always, by overriding the whole document.

So, client integration work should be done to fully complete this feat.  I see your proposal, my questions are:

- How to tackle default field initialization and validations with partial load?
- Do we should accept updates without a rg.load? I can imagine users creating records from scratch and setting/updating fields -> updated_fields will be automatically updated when the user set fields
- If we can upsert (update or create if does not exist) we should ALWAYS provide a minimal set of fields to prevent error or default behavior when updating

With this new feat. my feeling is that we should separate record creation and record validation. This can help to decide what to do depending on if we want to create new data or just partially update an existing record.

davidberenstein1957:
> So, client integration work should be done to fully complete this feat. I see your proposal, my questions are:

I agree.

> -     How to tackle default field initialization and validations with partial load?
I would say that a partial load only requires validation during item assignments/updates, which we are already doing. Correct? The default field will always be there, and in that case, we would not add them to the `updated_fields` yet.
> -     Do we should accept updates without a rg.load? I can imagine users creating records from scratch and setting/updating fields -> updated_fields will be automatically updated when the user set fields.
Yes, I see them allow to update record without `rg.load` that is why I proposed using `as_upsert`(stating if data is retrieved via load` and should this be considered as an upsert and `updated_fields` to determined the fields that need to be send, and have been updated. 
> -     If we can upsert (update or create if does not exist) we should ALWAYS provide a minimal set of fields to prevent error or default behaviour when updating
I agree with the minimum field, but during upsert, wouldn't it be possible to send {"id": x, "vector": y} and only send this info.
> With this new feat. my feeling is that we should separate record creation and record validation. This can help to decide what to do depending on if we want to create new data or just partially update an existing record.

Potentially we can but I think from a client-side, it can be inconvenient to decouple record validation from creation, because you will only find out your entire analysis is wrong, only after having finished it when you actually try logging your data.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- bulk endpoints will upsert data when record `id` is present. See [#2576](https://github.com/argilla-io/argilla/pull/2576)
```
```suggestion
- bulk endpoints will upsert data when record `id` is present. Closes [#2535](https://github.com/argilla-io/argilla/issues/2535)
```

src/argilla/server/daos/backend/client_adapters/opensearch.py
```diff
@@ -554,23 +550,15 @@ def index_documents(
...
 
-        data = {
-            "_op_type": "index",
-            "_index": index,
-            "_routing": None,  # TODO(@frascuchon): Use a sharding routing
-            **doc,
-        }
+        data = (
+            {"_index": index, "_op_type": "index", **doc}
+            if doc_id is None
```
davidberenstein1957, line 558: @frascuchon I believe we are currently creating doc ID Python client side, so this might have unexpected behaviour, i.e., the doc_id will never be None. https://github.com/argilla-io/argilla/issues/2340

frascuchon: That means the upsert (update or insert) will be always applied. If the record already exists, will be updated with new info. If all the record info is sent, this approach still works for those cases

davidberenstein1957 (approved):

PR #2577, opened 2023-03-20
refactor: improve load with fields projection

frascuchon:
# Description

This PR adds changes suggested in https://github.com/argilla-io/argilla/pull/2540#issuecomment-1470484071.  Also, includes `event_timestamp`  and `status` as minimal required fields.

Refs: #2576  and #2540 

**Type of change**

**How Has This Been Tested**

New tests are included 

**Checklist**

### DIFF
--- a/src/argilla/client/client.py
+++ b/src/argilla/client/client.py
@@ -636,13 +636,21 @@ class Argilla:
         task = dataset.task
 
         task_config = {
-            TaskType.text_classification: (SdkTextClassificationRecord, DatasetForTextClassification),
-            TaskType.token_classification: (SdkTokenClassificationRecord, DatasetForTokenClassification),
-            TaskType.text2text: (SdkText2TextRecord, DatasetForText2Text),
+            TaskType.text_classification: (
+                SdkTextClassificationRecord,
+                DatasetForTextClassification,
+                {"id", "inputs", "multi_label"},
+            ),
+            TaskType.token_classification: (
+                SdkTokenClassificationRecord,
+                DatasetForTokenClassification,
+                {"id", "text", "tokens"},
+            ),
+            TaskType.text2text: (SdkText2TextRecord, DatasetForText2Text, {"id", "text"}),
         }
 
         try:
-            sdk_record_class, dataset_class = task_config[task]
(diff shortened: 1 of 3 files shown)


src/argilla/client/client.py
```diff
@@ -636,13 +636,21 @@ def _load_records_internal(
         task = dataset.task
 
         task_config = {
-            TaskType.text_classification: (SdkTextClassificationRecord, DatasetForTextClassification),
-            TaskType.token_classification: (SdkTokenClassificationRecord, DatasetForTokenClassification),
-            TaskType.text2text: (SdkText2TextRecord, DatasetForText2Text),
+            TaskType.text_classification: (
+                SdkTextClassificationRecord,
+                DatasetForTextClassification,
+                {"id", "event_timestamp", "status", "inputs", "multi_label"},
```
davidberenstein1957, line 642: should this not also include the "text" parm? or will this be over-ruled by adding "inputs"?

davidberenstein1957:
@frascuchon Intuitively, I agree with having `event_timestamp` and `status` as minimal requirements too, but given the aim of de-coupling of records from their corresponding predictions, maybe `status` ought to be removed? Or will it then always default to `Default`?

src/argilla/client/client.py
frascuchon: For text classification, the `text` param is computed from inputs. Indeed we have a gap between fields for client records and projected `fields` for searches. But this should be tackled in a separate PR.

frascuchon:
> Intuitively, I agree with having `event_timestamp` and `status` as minimal requirements too, but given the aim of de-coupling of records from their corresponding predictions, maybe `status` ought to be removed? Or will it then always default to `Default`?

Since we have default initializers when creating client records providing some default values, if the status is not provided, the default value will be provided. It's weird to have records with a different status value than the store one depending on the field's projection.

Ideally, the load with projection should return a raw bunch of data, instead of a set of records, but the problem here is how to map between the API record model and the client data model for those cases. For example, the  `annotation` and `annotation_agent` are computed fields from the `annotation` field fetched from API.

frascuchon:
Closing in favour of https://github.com/argilla-io/argilla/pull/2668

PR #2579, opened 2023-03-21
Use admin workspace by default on Quickstart Docker image

jfcalvo:
On our Q&A session we found that some tutorials were failing due to missing `admin` workspace.

Argilla tutorials are using `admin` user but `admin` workspace was not created on the Quicktstart Docker image making tutorials to not work. This PR sets `ARGILLA_WORKSPACE` to use `ADMIN_USERNAME` by default so `admin` workspace is created.

<del>Once I have checked on HF that everything is working I will change this PR from draft to regular.</del> The [tutorial](https://docs.argilla.io/en/latest/tutorials/notebooks/training-textclassification-transformers-pretrained.html) that was failing because the `admin` workspace was missing is now working fine.

I have found others tutorials like [this one](https://docs.argilla.io/en/latest/tutorials/notebooks/labelling-textclassification-setfit-zeroshot.html) failing too but for different reasons. In this case the tutorial is using an old version of `argilla` Python package and is not sending the workspace. I have created a new issue to address these problems with our tutorials: https://github.com/argilla-io/argilla/issues/2583


frascuchon (approved):

frascuchon:
Maybe we can add a comment in the quickstart readme regarding the default how the default workspace will be created?

https://github.com/argilla-io/argilla/blob/create-admin-workspace-on-quickstart-docker/quickstart.README.md#environment-variables

jfcalvo:
> Maybe we can add a comment in the quickstart readme regarding the default how the default workspace will be created?
> 
> https://github.com/argilla-io/argilla/blob/create-admin-workspace-on-quickstart-docker/quickstart.README.md#environment-variables

Yep, let me add that.

frascuchon (approved):

PR #2601, opened 2023-03-24
Add new ADMIN_ENABLED environment variable to release.Dockerfile

jfcalvo:
This PR adds a new `ADMIN_ENABLED` environment variable with a default value of `true`. This environment variable is checked to create or not the default admin user for Docker image generated by `release.Dockerfile`.

We will set this environment variable to `false` later for `PRE` and `PRO` Docker compose file so the default admin user is not created on these environments.

### DIFF
--- a/release.Dockerfile
+++ b/release.Dockerfile
@@ -2,6 +2,7 @@ FROM python:3.9-slim
 
 # Environment Variables
 ENV ARGILLA_HOME_PATH=/var/lib/argilla
+ENV ADMIN_ENABLED=true
 ENV ADMIN_PASSWORD=1234
 ENV ADMIN_API_KEY=argilla.apikey
 ENV USERS_DB=/config/.users.yml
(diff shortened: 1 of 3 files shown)


release.Dockerfile
```diff
@@ -2,6 +2,7 @@ FROM python:3.9-slim
 
 # Environment Variables
 ENV ARGILLA_HOME_PATH=/var/lib/argilla
+ENV ADMIN_ENABLED=true
```
frascuchon: Since this variable is related to the default user creation, maybe we can rename it to `DEFAULT_USER_ENABLED`?
```suggestion
ENV DEFAULT_USER_ENABLED=true
```

scripts/start_argilla_server.sh
```diff
@@ -5,7 +5,9 @@ set -e
 python -m argilla.tasks.database.migrate
 
 # Create default user
-python -m argilla.tasks.users.create_default --password $ADMIN_PASSWORD --api-key $ADMIN_API_KEY
+if [ "$ADMIN_ENABLED" = "true" ]; then
```
frascuchon:
```suggestion
if [ "$DEFAULT_USER_ENABLED" = "true" ]; then
```

frascuchon (approved):

PR #2574, opened 2023-03-20
Feat/2504 add an argillatraining module to the core library with training workflows

davidberenstein1957:
# Description

Hi, this PR contains some preliminary work on the creating of an `Argilla.training` module, which should simplify training workflows for training using Argilla datasets for several common tasks using several common libraries. Initially, it only contains a `ArgillaSetFitTrainer` and an `ArgillaBaseTrainer`.

Closes #2504 

**Type of change**

(Please delete options that are not relevant. Remember to title the PR according to the type of change)

- [X] New feature (non-breaking change which adds functionality)

**How Has This Been Tested**

Not yet implemented tests

**Checklist**

- [ ] I have merged the original branch into my forked branch
- [ ] I added relevant documentation
- [ ] follows the style guidelines of this project
- [ ] I did a self-review of my code
- [ ] I made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added relevant notes to the CHANGELOG.md file (See https://keepachangelog.com/)

PR #2610, opened 2023-03-28
fix: removed "context" reference from datasets prepare_for_training #…

### LINKED ISSUE #2606: Bug in `DatasetForTextClassification._prepare_for_training_with_transformers` when `multi_label=True`
**Describe the bug**

When calling `.prepare_for_training()` with the default arguments on a `DatasetForTextClassification` dataset with `multi_label=True` records, the `_prepare_for_training_with_transformers` function fails. This is due to the recent addition of the `context` key when building the 🤗`Dataset`.

Then, IMO the following line should either be removed or made optional just in case the `context` key is available.

https://github.com/argilla-io/argilla/blame/ba4ae63dc43ae1d36aa0efcf03896bcf5d9206ee/src/argilla/client/datasets.py#L764.

So on, the bug was introduced in 1.5.0, as in 1.4.0 it works fine and the CI seems to be passing.

**To Reproduce**

Run the following script to reproduce:

```python
import argilla as rg

rb_dataset = rg.DatasetForTextClassification([
    rg.TextClassificationRecord(
        text="This is a sample text",
        annotation="pos",
        multi_label=True,
    )
])
rb_dataset.prepare_for_training()
```

**Expected behavior**

The `context` key shouldn't be mandatory, so as to ensure consistency with the previous datasets. So on, the code above should work and the CI should pass.

 - OS [e.g. iOS]: macOS
 - Argilla Version [e.g. 1.0.0]: 1.5.0

davidberenstein1957:
# Description
Fixed missing context key.

Closes #2606 

**Type of change**

**How Has This Been Tested**

N.A.
**Checklist**

N.A.

### DIFF
--- a/src/argilla/client/datasets.py
+++ b/src/argilla/client/datasets.py
@@ -761,7 +761,6 @@ class DatasetForTextClassification(DatasetBase):
                 {
                     "id": ds["id"],
                     "text": ds["text"],
-                    "context": ds_dict["context"],
                     "label": labels,
                     "binarized_label": binarized_labels,
                 },


alvarobartt:
Hi @davidberenstein1957! I'd say you also need to remove the `context` references in the unit tests at https://github.com/argilla-io/argilla/blob/29972416cf83869fb664d78f75e150c99f45ed51/tests/client/test_dataset.py#L346-L355 👍🏻

src/argilla/client/datasets.py
```diff
@@ -761,7 +761,6 @@ def _prepare_for_training_with_transformers(
                 {
                     "id": ds["id"],
                     "text": ds["text"],
-                    "context": ds_dict["context"],
```
alvarobartt, line 764: Hi @davidberenstein1957! I'd say you also need to remove the `context` references in the unit tests at https://github.com/argilla-io/argilla/blob/29972416cf83869fb664d78f75e150c99f45ed51/tests/client/test_dataset.py#L346-L355 👍🏻

alvarobartt: As that's just failing for the `multilabel_textclassification_records` when testing the `DatasetForTextClassification`

frascuchon, line 764: Thanks @alvarobartt 

Also, something to add. For single-label records, we're adding all defined inputs in records instead of only the `text` one See https://github.com/argilla-io/argilla/blob/29972416cf83869fb664d78f75e150c99f45ed51/src/argilla/client/datasets.py#L744-L748

Not sure what should be the better approach, but I think we should align both cases

frascuchon (changes requested):
See comments

davidberenstein1957:
will be included here https://github.com/argilla-io/argilla/pull/2574

PR #2617, opened 2023-03-29
Add datasets database table and Dataset SQLAlchemy model

jfcalvo:
# Description

This PR include early changes to add the new `datasets` database table and the `Dataset` SQLAlchemy model. I have additionally added also a factory for dataset so we can create tests later.

There are some points to review like:

* Are dataset names unique?.
* Are dataset guidelines optional?.
* Dataset `status` column could be a boolean column with the name `draft` in the case that we are sure that we only want to have draft/not-draft datasets (of course `status` is more flexible if we want to have different dataset statuses in the future).

### DIFF
--- a/src/argilla/server/alembic/versions/b9099dc08489_create_datasets_table.py
+++ b/src/argilla/server/alembic/versions/b9099dc08489_create_datasets_table.py
@@ -33,10 +33,14 @@ def upgrade() -> None:
     op.create_table(
         "datasets",
         sa.Column("id", sa.Uuid, primary_key=True),
-        sa.Column("name", sa.String, nullable=False, index=True),  # TODO: Should the name of a dataset be unique?.
-        sa.Column("guidelines", sa.Text),  # TODO: I understand the guidelines are optional.
+        sa.Column("name", sa.String, nullable=False, index=True),
+        sa.Column("guidelines", sa.Text),
+        sa.Column(
+            "workspace_id", sa.Uuid, sa.ForeignKey("workspaces.id", ondelete="CASCADE"), nullable=False, index=True
+        ),
         sa.Column("inserted_at", sa.DateTime, nullable=False),
         sa.Column("updated_at", sa.DateTime, nullable=False),
+        sa.UniqueConstraint("name", "workspace_id", name="dataset_name_workspace_id_uq"),
     )
 
 
--- a/src/argilla/server/alembic/versions/b9099dc08489_create_datasets_table.py
+++ b/src/argilla/server/alembic/versions/b9099dc08489_create_datasets_table.py
@@ -35,7 +35,6 @@ def upgrade() -> None:
         sa.Column("id", sa.Uuid, primary_key=True),
         sa.Column("name", sa.String, nullable=False, index=True),  # TODO: Should the name of a dataset be unique?.
         sa.Column("guidelines", sa.Text),  # TODO: I understand the guidelines are optional.
-        sa.Column("status", sa.String, nullable=False, index=True),
         sa.Column("inserted_at", sa.DateTime, nullable=False),
         sa.Column("updated_at", sa.DateTime, nullable=False),
     )
--- a/src/argilla/server/alembic/versions/b9099dc08489_create_datasets_table.py
+++ b/src/argilla/server/alembic/versions/b9099dc08489_create_datasets_table.py
@@ -1,0 +1,45 @@
+#  Copyright 2021-present, the Recognai S.L. team.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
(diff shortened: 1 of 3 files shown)


src/argilla/server/alembic/versions/b9099dc08489_create_datasets_table.py
```diff
@@ -0,0 +1,45 @@
...
+down_revision = "1769ee58fbb4"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "datasets",
+        sa.Column("id", sa.Uuid, primary_key=True),
+        sa.Column("name", sa.String, nullable=False, index=True),  # TODO: Should the name of a dataset be unique?.
```
frascuchon: Since dataset id will be generated without using the `name`, and record indexes in elasticsearch are using the dataset.id for the index naming, we can allow modifying the dataset name.

src/argilla/server/alembic/versions/b9099dc08489_create_datasets_table.py
```diff
@@ -0,0 +1,45 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "b9099dc08489"
+down_revision = "1769ee58fbb4"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.create_table(
```
frascuchon, line 33: The workspace should be related to the dataset. Maybe in the future, we can have the same dataset linked to different workspaces, so a new table can be used instead, but I think the simplest way is using a foreign key here.

src/argilla/server/alembic/versions/b9099dc08489_create_datasets_table.py
```diff
@@ -0,0 +1,45 @@
...
+        "datasets",
+        sa.Column("id", sa.Uuid, primary_key=True),
+        sa.Column("name", sa.String, nullable=False, index=True),  # TODO: Should the name of a dataset be unique?.
+        sa.Column("guidelines", sa.Text),  # TODO: I understand the guidelines are optional.
+        sa.Column("status", sa.String, nullable=False, index=True),
```
frascuchon: Regarding the iteration workflow, I think we can skip this field for now. We can consider adding a status to datasets later.

src/argilla/server/models.py
```diff
@@ -34,11 +34,31 @@ def default_inserted_at(context):
...
     admin = "admin"
     annotator = "annotator"
 
 
+class Dataset(Base):
+    __tablename__ = "datasets"
+
+    id: Mapped[UUID] = mapped_column(primary_key=True, default=uuid4)
+    name: Mapped[str]
+    guidelines: Mapped[Optional[str]]
```
frascuchon, line 47: Yes! The guidelines can be optional

frascuchon (changes requested):
> Are dataset names unique?.

We can allow editing dataset names.

> Are dataset guidelines optional?.

Yes

> Dataset status column could be a boolean column with the name draft in the case that we are sure that we only want to have draft/not-draft datasets (of course status is more flexible if we want to have different dataset statuses in the future).

Maybe we can skip for now this field, and think about it once the workflow is clearer for us.

Also, I've commented about the missing workspace_id field. Please, can you include it?

src/argilla/server/alembic/versions/b9099dc08489_create_datasets_table.py
jfcalvo: Ok, so a dataset belongs to a workspace.

src/argilla/server/alembic/versions/b9099dc08489_create_datasets_table.py
jfcalvo: Ok, we will allow to modify it but do we have any reason to make it unique?

src/argilla/server/alembic/versions/b9099dc08489_create_datasets_table.py
jfcalvo: Sure, let's remove it.

src/argilla/server/alembic/versions/b9099dc08489_create_datasets_table.py
jfcalvo: Done.

src/argilla/server/alembic/versions/b9099dc08489_create_datasets_table.py
jfcalvo: Added unique constraint between `name` and `workspace_id`.

src/argilla/server/alembic/versions/b9099dc08489_create_datasets_table.py
jfcalvo: Done.

frascuchon (approved):

PR #2624, opened 2023-03-30
Add new API v1 datasets list endpoint

jfcalvo:
# Description

This PR adds the first v1 endpoint for list datasets.

Some changes that we need to discuss:
- I have added `DatasetPolicyV1` class into `policies.py`. We can think into alternatives to split the policies into different versions if required.
- I have not used `response_model_exclude_none=True` into the new endpoint. I think it's not a good idea to remove attributes becase they are empty and I have taken the chance of not using it with this new v1.
- I have changed how database sessions are managed with the factories. Now it's possible to work with relationships so for example now the workspace is automatically created when a dataset factory is used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- `/api/v1/datasets` new endpoint to list datasets ([#2615]).
+
+[#2615]: https://github.com/argilla-io/argilla/issues/2615
+
+## [Unreleased]
+
+### Added
+
 - `ARGILLA_HOME_PATH` new environment variable ([#2564]).
 - `ARGILLA_DATABASE_URL` new environment variable ([#2564]).
 - Basic support for user roles with `admin` and `annotator` ([#2564]).
@@ -37,7 +45,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Copying datasets between workspaces with proper owner/workspace info. Closes [#2562](https://github.com/argilla-io/argilla/issues/2562)
 - Using elasticsearch config to request backend version. Closes [#2311](https://github.com/argilla-io/argilla/issues/2311)
 
-
 ### Removed
 
 - `email` user field ([#2564]).
@@ -46,7 +53,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `ARGILLA_LOCAL_AUTH_DEFAULT_APIKEY` and `ARGILLA_LOCAL_AUTH_DEFAULT_PASSWORD` environment variables. Use `python -m argilla.tasks.users.create_default` instead ([#2564]).
 
 [#2564]: https://github.com/argilla-io/argilla/issues/2564
-
 
 ## [1.5.0](https://github.com/recognai/rubrix/compare/v1.4.0...v1.5.0) - 2023-03-21
 
--- a/src/argilla/server/apis/v1/handlers/datasets.py
+++ b/src/argilla/server/apis/v1/handlers/datasets.py
@@ -31,4 +31,7 @@ router = APIRouter(tags=["datasets"])
 def list_datasets(*, db: Session = Depends(get_db), current_user: User = Security(auth.get_current_user)):
     authorize(current_user, DatasetPolicyV1.list)
 
-    return datasets.list_datasets(db)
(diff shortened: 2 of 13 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- `/api/v1/datasets` new endpoint to list datasets ([#2615]).
+
+[#2615]: https://github.com/argilla-io/argilla/issues/2615
+
+## [Unreleased]
```
frascuchon, line 16: This is to prevent problems when launching the WIP release, right?

src/argilla/server/apis/v1/handlers/datasets.py
```diff
@@ -0,0 +1,34 @@
...
+from argilla.server.security.model import User
+
+router = APIRouter(tags=["datasets"])
+
+
+@router.get("/datasets", response_model=List[Dataset])
+def list_datasets(*, db: Session = Depends(get_db), current_user: User = Security(auth.get_current_user)):
+    authorize(current_user, DatasetPolicyV1.list)
+
+    return datasets.list_datasets(db)
```
frascuchon: Take into account that for no admin users, returned datasets must be filtered by user workspaces

CHANGELOG.md
jfcalvo: Yes, exactly.

frascuchon (commented):
> I have added DatasetPolicyV1 class into policies.py. We can think into alternatives to split the policies into different versions if required.

👍 


> I have not used response_model_exclude_none=True into the new endpoint. I think it's not a good idea to remove attributes becase they are empty and I have taken the chance of not using it with this new v1.

So now the API response will ALWAYS contain the same number of fields, right?



> I have changed how database sessions are managed with the factories. Now it's possible to work with relationships so for example now the workspace is automatically created when a dataset factory is used.

This is only for tests, right? We didn't tackle transactions yet

src/argilla/server/apis/v1/handlers/datasets.py
jfcalvo: I totally forgot that.

jfcalvo:
> So now the API response will ALWAYS contain the same number of fields, right?

That's right.

> This is only for tests, right? We didn't tackle transactions yet

Only for factories yes.

src/argilla/server/apis/v1/handlers/datasets.py
```diff
@@ -0,0 +1,37 @@
...
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+from typing import List
+
+from fastapi import APIRouter, Depends, Security
+from sqlalchemy.orm import Session
```
frascuchon, line 18:
```suggestion
from sqlalchemy.orm import Session
from pydantic import parse_obj_as
```

src/argilla/server/apis/v1/handlers/datasets.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+@router.get("/datasets", response_model=List[Dataset])
+def list_datasets(*, db: Session = Depends(get_db), current_user: User = Security(auth.get_current_user)):
+    authorize(current_user, DatasetPolicyV1.list)
+
+    if current_user.is_admin:
+        return datasets.list_datasets(db)
+    else:
+        return current_user.datasets
```
frascuchon, line 37: This should be a List of Dataset schema instead of Dataset models:
```suggestion
        return parse_obj_as(List[Dataset], current_user.datasets)
```

frascuchon (commented):
Minimal comment about the schema data mapping

src/argilla/server/apis/v1/handlers/datasets.py
jfcalvo: I have found that is not necessary. If you have the `orm_mode = True` in the Pydantic schema and a correct `response_model` in the endpoint you have the serialization for free.

src/argilla/server/apis/v1/handlers/datasets.py
jfcalvo: Is not really necessary as I have commented.

frascuchon (approved):

PR #2628, opened 2023-03-31
Add new API v1 datasets create endpoint

jfcalvo:
# Description

This PR adds a new v1 endpoint to create datasets.

Some missing things that we need to discuss:
* Do we need to create something on ElasticSearch at the same time the dataset is created on the database?.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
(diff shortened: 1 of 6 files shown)


frascuchon (approved):
> * Do we need to create something on ElasticSearch at the same time the dataset is created on the database?.

Normally yes. But since the index creation is highly related to the `Record` and "questions/annotations/..."  definitions, I propose to create this atomic endpoint version now and add functionality once we start adding entities.

For example, the creation dataset endpoint should add info about requested questions for each record. So, once we introduce them to the creation endpoint, we can handle the es index creation.

PR #2634, opened 2023-04-02
fix Install doc spell error

PhilipMay:
this is just a small spelling  fix in the doc


dvsrepo (approved):
Thanks so much @PhilipMay 🙏

tomaarsen:
Thanks @PhilipMay, good to see you here too!

Issue #2533, opened 2023-03-14
add `retry` and `backoff` behavior to `rg.log`/ `rg.load`


davidberenstein1957:
**Is your feature request related to a problem? Please describe.**
logging is a bit fault prone but in a lot of my cases this logging could have been resolved using basic mitigation techniques.

**Describe the solution you'd like**
N.A.

**Describe alternatives you've considered**
N.A.

**Additional context**
https://github.com/encode/httpx/issues/108#issuecomment-1132753155

davidberenstein1957:
was handled in #2640

PR #2640, opened 2023-04-03
feat: Improve `rg.log` function

frascuchon:
# Description

Allow log data batches concurrently

- Accept `num_threads` to log batches concurrently    
- Add retries when an `httpx.TransportError` occurs

Close partially #2533

**Type of change**

**How Has This Been Tested**

TDB

**Checklist**


tomaarsen:
Works for me locally. I'm glad to see that the hacky `_ArgillaLogAgent` could be removed.
The default of `num_threads: int = 1` differs slightly to how I usually see threading, where 0 implies no threading and 1 implies one thread created. My understanding is that in this PR, 1 implies "no new thread created", is that right? I must say that I think I prefer this PR's approach, although it does differ from what I tend to see (e.g. in `torch` DataLoader `num_workers`).

Lastly, I think we're still missing tests for `num_threads > 1`.

frascuchon:
Thanks for your feedback @tomaarsen!!


> Works for me locally. I'm glad to see that the hacky `_ArgillaLogAgent` could be removed. The default of `num_threads: int = 1` differs slightly to how I usually see threading, where 0 implies no threading and 1 implies one thread created. My understanding is that in this PR, 1 implies "no new thread created", is that right? I must say that I think I prefer this PR's approach, although it does differ from what I tend to see (e.g. in `torch` DataLoader `num_workers`).
> 

Since the `rg.log` is a blocking function when `background=False`, the advantage of using a separate thread instead of the main one to process batches has no real improvement. Anyway, it's true that can raise confusion among users about the real number of separate threads used for computation. 

I think the approach of launching a separate one when `num_thread = 1` and using the main one when `num_thread=0` is more accurate to the param purpose. So, I will change it to fit that behavior

> Lastly, I think we're still missing tests for `num_threads > 1`.

Yes, I need to add some tests to check this flow.

tomaarsen:
Indeed, `num_threads=1` (or `num_workers=1`, etc) tends to be a performance decrease relative to using the main thread only (via 0). It's still offered usually, though.

src/argilla/client/client.py
```diff
@@ -391,27 +349,41 @@ async def log_async(
...
                     ),
                 )
 
-                processed += response.parsed.processed
-                failed += response.parsed.failed
-
                 progress_bar.update(task, advance=len(batch))
+                return bulk_result.processed, bulk_result.failed
+
+            if num_threads > 1:
```
```suggestion
            if num_threads >= 1:
```

src/argilla/client/api.py
```diff
@@ -112,10 +112,11 @@ def log(
     workspace: Optional[str] = None,
     tags: Optional[Dict[str, str]] = None,
     metadata: Optional[Dict[str, Any]] = None,
-    batch_size: int = 500,
+    batch_size: int = 100,
     verbose: bool = True,
     background: bool = False,
     chunk_size: Optional[int] = None,
+    num_threads: int = 0,
```
davidberenstein1957, line 121: Hi, thanks for fixing this. I would also like to see `max_retries` and the `backoff` variables here.

src/argilla/client/client.py
```diff
@@ -239,10 +220,7 @@ def copy(self, dataset: str, name_of_copy: str, workspace: str = None):
...
             client=self._client,
             name=dataset,
-            json_body=CopyDatasetRequest(
-                name=name_of_copy,
-                target_workspace=workspace,
-            ),
+            json_body=CopyDatasetRequest(name=name_of_copy, target_workspace=workspace),
         )
 
     def delete(self, name: str, workspace: Optional[str] = None):
```
davidberenstein1957, line 226: I would also like to see the `backoff` variables here.

src/argilla/client/client.py
```diff
@@ -423,6 +395,53 @@ async def log_async(
...
+        self,
+        records: Union[Record, Iterable[Record], Dataset],
+        name: str,
+        workspace: Optional[str] = None,
+        tags: Optional[Dict[str, str]] = None,
+        metadata: Optional[Dict[str, Any]] = None,
+        batch_size: int = 100,
+        verbose: bool = True,
+        chunk_size: Optional[int] = None,
+        num_threads: int = 1,
```
davidberenstein1957: I would also like to see `max_retries` and the `backoff` variables here.

davidberenstein1957 (commented):
Hi @frascuchon looks good. I would like to see the `max_retries`, and `backoff` in the other log functions, but ideally, they should also be added to the `rg.load`.

src/argilla/client/api.py
frascuchon: Agree with `max_retries`, I forget to add it. For the backoff not sure what you want to expose really? I would like to start with an MVP allowing the setup of at least the retries before stopping. We can consider exposing other parameters later. wdyt @davidberenstein1957 ?

src/argilla/client/client.py
frascuchon: We should apply changes step by step. We can consider adding a backoff mechanism to another method in a separate PR. Otherwise, a lot of changes will be included in the same PR, which can be a great bug farm. :-)

src/argilla/client/api.py
tomaarsen: `backoff` likely involves e.g. "expo", "constant", "fibo", etc. from https://github.com/litl/backoff
I think this is very overkill. Few people would ever change this from the default. I'd stick with `max_retries` as that one's easy to 1) use, 2) understand and 3) implement.

davidberenstein1957: IMO, we could add max_retries in combination with a default backoff 10, 20, 30

src/argilla/client/client.py
davidberenstein1957: https://github.com/argilla-io/argilla/issues/2687

davidberenstein1957: I added it to `rg.load` which seam the most relevant to me.

frascuchon: For `rg.load`, things could be a bit different. For instance, we should decrease the batch size, or we should prefetch some data before splitting and parallelizing the data loading. But yes. we can have a similar approach to improve also that method

frascuchon:
> Hi @frascuchon looks good. I would like to see the `max_retries`, and `backoff` in the other log functions, but ideally, they should also be added to the `rg.load`.

As in my previous comment, if it works for you, we can focus on the `rg.log` improvement in this PR and tackle the `rg.load` in a separate one.

src/argilla/client/client.py
tomaarsen: Decreasing batch size on failure seems very smart for `rg.load` in particular.

CHANGELOG.md
```diff
@@ -7,9 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
-### Changed
+## Added
+
+- add `max_retries` and `num_threads` parameters to `rg.log` to run data logging request concurrently with backoff retry policy. See [#2458](https://github.com/argilla-io/argilla/issues/2458) and [#2533](https://github.com/argilla-io/argilla/issues/2533)
+
+## Changed
 
 - Argilla quickstart image dependencies are externalized into `quickstart.requirements.txt`. See [#2666](https://github.com/argilla-io/argilla/pull/2666)
+- The `rg.log_async` is deprecated and will be removed in next minor release.
```
tomaarsen: Doesn't this fall under a "Deprecated" section?
image

CHANGELOG.md
```diff
@@ -7,9 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- The `rg.log` will compute all batches and raise an error for all failed batches.
```
tomaarsen, line 26: Perhaps mention that the batch size is now 100 under "Changed"? May be too small of a change to mention, though.

src/argilla/client/client.py
```diff
@@ -263,10 +241,12 @@ def log(
         workspace: Optional[str] = None,
         tags: Optional[Dict[str, str]] = None,
         metadata: Optional[Dict[str, Any]] = None,
-        batch_size: int = 500,
+        batch_size: int = 100,
         verbose: bool = True,
         background: bool = False,
         chunk_size: Optional[int] = None,
+        num_threads: int = 0,
+        max_retries: int = 3,
```
tomaarsen:
```suggestion
        num_threads: int = 0,
        max_retries: int = 3,
        chunk_size: Optional[int] = None,
```
Let's keep deprecated parameters at the bottom.

src/argilla/client/client.py
```diff
@@ -283,59 +263,38 @@ def log(
                 an ``asyncio.Future`` object. You probably want to set ``verbose`` to False
                 in that case.
             chunk_size: DEPRECATED! Use `batch_size` instead.
+            num_threads: If > 0, will use num_thread separate number threads to batches, sending data concurrently.
+                Default to `0`, which means no threading at all.
+            max_retries: Number of retries when logging a batch of records if a `httpx.TransportError` occurs.
+                Default `3`
```
tomaarsen:
```suggestion
            num_threads: If > 0, will use num_thread separate number threads to batches, sending data concurrently.
                Default to `0`, which means no threading at all.
            max_retries: Number of retries when logging a batch of records if a `httpx.TransportError` occurs.
                Default `3`
            chunk_size: DEPRECATED! Use `batch_size` instead.
```
Moving the deprecated parameter to the bottom.

src/argilla/client/client.py
```diff
@@ -283,59 +263,38 @@ def log(
...
+                verbose=verbose,
+                chunk_size=chunk_size,
+                num_threads=num_threads,
+                max_retries=max_retries,
+            )
 
-        Returns:
-            Summary of the response from the REST API
+        if workspace is not None:
+            self.set_workspace(workspace)
```
tomaarsen:
```suggestion
```
We already have this a few lines below:
https://github.com/argilla-io/argilla/blob/a23f9321cfa819f1e3a5fab7dd72a1047c5fbd15/src/argilla/client/client.py#L301-L302

src/argilla/client/client.py
```diff
@@ -389,27 +348,41 @@ async def log_async(
...
+            if num_threads >= 1:
+                with ThreadPoolExecutor(max_workers=num_threads) as executor:
+                    results = list(executor.map(log_batch, enumerate(batches)))
+            else:
+                results = list(map(log_batch, enumerate(batches)))
+
+            processed, failed = 0, 0
+            for processed_batch, failed_batch in results:
+                processed += processed_batch
+                failed += failed_batch
```
tomaarsen: I feel like this could be written more concisely, but it ends up being a mess of `zip`'s and `sum`'s 😅 
So we can just keep it like this.

tomaarsen (changes requested):
Some small nitpicks. Looks good otherwise, I'm always glad to see a PR that removes more code than it adds.

CHANGELOG.md
frascuchon: Totally

CHANGELOG.md
```diff
@@ -7,9 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-### Changed
+## Added
+
+- add `max_retries` and `num_threads` parameters to `rg.log` to run data logging request concurrently with backoff retry policy. See [#2458](https://github.com/argilla-io/argilla/issues/2458) and [#2533](https://github.com/argilla-io/argilla/issues/2533)
+
+## Changed
 
 - Argilla quickstart image dependencies are externalized into `quickstart.requirements.txt`. See [#2666](https://github.com/argilla-io/argilla/pull/2666)
+- The `rg.log` computes all batches and raise an error for all failed batches.
+- The `rg.log` default batch size is now 100.
```
```suggestion
- The default batch size for `rg.log` is now 100.
```

src/argilla/client/client.py
frascuchon: Done

src/argilla/client/client.py
frascuchon: Yes, I've refactored a bit this code block. Thanks for remarking it

CHANGELOG.md
```diff
@@ -7,14 +7,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-###  Added
+## Added
```
```suggestion
### Added
```

CHANGELOG.md
```diff
@@ -7,14 +7,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - `rg.load` accepts `exclude_vectors` and `exclude_metrics` when loading data. Closes [#2398](https://github.com/argilla-io/argilla/issues/2398)
 
 ### Changed
 
 - Argilla quickstart image dependencies are externalized into `quickstart.requirements.txt`. See [#2666](https://github.com/argilla-io/argilla/pull/2666)
 - bulk endpoints will upsert data when record `id` is present. Closes [#2535](https://github.com/argilla-io/argilla/issues/2535)
+- The `rg.log` computes all batches and raise an error for all failed batches.
+- The default batch size for `rg.log` is now 100.
+
+## Deprecated
```
```suggestion
### Deprecated
```

tomaarsen (approved):
Looks great, I think this is all set

PR #2641, opened 2023-04-03
Add datasets get endpoint and refactor API v1 errors and status codes

jfcalvo:
# Description

This PR adds a new endpoint useful to get datasets to API v1.

I have added some changes too:
* Now instead of raising custom exceptions I'm using FastAPI `HTTPException`.
* `status_code` is changed for the creation route from `200` to `201`.

@frascuchon please take a look to the v1 policies for Datasets and check that everything is correct.


frascuchon (approved):
The policies are fine for now.

PR #2647, opened 2023-04-04
Add new annotations table and ORM model

jfcalvo:
# Description

* A new migration to create `annotations` table. This is a temporal name if we think a better one is possible we can change it later.
* A new `Annotation` ORM model with the required relationships.
* Some missing details from previous PRs.

### DIFF
--- a/src/argilla/server/models.py
+++ b/src/argilla/server/models.py
@@ -50,9 +50,9 @@ class Annotation(Base):
     id: Mapped[UUID] = mapped_column(primary_key=True, default=uuid4)
     name: Mapped[str]
     title: Mapped[str] = mapped_column(Text)
-    type: Mapped[Optional[AnnotationType]] = mapped_column(default=AnnotationType.text)
-    required: Mapped[Optional[bool]] = mapped_column(default=False)
-    settings: Mapped[Optional[dict]] = mapped_column(JSON, default={})
+    type: Mapped[AnnotationType] = mapped_column(default=AnnotationType.text)
+    required: Mapped[bool] = mapped_column(default=False)
+    settings: Mapped[dict] = mapped_column(JSON, default={})
     dataset_id: Mapped[UUID] = mapped_column(ForeignKey("datasets.id"))
 
     inserted_at: Mapped[datetime] = mapped_column(default=datetime.utcnow)
--- a/src/argilla/server/models.py
+++ b/src/argilla/server/models.py
@@ -36,7 +36,7 @@ def default_inserted_at(context):
 
 class AnnotationType(str, Enum):
     text = "text"
-    single_label = "single_label"
+    rating = "rating"
 
 
 class UserRole(str, Enum):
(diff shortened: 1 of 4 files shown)


src/argilla/server/models.py
```diff
@@ -34,26 +34,54 @@ def default_inserted_at(context):
...
     annotator = "annotator"
 
 
+class Annotation(Base):
+    __tablename__ = "annotations"
+
+    id: Mapped[UUID] = mapped_column(primary_key=True, default=uuid4)
+    name: Mapped[str]
+    title: Mapped[str] = mapped_column(Text)
+    type: Mapped[Optional[AnnotationType]] = mapped_column(default=AnnotationType.text)
```
frascuchon: Why this attribute is optional? As far as I understand, optional attributes can be nullable, which differs from the table definition in this case

src/argilla/server/models.py
```diff
@@ -34,26 +34,54 @@ def default_inserted_at(context):
...
+    required: Mapped[Optional[bool]] = mapped_column(default=False)
+    settings: Mapped[Optional[dict]] = mapped_column(JSON, default={})
```
frascuchon: Same for these fields
```suggestion
    required: Mapped[bool] = mapped_column(default=False)
    settings: Mapped[dict] = mapped_column(JSON, default={})
```

src/argilla/server/models.py
jfcalvo: We have a default `AnnotationType.text`, that's why it's optional. If you don't specify one then we are using the `text` annotation type.

src/argilla/server/models.py
jfcalvo: The same applies here.
* `required` has a default value of `False` so it's optional. If we don't specify a value for `required` the annotation then is not required. Of course we can do the opposite, if a value for `required` is not specified by default the annotation is required (using `True` as default).
* I was thinking that maybe we don't have a `settings` value for all kind of annotations so I'm using as default an empty JSON object here (or Python empty dict) that's why is optional too.

src/argilla/server/models.py
jfcalvo: I have been rechecking and with a `default` value the `Optional` is useless. We can remove it.

src/argilla/server/models.py
```diff
@@ -34,26 +34,54 @@ def default_inserted_at(context):
     return context.get_current_parameters()["inserted_at"]
 
 
+class AnnotationType(str, Enum):
+    text = "text"
+    single_label = "single_label"
```
frascuchon: So, after the slack conversation, we can change this to:
```suggestion
    free_text = "free_text"
    rating = "rating"
```

jfcalvo: I have added `rating` but didn't change `text` to `free_text`. Not sure if `free_text` it's better than simply `text`.

dvsrepo: I agree that simply `text` is better, the free doesn't add anything

frascuchon: Sorry, I reviewed this quickly, and I was thinking about single and multi-label when I did the changes. `text` fits well for me too.

frascuchon (approved):

PR #2651, opened 2023-04-05
Record class model

frascuchon:
The first iteration for the new record class definition. This data model will be used as API response for record-related endpoints.


jfcalvo (approved):
It's ok I guess but it is difficult for me to see something without integrating it with the rest of the logic.

PR #2653, opened 2023-04-05
Add datasets annotations get enpdoint

jfcalvo:
# Description

This PR adds a new v1 endpoint to get all the annotations belonging to a specific dataset.

Notes:
* I'm not returning the `dataset_id` on the `Annotation` Pydantic schema. The reason behind this is because we are requesting annotations for a dataset so the `dataset_id` is implicit.
* I have created the `Annotation` Pydantic schema inside `src/argilla/server/schemas/v1/datasets.py`. The reason behind this is because this is the schema for dataset annotations, if required we can have another set of schemas for annotations not related with the dataset (.i.e. we have an endpoint only for annotations).

### DIFF
--- a/src/argilla/server/apis/v1/handlers/datasets.py
+++ b/src/argilla/server/apis/v1/handlers/datasets.py
@@ -21,11 +21,22 @@ from sqlalchemy.orm import Session
 from argilla.server.contexts import datasets
 from argilla.server.database import get_db
 from argilla.server.policies import DatasetPolicyV1, authorize
-from argilla.server.schemas.v1.datasets import Dataset, DatasetCreate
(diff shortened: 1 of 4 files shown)


frascuchon (approved):
> I'm not returning the dataset_id on the Annotation Pydantic schema. The reason behind this is because we are requesting annotations for a dataset so the dataset_id is implicit.

👍 

>I have created the Annotation Pydantic schema inside src/argilla/server/schemas/v1/datasets.py. The reason behind this is because this is the schema for dataset annotations, if required we can have another set of schemas for annotations not related with the dataset (.i.e. we have an endpoint only for annotations).

👍

PR #2656, opened 2023-04-05
Dataset index creation

frascuchon:
New elastic search engine component and `create_dataset_index` method.

Some notes:

- This component work using the official elasticsearch component, we should provide also opensearch compatibility as the current API endpoints. But we can tackle this in a separate PR.

- The engine is using the dataset id as part of the index name. The `rg` prefix is included to simplify index policies definition (you can define a snapshot policy by defining a wildcard. For Argilla indices this would not be possible only using the dataset id as index name)

### DIFF
--- a/src/argilla/server/elasticsearch.py
+++ b/src/argilla/server/elasticsearch.py
@@ -26,7 +26,7 @@ class ElasticSearchEngine:
     def __post_init__(self):
         self._client = ElasticsearchClient(vector_search_supported=False, index_shards=1, config_backend=self.config)
 
-    def create_records_index(self, dataset: Dataset) -> str:
+    def create_dataset_index(self, dataset: Dataset) -> str:
         fields = {}
 
         for annotation in dataset.annotations:
@@ -45,14 +45,11 @@ class ElasticSearchEngine:
         return index_name
 
     def _field_mapping_for_annotation(self, annotation_task: Annotation):
-        mapping = {}
         if annotation_task.type == AnnotationType.rating:
             # See https://www.elastic.co/guide/en/elasticsearch/reference/current/number.html
-            mapping.update({"type": "integer"})
+            return {"type": "integer"}
         elif annotation_task.type == AnnotationType.text:
             # See https://www.elastic.co/guide/en/elasticsearch/reference/current/text.html
-            mapping.update({"type": "text"})
+            return {"type": "text"}
         else:
             raise ValueError(f"Annotation of type {annotation_task.type} cannot be processed")
-
-        return mapping
--- a/src/argilla/server/elasticsearch.py
+++ b/src/argilla/server/elasticsearch.py
@@ -13,9 +13,10 @@
 #  limitations under the License.
 
 import dataclasses
-from typing import Any, Dict, Optional
+from typing import Any, Dict
 
-from argilla.server.daos.backend.client_adapters import ElasticsearchClient
+from elasticsearch import Elasticsearch
+
 from argilla.server.models import Annotation, AnnotationType, Dataset
 
 
@@ -24,7 +25,7 @@ class ElasticSearchEngine:
     config: Dict[str, Any]
 
     def __post_init__(self):
-        self._client = ElasticsearchClient(vector_search_supported=False, index_shards=1, config_backend=self.config)
+        self._client = Elasticsearch(**self.config)
 
     def create_dataset_index(self, dataset: Dataset) -> str:
         fields = {}
@@ -40,7 +41,7 @@ class ElasticSearchEngine:
         }
 
         index_name = f"rg.{dataset.id}"
-        self._client.create_index(index_name, mappings=mappings)
+        self._client.indices.create(index=index_name, body={"mappings": mappings})
 
         return index_name
 
--- a/src/argilla/server/elasticsearch.py
+++ b/src/argilla/server/elasticsearch.py
@@ -1,0 +1,58 @@
+#  Copyright 2021-present, the Recognai S.L. team.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+import dataclasses
+from typing import Any, Dict, Optional
+
+from argilla.server.daos.backend.client_adapters import ElasticsearchClient
+from argilla.server.models import Annotation, AnnotationType, Dataset
+
+
+@dataclasses.dataclass
+class ElasticSearchEngine:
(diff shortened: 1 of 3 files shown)


jfcalvo:
We need to think about these changes and make it match what we have on PR https://github.com/argilla-io/argilla/pull/2657

src/argilla/server/elasticsearch.py
```diff
@@ -0,0 +1,56 @@
...
+
+        # See https://www.elastic.co/guide/en/elasticsearch/reference/current/explicit-mapping.html
+        mappings = {
+            # See https://www.elastic.co/guide/en/elasticsearch/reference/current/dynamic.html#dynamic-parameters
+            "dynamic": "strict",
+            "properties": fields,
+        }
+
+        index_name = f"rg.{dataset.id}"
+        self._client.indices.create(index=index_name, body={"mappings": mappings})
```
jfcalvo, line 44: This is more like a mapping for a record and not a dataset right?

frascuchon: This is the setup for the index where the records will be indexed, yes. But here, you can set up other things than the record mapping. See the [docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-create-index.html#indices-create-api-request-body)

In terms of argilla concepts, this setup is related to records of a dataset instead of the dataset itself. Saying this, we can review the method name, adding the record-dataset relationship.

src/argilla/server/elasticsearch.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+@dataclasses.dataclass
+class ElasticSearchEngine:
+    config: Dict[str, Any]
+
+    def __post_init__(self):
+        self._client = Elasticsearch(**self.config)
+
+    def create_dataset_index(self, dataset: Dataset) -> str:
```
frascuchon:
```suggestion
    def setup_dataset_records_index(self, dataset: Dataset) -> str:
```

tests/server/test_elasticsearch.py
```diff
@@ -0,0 +1,117 @@
...
+    return ElasticSearchEngine(config=es_config)
+
+
+@pytest.mark.skipif(
+    condition=not _running_with_elasticsearch(),
+    reason="Test only running with elasticsearch backend",
+)
+def test_create_index_for_dataset(search_engine: ElasticSearchEngine, elasticsearch: Elasticsearch):
+    dataset = DatasetFactory.create()
+    index_name = search_engine.create_dataset_index(dataset)
```
frascuchon:
```suggestion
    index_name = search_engine.setup_dataset_records_index(dataset)
```

tests/server/test_elasticsearch.py
```diff
@@ -0,0 +1,117 @@
...
+    db: Session,
+    text_ann_size: int,
+    rating_ann_size: int,
+):
+    text_annotations = AnnotationFactory.create_batch(size=text_ann_size, type=AnnotationType.text)
+    rating_annotations = AnnotationFactory.create_batch(size=rating_ann_size, type=AnnotationType.rating)
+
+    dataset = DatasetFactory.create(annotations=text_annotations + rating_annotations)
+
+    index_name = search_engine.create_dataset_index(dataset)
```
frascuchon:
```suggestion
    index_name = search_engine.setup_dataset_records_index(dataset)
```

tests/server/test_elasticsearch.py
```diff
@@ -0,0 +1,117 @@
...
+def test_create_index_with_existing_index(
+    search_engine: ElasticSearchEngine, elasticsearch: Elasticsearch, db: Session
+):
+    dataset = DatasetFactory.create()
+    index_name = search_engine.create_dataset_index(dataset)
+
+    assert elasticsearch.indices.exists(index=index_name)
+
+    with pytest.raises(BadRequestError, match="'resource_already_exists_exception', 'index"):
+        search_engine.create_dataset_index(dataset)
```
frascuchon:
```suggestion
        search_engine.setup_dataset_records_index(dataset)
```

src/argilla/server/elasticsearch.py
```diff
@@ -0,0 +1,56 @@
...
+        mappings = {
+            # See https://www.elastic.co/guide/en/elasticsearch/reference/current/dynamic.html#dynamic-parameters
+            "dynamic": "strict",
+            "properties": fields,
+        }
+
+        index_name = f"rg.{dataset.id}"
+        self._client.indices.create(index=index_name, body={"mappings": mappings})
+
+        return index_name
```
jfcalvo: Why do we need to return the index name?

frascuchon: Yes, we really don't need to return it.

tests/server/test_elasticsearch.py
```diff
@@ -0,0 +1,117 @@
...
+    argvalues=[(random.randint(1, 9), random.randint(1, 9)) for _ in range(1, 5)],
+)
+def test_create_index_for_dataset_with_annotations(
+    search_engine: ElasticSearchEngine,
+    elasticsearch: Elasticsearch,
+    db: Session,
+    text_ann_size: int,
+    rating_ann_size: int,
+):
+    text_annotations = AnnotationFactory.create_batch(size=text_ann_size, type=AnnotationType.text)
```
jfcalvo, line 86: You can use `TextAnnotationFactory` instead if you wish.

tests/server/test_elasticsearch.py
```diff
@@ -0,0 +1,117 @@
...
+    rating_annotations = AnnotationFactory.create_batch(size=rating_ann_size, type=AnnotationType.rating)
```
jfcalvo, line 87: You can use `RatingAnnotationFactory` instead if you wish.

tests/server/test_elasticsearch.py
```diff
@@ -0,0 +1,117 @@
...
+@pytest.fixture(scope="session")
+def search_engine(es_config):
+    return ElasticSearchEngine(config=es_config)
+
+
+@pytest.mark.skipif(
+    condition=not _running_with_elasticsearch(),
+    reason="Test only running with elasticsearch backend",
+)
+def test_create_index_for_dataset(search_engine: ElasticSearchEngine, elasticsearch: Elasticsearch):
```
jfcalvo, line 60: We should not allow to create datasets without annotations so not sure if this case it's important.

tests/server/test_elasticsearch.py
frascuchon: Thanks. Those factories did not exist when I implemented the tests. I will update from the main feat branch.

tests/server/test_elasticsearch.py
frascuchon: ditto

tests/server/test_elasticsearch.py
frascuchon: This validation should be placed at the dataset creation level right? Otherwise, we should replicate the validation for each engine implementation.

frascuchon: I can imagine this as a validation step when a dataset becomes ready.

jfcalvo: > I can imagine this as a validation step when a dataset becomes ready.

Exactly, when a request to publish the dataset is done if the dataset doesn't has annotations the publish procedure should fail.

jfcalvo (approved):

PR #2657, opened 2023-04-05
Add datasets annotations create endpoint

jfcalvo:
# Description

This PR adds a new v1 endpoint to create annotations for a specific dataset.

Notes:
* It includes the `text` and `rating` annotation types.
* `text` annotation settings is empty right now, we can add possible settings in the future if required. At the same time if a users specifies some settings values for this type of annotation they are discarded.
* `rating` annotation settings allow a `options` attribute that can store a list of objects. Every object has the structure `{"value": int}` only allowing integer values.
* `rating` annotation settings allow a `options` attribute with a minimum of 2 items and a maximum of 100. If it doesn't has many sense we can change these values.
* I have tried to use Pydantic's [discriminated unions](https://docs.pydantic.dev/usage/types/#discriminated-unions-aka-tagged-unions) to select between different models based on the `type` attribute for the annotations but without any luck. Looks like it has a bug and JSON serialization is not allowed for subclases using this. I have solved this using a dynamic validator but please tell me if there is another better way to solve this.

### DIFF
--- a/src/argilla/server/apis/v1/handlers/datasets.py
+++ b/src/argilla/server/apis/v1/handlers/datasets.py
@@ -21,7 +21,12 @@ from sqlalchemy.orm import Session
 from argilla.server.contexts import datasets
 from argilla.server.database import get_db
 from argilla.server.policies import DatasetPolicyV1, authorize
-from argilla.server.schemas.v1.datasets import Annotation, Dataset, DatasetCreate
+from argilla.server.schemas.v1.datasets import (
+    Annotation,
+    AnnotationCreate,
+    Dataset,
+    DatasetCreate,
+)
 from argilla.server.security import auth
 from argilla.server.security.model import User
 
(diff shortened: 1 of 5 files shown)


src/argilla/server/apis/v1/handlers/datasets.py
```diff
@@ -99,6 +104,27 @@ def create_dataset(
...
 
+@router.post("/datasets/{dataset_id}/annotations", status_code=status.HTTP_201_CREATED, response_model=Annotation)
+def create_dataset_annotation(
+    *,
+    db: Session = Depends(get_db),
+    dataset_id: UUID,
+    annotation_create: AnnotationCreate,
+    current_user: User = Security(auth.get_current_user),
+):
+    authorize(current_user, DatasetPolicyV1.create)
```
frascuchon, line 115: The proper policy should be `create_annotation`, right?

src/argilla/server/schemas/v1/datasets.py
```diff
@@ -51,3 +54,34 @@ class Annotation(BaseModel):
...
+        max_items=RATING_OPTIONS_MAX_ITEMS,
+    )
+
+
+class AnnotationCreate(BaseModel):
+    name: str
+    title: str
+    type: AnnotationType
+    required: Optional[bool]
+    settings: Optional[dict] = {}
```
frascuchon, line 76: I think by using Pydantic discriminators we should define the annotation type field at the settings level. For me, it's fine your solution. Just one thing. If settings are defined as `Optional` that means that we can accept `None` values, but the defined validator will always try to create a new settings instance.
```suggestion
    settings: dict = {}
```

src/argilla/server/schemas/v1/datasets.py
```diff
@@ -51,3 +54,34 @@ class Annotation(BaseModel):
...
+
+    @validator("settings", always=True)
```
frascuchon, line 78: Adding `pre=True` we can be sure to apply settings parsing before other validations are done
```suggestion
    @validator("settings", pre=True, always=True)
```

src/argilla/server/schemas/v1/datasets.py
```diff
@@ -51,3 +54,34 @@ class Annotation(BaseModel):
...
+    def validate_settings(cls, settings: dict, values):
+        type = values.get("type")
+
+        if type == AnnotationType.text:
+            return {}
+        if type == AnnotationType.rating:
+            return RatingAnnotationSettingsCreate(**settings).dict()
```
frascuchon, line 85: So, the *SettingsCreate will be used only for validation purposes?

It's a bit weird to me this logic. Even more later when we'll start to use the settings classes for record data validation.

src/argilla/server/apis/v1/handlers/datasets.py
jfcalvo: Yes, you're right. `create_annotation` will be present for dataset status checks.

frascuchon (approved):

jfcalvo:
We will review @frascuchon feedback on another PR so we can advance making changes.

PR #2667, opened 2023-04-09
Fix GPU usage & other minor improvements in `ArgillaSpaCyTrainer`

alvarobartt:
# Description

After extensively testing `ArgillaTrainer` as of the latest release for `spaCy`, I've found some things that had to be fixed/improved/refactored:

* Since `gpu_id` was not provided in `ArgillaTrainer`, then it was not provided to `ArgillaSpaCyTrainer` so the GPU could only be used when using `ArgillaSpaCyTrainer` directly
* `ArgillaSpaCyTrainer.train` was still using `path` as the param instead of `output_dir`
* Remove unnecessary `requires_version` function calls in `ArgillaSpaCyTrainer`, and add `requires_version("spacy-transformers")` just when required
* Improve `gpu_id` param docstrings in both `ArgillaTrainer` and `ArgillaSpaCyTrainer`

**Type of change**

**How Has This Been Tested**

- [X] Add `test_init_with_gpu_id` integration tests to check that providing `gpu_id` to `ArgillaTrainer` with `ArgillaSpaCyTrainer` works as expected
- [X] Train a token-classification/text-classification model using `ArgillaSpaCyTrainer` with and without GPU in Google Colab via the `gpu_id` param being 0 and 1, respectively. _I can upload those either via sharing the link or just adding those to the tutorials directory if needed._

**Checklist**

### DIFF
--- a/src/argilla/training/spacy.py
+++ b/src/argilla/training/spacy.py
@@ -77,14 +77,14 @@ class ArgillaSpaCyTrainer:
         self._record_class = record_class
         if self._record_class == rg.TokenClassificationRecord:
             self._column_mapping = {"text": "text", "token": "tokens", "ner_tags": "ner_tags"}
-            self._pipeline_name = "ner"
+            self._pipeline = ["ner"]
         elif self._record_class == rg.TextClassificationRecord:
             if self._multi_label:
                 self._column_mapping = {"text": "text", "binarized_label": "label"}
-                self._pipeline_name = "textcat_multilabel"
+                self._pipeline = ["textcat_multilabel"]
             else:
                 self._column_mapping = {"text": "text", "label": "label"}
-                self._pipeline_name = "textcat"
+                self._pipeline = ["textcat"]
         else:
             raise NotImplementedError("`rg.Text2TextRecord` is not supported yet.")
 
@@ -95,21 +95,32 @@ class ArgillaSpaCyTrainer:
         self._eval_dataset_path = "./dev.spacy" if self._eval_dataset else None
 
         self.language = language or "en"
+        self.model = model
         self.gpu_id = gpu_id
-        if self.gpu_id != -1:
+        self.use_gpu = self.gpu_id != -1
+        if self.use_gpu:
             try:
+                require_version("spacy-transformers")
                 spacy.prefer_gpu(self.gpu_id)
             except:
                 self.gpu_id = -1
+                self.use_gpu = False
+                if self.model is not None:
+                    self.model = None
 
         self.config = init_config(
             lang=self.language,
-            pipeline=[self._pipeline_name],
+            pipeline=self._pipeline,
+            optimize="accuracy",
+            gpu=self.use_gpu,
         )
         self.config["paths"]["train"] = self._train_dataset_path
         self.config["paths"]["dev"] = self._eval_dataset_path or self._train_dataset_path
-        self.config["paths"]["vectors"] = model
         self.config["system"]["seed"] = seed or 42
+        if self.use_gpu:
+            self.config["components"]["transformer"]["model"]["name"] = model or "roberta-base"
+        else:
+            self.config["paths"]["vectors"] = model or "en_core_web_lg"
 
         self._nlp = None
 
@@ -212,7 +223,7 @@ class ArgillaSpaCyTrainer:
         docs = self._nlp.pipe(text)
         if as_argilla_records:
             for doc in docs:
-                if self._pipeline_name == "ner":
+                if "ner" in self._pipeline:
                     entities = [(ent.label_, ent.start_char, ent.end_char) for ent in doc.ents]
                     pred = {
                         "text": doc.text,
@@ -220,7 +231,7 @@ class ArgillaSpaCyTrainer:
                         "prediction": entities,
                     }
                     pred = self._record_class(**pred)
-                elif self._pipeline_name in ["textcat", "textcat_multilabel"]:
+                elif any([p in self._pipeline for p in ["textcat", "multilabel_textcat"]]):
                     pred = {
                         "text": doc.text,
                         "prediction": [(k, v) for k, v in doc.cats.items()],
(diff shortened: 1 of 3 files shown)


src/argilla/training/spacy.py
```diff
@@ -30,9 +30,6 @@ class ArgillaSpaCyTrainer:
     _logger = logging.getLogger("ArgillaSpaCyTrainer")
     _logger.setLevel(logging.INFO)
 
-    require_version("torch")
```
davidberenstein1957, line 31: I believe It is also possible to use CPU to train with spacy-transformers and therefore torch.

src/argilla/training/spacy.py
```diff
@@ -98,21 +99,45 @@ def __init__(
...
-            pipeline=[self._pipeline_name],
+            pipeline=self._pipeline,
+            optimize="efficiency",
+            gpu=self.use_gpu,
         )
         self.config["paths"]["train"] = self._train_dataset_path
         self.config["paths"]["dev"] = self._eval_dataset_path or self._train_dataset_path
-        self.config["paths"]["vectors"] = model
         self.config["system"]["seed"] = seed or 42
+        if self.use_gpu:
```
davidberenstein1957, line 149: I believe it makes more sense to default to certain models in a more explicit way. Also because we potentially do not only want to reuse the transformer/vectors but also other model components.

davidberenstein1957 (commented):
hi @alvarobartt, thanks for the PR! I have some small remarks:)

src/argilla/training/spacy.py
alvarobartt: Yes I agree, I'll be investigating this these days, so I'll move this back to draft since there are still some things I want/need to change before merging. Thanks!

davidberenstein1957:
Hi @alvarobartt could you merge your changes into this branch first?
https://github.com/argilla-io/argilla/pull/2691

src/argilla/training/spacy.py
alvarobartt: Hmm so it's weird because the train configuration when selecting the GPU is generated just for `spacy-transformers`, but I've tested it and it also works fine with CPU. Same thing for training `spaCy` models, so either we temporarily drop support for `spacy-transformers` models or just figure out the best way to combine both.

alvarobartt:
Hi @davidberenstein1957 so I've spotted some things during these days:

* `gpu_id` was missing in `ArgillaTrainer` so the GPU could never be used to train `spaCy` models via `ArgillaTrainer` while those could via `ArgillaSpaCyTrainer`
* `gpu_id` is useless when using MPS, but useful when using CUDA
* Both `spaCy` and `spacy-transformers` models can be trained in CPU and GPU
* `torch` is not required, since it's just needed when the GPU is used (same applies to `tensorflow` that can also be used as the GPU allocator)
* `ArgillaSpaCyTrainer` config is being generated for `spaCy` when using the CPU and for `spacy-transformers` when using the GPU by default, even though to use the GPU to train a `spaCy` model just the allocator needs to be defined e.g. `torch`
* It's not possible to infer whether the model is a `spaCy` model or a `spacy-transformers` model, and since those can be trained in CPU and GPU, it's impossible to determine which model to train unless the classes are separated in `ArgillaSpaCyTrainer` and `ArgillaSpaCyTransformersTrainer`, or a boolean flag is provided e.g. `use_spacy_transformers`

So on, for the moment `spaCy` is always used by default for both CPU and GPU training loops, while `spacy-transformers` is just used when the GPU is specified and `spacy-transformers` is installed. Anyway, my take on this is that I would just disable `spacy-transformers` support for the moment until some of the points mentioned above are properly discussed. And, IMO, the most compliant approach based on the current, is to just define `ArgillaSpaCyTransformersTrainer` and add `spacy-transformers` as a separate framework.

LMKWYT @davidberenstein1957 👍🏻

davidberenstein1957:
Hi @davidberenstein1957 so I've spotted some things during these days:

>     gpu_id was missing in ArgillaTrainer so the GPU could never be used to train spaCy models via ArgillaTrainer while those could via ArgillaSpaCyTrainer`

Will you pick this up in this PR?

>     gpu_id is useless when using MPS, but useful when using CUDA

Exactly, but i believe there is `spacy[metal]` for m1 support.

>     Both spaCy and spacy-transformers models can be trained in CPU and GPU
>     torch is not required, since it's just needed when the GPU is used (same applies to tensorflow that can also be used as the GPU allocator)

>     ArgillaSpaCyTrainer config is being generated for spaCy when using the CPU and for spacy-transformers when using the GPU by default, even though to use the GPU to train a spaCy model just the allocator needs to be defined e.g. torch
>     It's not possible to infer whether the model is a spaCy model or a spacy-transformers model, and since those can be trained in CPU and GPU, it's impossible to determine which model to train unless the classes are separated in ArgillaSpaCyTrainer and ArgillaSpaCyTransformersTrainer, or a boolean flag is provided e.g. use_spacy_transformers

I believe each saved spacy model holds a `config.cfg` which is created when saving a model. So, this contains info about the model, and additionally all default models end with `_trf`. I prefer having one ArgillaSpacyTrainer that infers knowledge from there instead of having a separate trainer handle this logic by loading this `.cfg` or by simple loading the model along with some `try-except-logic`. Not sure if you agree?

So on, for the moment spaCy is always used by default for both CPU and GPU training loops, while spacy-transformers is just used when the GPU is specified and spacy-transformers is installed. Anyway, my take on this is that I would just disable spacy-transformers support for the moment until some of the points mentioned above are properly discussed. And, IMO, the most compliant approach based on the current, is to just define ArgillaSpaCyTransformersTrainer and add spacy-transformers as a separate framework. 

LMKWYT @davidberenstein1957 👍🏻

alvarobartt:
Hi @davidberenstein1957, so the `gpu_id` param has already been included and that works fine, do you want me to add anything else on top of this PR? Maybe we can have a call to discuss it? Because to me it's not straightforward how to support both `spaCy` and `spacy-transformers` without any hint from the user, so maybe you have an idea on that 🤗

alvarobartt:
Hi @davidberenstein1957 I've temporarily removed support for `spacy-transformers` as I already told you via Slack, so this PR is ready to be merged. I'll tackle the `spacy-transformers` integration separately.

davidberenstein1957 (approved):

PR #2668, opened 2023-04-09
feat: Allow exclude vectors when loading records

### LINKED ISSUE #2398: rg.load takes too long because of the vector field

frascuchon:
# NOTE

~~PR #2576 must be merged first to make this PR "mergeable"~~

# Description

Work implemented in https://github.com/argilla-io/argilla/pull/2577 and https://github.com/argilla-io/argilla/pull/2540 tries to introduce a mechanism for:

1. Loading data with field projections (selecting which record fields will be loaded)
2. Allow partial updates when uploading projected records

Some problems of this flow are raised during the work:

- There is a field mismatch between client and API records. The proposed field projection is referring to the Record API data model, and aligning it with client representation can be messy and give an unexpected behavior
- The update flag mechanism is a bit cumbersome because of the client data projection and the record class validation alignments. So, even for a single field update, you need to retrieve extra fields to keep aligned the validations.

The field projection in `rg.load` function can introduce confusion because both record models, client, and API, are not aligned. So, I prefer to remove this functionality. Instead of that, we can introduce some extra parameters to load for setup if users want to exclude `vectors` or `metrics` when loading data. This will help to mitigate problems founds in #2398 

This option, together with changes introduced in #2576, allows users to load partial record info and still update it without losing stored data.

For example, if users want to add some metadata to already stored records, they could try the following flow:

```python
ds = rg.load(name=dataset, exclude_vectors=True)

for record in ds:
   record.metadata = ...

rg.log(name=dataset, records=ds) # This will update records only by adding the metadata info
```

Closes [#2398](https://github.com/argilla-io/argilla/issues/2398)

Refs: https://github.com/argilla-io/argilla/pull/2577 and https://github.com/argilla-io/argilla/pull/2540

**Type of change**

**How Has This Been Tested**

Existing tests are been modified including this new behavior

**Checklist**


davidberenstein1957 (approved):
This approach makes sense to me.

src/argilla/client/client.py
```diff
@@ -490,6 +492,13 @@ def load(
                 can be used to load using batches.
             as_pandas: DEPRECATED! To get a pandas DataFrame do
                 ``rg.load('my_dataset').to_pandas()``.
+            batch_size: If provided, load `batch_size` samples per request. A lower batch
+                size may help avoid timeouts.
+            exclude_vectors: When set to `True`, indicates that the record data will be retrieved excluding its vectors,
+                if any. By default, this parameter is set to `False`, meaning that vector data will be included.
+            exclude_metrics: When set to `True`, indicates that the record data will be retrieved excluding its metrics.
+                By default, this parameter is set to `False`, meaning that metrics will be included.
```
tomaarsen:
```suggestion
            batch_size: If provided, load `batch_size` samples per request. A lower batch
                size may help avoid timeouts.
            exclude_vectors: When set to `True`, indicates that the record data will be retrieved excluding its vectors,
                if any. By default, this parameter is set to `False`, meaning that vector data will be included.
            exclude_metrics: When set to `True`, indicates that the record data will be retrieved excluding its metrics.
                By default, this parameter is set to `False`, meaning that metrics will be included.
            as_pandas: DEPRECATED! To get a pandas DataFrame do
                ``rg.load('my_dataset').to_pandas()``.
```
Let's stick with the ordering from the method signature.
Edit: Wow, the diff looks like a mess. I've just moved `as_pandas` down to below `exclude_metrics`.

src/argilla/client/client.py
```diff
@@ -646,6 +659,11 @@ def _load_records_internal(
             if sort is not None:
                 _LOGGER.warning("Results are sorted by vector similarity, so 'sort' parameter is ignored.")
 
+            if exclude_metrics or exclude_vectors:
+                _LOGGER.warning(
+                    "Metrics and vectors cannot be excluded when using vector search. Those params will be ignored"
```
tomaarsen:
```suggestion
                    "Metrics and vectors cannot be excluded when using vector search. These parameters will be ignored."
```

tomaarsen (approved):
The `exclude_vectors = False` is a bit of a double negative which I don't love, so I have preference for `include_vectors = True`. That one requires a bit less thought to understand. However, I'm cool with merging this with `exclude_...`, too.

I have a few nits, this is ready when they're tackled as far as I'm concerned.

frascuchon:
> The `exclude_vectors = False` is a bit of a double negative which I don't love, so I have preference for `include_vectors = True`. That one requires a bit less thought to understand. However, I'm cool with merging this with `exclude_...`, too.
> 
> I have a few nits, this is ready when they're tackled as far as I'm concerned.

Yes, makes sense to me also. It's more standard to use `include_...` instead of `exclude...`

tomaarsen (approved):
I think this is all set. When the tests go green anyways. Thanks for this and well done :)

PR #2670, opened 2023-04-10
Fix default user credentials

akaia-shadowfox:
# Description

`admin` / `12345678` login / password combination doesn't work, while `argilla` / `1234` from https://docs.argilla.io/en/latest/getting_started/installation/configurations/user_management.html#default-user page matches successfully

**Type of change**

- [x] Documentation update


frascuchon:
Thanks for the PR @near-akaia-root 

Indeed, the credentials provided in docs should work with the latest version of the argilla quickstart image. The template space is upgraded now (see [here](https://huggingface.co/spaces/argilla/argilla-template-space/blob/main/Dockerfile))

PR #2676, opened 2023-04-11
Refactor/typing settings properly

frascuchon:
Changes in this PR allow to expose setting classes to the API docs, instead of having a raw dictionary without the proper schema info.

### DIFF
--- a/src/argilla/server/schemas/v1/datasets.py
+++ b/src/argilla/server/schemas/v1/datasets.py
@@ -13,11 +13,11 @@
 #  limitations under the License.
 
 from datetime import datetime
-from typing import Any, Optional, Union
+from typing import Optional, Union
 from uuid import UUID
 
-from pydantic import BaseModel, conlist, validator
-from pydantic.utils import GetterDict
+from pydantic import BaseModel, Field, conlist
+from typing_extensions import Literal
 
 from argilla.server.models import AnnotationType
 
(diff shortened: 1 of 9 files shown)


src/argilla/server/schemas/v1/datasets.py
```diff
@@ -42,46 +42,57 @@ class DatasetCreate(BaseModel):
...
     name: str
     title: str
     type: AnnotationType
     required: bool
-    settings: dict
+    settings: Union[RatingAnnotationSettings, TextAnnotationSettings]
     inserted_at: datetime
     updated_at: datetime
 
+    _validate_settings = validator("settings", pre=True, always=True, allow_reuse=True)(_validate_settings)
```
jfcalvo: Correct me if I'm wrong but we are executing validations when serializing annotations. For example when we are listing a dataset annotations. We don't want that right?

frascuchon: I understand your point. Exactly, is not really a validation, but we still need to parse the settings dictionary into a valid settings instance.

Anyway, I think we can use the [`GetterDict`](https://docs.pydantic.dev/usage/models/#data-binding) mechanism that pydantic provides. I will push the changes.

Thanks for the insight

frascuchon: Done

jfcalvo (approved):

PR #2677, opened 2023-04-11
Add status to datasets and new publish endpoint

jfcalvo:
# Description

* Add a new `status` column to `datasets` table (non nullable and with a index).
* Add a new `PUT /api/v1/dataset/{dataset_id}/publish` endpoint that if possible will change the status of the dataset to `ready` to be annotated.
  * Only admins can publish a dataset.
  * Only datasets in `draft` status can be published (and changed to `ready`).
  * Only datasets with one or more annotations can be published.
* Add new `create_annotation` policy function allowing only to create annotations for a dataset if the user is admin and the dataset is in `draft` status.


frascuchon (approved):
Looks good to me.

Just one comment, maybe is useful to users, when an error occurs when trying to publish a dataset, contextualize the error with the reason:

- dataset has no annotations
- dataset already published

jfcalvo:
> Looks good to me.
> 
> Just one comment, maybe is useful to users, when an error occurs when trying to publish a dataset, contextualize the error with the reason:
> 
> * dataset has no annotations
> * dataset already published

Yes, that's really useful. Maybe we can move that logic from the policy to a Pydantic `DatasetPublish` schema with those validations.

frascuchon:
> > Looks good to me.
> > Just one comment, maybe is useful to users, when an error occurs when trying to publish a dataset, contextualize the error with the reason:
> > 
> > * dataset has no annotations
> > * dataset already published
> 
> Yes, that's really useful. Maybe we can move that logic from the policy to a Pydantic `DatasetPublish` schema with those validations.

I think we should separate between 403 errors and 422 errors. When users are not admin, we should return a 403 status, but when the dataset state is inconsistent to this operation, we should launch a 422 giving the status reason

jfcalvo:
> > > Looks good to me.
> > > Just one comment, maybe is useful to users, when an error occurs when trying to publish a dataset, contextualize the error with the reason:
> > > 
> > > * dataset has no annotations
> > > * dataset already published
> > 
> > 
> > Yes, that's really useful. Maybe we can move that logic from the policy to a Pydantic `DatasetPublish` schema with those validations.
> 
> I think we should separate between 403 errors and 422 errors. When users are not admin, we should return a 403 status, but when the dataset state is inconsistent to this operation, we should launch a 422 giving the status reason

Not really sure if that should be a 422 error, in any case we better return a specific detail about what is happening.

frascuchon:
> > > > Looks good to me.
> > > > Just one comment, maybe is useful to users, when an error occurs when trying to publish a dataset, contextualize the error with the reason:
> > > > 
> > > > * dataset has no annotations
> > > > * dataset already published
> > > 
> > > 
> > > Yes, that's really useful. Maybe we can move that logic from the policy to a Pydantic `DatasetPublish` schema with those validations.
> > 
> > 
> > I think we should separate between 403 errors and 422 errors. When users are not admin, we should return a 403 status, but when the dataset state is inconsistent to this operation, we should launch a 422 giving the status reason
> 
> Not really sure if that should be a 422 error, in any case we better return a specific detail about what is happening.

For me, it makes sense this error https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422, since the request is syntactically well-formed (400), even if this status is more related to processing the reques't content. In any case, the main idea should be separate access errors from validation errors, yes.

frascuchon (approved):

Issue #2615, opened 2023-03-29
Feedback Task Feature


jfcalvo:
This task will bring together all efforts related to the new feedback task feature.

We will create a new feature branch associated with this issue where we can merge all the code related to this task.

PR #2680, opened 2023-04-12
feat : front orm model instruction task

keithCuniah:
# Description
Vuex orm model for instruction task and change route for instruction task page (domain/datasetId/annotation-mode)

image

Please include a summary of the changes and the related issue. Please also include relevant motivation and context. List any dependencies that are required for this change.

Part of the main branch #2615 

**Type of change**

- [ ] Documentation update

**How Has This Been Tested**

- [x] Feedback task (no impact in current task)

**Checklist**


frontend/models/feedback-task-model/feedback-dataset/FeedbackDataset.model.js
```diff
@@ -0,0 +1,25 @@
...
+class FeedbackDataset extends Model {
+  static entity = "feedbackDatasets";
+
+  static fields() {
+    return {
+      id: this.uid(),
+      name: this.string(""),
+      guidelines: this.string(""),
+      workspace_id: this.attr(null),
+      workspace_name: this.attr(null),
```
frascuchon, line 14: Is possible to fetch the workspace through a relationship with the workspace entity?

frontend/models/feedback-task-model/feedback-dataset/FeedbackDataset.model.js
```diff
@@ -0,0 +1,25 @@
...
+      inserted_at: this.attr(null),
+      updated_at: this.attr(null),
+
+      // relationships
+      global_questions: this.hasMany(GlobalQuestion, "dataset_id"),
```
frascuchon: Regarding the naming, I think this could be renamed to `questions_settings`, and the entity to `QuestionSettings`. Please, correct me if I'm wrong.

frontend/models/feedback-task-model/global-question/GlobalQuestion.model.js
```diff
@@ -0,0 +1,22 @@
...
+  static entity = "globalQuestions";
+  static primaryKey = ["component_type", "question"];
+
+  static fields() {
+    return {
+      id: this.uid(),
+      dataset_id: this.attr(null),
+      order: this.number(0),
+      question: this.string(""),
+      outputs: this.attr([]),
```
frascuchon: What's the purpose of this field (`outputs`)?

frontend/models/feedback-task-model/feedback-dataset/FeedbackDataset.model.js
keithCuniah: Do you mean updating the orm like this ? 
image

frontend/models/feedback-task-model/feedback-dataset/FeedbackDataset.model.js
frascuchon: Or maybe just, `questions` and `Question`. Do you have a `LocalQuestion` equivalent model?

frontend/models/feedback-task-model/global-question/GlobalQuestion.model.js
```diff
@@ -0,0 +1,22 @@
+import { Model } from "@vuex-orm/core";
+
+class GlobalQuestion extends Model {
+  static entity = "globalQuestions";
+  static primaryKey = ["component_type", "question"];
+
+  static fields() {
```
frascuchon: Here I'm missing the question name. A value that will be used to link a record response value with the corresponding question. For instance, If we have a record containing the following responses:

```json
{ 
  "question_01": { "value": 3 },
  "question_02": { "value": "NEGATIVE" }
}
```

Then, that means the dataset contains at least 2 questions as follows:
```json
[
  { "id": "...", "name": "question_01", "question": "Rate the output from blablablab...", ... },
  { "id": "...", "name": "question_02", "question": "Classify based on following blablabla...", ... }
]
```

frontend/models/feedback-task-model/feedback-dataset/FeedbackDataset.model.js
keithCuniah: - In the point of view of the front app, I think it's a little confusing to use "settings" here, since we have differents component using this prefix too. 
- The "global" here refer to a model attach to the dataset and not in nested entity (like a global variable) but maybe there are better naming. if I put only "question" as the name of the entity, I feel a little confusing if an entity called "Question" have also attribut called "question"

=> I try to use as possible the JS convention to have two word in naming, so a compromise could be "dataset_questions" instead of "global_questions" ?

frontend/models/feedback-task-model/record-field/RecordField.model.js
```diff
@@ -0,0 +1,16 @@
+import { Model } from "@vuex-orm/core";
+
+class RecordField extends Model {
+  static entity = "recordFields";
+
+  static fields() {
+    return {
+      id: this.uid(),
+      title: this.string(null).nullable(),
+      text: this.string(null).nullable(),
```
frascuchon, line 10: I can imagine this is the name of the field, right? Maybe `name` is more specific than `text`. Please, correct me if I'm wrong in any case.

frontend/models/feedback-task-model/global-question/globalQuestion.queries.js
```diff
@@ -0,0 +1,26 @@
+import { GlobalQuestion as GlobalQuestionModel } from "../global-question/GlobalQuestion.model";
+
+// UPSERT
+const upsertGlobalQuestion = (globalQuestion) => {
+  GlobalQuestionModel.insertOrUpdate({ data: globalQuestion });
+};
```
leiyre: Why we have upsertGlobalQuestion and upsertGlobalQuestions?

frontend/models/feedback-task-model/global-question/GlobalQuestion.model.js
keithCuniah: The purpose of outputs It's to initiate the kind of answer(s) used by default in the components of the question form.
Each component (singleLabel, textArea, Rating, etc) have the same shape of outputs : I mean a list of object, but each object can have different attribute depending of the componentType.

frontend/models/feedback-task-model/record-response/RecordResponse.model.js
```diff
@@ -0,0 +1,17 @@
...
+class RecordResponse extends Model {
+  static entity = "recordResponses";
+  static primaryKey = ["question_id", "record_id"];
+
+  static fields() {
+    return {
+      id: this.uid(),
+      outputs: this.attr([]),
+      question_id: this.attr(null),
+      record_id: this.attr(null),
```
frascuchon, line 12: A response is given by a user. So, to support multi-user response, I think `user_id` is a required field here (even if will correspond to the connected user)

frontend/models/feedback-task-model/global-question/GlobalQuestion.model.js
keithCuniah: To do the connection between question/recordAnswer internally I use the id of the entity GlobalQuestion. This mean I can update the table with the 'name' in the field 'id'. in that case, it's may be better to remove the primary key and keep only the id.

frontend/models/feedback-task-model/global-question/globalQuestion.queries.js
keithCuniah: upsertGlobalQuestion => upsert one single question in the globalQuestion entity
upsertGlobalQuestions => upsert a table of questions in the globalQuestion entity

frontend/models/feedback-task-model/record-field/RecordField.model.js
keithCuniah: I am not sure it's better in the point of view of front app, since name need to keep generic for the "stupid" component. 
More specifically, this recordField entity is used in the RecordFeedback.component => this component is only to render bloc of title with a specific text. So to not have business naming in stupid component, I need to convert some of the naming in generic naming

frontend/models/feedback-task-model/record-response/RecordResponse.model.js
keithCuniah: Ok make sense. I will add the user_id

frontend/models/feedback-task-model/global-question/globalQuestion.queries.js
keithCuniah: oups. done.

frontend/models/feedback-task-model/record/Record.model.js
```diff
@@ -0,0 +1,30 @@
+import { Model } from "@vuex-orm/core";
+import { RecordField as RecordFieldModel } from "../record-field/RecordField.model";
+import { RecordResponse as RecordResponseModel } from "../record-response/RecordResponse.model";
+
+class Record extends Model {
+  static entity = "records";
+  static primaryKey = ["record_id", "dataset_id"];
```
frascuchon: for this new data model the, record id is quite enough to identify a single record. Composite primary key is not needed

frontend/models/feedback-task-model/record/Record.model.js
```diff
@@ -0,0 +1,30 @@
...
+
+  static fields() {
+    return {
+      id: this.uid(),
+      dataset_id: this.attr(null),
+      record_status: this.string(null).nullable(),
```
frascuchon, line 12: This is a local status for the web app, right?

frascuchon (commented):
This PR adds more things than the ORM model definition, right? Or there is another PR affecting the rest of the changes?

frontend/models/feedback-task-model/record/Record.model.js
keithCuniah: this is the status of the record regarding the interaction of the user with the corresponding formular

frontend/models/feedback-task-model/feedback-dataset/FeedbackDataset.model.js
frascuchon: > * The "global" here refer to a model attach to the dataset and not in nested entity (like a global variable) but maybe there are better naming. if I put only "question" as the name of the entity, I feel a little confusing if an entity called "Question" have also attribut called "question"
> 
For our model, we have an attribute called `title`, instead of `question`.  It's more general and can fit better if we have feedback components without an explicit question (for example a text area for `Notes`)
```json
{
    "id": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
    "name": "string",
    "title": "string",
    "required": true,
    "settings": {
      "type": "text"
    },
    "inserted_at": "2023-04-14T12:16:21.823Z",
    "updated_at": "2023-04-14T12:16:21.823Z"
  }
```
> => I try to use as possible the JS convention to have two word in naming, so a compromise could be "dataset_questions" instead of "global_questions" ?

Sounds good to me.

frontend/models/feedback-task-model/global-question/GlobalQuestion.model.js
keithCuniah: As discussed, I rename all "output" by "option" to remove confusion

frontend/models/feedback-task-model/global-question/GlobalQuestion.model.js
keithCuniah: I add the name attribute and remove the primary key.

leiyre (approved):

PR #2699, opened 2023-04-14
[Tests] Update tests using Rubrix datasets to use Argilla datasets

tomaarsen:
# Description

I went through all tests that use datasets from the `rubrix` namespace, and updated them to use `argilla` **if** an equivalently named Argilla dataset existed (that wasn't always the case). I don't have the HF token locally, so most of these tests were skipped locally, but I hope that they'll run in the CI.
Otherwise I'll try to get a token that works.

**Type of change**

**How Has This Been Tested**
Ran `pytest` locally, but most of these tests are under a `skipif`, so hopefully the CI has the token (I haven't looked into this yet).

**Checklist**

- ~I added relevant documentation~
- ~I made corresponding changes to the documentation~
- ~I have added tests that prove my fix is effective or that my feature works~
- ~I have added relevant notes to the CHANGELOG.md file (See https://keepachangelog.com/)~

### DIFF
--- a/tests/client/conftest.py
+++ b/tests/client/conftest.py
@@ -57,7 +57,6 @@ def gutenberg_spacy_ner(mocked_client):
     from datasets import load_dataset
 
     dataset = "gutenberg_spacy_ner"
-    # TODO(@frascuchon): Move dataset to new organization
     dataset_ds = load_dataset(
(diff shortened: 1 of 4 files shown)


frascuchon (approved):
thanks @tomaarsen

PR #2700, opened 2023-04-14
Create es index when a dataset published

frascuchon:
A basic integration with elasticsearch when a dataset is published. At this point, the elasticsearch index will be created based on dataset configuration.

A DB rollback is executed if some problem is found creating the index. If the error is raised from the commit statement, index creation should be reverted, or another mechanism to recreate the index should be provided through the API endpoint (and extra param with `force_recreate` or just `force`). But this won't be tackled in this PR.

Edit: Also, the elasticsearch client deps are reviewed. Now, the named 8x version is used. This allows us more control over the `es` client version to use, and it can suppress installation conflicts in an environment with a preview es client package installed.

### DIFF
--- a/src/argilla/server/elasticsearch.py
+++ b/src/argilla/server/elasticsearch.py
@@ -15,9 +15,10 @@
 import dataclasses
 from typing import Any, Dict
 
-from elasticsearch import Elasticsearch
+from elasticsearch8 import AsyncElasticsearch as AsyncElasticsearch8x
 
 from argilla.server.models import Annotation, AnnotationType, Dataset
+from argilla.server.settings import settings
 
 
 @dataclasses.dataclass
@@ -25,9 +26,9 @@ class ElasticSearchEngine:
     config: Dict[str, Any]
 
     def __post_init__(self):
-        self._client = Elasticsearch(**self.config)
+        self.client = AsyncElasticsearch8x(**self.config)
 
-    def create_index(self, dataset: Dataset):
+    async def create_index(self, dataset: Dataset):
         fields = {}
 
         for annotation in dataset.annotations:
@@ -41,7 +42,10 @@ class ElasticSearchEngine:
         }
 
         index_name = f"rg.{dataset.id}"
-        self._client.indices.create(index=index_name, body={"mappings": mappings})
+        await self.client.indices.create(index=index_name, mappings=mappings)
+
+    def close(self):
+        self.client.close()
 
     def _field_mapping_for_annotation(self, annotation_task: Annotation):
         settings_type = annotation_task.settings.get("type")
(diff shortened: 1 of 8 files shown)


src/argilla/server/elasticsearch.py
```diff
@@ -53,4 +54,19 @@ def _field_mapping_for_annotation(self, annotation_task: Annotation):
             # See https://www.elastic.co/guide/en/elasticsearch/reference/current/text.html
             return {"type": "text"}
         else:
-            raise ValueError(f"Annotation of type {settings_type} cannot be processed")
+            raise ValueError(f"ES mappings for Annotation of type {settings_type} cannot be generated")
+
+
+async def get_engine():
```
jfcalvo: Looks like we are using search engine everywhere. Maybe we should use `get_search_engine` instead:

```suggestion
async def get_search_engine():
```

tests/conftest.py
```diff
@@ -48,6 +51,29 @@ def db():
...
+    open_search = OpenSearch(hosts=settings.elasticsearch)
+
+    info = open_search.info(format="json")
+    version_info = info["version"]
+
+    return "distribution" not in version_info
+
+
+@pytest.fixture(scope="session")
+def es_config():
```
jfcalvo:
```suggestion
def elasticsearch_config():
```

tests/server/api/v1/test_datasets.py
```diff
@@ -1012,9 +1014,15 @@ def test_create_dataset_records_with_nonexistent_dataset_id(client: TestClient,
     assert db.query(Response).count() == 0
 
 
-def test_publish_dataset(client: TestClient, db: Session, admin_auth_header: dict):
+@pytest.mark.skipif(condition=not is_running_elasticsearch(), reason="Test only running with elasticsearch backend")
+def test_publish_dataset(
+    client: TestClient,
+    db: Session,
+    elasticsearch: Elasticsearch,
```
jfcalvo, line 1021: If we change this to `search_engine` on handlers maybe we should change it here too:

```suggestion
    search_engine: Elasticsearch,
```

Remember to change it for the rests of tests.

tests/server/api/v1/test_datasets.py
```diff
@@ -1024,6 +1032,26 @@ def test_publish_dataset(client: TestClient, db: Session, admin_auth_header: dic
...
+
+@pytest.mark.skipif(condition=not is_running_elasticsearch(), reason="Test only running with elasticsearch backend")
+def test_publish_dataset_with_error_on_index_creation(
+    client: TestClient,
+    db: Session,
+    elasticsearch: Elasticsearch,
+    admin_auth_header: dict,
+):
+    dataset = DatasetFactory.create()
+    AnnotationFactory.create(dataset=dataset)
```
jfcalvo:
```suggestion
    AnnotationFactory.create(settings={"type": "invalid"}, dataset=dataset)
```

jfcalvo (approved):

tests/server/api/v1/test_datasets.py
frascuchon: This is not the search engine, is just the official elasticsearch client used to check test results.

jfcalvo (approved):

PR #2702, opened 2023-04-17
Add records and responses

jfcalvo:
# Description

This PR adds several changes related with records and responses entities:
* Create `records` table and database model.
* Create `responses` table and database model.
* Add some new endpoints:
  * List records for a dataset with `GET` `/api/v1/datasets/{dataset_id}/records`.
    * Right now `admin` users will get all the responses for every record and `annotator` users the responses created by them. (This is probably gonna change in the future).
  * Create records for a dataset in bulk with `POST` `/api/v1/datasets/{dataset_id}/records`.
  * Create record responses with `POST` `/api/v1/records/{record_id}/responses`.

### DIFF
--- a/src/argilla/server/schemas/v1/datasets.py
+++ b/src/argilla/server/schemas/v1/datasets.py
@@ -81,11 +81,21 @@ class AnnotationCreate(BaseModel):
     settings: Union[TextAnnotationSettings, RatingAnnotationSettings] = Field(..., discriminator="type")
 
 
+class Response(BaseModel):
+    id: UUID
+    values: Dict[str, Any]
+    inserted_at: datetime
+    updated_at: datetime
+
+    class Config:
+        orm_mode = True
+
+
 class Record(BaseModel):
     id: UUID
     fields: Dict[str, Any]
     external_id: Optional[str]
-    dataset_id: UUID  # TODO: Maybe delete this field because we are returning records for a specific dataset (same that with Annotation schema)
+    responses: List[Response]
     inserted_at: datetime
     updated_at: datetime
 
--- a/src/argilla/server/schemas/v1/datasets.py
+++ b/src/argilla/server/schemas/v1/datasets.py
@@ -13,7 +13,7 @@
 #  limitations under the License.
 
 from datetime import datetime
-from typing import Optional, Union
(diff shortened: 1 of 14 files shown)


src/argilla/server/schemas/v1/records.py
```diff
@@ -12,39 +12,22 @@
...
-    predictions: Optional[Dict[str, Prediction]]
+    class Config:
+        orm_mode = True
 
-    metadata: Optional[Dict[str, Any]]
-    vectors: Optional[Dict[str, List[float]]]
 
-    validated_annotation: Optional[ValidatedAnnotation]
+class ResponseCreate(BaseModel):
+    values: Dict[str, Any]
```
frascuchon, line 33: This model is already defined in dataset's schema module
```suggestion
```

src/argilla/server/schemas/v1/datasets.py
```diff
@@ -79,3 +82,43 @@ class AnnotationCreate(BaseModel):
...
+class Record(BaseModel):
+    id: UUID
+    fields: Dict[str, Any]
+    external_id: Optional[str]
+    responses: List[Response]
+    inserted_at: datetime
+    updated_at: datetime
+
+    class Config:
+        orm_mode = True
```
frascuchon, line 106: Since we have a records schema module, do we should move this definition to that module?

tests/server/api/v1/test_datasets.py
```diff
@@ -230,14 +190,310 @@ def test_get_dataset_annotations_as_annotator_from_different_workspace(client: T
...
-def test_get_dataset_annotations_with_nonexistent_dataset_id(client: TestClient, db: Session, admin_auth_header: dict):
+def test_list_dataset_annotations_with_nonexistent_dataset_id(client: TestClient, db: Session, admin_auth_header: dict):
     DatasetFactory.create()
 
     response = client.get(f"/api/v1/datasets/{uuid4()}/annotations", headers=admin_auth_header)
 
     assert response.status_code == 404
 
 
+def test_list_dataset_records(client: TestClient, admin_auth_header: dict):
```
frascuchon, line 201: I would include a test listing records as an admin with annotator responses.

frascuchon (approved):

PR #2708, opened 2023-04-17
Remove responses from records

frascuchon:
This PR removes the responses from the record schemas. A new endpoint will be provided to retrieve the responses for a given record id.

### DIFF
--- a/src/argilla/server/contexts/datasets.py
+++ b/src/argilla/server/contexts/datasets.py
@@ -11,7 +11,6 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
-from typing import List, Optional
 from uuid import UUID
 
 from argilla.server.models import Annotation, Dataset, DatasetStatus, Record, Response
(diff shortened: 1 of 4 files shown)


src/argilla/server/contexts/datasets.py
```diff
@@ -120,20 +120,6 @@ def get_record_by_id(db: Session, record_id: UUID):
 def list_records(db: Session, dataset: Dataset, offset: int = 0, limit: int = LIST_RECORDS_LIMIT):
     return (
         db.query(Record)
-        .options(joinedload(Record.responses))
```
jfcalvo, line 123: I believe you can remove the imported `joinedload`, `_and`,  `constains_eager` if they are not used in the rest of the code.

tests/server/api/v1/test_datasets.py
```diff
@@ -200,32 +200,13 @@ def test_list_dataset_annotations_with_nonexistent_dataset_id(client: TestClient
...
-        record=record_b,
-    )
-    response_b_2 = ResponseFactory.create(
-        values={
-            "input_ok": {"value": "no"},
-            "output_ok": {"value": "no"},
-        },
-        record=record_b,
-    )
+    ResponseFactory.create(values={"input_ok": {"value": "yes"}, "output_ok": {"value": "yes"}}, record=record_a)
```
jfcalvo: I believe we should remove the created responses. They are not necessary. (here and in the rests of affected tests).

jfcalvo (changes requested):

jfcalvo (approved):

PR #2711, opened 2023-04-18
Add new APIv1 update response endpoint

jfcalvo:
# Description

This PR adds a new endpoint to update a previously created record response:
* `PUT` `/api/v1/responses/{response_id}`

Notes:
* Admin users can update any response (even if that admin user didn't create the response in first place).
* Annotators can only update a response if the response was created by them.

### DIFF
--- a/src/argilla/server/schemas/v1/responses.py
+++ b/src/argilla/server/schemas/v1/responses.py
@@ -22,6 +22,8 @@ from pydantic import BaseModel
 class Response(BaseModel):
     id: UUID
     values: Dict[str, Any]
+    record_id: UUID
+    user_id: UUID
     inserted_at: datetime
     updated_at: datetime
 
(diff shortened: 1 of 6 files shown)


src/argilla/server/schemas/v1/responses.py
```diff
@@ -0,0 +1,33 @@
...
+from uuid import UUID
+
+from pydantic import BaseModel
+
+
+class Response(BaseModel):
+    id: UUID
+    values: Dict[str, Any]
+    inserted_at: datetime
+    updated_at: datetime
```
frascuchon, line 28: Aligning the entity definition, I think we should add the user_id in the schema.

frascuchon: And the `record_id` for these endpoints.

jfcalvo: Yes, it makes sense.

jfcalvo: Done.

frascuchon (approved):

PR #2714, opened 2023-04-18
Add attribute total counting number of records to dataset records APIv1 endpoint response

jfcalvo:
# Description

This PR add a new `total` value the listing records for datasets on APIv1 endpoint.

With this change we are allowing classic pagination on the frontend so the pages can be calculated.

Notes:
* I'm adding `RecordFactory.create_batch(size=2)` to some tests so we can be sure that we have records for other datasets apart from the one that we are requesting.

### DIFF
--- a/tests/server/api/v1/test_datasets.py
+++ b/tests/server/api/v1/test_datasets.py
@@ -42,7 +42,6 @@ from tests.factories import (
     DatasetFactory,
     RatingAnnotationFactory,
     RecordFactory,
-    ResponseFactory,
     TextAnnotationFactory,
     WorkspaceFactory,
 )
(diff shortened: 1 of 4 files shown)


tests/server/api/v1/test_datasets.py
```diff
@@ -206,6 +205,8 @@ def test_list_dataset_records(client: TestClient, admin_auth_header: dict):
     record_b = RecordFactory.create(fields={"record_b": "value_b"}, dataset=dataset)
     record_c = RecordFactory.create(fields={"record_c": "value_c"}, dataset=dataset)
 
+    RecordFactory.create_batch(size=2)
```
frascuchon:
```suggestion
    RecordFactory.create_batch(size=2, dataset=other_dataset)
```

jfcalvo: Done for all the examples.

frascuchon (approved):

PR #2716, opened 2023-04-18
Add items to all APIv1 list endpoints

jfcalvo:
# Description

With this PR we have all the APIv1 list endpoints using the same response structure with `{"items": []}`.

We are of course only adding `total` when necessary but it's something that we can add later if necessary.


frascuchon:
Could we add here changes to the affected part in the frontend @jfcalvo?

jfcalvo:
> Could we add here changes to the affected part in the frontend @jfcalvo?

Yes, it's something that we can do. @keithCuniah @leiyre

jfcalvo:
@frascuchon I gonna to self-approve my PR because we want to move the changes to DEV. Please take a look and if you think something needs to be changed I will do it in a new branch.

PR #2718, opened 2023-04-19
Add regex constraint for annotation name

jfcalvo:
# Description

Annotation names will be used on search queries and written into our search engine. For this reason we need to add a constraint to `name` attribute for annotations so only some patterns are allowed.

I have added the following constraints for annotation `name` attribute when they are created:
* The minimum number of characters is `1`.
* The maximum number of characters is `200`.
* It needs to match a regular expression with the following rules:
  * Can only contains lowercase letters, digits, hyphens or underscores.
  * It must contains at least one lowercase letter or digit.


tomaarsen (approved):
It is my understanding that the regex matches your textual description, so that looks good.

PR #2751, opened 2023-04-21
Add dataset fields

### LINKED ISSUE #2750: Add dataset fields
This issue should include the following changes:

* Add a new database migration creating `fields` table.
* Add a SQLAlchemy `Field` model.
* Add a new `FieldFactory` class so we can use it for testing.
* Add a first endpoint to list fields for a specific dataset.

jfcalvo:
# Description

* Add a new database migration creating `fields` table.
* Add a SQLAlchemy `Field` model.
* Add a first endpoint to list fields for a specific dataset `GET /api/v1/datasets/{dataset_id}/fields`

Notes:
* I have finally used `fields` instead of `field_settings` because the naming started to feel strange when creating things like Pydantic schemas. I'm open to suggestions for changing it.
* I didn't add an `Union` for `settings` on `Field` schema. It's because right now it only has one type that is `text`. We can add the `Union` and discriminator once that we have more than one type of field.

Closes #2750

### DIFF
--- a/src/argilla/server/alembic/versions/ae5522b4c674_create_fields_table.py
+++ b/src/argilla/server/alembic/versions/ae5522b4c674_create_fields_table.py
@@ -1,0 +1,49 @@
+#  Copyright 2021-present, the Recognai S.L. team.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


frascuchon (approved):
It looks good to me. We will have very soon support to other kind of field, as ImageUrl, for example. But it's fine to keep focused on Text fields for now.

PR #2756, opened 2023-04-23
feat: add v1 enpoint to set `Response` status

### LINKED ISSUE #2733: Allow discarding/submitting responses for a record
## Description 

We need to add an APIv1 feature allowing users to discard or submit (confirm) a response for a single record.  This response can be discarded by admin users or the user which created the response previously. 

After that action, the status of the response will become `Discarded` / `Submitted` depending on the action. 

This issue can be tackled in two separate tasks: 
- Submit responses and
- Discard responses

You could tackle one of them and is not necessary to finish the whole feat.

# Notes

Feel free to propose the endpoint definition that you consider better for this. Also, feel free to modify the database schema/table to support this new feature.

gabrielmbmb:
# Description

I've added a new endpoint `PUT /api/v1/responses/{response_id}/status` that allows setting the status of a `Response`.
 
To do so, I've added a new column to the `Response` model/table called `status`, which right now has 3 possible values: `pending` (which is the default value), `submitted`, and `discarded`. I've also updated the `alembic` migrations files to include this new column, which is not `nullable` (it will have a default value always), and also an `index` will be created for it, as I suppose that we would like to filter by it and having an index will allow querying using a filter on this column faster.

Regarding the endpoint definition, I've decided to just add a new `PUT` endpoint to update the `status` of the `Response` instead of using the `PUT /v1/response/{response_id}` endpoint, because I think:

1. It makes the API's intent more explicit and easier to understand.
2. It has better separation of concerns as we can more easily optimize and scale this specific operation independently from the update response operation. Also, if in the future is needed to do something after status has been set, it will be easier this way than having an if condition in the update response endpoint that checks if the status has been updated, and if so then do something specific for when the status has changed.

As an alternative to this new endpoint, we could have:

1. Created two endpoints (`PUT /api/v1/responses/{response_id}/submit` and `PUT /api/v1/responses/{response_id}/discard`), one for submitting and one for discarding, but I think this is not a good idea, because if a new possible status is introduced, then a new endpoint has to be created and maintained. The logic of both endpoints will be exactly the same, so we would have two endpoints almost identical. Also, this approach would require the client to send a PUT request to a different endpoint depending on the status, and this can make the API harder to use.
2.  Create one endpoint very similar to the one implemented but passing the status via URL path `PUT /api/v1/responses/{response_id}/status/{status}`. This approach has the inconvenience that the client would have to send a PUT request to a different endpoint each time the status has to be changed.

Closes #2733 

**Type of change**

- [ ] Documentation update

**How Has This Been Tested**

- I've added unit tests for all the possible scenarios when calling this endpoint.

**Checklist**

### DIFF
--- a/src/argilla/server/alembic/versions/e402e9d9245e_create_responses_table.py
+++ b/src/argilla/server/alembic/versions/e402e9d9245e_create_responses_table.py
@@ -36,6 +36,7 @@ def upgrade() -> None:
         sa.Column("values", sa.JSON, nullable=False),
         sa.Column("record_id", sa.Uuid, sa.ForeignKey("records.id", ondelete="CASCADE"), nullable=False, index=True),
         sa.Column("user_id", sa.Uuid, sa.ForeignKey("users.id", ondelete="SET NULL"), index=True),
+        sa.Column("status", sa.String, nullable=False, index=True),
         sa.Column("inserted_at", sa.DateTime, nullable=False),
         sa.Column("updated_at", sa.DateTime, nullable=False),
         sa.UniqueConstraint("record_id", "user_id", name="response_record_id_user_id_uq"),
--- a/src/argilla/server/models.py
+++ b/src/argilla/server/models.py
@@ -49,6 +49,12 @@ class UserRole(str, Enum):
     annotator = "annotator"
 
 
+class ResponseStatus(str, Enum):
+    pending = "pending"
+    submitted = "submitted"
+    discarded = "discarded"
+
+
 class Annotation(Base):
     __tablename__ = "annotations"
 
@@ -79,6 +85,7 @@ class Response(Base):
     values: Mapped[dict] = mapped_column(JSON)
     record_id: Mapped[UUID] = mapped_column(ForeignKey("records.id"))
     user_id: Mapped[UUID] = mapped_column(ForeignKey("users.id"))
+    status: Mapped[ResponseStatus] = mapped_column(default=ResponseStatus.pending, index=True, nullable=False)
 
     inserted_at: Mapped[datetime] = mapped_column(default=datetime.utcnow)
     updated_at: Mapped[datetime] = mapped_column(default=default_inserted_at, onupdate=datetime.utcnow)
@@ -89,7 +96,7 @@ class Response(Base):
     def __repr__(self):
         return (
             f"Response(id={str(self.id)!r}, record_id={str(self.record_id)!r}, user_id={str(self.user_id)!r}, "
-            f"inserted_at={str(self.inserted_at)!r}, updated_at={str(self.updated_at)!r})"
+            f"status={self.status!r}, inserted_at={str(self.inserted_at)!r}, updated_at={str(self.updated_at)!r})"
         )
 
 
--- a/src/argilla/server/models.py
+++ b/src/argilla/server/models.py
@@ -123,7 +123,7 @@ class Response(Base):
     def __repr__(self):
         return (
             f"Response(id={str(self.id)!r}, record_id={str(self.record_id)!r}, user_id={str(self.user_id)!r}, "
-            f"status={self.status!r}, inserted_at={str(self.inserted_at)!r}, updated_at={str(self.updated_at)!r})"
+            f"status={self.status}, inserted_at={str(self.inserted_at)!r}, updated_at={str(self.updated_at)!r})"
         )
 
 
--- a/src/argilla/server/schemas/v1/responses.py
+++ b/src/argilla/server/schemas/v1/responses.py
@@ -18,12 +18,15 @@ from uuid import UUID
 
 from pydantic import BaseModel
 
+from argilla.server.models import ResponseStatus
+
 
 class Response(BaseModel):
     id: UUID
     values: Dict[str, Any]
     record_id: UUID
     user_id: UUID
+    status: ResponseStatus
     inserted_at: datetime
     updated_at: datetime
 
@@ -33,3 +36,7 @@ class Response(BaseModel):
 
 class ResponseUpdate(BaseModel):
     values: Dict[str, Any]
+
+
+class ResponseStatusUpdate(BaseModel):
+    status: ResponseStatus
--- a/tests/server/api/v1/test_responses.py
+++ b/tests/server/api/v1/test_responses.py
@@ -225,6 +225,16 @@ def test_update_response_status_with_nonexistent_response_id(client: TestClient,
     assert resp.status_code == 404
 
 
+def test_update_response_status_with_invalid_status(client: TestClient, db: Session, admin_auth_header: dict):
+    response = ResponseFactory.create()
+    response_json = {"status": "invalid"}
+
+    resp = client.put(f"/api/v1/responses/{response.id}/status", headers=admin_auth_header, json=response_json)
+
+    assert resp.status_code == 422
+    assert db.get(Response, response.id).status == "pending"
+
+
 def test_delete_response(client: TestClient, db: Session, admin_auth_header: dict):
     response = ResponseFactory.create()
 
--- a/tests/server/api/v1/test_responses.py
+++ b/tests/server/api/v1/test_responses.py
@@ -16,7 +16,7 @@ from datetime import datetime
 from uuid import uuid4
 
 from argilla._constants import API_KEY_HEADER_NAME
-from argilla.server.models import Response
+from argilla.server.models import Response, ResponseStatus
 from fastapi.testclient import TestClient
 from sqlalchemy.orm import Session
 
@@ -54,6 +54,7 @@ def test_update_response(client: TestClient, db: Session, admin_auth_header: dic
         },
         "record_id": str(response.record_id),
         "user_id": str(response.user_id),
+        "status": ResponseStatus.pending,
         "inserted_at": response.inserted_at.isoformat(),
         "updated_at": datetime.fromisoformat(resp_body["updated_at"]).isoformat(),
     }
@@ -117,6 +118,7 @@ def test_update_response_as_annotator(client: TestClient, db: Session):
         },
         "record_id": str(response.record_id),
         "user_id": str(response.user_id),
+        "status": ResponseStatus.pending,
         "inserted_at": response.inserted_at.isoformat(),
         "updated_at": datetime.fromisoformat(resp_body["updated_at"]).isoformat(),
     }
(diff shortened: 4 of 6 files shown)


gabrielmbmb:
A review @frascuchon and @jfcalvo would be much appreciated! I would like to also hear your opinion. Also, I've some questions:

- Do we need to save the id of the user that set the status for a `Response`? Right now only the user that created the response or an admin can do it, so I guess is not that important.
- As only the user that created the response and the admin can set the status for a `Response`, I guess that the `ResponsePolicyV1.update` is enough, but in the future if users can set the status for a response that has not been created by them, then I guess that a new policy would be needed.

jfcalvo:
Hi @gabrielmbmb, thanks for the PR. I will take a look to it.

Regarding your questions:

> Do we need to save the id of the user that set the status for a Response? Right now only the user that created the response or an admin can do it, so I guess is not that important.

Although is not within the immediate scope of the issue, if you have the time and desire to work on it, you're welcome to do so.

> As only the user that created the response and the admin can set the status for a `Response`, I guess that the `ResponsePolicyV1.update` is enough, but in the future if users can set the status for a response that has not been created by them, then I guess that a new policy would be needed.

I think it's correct to use `ResponsePolicyV1.update` for now.

src/argilla/server/alembic/versions/e402e9d9245e_create_responses_table.py
```diff
@@ -36,6 +36,7 @@ def upgrade() -> None:
         sa.Column("values", sa.JSON, nullable=False),
         sa.Column("record_id", sa.Uuid, sa.ForeignKey("records.id", ondelete="CASCADE"), nullable=False, index=True),
         sa.Column("user_id", sa.Uuid, sa.ForeignKey("users.id", ondelete="SET NULL"), index=True),
+        sa.Column("status", sa.String, nullable=False, index=True),
```
jfcalvo, line 39: Can you explain your rationale for modifying this database migration instead of opting to create a new one?

src/argilla/server/models.py
```diff
@@ -116,7 +123,7 @@ class Response(Base):
     def __repr__(self):
         return (
             f"Response(id={str(self.id)!r}, record_id={str(self.record_id)!r}, user_id={str(self.user_id)!r}, "
-            f"inserted_at={str(self.inserted_at)!r}, updated_at={str(self.updated_at)!r})"
+            f"status={self.status!r}, inserted_at={str(self.inserted_at)!r}, updated_at={str(self.updated_at)!r})"
```
jfcalvo: By using `self.status`, you will get the Enum object itself, which may not be the desired representation. Accessing `self.status.value` will return the corresponding value of the Enum member as a string.

```suggestion
            f"status={self.status.value!r}, inserted_at={str(self.inserted_at)!r}, updated_at={str(self.updated_at)!r})"
```

tests/server/api/v1/test_responses.py
jfcalvo, line 1: I miss a new test case to check the behavior when the response status is invalid, such as attempting to update the status to an unsupported value like `invalid`.

tests/server/api/v1/test_responses.py
```diff
@@ -171,6 +173,58 @@ def test_update_response_with_nonexistent_response_id(client: TestClient, db: Se
...
+def test_update_response_status(client: TestClient, db: Session, admin_auth_header: dict):
+    response = ResponseFactory.create()
+    response_json = {"status": "submitted"}
+
+    resp = client.put(f"/api/v1/responses/{response.id}/status", headers=admin_auth_header, json=response_json)
+
+    assert resp.status_code == 200
+    assert db.get(Response, response.id).status == "submitted"
+
+
```
jfcalvo: It would be nice to check the request body response in the same way as `test_update_response`.

src/argilla/server/schemas/v1/responses.py
```diff
@@ -18,12 +18,15 @@
...
 
+from argilla.server.models import ResponseStatus
+
 
 class Response(BaseModel):
     id: UUID
     values: Dict[str, Any]
     record_id: UUID
     user_id: UUID
+    status: ResponseStatus
```
jfcalvo, line 29: In `src/argilla/server/schemas/v1/datasets.py`, there exists a distinct `Response` schema. Would you suggest including the `status` field in this schema as well?

jfcalvo (changes requested):
I have added some questions that you can answer at your convenience and also included some suggestions to help improve the solution. Please feel free to respond to the questions whenever you have time.

src/argilla/server/alembic/versions/e402e9d9245e_create_responses_table.py
gabrielmbmb: So in the current latest version `1.6.0`, only three tables have been created and managed by alembic `workspaces`, `users`, and `workspaces_users`.

In the branch `feat/2615-instructions-task`, migrations files have been created for some other tables. I've opted to modify the migration file instead of generating a new one because I've assumed that as these changes have not been used yet in any production environment, it is better to modify the migration file directly to add this new column, instead of creating a new migration file for just adding one column. One more file, but a little bit more noise.

Of course, if the migrations files in `feat/2615-instruction-task` were already applied in any production DB, then the logical thing to do would have been to generate a new migration script to add this new column, so alembic would have detected that there is a revision that has not been applied and that it needs to apply to the DB.

Having that said, for me it was not super clear how are you planning to handle the migrations with `alembic`. Maybe it's a good idea to have one single alembic revision/script to update the DB per the released version of the package.

Also, I've realized that `alembic revision --autogenerate` command cannot be used to generate [migrations automatically](https://alembic.sqlalchemy.org/en/latest/autogenerate.html) because https://github.com/argilla-io/argilla/blob/5e3ecd7d0e8a99b7be8d4412d9f3f94410afce43/src/argilla/server/alembic/env.py#L37 the `target_metadata` has not been set to https://github.com/argilla-io/argilla/blob/5e3ecd7d0e8a99b7be8d4412d9f3f94410afce43/src/argilla/server/database.py#L51 `Base.metadata`.

Another doubt that I have, is that I've seen that the attributes `nullable`, `index`, etc are being set in the migrations files, but not in the `mapped_column` function of `sqlalchemy`.

src/argilla/server/models.py
gabrielmbmb: Yes, I saw that, but I thought that it would be a good idea to see where the value is coming from. Updating this!

tests/server/api/v1/test_responses.py
gabrielmbmb: True! Missed that case

src/argilla/server/models.py
gabrielmbmb: Also, I've run into an issue with the `Enum` when using Python 3.11, because there is a [breaking change in Python 3.11](https://github.com/python/cpython/issues/100458) when using `Enum`s like `MyEnum(str, Enum): ...`

This is a problem because the endpoint generated here https://github.com/argilla-io/argilla/blob/5e3ecd7d0e8a99b7be8d4412d9f3f94410afce43/src/argilla/server/apis/v0/handlers/token_classification.py#L61 using Python 3.11 is `/api/datasets/{dataset_id}/TaskType.token_classification` instead of `/api/datasets/{dataset_id}/TokenClassification` and when calling the API from the client, for example to log some records in the server with `rg.log`, you will get a 404.

If supporting Python 3.11 is required then maybe this could be done:

```python
import sys
from enum import Enum

if sys.version_info >= (3, 11):
    from enum import StrEnum
else:
    class StrEnum(str, Enum):
        pass

class MyEnum(StrEnum):
    FOO = "foo"
    BAR = "bar"

print(f"{MyEnum.FOO}")
```

tests/server/api/v1/test_responses.py
gabrielmbmb: True, the `status` value could have been updated in the DB, but not returned in the endpoint.

src/argilla/server/alembic/versions/e402e9d9245e_create_responses_table.py
jfcalvo: > Also, I've realized that alembic revision --autogenerate command cannot be used to generate [migrations automatically](https://alembic.sqlalchemy.org/en/latest/autogenerate.html)

In your opinion, what are the benefits of utilizing automatic migrations?

> Another doubt that I have, is that I've seen that the attributes nullable, index, etc are being set in the migrations files, but not in the mapped_column function of sqlalchemy.

In the same vein as the previous question, what are your views on the advantages of including these attributes in models?

src/argilla/server/models.py
jfcalvo: I understand. We are currently using Python 3.7, so we may need to revisit this before upgrading to version 3.11.

src/argilla/server/alembic/versions/e402e9d9245e_create_responses_table.py
gabrielmbmb: I think that I can answer both questions at the same time 😁

So if we include these attributes in the models, then `alembic` will be available to automatically pick up them and it will generate a statement in the migration script with all the columns constraints, index, etc.

I think that it's really beneficial to use `alembic` automatic migrations, because `alembic` is able to automatically detect the updates you have done in your model class, or in other words, it is able to generate a `diff` between the schema generated with all the previous revisions and the one you're about to generate, and it will only create statements for the things that need to be updated.

So in my opinion, it is better to let `alembic` generate the migration file automatically with the `SQLAlchemy` model metadata than to let a human generate it manually and increase the chances of forgetting to include something in the migration script.

src/argilla/server/schemas/v1/responses.py
gabrielmbmb: Oh, I didn't realize there was this `Response` schema. Yes, we should include the `status` field here too, otherwise when calling the listing records `/api/v1/datasets/{dataset_id}/records` the status for each response won't be there, and the frontend would have to send an additional request to get the status for getting the responses status, which is not ideal.

I think it makes sense to have a `ResponseBase` class in `src/argilla/server/schemas/v1/responses.py`, this way we avoid code duplication. 

- We can create a `Response`class inheriting from `ResponseBase` class with additional attributes that have to be exposed when listing responses (like the `record_id`) in `src/argilla/server/schemas/v1/responses.py`.
- We can create a `Response` class inheriting from `ResponseBase` with no additional attributes that can be used when listing records in `src/argilla/server/schemas/v1/datasets.py`.

gabrielmbmb:
hey @jfcalvo, I've changed the status of the PR to ready for review. Is there anything else missing that I should add?

jfcalvo:
> hey @jfcalvo, I've changed the status of the PR to ready for review. Is there anything else missing that I should add?

Thank you, @gabrielmbmb, for your answers. The PR looks fine, and @frascuchon should be reaching out to you soon for the next steps.

PR #2770, opened 2023-04-24
Add a new APIv1 endpoint to create dataset fields

### LINKED ISSUE #2768: Allow creation of dataset fields via API
Create a new APIv1 endpoint to create dataset fields.

`POST /api/v1/datasets/{dataset_id}/fields`

jfcalvo:
# Description

This PR add changes to allow the creation of fields for a dataset. These fields will modelate how are the valid content of record fields accepted for this dataset.

The endpoint is the following:
* `POST` `/api/v1/datasets/{dataset_id}/fields`

Notes:
* I didn't make any change related with ElasticSearch indexing. It's something that we need to do in a different issue.

Closes #2768

### DIFF
--- a/tests/server/api/v1/test_datasets.py
+++ b/tests/server/api/v1/test_datasets.py
@@ -21,12 +21,14 @@ from argilla.server.models import (
     Annotation,
     Dataset,
     DatasetStatus,
+    Field,
     Record,
     Response,
     User,
 )
 from argilla.server.schemas.v1.datasets import (
     ANNOTATION_CREATE_NAME_MAX_LENGTH,
+    FIELD_CREATE_NAME_MAX_LENGTH,
     RATING_OPTIONS_MAX_ITEMS,
     RATING_OPTIONS_MIN_ITEMS,
     RECORDS_CREATE_MAX_ITEMS,
@@ -42,6 +44,7 @@ from tests.factories import (
     AnnotationFactory,
     AnnotatorFactory,
     DatasetFactory,
+    FieldFactory,
     RatingAnnotationFactory,
     RecordFactory,
     ResponseFactory,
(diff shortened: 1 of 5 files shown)


tests/server/api/v1/test_datasets.py
```diff
@@ -749,6 +752,165 @@ def test_create_dataset_with_existent_name(client: TestClient, db: Session, admi
...
+    dataset = DatasetFactory.create()
+    field_json = {
+        "name": invalid_name,
+        "title": "title",
+        "settings": {"type": "text"},
+    }
+
+    response = client.post(f"/api/v1/datasets/{dataset.id}/fields", headers=admin_auth_header, json=field_json)
+
+    assert response.status_code == 422
```
frascuchon, line 827: If the response content contains a more descriptive error, maybe makes sense to add a check of the expected info.

tests/server/api/v1/test_datasets.py
```diff
@@ -749,6 +752,165 @@ def test_create_dataset_with_existent_name(client: TestClient, db: Session, admi
...
+    dataset = DatasetFactory.create()
+    field_json = {
+        "name": "a" * (FIELD_CREATE_NAME_MAX_LENGTH + 1),
+        "title": "title",
+        "settings": {"type": "text"},
+    }
+
+    response = client.post(f"/api/v1/datasets/{dataset.id}/fields", headers=admin_auth_header, json=field_json)
+
+    assert response.status_code == 422
```
frascuchon, line 841: ditto

frascuchon (approved):

tests/server/api/v1/test_datasets.py
jfcalvo: It's not custom, it's the default Pydantic error message. We can check it but we have not been checking it for the rest of fields.

PR #2773, opened 2023-04-24
Add a new APIv1 endpoint to delete dataset fields

### LINKED ISSUE #2771: Allow deletion of dataset fields via API
We should allow deletion of dataset fields before the dataset has been published. With this users can easily delete dataset fields with errors.

jfcalvo:
# Description

This PR add changes to allow the deletion of fields specifying the id of a field resource.

Notes:
* I have modified annotations deletion endpoint to return an `Annotation` schema including the `dataset_id` as we are doing with `Field`. 

Closes #2771 

**WARNING:** You should review and merge PR #2770 before review this one.

### DIFF
--- a/src/argilla/server/schemas/v1/annotations.py
+++ b/src/argilla/server/schemas/v1/annotations.py
@@ -13,9 +13,26 @@
 #  limitations under the License.
 
 from datetime import datetime
+from typing import Union
 from uuid import UUID
 
-from pydantic import BaseModel
+from pydantic import BaseModel, Field, conlist
+from typing_extensions import Literal
+
+from argilla.server.models import AnnotationType
+
+
+class TextAnnotationSettings(BaseModel):
+    type: Literal[AnnotationType.text]
+
+
+class RatingAnnotationSettingsOption(BaseModel):
+    value: int
+
+
+class RatingAnnotationSettings(BaseModel):
+    type: Literal[AnnotationType.rating]
+    options: conlist(item_type=RatingAnnotationSettingsOption)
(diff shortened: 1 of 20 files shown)


src/argilla/server/schemas/v1/annotations.py
```diff
@@ -0,0 +1,32 @@
...
+
+from pydantic import BaseModel
+
+
+class Annotation(BaseModel):
+    id: UUID
+    name: str
+    title: str
+    required: bool
+    settings: dict
```
frascuchon: Should not annotation settings be a compound attribute of available question settings that we have already defined?

src/argilla/server/schemas/v1/fields.py
```diff
@@ -0,0 +1,32 @@
...
+
+from pydantic import BaseModel
+
+
+class Field(BaseModel):
+    id: UUID
+    name: str
+    title: str
+    required: bool
+    settings: dict
```
frascuchon: `TextFieldSettings` instead? Regarding generated from pydantic schema, makes sense to provide this info to API consumers.

src/argilla/server/schemas/v1/annotations.py
jfcalvo: Done.

src/argilla/server/schemas/v1/fields.py
jfcalvo: Done.

frascuchon (approved):

Issue #2763, opened 2023-04-23
Show the annotation progress for the user


frascuchon:
The user can access her annotation progress and see:

* % of annotated records
* % of remaining records
* % of discarded records

Issues:
- [x] Frontend: #2822
- [x] Backend: https://github.com/argilla-io/argilla/pull/2811

Refs #2761

PR #2811, opened 2023-04-28
Add new endpoint to get current user dataset metrics

jfcalvo:
# Description

This endpoint adds a new endpoint that will return the metrics for the current user and for an specific dataset.

The new endpoint will be the first one for APIv1 where we are using the prefix `/me` to notice that we are returning information in the context of the current user doing the request.

The returned schema will be something like the following:
```json
{
        "records": {
            "count": 5,
        },
        "responses": {
            "count": 3,
            "submitted": 2,
            "discarded": 1,
        },
}
```

Associated issue: #2763

### DIFF
--- a/src/argilla/server/apis/v1/handlers/datasets.py
+++ b/src/argilla/server/apis/v1/handlers/datasets.py
@@ -29,6 +29,7 @@ from argilla.server.schemas.v1.datasets import (
     Field,
     FieldCreate,
     Fields,
+    Metrics,
     Question,
     QuestionCreate,
     Questions,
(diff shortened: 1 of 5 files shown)


src/argilla/server/apis/v1/handlers/datasets.py
```diff
@@ -140,6 +141,29 @@ def get_dataset(
...
+    dataset = _get_dataset(db, dataset_id)
+
+    authorize(current_user, DatasetPolicyV1.get(dataset))
+
+    return {
+        "records": {
+            "count": datasets.count_records_by_dataset_id(db, dataset_id),
+        },
+        "responses": {
+            "count": datasets.count_responses_by_dataset_id_and_user_id(db, dataset_id, current_user.id),
```
frascuchon, line 160: Do we really need this count? Normally, discarded + submitted = count. Once we introduce a new status, we can think about if we need a new `new_status` count or not.

jfcalvo: Yes we can avoid that count right now but I prefer to be explicit about it and don't involve status logic.

frascuchon (approved):

PR #2827, opened 2023-05-03
Add new `FeedbackDataset` class using `/api/v1` endpoints

### LINKED ISSUE #2758: Allow `FeedbackDataset` creation via Python
The Python client must support this new task by creating a new set of functions to create, configure and publish a new dataset for this task.

[Here](https://gist.github.com/frascuchon/2e11db8a00559746e4b4c2dff687daaf) a code snippet showing the flow. Please, take it as an example. The final solution does not have to be this way.

### LINKED ISSUE #2764: Load a dataset for the new `FeedbackTask`
Users can load the dataset using the Python client. The current path to do this is using the `rg.load` method, but could make sense (for technical or other implications) to use a different way to do it.

alvarobartt:
## Description

This PR addresses the creation of a new `FeedbackDataset` class to handle the new datasets provided in Argilla. These datasets define some fields and questions for every dataset. The fields are the inputs, while the questions are the "annotations" which in this case correspond to questions over the fields that can either be numerical or textual, so that the annotator has to answer those questions in order to label the data.

This PR closes #2758, and #2764

**Type of change**

**How Has This Been Tested**

- [X] Included unit tests for `/api/v1/datasets` endpoint wrappers
- [X] Included unit tests for `rg.FeedbackDataset` class and any other related method/function

**Checklist**

## Example

```python
import argilla as rg

rg.init(
    api_url="<ARGILLA_URL_GOES_HERE>",
    api_key="<ARGILLA_API_KEY_GOES_HERE>",
)

dataset = rg.FeedbackDataset(
  guidelines="...",
  fields=[
    rg.TextField(name="context", required=False),
    rg.TextField(name="completion", required=True),
  ],
  questions=[
    rg.RatingQuestion(name="prompt-quality", title="Assess the quality of the prompt", required=True),
    rg.TextQuestion(name="prompt-edit", title="Edit the prompt if applicable", required=False),
  ]
)

dataset.add_records(records=[
  rg.FeedbackRecord(
    fields={"prompt": ..., "context": ..., "completion": ...}
  ),
])
print(dataset.records[0])
dataset.push_to_argilla(name="my-dataset", workspace="my-workspace")
```


src/argilla/client/sdk/datasets/v1/models.py
```diff
@@ -0,0 +1,34 @@
...
+from datetime import datetime
+from typing import Any, Dict
+
+from pydantic import BaseModel, Field
+
+
+class BaseDatasetModel(BaseModel):
+    tags: Dict[str, str] = Field(default_factory=dict)
+    metadata: Dict[str, Any] = Field(default_factory=dict)
+    name: str = Field(regex="^(?!-|_)[a-zA-Z0-9-_ ]+$")
```
alvarobartt: @jfcalvo I saw that datasets can be created using spaces, which was not supported before, is this correct?

src/argilla/client/sdk/datasets/v1/api.py
```diff
@@ -0,0 +1,152 @@
...
+        return Response(
+            status_code=response.status_code,
+            content=response.content,
+            headers=response.headers,
+            parsed=response.json(),
+        )
+    return handle_response_error(response)
+
+
+def add_record(
```
alvarobartt: Also it seems that we're no longer able to add records until the dataset is published, is this correct? @jfcalvo

src/argilla/client/sdk/datasets/models.py
```diff
@@ -36,7 +36,7 @@ def _missing_(cls, value):
 class BaseDatasetModel(BaseModel):
     tags: Dict[str, str] = Field(default_factory=dict)
     metadata: Dict[str, Any] = Field(default_factory=dict)
-    name: str = Field(regex="^(?!-|_)[a-z0-9-_]+$")
+    name: str = Field(regex="^(?!-|_)[a-zA-Z0-9-_]+$")
```
alvarobartt: I also updated this regex, but didn't include the space just in case

src/argilla/client/nightly.py
```diff
@@ -0,0 +1,231 @@
...
+            self.records.append(self.schema(**record))
+
+    def fetch_one(self) -> Union[Dict[str, Any], List[str, Any]]:
+        if self.__records is None or len(self.__records) < 1:
+            # TODO: handle exception if there are no records
+            return self.client.get_records(id=self.id, offset=0, limit=1).parsed["items"][0]
+        return self.__records[0]
+
+    # TODO: we could fetch those on iter, maybe we can create an `streaming` flag or something similar
+    def iter(self, batch_size: int = 32) -> List[BaseModel]:
```
alvarobartt: Here I'll probably create another class named `StreamingDataset` to stream the data during training if that's something we need/want, but still need to discuss it

src/argilla/client/nightly.py
```diff
@@ -0,0 +1,231 @@
...
+
+
+class QuestionSchema(BaseModel):
+    name: str
+    title: str
+    required: bool
+    settings: Dict[str, Any]
+
+
+class Dataset:
```
alvarobartt: I'm planning to combine the current with some ideas that @frascuchon left at https://gist.github.com/frascuchon/2e11db8a00559746e4b4c2dff687daaf

src/argilla/client/sdk/datasets/v1/models.py
jfcalvo: That's right. We don't have any limitation about what a dataset name can be. We will have soon a limitation in the number of characters but that's all.

src/argilla/client/sdk/datasets/v1/api.py
jfcalvo: Exactly. Right now the flow should be the following:
1. Create the Dataset.
2. Add at least one Field to the Dataset.
3. Add at least one Question to the Dataset.
4. Publish the Dataset.
5. Add records to the dataset that matches the defined fields and questions for the dataset.

Before publishing the dataset is not possible to add records.

src/argilla/client/sdk/datasets/models.py
jfcalvo: If that's the old Dataset model I think we should not change it. Correct me if I'm wrong @frascuchon

src/argilla/client/sdk/datasets/v1/models.py
jfcalvo: With that said maybe you don't need to specify any regex. That should be transparent to you.

src/argilla/client/sdk/datasets/models.py
alvarobartt: Makes sense, but was that intended originally? I mean, not allowing upper-case letters in the dataset name

src/argilla/client/sdk/datasets/v1/models.py
alvarobartt: Since I was validating the schema of the dataset uploaded to Argilla, I changed that since the dataset I was using was named `"Big Dataset"` and the validation was failing, but we can make it just a string with no regex validation.

src/argilla/client/sdk/datasets/v1/api.py
alvarobartt: Cool, I'll push both the `fields` and `questions` methods today to cover all the `/api/v1` endpoints 😄

src/argilla/client/sdk/datasets/models.py
jfcalvo: The idea was to format the dataset name so it can be used as ID for ElasticSearch indexes.

alvarobartt:
As mentioned today, there are some things still missing to move this PR out of draft:

* handle every possible exception in both `rg.create_feedback_dataset` and `rg.FeedbackDataset`
* delete dataset if any validation fails during `rg.create_feedback_dataset`
* add docstrings to every class, method, and function
* add unit tests to make sure that the introduced features work as expected
* collect more feedback on whether the current approach works as expected

---

Besides that, feel free to start playing around with it cc. @nataliaElv @tomaarsen @davidberenstein1957 

Install it from source as `pip install git+https://github.com/argilla-io/argilla.git@feat/new-datasets`

Then run the following:

```python
import argilla as rg
from argilla.client.nightly import TextField, TextQuestion, RatingQuestion

rg.init(
    api_url="...",
    api_key="..."
)

dataset = rg.create_feedback_dataset(
    name="feedback-task-dataset-via-sdk",
    workspace="workspace-0",
    guidelines="These are the guidelines",
    fields=[
        TextField(name="prompt"),
        TextField(name="completion"),
    ],
    questions=[
        TextQuestion(
            name="question-1",
            description="This is the first question",
            required=True,
        ),
        RatingQuestion(
            name="question-2",
            description="This is the second question",
            required=True,
            values=[1, 2, 3, 4, 5],
        ),
    ],
)

dataset.add_record(
    record={
        "prompt": "This is the prompt",
        "completion": "This is the completion",
    },
    response={
        "values": {
            "question-1": {"value": "This is the answer to question bla bla bla"},
            "question-2": {"value": 2},
        },
        "status": "submitted",
    },
)
 
dataset = rg.FeedbackDataset(
    name="feedback-task-dataset-via-sdk",
    workspace="workspace-0",
)

print(dataset)
print(dataset.fields)
print(dataset.questions)
print(dataset.records)
```

src/argilla/client/client.py
```diff
@@ -208,6 +215,50 @@ def get_workspace(self) -> str:
...
+
+    def add_field(self, id: str, field: Dict[str, Any]) -> httpx.Response:
+        return datasets_api_v1.add_field(client=self._client, id=id, field=field)
+
+    def get_questions(self, id: str) -> List[Dict[str, Any]]:
+        return datasets_api_v1.get_questions(client=self._client, id=id).parsed
+
+    def add_question(self, id: str, question: Dict[str, Any]) -> httpx.Response:
+        return datasets_api_v1.add_question(client=self._client, id=id, question=question)
+
```
frascuchon, line 261: All these methods are referring to the new apiV1 endpoints and could be confusing to keep them here as is. For example, the `list_datasets` method will return only new kinds of datasets. 

Could we move them to a DatasetsV1 class and work with them from there?

```python 
class DatasetsV1:

   def __init__(client: AuthenticatedCliewnt):
       self.client = client

  def list_datasets(self) -> List[FeedbackDatasetModel]:
        return datasets_api_v1.list_datasets(client=self._client).parsed
```

And then use as:

```python
argilla_client.datasets_v1.list_datasets()
```

src/argilla/client/client.py
```diff
@@ -208,6 +215,50 @@ def get_workspace(self) -> str:
...
+        return datasets_api_v1.get_fields(client=self._client, id=id).parsed
+
+    def add_field(self, id: str, field: Dict[str, Any]) -> httpx.Response:
+        return datasets_api_v1.add_field(client=self._client, id=id, field=field)
+
+    def get_questions(self, id: str) -> List[Dict[str, Any]]:
+        return datasets_api_v1.get_questions(client=self._client, id=id).parsed
+
+    def add_question(self, id: str, question: Dict[str, Any]) -> httpx.Response:
+        return datasets_api_v1.add_question(client=self._client, id=id, question=question)
```
frascuchon: All these methods are referring to the new apiV1 endpoints and could be confusing to keep them here as is. For example, the `list_datasets` method will return only new kinds of datasets. 

Could we move them to a DatasetsV1 class and work with them from there?

```python 
class DatasetsV1:

   def __init__(client: AuthenticatedCliewnt):
       self.client = client

  def list_datasets(self) -> List[FeedbackDatasetModel]:
        return datasets_api_v1.list_datasets(client=self._client).parsed
```

And then use as:

```python
argilla_client.datasets_v1.list_datasets()
```

src/argilla/client/nightly.py
frascuchon, line 1: If this this module defines the user layer data models, we should rename it to `client/feedback.py` and work from that

src/argilla/client/nightly.py
```diff
@@ -0,0 +1,367 @@
...
+
+class FeedbackDataset:
+    def __init__(
+        self,
+        name: Optional[str] = None,
+        *,
+        workspace: Optional[Union[rg.Workspace, str]] = None,
+        id: Optional[str] = None,
+    ) -> None:
+        self.client: "Argilla" = rg.active_client()
```
frascuchon: I prefer to let users work locally with these feedback datasets, and do explicit interactions with the argilla server, as `dataset.save` or `dataset.push`

Similar for loading. We should load datasets from argilla with an explicit method.
````
python
ds = FeedbackDataset.from_name_and_workspace(name, workspace)
...
````

File: src/argilla/client/nightly.py [hunk 9]
```diff
@@ -0,0 +1,367 @@
...
+    workspace: Optional[Union[str, rg.Workspace]] = None,
+    guidelines: Optional[str] = None,
+    fields: Optional[List[FieldSchema]] = None,
+    questions: Optional[List[QuestionSchema]] = None,
+) -> FeedbackDataset:
+    client = rg.active_client()
+
+    if workspace is None or isinstance(workspace, str):
+        try:
+            workspace = rg.Workspace.from_name(workspace)
```
frascuchon: If the `workspace` is `None` this statement should fail. Could we use the `client.get_current_workspace()` to setup a default value?
```python
workspace = workspace or client.get_current_workspace()
```

src/argilla/client/sdk/datasets/models.py
frascuchon: For old datasets this is not a valid name (uppercase letters are not allowed), we should revert this change here

src/argilla/client/sdk/datasets/v1/__init__.py
frascuchon, line 1: I think it makes sense changing a bit the module structure.  Instead of 
````
sdk/
  datasets/
    v1/
````
we can group by sdk API versions (at least for the v1 one):
````
sdk/
  v1/
    datasets/
````

File: src/argilla/client/sdk/datasets/v1/api.py [hunk 10]
```diff
@@ -0,0 +1,301 @@
...
+        return Response(
+            status_code=response.status_code,
+            content=response.content,
+            headers=response.headers,
+            parsed=parsed_response,
+        )
+    return handle_response_error(response)
+
+
+@lru_cache(maxsize=None)
```
frascuchon: we can remove this for now, and introduce a cache mechanism later. This method can be called several times and invalidating the cache can be a hell

alvarobartt: Sure, I used it because it was already used in the former `datasets/api.py`

src/argilla/client/sdk/datasets/v1/__init__.py
alvarobartt: Sure, I'll change this rn!

src/argilla/client/sdk/datasets/models.py
alvarobartt: You're right! I'll rever the `regex` validation update there!

src/argilla/client/nightly.py
alvarobartt: Indeed the `rg.Workspace.from_name` uses the default if the provided is None, so it won't fail 👍🏻

alvarobartt: Unless you prefer to set `name` in `rg.Workspace.from_name` as a mandatory value, and get the default in the line you mentioned instead.

src/argilla/client/nightly.py
alvarobartt: Yes, I was already preparing `push_to_argilla` and `from_argilla` as we discussed! 😄

src/argilla/client/nightly.py
alvarobartt: Yes, I was keeping the nightly just until we decided this, so I'll rename it now!

src/argilla/client/client.py
alvarobartt: Yes, you're right! I'll update this too, thanks 👍🏻

src/argilla/client/feedback.py
```diff
@@ -0,0 +1,454 @@
...
+class ResponseSchema(BaseModel):
+    values: Dict[str, ValueSchema]
+    status: Literal["submitted", "missing", "discarded"]
+
+
+class OfflineFeedbackResponse(ResponseSchema):
+    user_id: Optional[UUID] = None
+
+    @validator("user_id", always=True)
+    def user_id_must_have_value(cls, v):
```
frascuchon, line 65: This won't work for now. The current endpoint accepts responses only for connected user. I think we can remove the user_id attribute for now in the `OfflineFeedbackResponse` class

src/argilla/client/feedback.py
```diff
@@ -0,0 +1,454 @@
...
+
+        self.__records = []
+
+    def __len__(self) -> int:
+        return len(self.__records)
+
+    def __getitem__(self, key: Union[slice, int]) -> Union["FeedbackItemModel", List["FeedbackItemModel"]]:
+        if len(self.__records) < 1:
+            raise RuntimeError(
+                "In order to get items from `rg.FeedbackDataset` you need to either add them first with `add_records` or fetch them from Argilla or HuggingFace with `fetch_records`."
```
frascuchon:
```suggestion
                "In order to get items from `rg.FeedbackDataset` you need to either add them first with `add_records` "
                "or fetch them from Argilla or HuggingFace with `fetch_records`."
```

src/argilla/client/feedback.py
```diff
@@ -0,0 +1,454 @@
...
+
+    @property
+    def questions(self) -> List[FeedbackQuestionModel]:
+        return self.__questions
+
+    @property
+    def records(self) -> List["FeedbackItemModel"]:
+        return self.__records
+
+    def fetch_records(self, overwrite: bool = False) -> List["FeedbackItemModel"]:
```
frascuchon: So users can refresh info from argilla using this method. How users can fetch a chunk of data `ds.records[100:1000]` ?

src/argilla/client/feedback.py
```diff
@@ -0,0 +1,454 @@
...
+            first_batch = client.get_records(id=self.argilla_id, offset=0, limit=FETCHING_BATCH_SIZE)
+            self.__records = first_batch.items
+            total_batches = first_batch.total // FETCHING_BATCH_SIZE
+            current_batch = 1
+            with tqdm(initial=current_batch, total=total_batches, desc="Fetching records from Argilla") as pbar:
+                while current_batch <= total_batches:
+                    batch = client.get_records(
+                        id=self.argilla_id, offset=FETCHING_BATCH_SIZE * current_batch, limit=FETCHING_BATCH_SIZE
+                    )
+                    self.__records += batch.items
```
frascuchon: If we do:

```python
ds.fetch_records(override=True)
```
and later 
```python
ds.fetch_records(override=False) # <- warning here!
```

This will show a message that can be confusing regarding the non-pushed records (which are records fetched previously)

src/argilla/client/feedback.py
alvarobartt: The thing is that the only consistent solution is to fetch all the records from Argilla when calling `from_argilla`, as otherwise relying on the API for every single record will end up on a consistency issue... We can discuss this later today with a demo!

src/argilla/client/sdk/v1/datasets/models.py
```diff
@@ -0,0 +1,87 @@
...
+
+import sys
+from datetime import datetime
+from typing import Any, Dict, List, Optional, Union
+from uuid import UUID
+
+if sys.version_info >= (3, 8):
+    from typing import Literal
+else:
+    from typing_extensions import Literal
```
tomaarsen: The most robust solution may be to use the Python [EAFP](https://docs.python.org/3/glossary.html#term-EAFP) coding style:
```python
try:
    from typing import Literal
except ImportError:
    from typing_extensions import Literal
```
That way, you'll never accidentally list the wrong `sys` version, and it won't break if in the future the Python devs decide to moving something from `typing` back to `typing_extensions` or something. (I assume they won't, but still).

This extends to the rest of the codebase here. I'm open to your thoughts here.

alvarobartt: Well since that certainly would work, I guess that if we can catch it better right? Anyway it's a minor issue so I'm also open to catch that on `ImportError` and fallback to `typing_extensions` if `typing` doesn't work for a certain Python version e.g. `typing.Annotated` introduced in Python 3.9, or `typing.Literal` introduced in Python 3.8.

src/argilla/client/feedback.py
```diff
@@ -0,0 +1,809 @@
...
+class RatingQuestion(QuestionSchema):
+    settings: Dict[str, Any] = Field({"type": "rating"})
+    values: List[int]
+
+    @validator("values", always=True)
+    def update_settings_with_values(cls, v, values):
+        if v:
+            values["settings"]["options"] = [{"value": value} for value in v]
+        return v
+
```
frascuchon, line 150: If we want to keep updated values and settings once the question object is created, we should enable the `validate_assignment` config attribute. See [docs](https://docs.pydantic.dev/latest/usage/model_config/#options) 
```suggestion
class RatingQuestion(QuestionSchema):
    settings: Dict[str, Any] = Field({"type": "rating"})
    values: List[int]

    @validator("values", always=True)
    def update_settings_with_values(cls, v, values):
        if v:
            values["settings"]["options"] = [{"value": value} for value in v]
        return v
    
    class Config:
        validate_assignment= True
```

src/argilla/client/feedback.py
```diff
@@ -0,0 +1,807 @@
...
+        return self.records[key]
+
+    def __del__(self) -> None:
+        """When the dataset object is deleted, delete all the records as well to avoid memory leaks."""
+        if hasattr(self, "__records") and self.__records is not None:
+            del self.__records
+        if hasattr(self, "__new_records") and self.__new_records is not None:
+            del self.__new_records
+
+    def __enter__(self) -> "FeedbackDataset":
```
frascuchon, line 339: What the use case you're thinking about the context manager?

alvarobartt: I used it a couple times while testing everything, but probably there's no use-cases, so we can remove this

src/argilla/client/feedback.py
```diff
@@ -0,0 +1,807 @@
...
+            first_batch = client.get_records(id=self.argilla_id, offset=0, limit=FETCHING_BATCH_SIZE)
+            self.__records = first_batch.items
+            total_batches = first_batch.total // FETCHING_BATCH_SIZE
+            current_batch = 1
+            with tqdm(
+                initial=current_batch,
+                total=total_batches,
+                desc="Fetching records from Argilla",
+            ) as pbar:
+                while current_batch <= total_batches:
```
frascuchon: Since the `total` is optional (See the [API ref](https://tqdm.github.io/docs/tqdm/#__init__)), we still use the tqdm progress bar and prepare changes for #2917

```suggestion
            current_batch = 1
            with tqdm(
                initial=current_batch,
                desc="Fetching records from Argilla",
            ) as pbar:
                while current_batch <= total_batches:
```

src/argilla/client/feedback.py
alvarobartt: You mean to do something like e.g.

```python
import argilla as rg

question = rg.RatingQuestion(name="rating", required=True, values=[1, 2, 3])
question.values = [1, 2, 3, 4, 5]
```

Not sure if that's what you mean

frascuchon: Exactly

src/argilla/client/feedback.py
alvarobartt: Then I'll need to change the whole loop I guess, since I was also using the computed `total_batches` for updating the `offset` on every request to Argilla API

alvarobartt: Is there any way to retrieve the total number of records from the metrics?

src/argilla/client/feedback.py
alvarobartt: Ok then I'll update that!

src/argilla/client/client.py
```diff
@@ -208,6 +217,143 @@ def get_workspace(self) -> str:
...
+            guidelines: The guidelines of the dataset. Defaults to `None`.
+
+        Returns:
+            A `FeedbackDatasetModel` object, containing the dataset attributes: id, name,
+            workspace_id, guidelines, status, created_at, and last_updated.
+        """
+        return datasets_api_v1.create_dataset(
+            client=self._client.httpx,
+            name=name,
+            workspace_id=workspace_id,
```
frascuchon: Since all these methods are referring the new FeedbackDataset and are used in the `feedback.py`, I would move to the module as private functions, for 2 main reasons:

1. Keep current/old code as invariant as possible -> (new things in a different place will help us with the migration progress)
2. *Dataset* is a concept already defined in Argilla, so can be confusing having a `list_datasets` method returning only V1 ones.

alvarobartt: Yes, I remember you mentioned this some days ago, can we schedule a quick call to see the approach you're thinking of? Thanks 😄

frascuchon (approved):

PR #2837, opened 2023-05-04
Response creation/update must receive a value for status

### LINKED ISSUE #2762: Allow discarding a record response for a user
Users can discard a record response for a record. The content of the response can be filled in even if it's discarded.

# TODO

- [ ] Connect front and back to discard records properly (See https://github.com/argilla-io/argilla/pull/2837 for more details)

jfcalvo:
# Description

**Warning:** review and merge PR https://github.com/argilla-io/argilla/pull/2826 before this one.

This PR add changes to record responses so it must receive a value for `status` when they are created or updated.

Related changes:
* Now when a response is created a `status` value is mandatory to be used (previously it was optional with `submitted` as default).
  * This change affects frontend @keithCuniah
* Now response `values` is optional. You can create a response with `status` to `discarded` without specifying a value for `values` field.
  * This change affect database migrations. We need to rollback responses table migration on `dev` environment and running it again after deployment.

This PR affects the following endpoints:
* `GET  /datasets/{dataset_id}/records`
* `POST /datasets/{dataset_id}/records`
* `POST /records/{record_id}/responses`
* `PUT  /responses/{response_id}`

@keithCuniah feel free to add frontend changes if the PR is approved.

Closes #2762


src/argilla/server/schemas/v1/records.py
```diff
@@ -33,9 +42,14 @@ class Config:
...
+
+
+class DiscardedResponseCreate(BaseModel):
+    values: Optional[Dict[str, ResponseValueCreate]]
+    status: Literal[ResponseStatus.discarded]
 
 
-class Responses(BaseModel):
-    items: List[Response]
+ResponseCreate = Annotated[Union[SubmittedResponseCreate, DiscardedResponseCreate], Field(discriminator="status")]
```
frascuchon, line 55: Providing a default status is not easy from this definition. So, I would prefer to make it required and explore solutions in the future.

PR #2845, opened 2023-05-05
feat : 2839 front update me endpoints

keithCuniah:
# Description

The following PR add changes to support new `/api/v1/me/datasets` and  `/api/v1/datasets/:dataset_id/records` endpoints

Centralized url in url.properties.js

Refs: #2826


keithCuniah:
the change will be in the PR https://github.com/argilla-io/argilla/pull/2846

Issue #2840, opened 2023-05-05
Add response_status filter when listing records for a user


frascuchon:
Refs: #2742 and #2743

Issue #2743, opened 2023-04-20
Allow to switch the record list


frascuchon:
Allow changing the default record list to show

a. Already annotated records
b. Discarded records

From each view, users can still provide the questions feedback.

Refs #2742

Issue #2742, opened 2023-04-20
Show non annotated/pending records as default


frascuchon:
The default view of the dataset annotation view will show only the non-annotated records. 

From this view, users can annotate records and navigate over them. If the user refreshes the view or applies a query, the list will change, and previous annotated/discarded records won't be shown anymore

nataliaElv:
The default view will show records where status of the response is `status:Pending`

PR #2857, opened 2023-05-08
feat : front => allow switch record lists by status

keithCuniah:
# Description
Allow to swich the record lists regarding the status of the records.
By default, records with `reponses_status === "missing"` are fetched

See #2840 
See #2742  
See #2743 

**Type of change**

**How Has This Been Tested**

- Only feedback task : when selecting a different status from the filter => the records and the total record are updated. 

IMPORTANT : before fetching new records, the Record model is dropped and then the new records are inserted inside the orm model. This is important to not mess with the order of record to get from vuexorm by components

### DIFF
--- a/frontend/components/page-contents/feedback-task-content/CenterFeedbackTask.content.vue
+++ b/frontend/components/page-contents/feedback-task-content/CenterFeedbackTask.content.vue
@@ -1,15 +1,21 @@
 <template>
   <BaseLoading v-if="$fetchState.pending" />
   <RecordFeedbackTaskAndQuestionnaireContent
+    :key="rerenderChildren"
     v-else-if="!$fetchState.pending"
     :datasetId="datasetId"
     :recordOffset="recordOffset"
+    :recordStatusToFilterWith="recordStatusFilteringValue"
   />
 </template>
 
 <script>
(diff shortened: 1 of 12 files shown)


frontend/components/page-contents/feedback-task-content/CenterFeedbackTask.content.vue
```diff
@@ -50,6 +57,17 @@ export default {
...
     },
+    recordStatusFilteringValue() {
+      const { _status: recordStatus } = this.$route.query;
+      return recordStatus ?? RECORD_STATUS.PENDING.toLowerCase();
+    },
+  },
+  watch: {
+    async recordStatusFilteringValue() {
+      // NOTE - each time the filter change, clean records orm && rerender the children component
+      await deleteAllRecords();
```
frascuchon, line 68: Regarding the ORM refresh when the response status filter changes, we need to adapt the change introduced in https://github.com/argilla-io/argilla/pull/2851. We should prevent users if some data may get lost. 

If this change is big, you can tackle it in another PR. In that case, please, note this somewhere before merging the branch

keithCuniah: Ref : #2863

PR #2859, opened 2023-05-08
Add `Workspace` class & `list_workspaces`

alvarobartt:
# Description

This PR has been created on top of #2827 to add a `Workspace` class so that we can easily convert workspace names (human readable format) to workspace IDs (transparent to the user). This is just covering the basic functionality which is to `list_workspaces` to get all the available workspaces for the current user, so that when instantiating `Workspace` we can know whether that workspace exists and is available for the current user or not.

Besides that, the main reason behind this PR is because the `/api/v1` and `/api/v1/me` endpoints expect the workspace ID, not the workspace name, as opposed to the former `/api` (v0) which was using names not IDs.

> :warning: Note that this may be extended in the future to cover all the `Workspace` functionalities provided by Argilla, but for the moment, we're keeping it simple since it's just going to be used for name to ID conversation and workspace validation.

**Type of change**

**How Has This Been Tested**

- [X] Add unit tests for `list_datasets` and `Workspace`

**Checklist**

### DIFF
--- a/src/argilla/client/workspaces.py
+++ b/src/argilla/client/workspaces.py
@@ -13,6 +13,7 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+import warnings
 from typing import TYPE_CHECKING, Optional
 
 import argilla as rg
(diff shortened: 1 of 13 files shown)


src/argilla/client/workspaces.py
```diff
@@ -0,0 +1,64 @@
...
+    Raises:
+        ValueError: if the workspace does not exist in the current Argilla account.
+
+    Examples:
+        >>> from argilla import rg
+        >>> workspace = rg.Workspace("my-workspace")
+        >>> workspace.id
+    """
+
+    def __init__(self, name: Optional[str] = None) -> None:
```
frascuchon: As for the dataset initialization, I would prefer to remove this code from the __init__ class method. For me, make sense having a  class method to fetch the workspace by name:

```python
workspace = rg.Workspace.find_by_name(name="...")
...
```

Later, we can user the __init__ method to create an workspace instance and use it create a new workspace in argilla:

```python
new_workspace = rg.Workspace(name="new-workspace")

new_workspace.save()
# or:
new_workspace.create()
# or:
rg.Workspace.save(new_workspace)
# or:
rg.create_workspace(new_workspace)
# or (sorry for the alternatives,...)
from argilla import workspaces

workspaces.create(new_workspace)
```

alvarobartt: Ok it makes sense if we intended to extend it! Thanks for the feedback 👍🏻 Do you agree on naming the classmethod `rg.Workspace.from_name` instead of `rg.Workspace.find_by_name` since the first is probably the most common used in Python. Unless you have any special preference for `find_by_name` obviously.

PR #2893, opened 2023-05-12
feat: include simple transition 

### LINKED ISSUE #2892: Include transition when navigating between records
Includes transition when navigating between records to make pagination more explicit

leiyre:
# Description

This PR includes transition when navigating between records

Closes #2892

**Type of change**

**Checklist**

### DIFF
--- a/frontend/components/page-contents/feedback-task-content/RecordFeedbackTaskAndQuestionnaire.content.vue
+++ b/frontend/components/page-contents/feedback-task-content/RecordFeedbackTaskAndQuestionnaire.content.vue
@@ -4,7 +4,11 @@
     :key="recordOffset"
     class="wrapper"
   >
-    <transition name="record-page" v-if="recordId" appear>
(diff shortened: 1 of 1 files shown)


frontend/components/page-contents/feedback-task-content/RecordFeedbackTaskAndQuestionnaire.content.vue
```diff
@@ -86,6 +92,7 @@ export default {
   data() {
     return {
       rerenderQuestionnaire: 1,
+      transitionWhenNavigating: true,
```
keithCuniah: 1/ renaming : 'isTransition' (the component does not know that the transition is used on pagination,or other way)
2/ can you pass it as a props with a default value as true ? because when we will want to remove the transition, the logic for it need to come from the parent

keithCuniah (approved):

PR #2907, opened 2023-05-12
fix : update total record only once

keithCuniah:
# Description
Update total_record in FeedbackDataset orm only if value not exists
This is for performance, in order to not always update the total_record when fetching new records

**Type of change**

**How Has This Been Tested**

- Only feedback tasks and only fetching records (this endpoint contains tht total_records)

### DIFF
--- a/frontend/components/page-contents/feedback-task-content/RecordFeedbackTaskAndQuestionnaire.content.vue
+++ b/frontend/components/page-contents/feedback-task-content/RecordFeedbackTaskAndQuestionnaire.content.vue
@@ -208,8 +208,9 @@ export default {
       // FORMAT records for orm
       const formattedRecords = this.factoryRecordsForOrm(records);
 
-      // UPSERT total records && records in ORM
-      updateTotalRecordsByDatasetId(this.datasetId, totalRecords);
+      // UPSERT total records (only if it's not exists in ORM) && records in ORM
+      this.totalRecords ??
+        updateTotalRecordsByDatasetId(this.datasetId, totalRecords);
       upsertRecords(formattedRecords);
     },
     async getRecords(datasetId, recordOffset, numberOfRecordsToFetch = 5) {


frontend/components/page-contents/feedback-task-content/RecordFeedbackTaskAndQuestionnaire.content.vue
```diff
@@ -208,8 +208,9 @@ export default {
       // FORMAT records for orm
       const formattedRecords = this.factoryRecordsForOrm(records);
 
-      // UPSERT total records && records in ORM
-      updateTotalRecordsByDatasetId(this.datasetId, totalRecords);
+      // UPSERT total records (only if it's not exists in ORM) && records in ORM
+      this.totalRecords ??
```
frascuchon, line 212: Be careful with this condition. I can imagine that the total number of records should be updated at the same time as the record list itself. Does the condition control that?

keithCuniah: totalRecords is a computed, that means it will always reflect and be update with what the value from the orm.

The method initRecordsInDatabase have the purpose to fetch/add records to the orm and fetch/add total_records to the orm (not Record.model but Feedback.model)

The method `initRecordsInDatabase` is called at two places and only : 
 - on the fetch hook (so at the initialisation of the component)
 - on watch of recordOffset (this is the index of the current record)

This means that we update total_records only at the initialisation and on watch if there was not a value before. 

So in short words:
1/We control with reactivity that we have a total_records in the orm
2/the Nullish coalescing operator `??` control that the value is not null/undefined, only if it is then we update the total_records from the orm
3/ we upsert the records

frascuchon: If `totalRecords` is not reset when the records table is cleaned, this won't work really: changing the status filter, or refreshing data, brings unexpected results since the `totalRecords` is already defined.

PR #2919, opened 2023-05-16
Using an already setup httpx client in sdk modules

frascuchon:
# Description

Since the new SDK modules are creating a new httpx client on each request, some problems related to connection performance can be raised. In the [official docs](https://www.python-httpx.org/advanced/) you can read: 

> When you make requests using the top-level API as documented in the [Quickstart](https://www.python-httpx.org/quickstart/) guide, HTTPX has to establish a new connection for every single request (connections are not reused). As the number of requests to a host increases, this quickly becomes inefficient.
> 
> On the other hand, a Client instance uses [HTTP connection pooling](https://en.wikipedia.org/wiki/HTTP_persistent_connection). This means that when you make several requests to the same host, the Client will reuse the underlying TCP connection, instead of recreating one for every single request.

In order to better use the httpx client, this PR changes the `sdk` module to reuse the already set up httpx client in the `AuthenticatedClient` instance.

By using this client, you don't need to pass header or base URL configuration on each request, since is setup on client initialization.

### DIFF
--- a/src/argilla/client/client.py
+++ b/src/argilla/client/client.py
@@ -224,42 +224,45 @@ class Argilla:
             A list of `WorkspaceModel` objects, containing the workspace
             attributes: name, id, created_at, and updated_at.
         """
-        return workspaces_api.list_workspaces(client=self._client).parsed
+        return workspaces_api.list_workspaces(client=self._client.httpx).parsed
 
     def list_datasets(self) -> List[FeedbackDatasetModel]:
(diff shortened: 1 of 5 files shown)


alvarobartt:
Hi @frascuchon everything looks good to me! Really nice improvement on the usage of `httpx` 🤗

alvarobartt (approved):

PR #2932, opened 2023-05-17
2926 argilla server fails on python 310 and python 311

### LINKED ISSUE #2926: `argilla server` fails on Python 3.10 and Python 3.11
**Describe the bug**

As of the recent addition of the `Questions` schema for the `/api/v1`, the `typing.Annotated` has been introduced, and it seems to be failing under the following condition:

```python
from pydantic import BaseModel, Field
from typing import Union
from sqlalchemy import JSON
from sqlalchemy.orm import Mapped, relationship

class A(BaseModel):
  a: int

class B(BaseModel):
  b: int

AB = Annotated[Union[A, B], Field(..., discriminator="type")

class ABTable(Base):
    __tablename__ = "ab"

    settings: Mapped[dict] = mapped_column(JSON, default={})
    ab: Mapped["Whatever"] = relationship(back_populates="ab")

    @property
    def parsed_settings(self) -> AB:
        return parse_obj_as(AB, self.settings)
```

With the following error: `sqlalchemy.exc.InvalidRequestError: When initializing mapper Mapper[Question(questions)], expression 'Union' failed to locate a name ('Union'). If this is a class name, consider adding this relationship() to the <class 'argilla.server.models.questions.Question'> class after both dependent classes have been defined.`

Indeed, the code at https://github.com/argilla-io/argilla/blob/feat/2615-instructions-task/src/argilla/server/models/questions.py

**To Reproduce**

1. Install Argilla from source as `pip install git+https://github.com/argilla-io/argilla.git@feat/2615-intructions-task`
2. Run `python -m argilla server` from either Python 3.10 or 3.11 (as it works with other Python versions)

**Expected behavior**

`argilla server` should properly launch the Argilla Server in any Python version within the range of Python 3.7 and Python 3.11.

 - OS [e.g. iOS]: macOS
 - Argilla Version [e.g. 1.0.0]: 1.8.0-dev0
 - ElasticSearch Version [e.g. 7.10.2]: 8.5.3 (but fails regardless the ElasticSearch version)

frascuchon:
# Description

By moving the sqlalchemy `Question` class to the `models.py` module, the error is gone described in #2926 is gone.

Closes #2926

### DIFF
--- a/src/argilla/server/models/models.py
+++ b/src/argilla/server/models/models.py
@@ -18,11 +18,12 @@ from enum import Enum
 from typing import List, Optional
 from uuid import UUID, uuid4
 
+from pydantic import parse_obj_as
 from sqlalchemy import JSON, ForeignKey, Text, and_
 from sqlalchemy.orm import Mapped, mapped_column, relationship
 
 from argilla.server.database import Base
-from argilla.server.models import Question
+from argilla.server.models import QuestionSettings
 
 _USER_API_KEY_BYTES_LENGTH = 80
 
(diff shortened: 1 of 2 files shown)


alvarobartt:
Thanks for tackling this, I'll clone this branch locally and reproduce the tests that led to that issue to see if that's solved, and then happy to merge! 💪🏻

alvarobartt:
Already checked, and it works fine indeed with the latest Python versions 3.10.11, and 3.11.3, great job! 🤗

alvarobartt:
Can we increase the soft limit for the coverage percentage? Just not to see a lot of "failed runs" around when the only thing that may have failed is the coverage cc @frascuchon

alvarobartt (approved):

PR #2943, opened 2023-05-18
List all the `FeedbackDataset` responses not just the current user's

alvarobartt:
# Description

As recently spotted by @nataliaElv, we were just listing the current dataset records with their annotations, instead of all the available record annotations, as we were sending the request to `GET /api/v1/me/datasets/{id}/records` instead to `GET /api/v1/datasets/{id}/records`, so removing the prefix `/me` solves this issue.

Kudos to @nataliaElv 🎉

**Type of change**

**How Has This Been Tested**

- [X] Current unit tests are passing, and annotations are retrieved for more than one user when applicable

**Checklist**


frascuchon (approved):
Just to clarify, the difference between `/api/v1/datasets/:dataset_id/records` and /api/v1/datasets/:dataset_id/records` is only regarding the returned responses, not the records.

The number of records will be the same for both requests, the difference is with the prefixed `/me` endpoint, the record contains only the responses for the users requesting it

alvarobartt:
Thanks @frascuchon I've updated the PR title as it was a little bit confusing 😄

PR #2948, opened 2023-05-18
Docs/llm document structure

davidberenstein1957:
# Description

A new structure to facilitate LLM docs.

**Type of change**

- [X] Documentation update

**How Has This Been Tested**

N.A.

**Checklist**

N.A.


docs/_source/guides/llms/practical_guides/collect_responses.ipynb
alvarobartt, line 1: At the beginning we should mention that the user is supposed to `rg.init()` if using an Argilla Server different than `localhost`, as we're pulling a `FeedbackDataset` from Argilla. At the same time, we should mentioned that `rg.init` is not mandatory unless we call either `push_to_argilla` and/or `from_argilla`.

alvarobartt, line 1: I would remove the last section from this Jupyter Notebook, and rename the one `create_and_import_dataset.ipynb` to be `create_import_and_export.ipynb` to add the "Export or publish dataset" there, as this Jupyter Notebook already starts off with a previously published dataset in Argilla, as it's using the `from_argilla` class method.

alvarobartt, line 1: Additionally, regarding the export, the JSON-export mentioned in this Jupyter Notebook is not supported and will fail, so on, I'd suggest to recommend either `push_to_argilla` or `push_to_huggingface`, and additionally, if the user wants to store it locally, I'd recommend something like:

```python
hfds = ds.format_as("datasets")
hfds.save_to_disk("dataset")
hfds.to_csv("dataset.csv")
hfds.to_parquet("dataset.parquet")
hfds.to_pandas()
```

Which implies converting the `rg.FeedbackDataset` to a HuggingFace Dataset first, and then relying on their pre-implemented functions to export to different formats, to avoid re-inventing the wheel.

docs/_source/guides/llms/practical_guides/define_task.ipynb
alvarobartt, line 1: To keep the conversion to `pandas` within HuggingFace Datasets, I would just call `ds.to_pandas()`, since we're already importing the dataset from HuggingFace, and there's a built-in method for that

alvarobartt, line 1: I think here we're missing a section to define the fields, so we could create something like "Define fields".

alvarobartt, line 1: Also on "Define questions" I think we should link the matching `pydantic.BaseModel` for the rating and the text questions, as well as defining that both the `title` and the `required` params are optional, and those default to capitalized-name, and True, respectively.

alvarobartt (dismissed):
Just some minor comments, but I think the `rg.FeedbackDataset` creation could be more clear, as now we need to mix information from two sections to actually be able to create a `rg.FeedbackDataset`.

And also regarding the mentions to import/export, I'd mention first the export and then the import to/from, respectively, either Argilla's or HuggingFace's, since a user cannot import a dataset that hasn't been exported before. And as part of the HuggingFace export, we need to mention that we're also adding a config file named `argilla.cfg` which is required to load the dataset configuration back, the configuration being the fields, questions, and guidelines if applicable.

Also we're missing the mention that at least one field and one question must be `required=True`.

P.S. Sorry for the long review, but I couldn't comment on the files directly since those are Jupyter Notebooks.

docs/_source/guides/llms/practical_guides/define_task.ipynb
alvarobartt: e.g.

* Rating (`rg.RatingQuestion`): ...
* Text (`rg.TextQuestion`): ...

dvsrepo (commented):
This is looking good! 

As a quick comment: I feel we should remove the repetitive Next steps block (it's outdated and makes the user feel like going somewhere else, while they should keep reading the next section). @nataliaElv is this something you could tackle or is @davidberenstein1957 ?

nataliaElv:
> This is looking good!
> 
> As a quick comment: I feel we should remove the repetitive Next steps block (it's outdated and makes the user feel like going somewhere else, while they should keep reading the next section). @nataliaElv is this something you could tackle or is @davidberenstein1957 ?

I'm not sure where the block comes from or how to stop it from appearing in every page, so I'll leave that to @davidberenstein1957

tmp_trainer/runs/May30_12-43-52_Davids-MacBook-Pro.local/1685443432.482243/events.out.tfevents.1685443432.Davids-MacBook-Pro.local.25465.1
alvarobartt, line 1: docs/llm-document-structure

alvarobartt, line 1: Watch out @davidberenstein1957! You uploaded a lot of files under `tmp_trainer/runs` 😩

PR #2949, opened 2023-05-19
Add `format_as`, `push_to_huggingface`, & `from_huggingface` in `FeedbackDataset`

### LINKED ISSUE #2765: Allow users push an feedback task dataset to the HF hub using `push_to_hub`
Using the Python client to load a feedback task dataset, apply necessary changes to create a  HF Dataset which can be published on the HF hub

alvarobartt:
# Description

This PR adds some methods to the `rg.FeedbackDataset` as follows:

* `format_as` to format the records of the existing dataset as 🤗`datasets` format (i.e. `datasets.Dataset`) to be extended to support other formats like `pandas` or `numpy`
* `push_to_huggingface` to push the existing records and dataset configuration to the HuggingFace Hub (including an auto-generated dataset card! 🎉)
* `from_huggingface` is a classmethod to load a `rg.FeedbackDataset` from the HuggingFace Hub, pulling the dataset itself (records and annotations) and the configuration (guidelines, fields, and questions)

Closes #2765

**Type of change**

**How Has This Been Tested**

- [X] Added integration tests for `push_to_huggingface` and `from_huggingface`
- [x] Added integration tests for `format_as("datasets")`

**Checklist**


alvarobartt:
Here's an example on how to use the implemented features:

```python
import argilla as rg

fds = rg.FeedbackDataset(fields=[...], questions=[...])
fds.add_records([...])
fds.push_to_huggingface(repo_id="argilla/feedback-dataset", private=True, ...) # args and **kwargs correspond to the original `push_to_hub` method from 🤗HuggingFace
```

```python
import argilla as rg

fds = rg.FeedbackDataset.from_huggingface(repo_id="argilla/feedback-dataset", ...) # args and **kwargs correspond to the original `load_dataset` function from 🤗HuggingFace
```

alvarobartt:
I wanted to clarify what do we do when the dataset to be pushed to the HuggingFace Hub has no records, should we still upload the configuration and the dataset card? Or should we show a warning that the dataset cannot be created since there are no records? cc @frascuchon @dvsrepo

dvsrepo:
Thats

> I wanted to clarify what do we do when the dataset to be pushed to the HuggingFace Hub has no records, should we still upload the configuration and the dataset card? Or should we show a warning that the dataset cannot be created since there are no records? cc @frascuchon @dvsrepo

That's a very good question. 

On the one hand, I think is great to be able to push and read just the config.

On the other hand, what happens when people use the load_dataset (instead of our own from_huggingface)? I wonder if they'll get a weird error. If the error is not weird or there's no error, then I'd say we are fine using warnings on our side as discussed below.

From our side we can always warn users both when pushing and doing from_huggingface. We could even raise an exception first time and recommend the user to use a param (e.g., force_whatever) to push without records (maybe a dirty option)

All in all, the thing I find interesting is to be able to push and read configs so I let you think of we can cover this with these methods or there's a complete different workflow we can provide in the future.

alvarobartt:
> Thats
> 
> > I wanted to clarify what do we do when the dataset to be pushed to the HuggingFace Hub has no records, should we still upload the configuration and the dataset card? Or should we show a warning that the dataset cannot be created since there are no records? cc @frascuchon @dvsrepo
> 
> That's a very good question.
> 
> On the one hand, I think is great to be able to push and read just the config.
> 
> On the other hand, what happens when people use the load_dataset (instead of our own from_huggingface)? I wonder if they'll get a weird error. If the error is not weird or there's no error, then I'd say we are fine using warnings on our side as discussed below.
> 
> From our side we can always warn users both when pushing and doing from_huggingface. We could even raise an exception first time and recommend the user to use a param (e.g., force_whatever) to push without records (maybe a dirty option)
> 
> All in all, the thing I find interesting is to be able to push and read configs so I let you think of we can cover this with these methods or there's a complete different workflow we can provide in the future.

Thanks for the quick response! Then I guess for the moment we're safer with an exception being raised in case there are no records, and we can look into it in more detail when we tackle the `from_config` and `save_config` methods, so as to "standardize" a little bit more how we handle the configuration and, if applicable, upload just the configuration to the HuggingFace Hub 👍🏻

dvsrepo:
Sounds good!

src/argilla/client/feedback.py
```diff
@@ -672,6 +680,209 @@ def from_argilla(
...
+                for field in self.fields:
+                    dataset[field.name].append(record["fields"][field.name])
+                for question in self.questions:
+                    dataset[question.name].append(
+                        [
+                            {"user_id": r["user_id"], "value": r["values"][question.name]["value"]}
+                            for r in record["responses"]
+                        ]
+                        or None
+                    )
```
frascuchon, line 742: Maybe makes sense to move this code block into a private method `_format_as_datasets`

src/argilla/client/feedback.py
```diff
@@ -38,6 +39,7 @@
 
 if TYPE_CHECKING:
     import httpx
+    from datasets import Dataset
```
frascuchon, line 43: Do we need this lib as an explicit dep in pyproject.toml?

src/argilla/client/feedback.py
```diff
@@ -672,6 +680,209 @@ def from_argilla(
...
+        when a dataset is previously uploaded to the HuggingFace Hub.
+
+        Args:
+            repo_id: the ID of the HuggingFace Hub repo to push the `FeedbackDataset` to.
+            generate_card: whether to generate a dataset card for the `FeedbackDataset` in the HuggingFace Hub. Defaults
+                to `True`.
+            *args: the args to pass to `datasets.Dataset.push_to_hub`.
+            **kwargs: the kwargs to pass to `datasets.Dataset.push_to_hub`.
+        """
+        from huggingface_hub import DatasetCard, HfApi
```
frascuchon, line 765: I previously saw some problems with hf package updates regarding backward compatibility. Maybe it is more clear to users if we warn them when the hf package version is not the expected one, instead of letting the process fails

src/argilla/client/feedback.py
```diff
@@ -672,6 +680,209 @@ def from_argilla(
...
+                FeedbackDatasetConfig(fields=self.fields, questions=self.questions, guidelines=self.guidelines).json()
+            )
+            f.flush()
+
+            HfApi().upload_file(
+                path_or_fileobj=f.name,
+                path_in_repo="argilla.cfg",
+                repo_id=repo_id,
+                repo_type="dataset",
+                token=kwargs["token"] if "token" in kwargs else None,
```
frascuchon:
```suggestion
                token=kwargs.get("token"),
```

src/argilla/client/feedback.py
```diff
@@ -672,6 +680,209 @@ def from_argilla(
...
+            **kwargs: the kwargs to pass to `datasets.Dataset.push_to_hub`.
+        """
+        from huggingface_hub import DatasetCard, HfApi
+
+        if len(self) < 1:
+            raise ValueError(
+                "Cannot push an empty `rg.FeedbackDataset` to the HuggingFace Hub, please make sure to add at least one record, via the method `add_records`."
+            )
+
+        hfds = self.format_as("datasets")
```
frascuchon, line 778: If we move the previous referred code block into an private method, we can use it directly here
```suggestion
        hfds = self._format_as_datasets()
```

src/argilla/client/feedback.py
```diff
@@ -672,6 +680,209 @@ def from_argilla(
...
+            )
+            card = DatasetCard(
+                (
+                    f"## Guidelines\n\n{self.guidelines}\n\n"
+                    f"{explained_fields}\n\n"
+                    f"{explained_questions}\n\n"
+                    f"{loading_guide}\n\n"
+                )
+            )
+            card.push_to_hub(repo_id, repo_type="dataset", token=kwargs["token"] if "token" in kwargs else None)
```
frascuchon:
```suggestion
            card.push_to_hub(repo_id, repo_type="dataset", token=kwargs.get("token"))
```

src/argilla/client/feedback.py
```diff
@@ -672,6 +680,209 @@ def from_argilla(
...
+            )
+            f.flush()
+
+            HfApi().upload_file(
+                path_or_fileobj=f.name,
+                path_in_repo="argilla.cfg",
+                repo_id=repo_id,
+                repo_type="dataset",
+                token=kwargs["token"] if "token" in kwargs else None,
+            )
```
frascuchon: Again, we could encapsulate this block into a `_push_config_to_hugginface` method/local function. Just for the clarity of the flow.

src/argilla/client/feedback.py
frascuchon: Same here: `_push_dataset_card_to_hugginface`

frascuchon (approved):
Really nice! 

Just some code refactor suggestions. Feel free to apply them or not.

src/argilla/client/feedback.py
alvarobartt: So I left it like `format_as` to basically allow users to e.g. pull from Argilla, format it as they want (i.e. `datasets`), and then use it, without having to push anything, just to format the current dataset records in the format they want. The idea is to extend the `format_as` method to support `pandas`, `numpy`, etc.

src/argilla/client/feedback.py
alvarobartt: Sure, I'll add a decorator to check for certain versions of `huggingface_hub` and `datasets` 👍🏻

src/argilla/client/feedback.py
alvarobartt: As per the comment below: 

> I left it like format_as to basically allow users to e.g. pull from Argilla, format it as they want (i.e. datasets), and then use it, without having to push anything, just to format the current dataset records in the format they want. The idea is to extend the format_as method to support pandas, numpy, etc.

src/argilla/client/feedback.py
alvarobartt: Hmm at least for the `ArgillaTrainer` dependencies we're assuming everything is optional, so I can either add a `try-except ImportError` block there to enable the methods just if the required dependencies are available, add `datasets` as a core dependency, or just add the `requires_version` decorator to check the version before calling any of those methods.

src/argilla/client/feedback.py
alvarobartt: I agree with all these comments regarding the detach into smaller functions, but I was planning to move everything to a `MixIn` for the next release, so that everything's better organized, so I'd leave it like this for now, having in mind that I plan to refactor it into a `MixIn`.

src/argilla/client/feedback.py
alvarobartt: Same reply as above 😄

PR #2953, opened 2023-05-19
✨ Allow users to edit Descriptions and annotations guidelines WIP.

damianpumar:
# Description
**Allow users to create or edit the Description and annotation guidelines.**
I'm trying to implement the feature that allow the users to create or edit the descriptions and annotations.
I want to validate the first approach and look and feel of this request, and also I need some help to decide which markdown component to use here.

Description and annotation V1

**Type of change**

**How Has This Been Tested**
- [ ] DatasetDescriptionComponent unit test

**Checklist**

### DIFF
--- a/frontend/components/commons/dataset-description/DatasetDescription.component.vue
+++ b/frontend/components/commons/dataset-description/DatasetDescription.component.vue
@@ -1,43 +1,19 @@
 <template>
-  <div class="description">
+  <div>
     <h2 class="--heading5 --semibold description__title">{{ title }}</h2>
-    <div class="description__markdown__viewer" v-if="!isEditing">
-      <span
-        class="--body1 description__markdown__viewer-text"
-        v-html="compiledDescription"
-      />
-      <BaseButton name="edit-button" @on-click="onEdit">
-        <svgicon width="15" height="15" name="pen"></svgicon
-      ></BaseButton>
-    </div>
-    <div class="description__markdown__editor" v-if="isEditing">
-      <textarea
-        name="description-input"
-        class="description__markdown__editor-input"
-        v-model="newDescription"
-      ></textarea>
-
-      <div class="description__markdown__editor-button-area">
-        <BaseButton name="save-button" class="primary small" @on-click="onSave"
-          >Save</BaseButton
-        >
-        <BaseButton
-          name="close-button"
-          class="secondary small"
-          @on-click="onClose"
-          >Discard</BaseButton
-        >
-      </div>
-    </div>
+    <EditableDescriptionMarkdown :description="description" @onSave="onSave" />
   </div>
 </template>
 
 <script>
-import { marked } from "marked";
+import EditableDescriptionMarkdown from "./EditableDescriptionMarkdown.component.vue";
 
 import { getDatasetDescription, saveDescription } from "./MockedService";
 
 export default {
+  components: {
+    EditableDescriptionMarkdown,
+  },
   props: {
     datasetId: {
       type: Array,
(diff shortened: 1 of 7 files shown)


dvsrepo:
I'll let @leiyre and @keithCuniah comment, but in terms of look and feel is looking good! and I guess you can pick and propose one.

frontend/components/commons/dataset-description/DatasetDescription.component.vue
```diff
@@ -1,30 +1,96 @@
...
+      <div class="description__button-area">
+        <base-button name="save-button" class="primary small" @on-click="onSave"
+          >Save</base-button
+        >
+        <base-button
+          name="close-button"
+          class="secondary small"
+          @on-click="onClose"
+          >Discard</base-button
+        >
```
keithCuniah: consistency : base-button in camel case

frontend/components/commons/dataset-description/DatasetDescription.component.vue
```diff
@@ -1,30 +1,96 @@
...
   data: () => {
     return {
       title: "Description and annotation guidelines",
+      isEditing: false,
     };
   },
+  computed: {
+    description() {
+      return getDatasetDescription(this.datasetId, this.datasetTask, false);
+    },
```
keithCuniah, line 35: 1/ Instead of using the mock service, you can create the corresponding vuex orm query and call it here instead 
2/ performance : since the description will be update only in this component, it's not necesary to use a computed to get the description, you can use instead a static variable. You can get instead the data from the ORM in a hook lyfecycle instead (here an interesting [article](https://dev.to/pecus/when-to-avoid-vuejs-computed-properties-for-greater-performance-5anj)) 

Just to understand better, why get the description inside the component, instead of getting it from the parent component ?

keithCuniah:
About the question regarding the markdown, you can use an external plugin

frontend/components/commons/dataset-description/DatasetDescription.component.vue
damianpumar: Hi Keith, thanks for your suggestion, I was reading the code and I saw 51 components that are using the "base-button", so I agree with you and I will change it!.

Thanks!

frontend/components/commons/dataset-description/DatasetDescription.component.vue
damianpumar: Thanks Keith;
About the first point, I'm not sure which part of dataset entity I must show, because i don't have so much experience in this project, sorry, if you want, we can pair programming and i'll be happy to understand where the data is.

Related to second point, I created computed prop because the user can update that value; other approach is get the value as a prop, but in my humble opinion, when the components has a dependency from external services like a rest api, is more easy to maintain when state is in the current component. 

Wait your response!
Thanks,

frontend/components/commons/dataset-description/DatasetDescription.component.vue
damianpumar: I found 51 components are using BaseButton as kebab-case, maybe we can refactor it.

keithCuniah: Yes you are right. We try to be consistent in the new implementations but we didn't correct all yet :)

damianpumar:
> About the question regarding the markdown, you can use an external plugin

@keithCuniah about the markdown component, did you do any spike to decide which external library works better for Argilla?

davidberenstein1957:
Awesome @damianpumar

davidberenstein1957:
@damianpumar you might be able to align this here? 

https://github.com/argilla-io/argilla/pull/2981

Issue #2456, opened 2023-03-02
Normalization of deprecation warnings


tomaarsen:
**Is your feature request related to a problem? Please describe.**
I would like for all deprecation warnings to have:
1. The same text format.
2. Be thrown in the same way, either via `warnings.warn` or via `logger.warning`.
3. If `warnings.warn` is decided on, I would like for all deprecation warnings to have the same category. Currently, we use UserWarning, FutureWarning, DeprecationWarning

**Describe the solution you'd like**
I would like to see one approach used throughout Argilla.

**Describe alternatives you've considered**
Sticking with the current approach of mixing and matching `_LOGGER.warning(...)` and `warnings.warn`:
![image](https://user-images.githubusercontent.com/37621491/222385569-08de3564-4fa7-4508-a363-249ea2668d56.png)

**Additional context**
N/A

Note that the `warnings.warn` warnings are not formatted nicely using `rich`, while the logger ones obviously are.
I'd like to have a bit of a discussion on this topic, so we can reach a consensus on which approach we should use throughout the project.

cc: @dvsrepo @frascuchon @davidberenstein1957 

- Tom Aarsen

davidberenstein1957:
For me `_LOGGER.warning` seems like the way to go. And I would like to use `FutureWarning` to ensure the fact that it is being reported.

tomaarsen:
I don't think we can use a `Warning` class if we use `_LOGGER.warning`, can we? The `Warning` classes are only for `warning.warn(..., category=FutureWarning)` etc.

davidberenstein1957:
Ahh, I assumed we could pass that there too.

alvarobartt:
Hi @tomaarsen and @davidberenstein1957, so in order to handle this we can rely on https://docs.python.org/3/library/logging.html#logging.captureWarnings, so that we can still use the `warnings` library and keep the `DeprecationWarning`/`FutureWarning` messages.

I'm happy to tackle this.

PR #2962, opened 2023-05-22
Minor fixes and improvements around `rg.FeedbackDataset`

alvarobartt:
# Description

As requested by @davidberenstein1957, we are now mapping the `fields` and `questions` to their `pydantic.BaseModel` schema, instead of as the default schema for both `FieldSchema` and `QuestionSchema`, respectively; just when pulling from Argilla with `rg.FeedbackDataset.from_argilla`.

Now when pushing a `FeedbackDataset` to the HuggingFace Hub with `push_to_hub`, we are not just pushing the `user_id` and the `value` in the responses, but also the `status`, which can be "submitted", "discarded", or "missing".

Now there's no need to provide the `user_id` when adding record responses, as now the active Argilla user will be used just when pushing the records to Argilla. See the following example:

```python
import argilla as rg

rg.init(api_url="...", api_key="...")

dataset = rg.FeedbackDataset(
  fields=[
    rg.TextField(name="completion", required=True),
  ],
  questions=[
    rg.TextQuestion(
      name="question-1",
      description="This is the first question",
      required=True,
    ),
    rg.RatingQuestion(
      name="question-2",
      description="This is the second question",
      required=True,
      values=[1, 2, 3, 4, 5],
    ),
  ],
)

dataset.add_records([
  rg.FeedbackRecord(
    fields={
      "prompt": "The quick brown fox",
      "completion": "jumps over the lazy dog.",
    },
    responses=[
      {
        "user_id": None,
        "values": {"question-1": {"value": "Response 1"}, "question-2": {"value": 1}},
        "status": "submitted",
      },
    ],
  ),
])

dataset.push_to_argilla(name="my-dataset", workspace="my-workspace")
```

Note that in the snippet above, both `user_id` and `status` are optional, and their default values are `None` and `submitted`, respectively. So that the `user_id=None` is then transformed into the active Argilla `user_id`.

**Type of change**

**How Has This Been Tested**

- [x] Added integration tests to make sure that adding a response with `user_id=None` fulfils its value to be the active user
- [x] Added integration tests to make sure that a warning is raised in case there's more than one response with `user_id=None` for the same record, and that the response is not logged into Argilla

**Checklist**

### DIFF
--- a/src/argilla/client/feedback.py
+++ b/src/argilla/client/feedback.py
@@ -866,13 +866,18 @@ class FeedbackDataset:
                 f"Recommended `huggingface_hub` version is 0.14.0 or higher, and you have {huggingface_hub.__version__}, so in case you have any issue when pushing the dataset to the HuggingFace Hub upgrade it as `pip install huggingface_hub --upgrade`."
             )
 
-        token = kwargs.get("token") or kwargs.get("use_auth_token")
+        if "token" in kwargs:
+            token = kwargs.pop("token")
+        elif "use_auth_token" in kwargs:
+            token = kwargs.pop("use_auth_token")
+        else:
+            token = None
 
         config_path = hf_hub_download(
             repo_id=repo_id,
             filename="argilla.cfg",
             repo_type="dataset",
-            token=token if not isinstance(token, bool) else None,
+            token=token,
         )
         with open(config_path, "rb") as f:
             config = FeedbackDatasetConfig.parse_raw(f.read())
@@ -913,7 +918,7 @@ class FeedbackDataset:
                     fields={field.name: hfds[index][field.name] for field in cls.fields},
                     responses=list(responses.values()) or None,
                     external_id=hfds[index]["external_id"],
-                )
+                ).dict()
             )
         del hfds
         return cls
--- a/src/argilla/client/feedback.py
+++ b/src/argilla/client/feedback.py
@@ -58,21 +58,17 @@ class ValueSchema(BaseModel):
 
 
 class ResponseSchema(BaseModel):
-    values: Dict[str, ValueSchema]
-    status: Literal["submitted", "missing", "discarded"]
-
-
-class OfflineFeedbackResponse(ResponseSchema):
     user_id: Optional[UUID] = None
+    values: Dict[str, ValueSchema]
+    status: Literal["submitted", "missing", "discarded"] = "submitted"
 
     @validator("user_id", always=True)
     def user_id_must_have_value(cls, v):
         if not v:
             warnings.warn(
-                "`user_id` not provided for `OfflineFeedbackResponse`. So it will be"
-                " set to `None`. Which is not an issue, even though if you're planning"
-                " to log the record in Argilla, as it will be set automatically to the"
-                " current user's id."
+                "`user_id` not provided, so it will be set to `None`. Which is not an"
+                " issue, unless if you're planning to log the response in Argilla, as "
+                " it will be automatically set to the active `user_id`."
             )
         return v
 
@@ -97,7 +93,7 @@ class FeedbackRecord(BaseModel):
 class OfflineFeedbackRecord(BaseModel):
     id: Optional[str] = None
     fields: Dict[str, str]
-    responses: List[OfflineFeedbackResponse] = []
+    responses: List[ResponseSchema] = []
     external_id: Optional[str] = None
     inserted_at: Optional[str] = None
     updated_at: Optional[str] = None
--- a/src/argilla/client/feedback.py
+++ b/src/argilla/client/feedback.py
@@ -887,6 +887,7 @@ class FeedbackDataset:
                 raise ValueError(
                     f"Only one dataset can be loaded at a time, use `split` to select a split, available splits are: {', '.join(hfds.keys())}."
                 )
+            hfds = hfds[list(hfds.keys())[0]]
 
         for index in range(len(hfds)):
             responses = {}
--- a/src/argilla/client/feedback.py
+++ b/src/argilla/client/feedback.py
@@ -1061,5 +1061,5 @@ def feedback_dataset_in_argilla(
     else:
         try:
             return (True, datasets_api_v1.get_dataset(client=httpx_client, id=id).parsed)
-        except Exception as e:
+        except:
             return (False, None)
--- a/src/argilla/client/feedback.py
+++ b/src/argilla/client/feedback.py
@@ -12,8 +12,8 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+import logging
 import tempfile
-import warnings
 from typing import TYPE_CHECKING, Any, Dict, Iterator, List, Optional, Tuple, Union
 from uuid import UUID
 
@@ -52,6 +52,8 @@ if TYPE_CHECKING:
 FETCHING_BATCH_SIZE = 250
 PUSHING_BATCH_SIZE = 32
 
+_LOGGER = logging.getLogger(__name__)
+
 
 class ValueSchema(BaseModel):
     value: Union[StrictStr, StrictInt]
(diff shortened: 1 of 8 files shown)


src/argilla/client/sdk/v1/datasets/api.py
```diff
@@ -234,6 +234,22 @@ def add_records(
     """
     url = "/api/v1/datasets/{id}/records".format(id=id)
 
+    for record in records:
+        cleaned_responses = []
+        response_without_user_id = False
+        for response in record["responses"]:
+            if response["user_id"] is None:
+                if response_without_user_id is True:
```
gabrielmbmb, line 242: If `response_without_user_id` is always going to hold a `bool` value, then maybe we can remove the `is True` from the condition

src/argilla/client/sdk/v1/datasets/api.py
```diff
@@ -234,6 +234,22 @@ def add_records(
...
+                    warnings.warn(
+                        f"Multiple responses without `user_id` found in record {record}, so just the first one will be used while the rest will be ignored."
+                    )
+                    continue
+                else:
+                    response["user_id"] = client.get("api/me").json()["id"]
+                    response_without_user_id = True
```
gabrielmbmb:
```suggestion
            if response["user_id"] is None and response_without_user_id:
                warnings.warn(
                    f"Multiple responses without `user_id` found in record {record}, so just the first one will be used while the rest will be ignored."
                )
                continue
            else:
                response["user_id"] = client.get("api/me").json()["id"]
                response_without_user_id = True
```

alvarobartt: I'll apply the boolean change you mentioned above, but not this one, as we only need to retrieve the `user_id` from `/api/me` when the `user_id` is None, if it has any value, then we just keep it. It will fail anyway if the `user_id` doesn't exist in Argilla.

src/argilla/client/feedback.py
```diff
@@ -668,11 +668,17 @@ def from_argilla(
...
+                else FieldSchema.construct(**field.dict())
                 for field in datasets_api_v1.get_fields(client=httpx_client, id=existing_dataset.id).parsed
             ],
             questions=[
-                QuestionSchema.construct(**question.dict())
+                RatingQuestion.construct(**question.dict())
+                if question.settings["type"] == "rating"
+                else TextQuestion.construct(**question.dict())
+                if question.settings["type"] == "text"
+                else QuestionSchema.construct(**question.dict())
```
frascuchon: I think this line should raise an error instead of parsing an unexpected question type. Otherwise, we can have unexpected behavior. What do you think @alvarobartt?

alvarobartt: We can either raise and exception, or assume that the Argilla Server may have one version e.g. 1.9.0 with support for a new question type, while the Argilla Python Client is another version e.g. 1.8.0 with support for that question type, so raising and error may not be desired in that scenario.

Anyway, if you think we're safer raising the exception on `ValidationError`, then I'll change that 👍🏻 LMKWYT!

src/argilla/client/feedback.py
```diff
@@ -50,28 +50,25 @@
     )
 
 FETCHING_BATCH_SIZE = 250
+PUSHING_BATCH_SIZE = 32
```
davidberenstein1957, line 53: should we move these things to argilla._constants.py

src/argilla/client/sdk/v1/datasets/api.py
```diff
@@ -234,6 +234,26 @@ def add_records(
...
 
+    active_user_id = None
+
+    for record in records:
+        cleaned_responses = []
+        response_without_user_id = False
+        for response in record["responses"]:
+            if response["user_id"] is None:
+                if response_without_user_id is True:
+                    warnings.warn(
```
davidberenstein1957, line 247: @tomaarsen I believe we opted for using logger.warning() right? I don't fully recall the discussion.

davidberenstein1957 (commented):
These tiny thingies to talk through tentatively.

src/argilla/client/feedback.py
alvarobartt: Hmm I decided to keep those here for the moment since we don't want to mix v0 and v1 stuff 😩

src/argilla/client/sdk/v1/datasets/api.py
alvarobartt: Oh cool, then @tomaarsen if you confirm this I'll change this before merging, as well as the `pytest.warns` statements in the unit tests

tomaarsen: I think we discussed it a bit in #2456. However, I think in situations where you're not providing a special type of warning (e.g. DeprecationWarning), then I think logger.warning might suit Argilla best? In the end, we would just have to normalize throughout the full project.

alvarobartt: Ok for the moment I'm happy to make it `logger.warning` and we can come back to that issue once we release Argilla 1.8.0?

tomaarsen: Sounds like the best approach

src/argilla/client/feedback.py
frascuchon: I see. You mean allowing Python clients to provide question types without version upgrades, right? If we have no specific validation, could make sense. Anyway, also it makes sense to me also force a version upgrade to support new types.

src/argilla/client/feedback.py
```diff
@@ -50,28 +50,27 @@
...
 
 class ResponseSchema(BaseModel):
-    values: Dict[str, ValueSchema]
-    status: Literal["submitted", "missing", "discarded"]
-
-
-class OfflineFeedbackResponse(ResponseSchema):
     user_id: Optional[UUID] = None
+    values: Dict[str, ValueSchema]
+    status: Literal["submitted", "missing", "discarded"] = "submitted"
```
frascuchon: The `missing` value is not really a response status. This value is used as a filter to return records without user responses. As the response is not created yet, we cannot link this `missing` status to the response itself. So, you won't have responses with this status; it's a filter value.

alvarobartt: Thanks @frascuchon then I'll remove this for the moment! Also @nataliaElv asked about this, so now we have the confirmation, thanks 👍🏻

src/argilla/client/feedback.py
alvarobartt: Sure, then I'll just allow the current types and ignore anything that is not explicitly supported in the Argilla version used in the Python client 👍🏻

src/argilla/client/sdk/v1/datasets/api.py
```diff
@@ -234,6 +234,28 @@ def add_records(
...
+    active_user_id = None
+
+    for record in records:
+        cleaned_responses = []
+        response_without_user_id = False
+        if "responses" not in record:
+            continue
+        for response in record["responses"]:
+            if response["user_id"] is None:
+                if response_without_user_id is True:
```
frascuchon:
```suggestion
                if response_without_user_id:
```

src/argilla/client/sdk/v1/datasets/api.py
```diff
@@ -234,6 +234,28 @@ def add_records(
...
+                    warnings.warn(
+                        f"Multiple responses without `user_id` found in record {record}, so just the first one will be used while the rest will be ignored."
+                    )
+                    continue
+                else:
+                    if active_user_id is None:
+                        active_user_id = client.get("api/me").json()["id"]
+                    response["user_id"] = active_user_id
```
frascuchon, line 254: maybe we can move this to the top of the method to avoid fetching the user_id for each record.

src/argilla/client/sdk/v1/datasets/api.py
```diff
@@ -234,6 +234,28 @@ def add_records(
...
+    for record in records:
+        cleaned_responses = []
+        response_without_user_id = False
+        if "responses" not in record:
+            continue
+        for response in record["responses"]:
+            if response["user_id"] is None:
+                if response_without_user_id:
+                    warnings.warn(
+                        f"Multiple responses without `user_id` found in record {record}, so just the first one will be used while the rest will be ignored."
```
frascuchon, line 248: This warning will show each time we found a response without a user id and we already setup one for the active user, right?

alvarobartt: I think the `UserWarning` just shows once, as the default is to show just once the same warning in the same line, see https://docs.python.org/3/library/warnings.html#the-warnings-filter

frascuchon (approved):

PR #2979, opened 2023-05-24
Refactor : Stop using total records in front end

### LINKED ISSUE #2917: Remove the record counts from pagination
- Frontend: remove the record count (1 of ...) 
- Frontend: Hide from pagination (OK)
- Backend: stop computing the total number of records when fetching records.

keithCuniah:
# Description
Since the `total number records` is not include anymore in the call to fetch records (See #2917) , the value have to be removed from the feedback task.

List of impacts by components/model :

1/ FeedbackPagination.component.vue :
the total_records is used to compute the number of total page
it's also used to render the pagination component (if no total_records, the component is not rendered)

2/ CenterFeedbackTask.content.vue :
when total_records change, we reinitiate the flag areResponsesUntouched to true (this help us to show are not a toast on change route or on change filter)
also used to know in which page the user is in order to allow to go to next record or not on submit/discard

3/ RecordFeedbackTaskAndQuestionnaire.content.vue :
it's where the total_records is fetch and insert in the orm
it's used as a flag to show or not a message in case there is no record

4/ FeedbackTask.model.js && feedbackTask.queries.js :
where there is the attribute total_record && queries to update/get the value from the orm

=> When the total record will be removed from the api, we have to clean the front orm and adapt the 3 components to get the toal record from the metrics instead of the previous way

**Type of change**

**How Has This Been Tested**

- Feedback task only

### DIFF
--- a/frontend/components/feedback-task/footer/Pagination.component.vue
+++ b/frontend/components/feedback-task/footer/Pagination.component.vue
@@ -1,57 +1,34 @@
 <template>
   <div class="pagination">
-    <div class="number-of-records-by-page-area"></div>
     <div class="pagination__buttons">
       <BaseButton
         class="pagination__button"
         ref="prevButton"
-        @click="onPaginate(onClickPrev)"
-        :disabled="currentPage === 1"
+        @click="onClickPrev"
+        :disabled="isFirstPage"
       >
         <svgicon name="chevron-left" width="8" height="8" />
         {{ prevButtonMessage }}
       </BaseButton>
 
-      <div class="pagination__page-number-area" v-if="showPageNumber">
-        <BaseButton
-          v-for="page in totalPages"
-          :key="page"
-          class="pagination__button"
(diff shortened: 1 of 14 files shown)


frontend/components/feedback-task/footer/Pagination.component.vue
```diff
@@ -63,59 +40,28 @@ export default {
...
-    totalOfRecordMessage() {
-      return `${this.currentPage} of ${this.totalItems} records`;
+    isLastPage() {
+      // TODO: Expose as a prop
+      return false;
     },
     isPageAvailable() {
       return this.pageFromRoute <= this.totalPages;
     },
+    // TODO: Move to the parent
```
frascuchon, line 48: Do we still need this computed value? If not, we should remove it

leiyre (approved):

frascuchon (approved):

PR #2980, opened 2023-05-24
Show info to users when the version for argilla client and server are misaligned

### LINKED ISSUE #2967: Show info to users when the version for Argilla client and server are misaligned

frascuchon:
# Description

This PR checks the Argilla client and server versions and shows a compatibility warning if minor versions are not aligned.

Also, some unused functions/methods and variables have been dropped out.

Closes #2967 

**Type of change**

**How Has This Been Tested**

Test checking this behaviour are included 

**Checklist**


davidberenstein1957 (commented):
Look good to me. 

Thinking about this kind of error. 
I think it would also be good if we could provide a more explicit error when database migration/no-users a present in the database.

tests/client/test_api.py
```diff
@@ -747,3 +748,34 @@ def test_load_sort(mocked_client):
...
+    def mock_get_info(*args, **kwargs):
+        return ApiInfo(version="1.0.0")
+
+    def mock_whoami(*args, **kwargs):
+        return User(username="mock")
+
+    monkeypatch.setattr(Status, "get_info", mock_get_info)
+    monkeypatch.setattr(users_api, "whoami", mock_whoami)
+
+    with pytest.warns(UserWarning, match="You're connecting to Argilla Server 1.0.0 using a different client version"):
```
davidberenstein1957: Should wee not match the actual used version specifically here?

davidberenstein1957 (commented):
We have been getting a lot of model and dataset errors from Hugging Face lately, does it make sense to host a copy of the used datasets and models ourselves on GitHub for example?

frascuchon:
> Look good to me.
> 
> Thinking about this kind of error. I think it would also be good if we could provide a more explicit error when database migration/no-users a present in the database.

Sure! Unfortunately, this could not be tackled by the python client. This kind of error should be shown in the server startup.

tests/client/test_api.py
frascuchon: Yes! I've also included the client version in the match pattern. Thanks!

src/argilla/client/client.py
```diff
@@ -140,6 +139,25 @@ def __init__(
...
+
+    def _check_argilla_versions(self):
+        from argilla import __version__ as rg_version
+
+        api_info = Status(self.http_client).get_info()
+
+        client_version = rg_version.split(".")[:2]
+        server_version = api_info.version.split(".")[:2]
+        if client_version != server_version:
+            warnings.warn(
```
alvarobartt, line 152: After some slight discussion, we decided to use `logging.warning` instead of `warnings.warn`, I don't know if your opinionated on this @frascuchon

alvarobartt: But we can `import logging; _LOGGER = logging.getLogger(__name__); _LOGGER.warning(...)`, no need to specify the category, we can just provide the message to `_LOGGER.warning`

alvarobartt: But note that this still pending a detailed discussion cc @tomaarsen @davidberenstein1957

tests/client/test_api.py
```diff
@@ -747,3 +754,39 @@ def test_load_sort(mocked_client):
...
+    def mock_get_info(*args, **kwargs):
+        return ApiInfo(version=rg_version)
+
+    def mock_whoami(*args, **kwargs):
+        return User(username="mock")
+
+    monkeypatch.setattr(Status, "get_info", mock_get_info)
+    monkeypatch.setattr(users_api, "whoami", mock_whoami)
+
+    with pytest.warns(None) as record:
```
alvarobartt, line 790: And here I guess there's no need to catch the warning, since no warning will be raised, right?

alvarobartt (commented):
LGTM! Just some minor nits

src/argilla/client/client.py
frascuchon: If me it's okay to use logging instead of warnings. Not sure about what is the convention for this kind of message.

Maybe this doc section can help:
Ref: https://docs.python.org/3/howto/logging.html#when-to-use-logging

alvarobartt: Yes, indeed in this case seems more suitable to use `warnings.warn` that's true, as the user will need to do something in the client to fix the issue i.e. upgrade it

tests/client/test_api.py
frascuchon: Yes, this test is checking exactly this. The test condition is `len(record) == 0`, which means no warnings are raised.

davidberenstein1957 (approved):

PR #2988, opened 2023-05-25
Update toast message on queue

keithCuniah:
# Description
Update the toast message when user try to see record after the last one from a queue.
The message is : `You've reached the end of the data for the ${status} queue.` Where status is pending/discarded/submitted

**Type of change**

- Update toast message

**How Has This Been Tested**

- only feedback task


frascuchon (commented):
Just change the base branch to `releases/1.8.0` please.

frascuchon (approved):

PR #2989, opened 2023-05-25
Relax `codecov/patch` threshold to avoid failures

### LINKED ISSUE #2985: Update the `.codecov.yml` also for `patch`
**Describe the bug**

This is not a bug actually, but as of https://github.com/argilla-io/argilla/pull/2934 we increased the `codecov` threshold, but that was just affecting the `codecov/project` while the `codecov/patch` is still failing most of the times due to still having a pretty constrained margin, so we need to update the configuration for it in https://github.com/argilla-io/argilla/blob/develop/.codecov.yml

alvarobartt:
# Description

Relax the `codecov/patch` threshold to avoid CI from failing too often due to small code coverage percentage changes.

Closes #2985

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -6,3 +6,7 @@ coverage:
       default:
         target: auto
         threshold: 2%
+    patch:
+      default:
+        target: auto
+        threshold: 2%


gabrielmbmb (approved):
LGTM!

PR #2991, opened 2023-05-25
fix : copy feedback/old  dataset link

keithCuniah:
# Description
Since link attribute have been update in the dataset object, the copy link was not working in home page

I also applied the same function to detect if it's an old or not dataset (since the redirection is not the same for feedback task and the other old datasets)

**Type of change**

**How Has This Been Tested**

- Datasets lists page


frontend/pages/datasets/index.vue
```diff
@@ -252,6 +259,14 @@ export default {
     ...mapActions({
       fetchDatasets: "entities/datasets/fetchAll",
     }),
+    isOldTask(task) {
+      // NOTE - we need to detect ald/new task because the redirection correcponding pages does not have same route
```
frascuchon:
```suggestion
      // NOTE - we need to detect old/new task because the redirection corresponding pages does not have the same route
```

frontend/pages/datasets/index.vue
```diff
@@ -352,31 +367,33 @@ export default {
...
+    goToSetting({ id, workspace, name, task }) {
+      const isOldTask = this.isOldTask(task);
+
+      if (isOldTask) {
+        this.$router.push({
+          path: `/datasets/${workspace}/${name}/settings`,
+        });
+      } else {
+        this.$router.push({
+          path: `/dataset/${id}/settings`,
```
frascuchon: Could we use named routes here? This can help if we decide to change routes for datasets

See https://router.vuejs.org/guide/essentials/named-routes.html#named-routes

frascuchon (approved):
Just a question regarding the code maintainability

frontend/pages/datasets/index.vue
```diff
@@ -352,31 +367,38 @@ export default {
...
     },
-    copyUrl(dataset) {
-      const route = `${window.origin}${dataset.link}`;
-      this.copy(route);
+    copyUrl({ task, id, workspace, name }) {
+      let url = `${window.origin}`;
+      const isOldTask = this.isOldTask(task);
+
+      if (isOldTask) {
+        url += `/datasets/${workspace}/${name}`;
```
frascuchon, line 397: Include a TODO To don't forget it when URL will be changed

frascuchon (approved):

PR #2992, opened 2023-05-25
fix : missmatch status on navigation

keithCuniah:
# Description
Since the Record orm was not cleaned up before fetching new records on $fetch, a mismatch on the current recordStatus could appear

See next video at 0:18
https://github.com/argilla-io/argilla/assets/88380932/ac08df29-8d7f-4dfa-8e14-b6af71cdd89e

**Type of change**

**How Has This Been Tested**

- Only feedback task (when fetching record on fetch hook, we need to ensure that the orm is empty)


frontend/components/page-contents/feedback-task-content/RecordFeedbackTaskAndQuestionnaire.content.vue
```diff
@@ -50,7 +50,8 @@ import {
   isRecordWithRecordIndexByDatasetIdExists,
   isAnyRecordByDatasetId,
 } from "@/models/feedback-task-model/record/record.queries";
-
+import { deleteRecordResponsesByUserIdAndResponseId } from "@/models/feedback-task-model/record-response/recordResponse.queries";
+import { deleteAllRecordFields } from "@/models/feedback-task-model/record-field/recordField.queries";
```
frascuchon, line 54: Be sure that these functions are returning a promise. I had problems with `deleteAllRecords` functions with had no `async-await` definition.

frascuchon (approved):

PR #2993, opened 2023-05-25
feat: use different database for unit tests

### LINKED ISSUE #2987: Use different database to run the unit tests
The current setup for the unit tests uses the same engine with the same URL for connecting to the database that gets deployed when executing the Argilla server locally. This is not ideal, because if you have played with the server, added datasets, annotated, etc, and then run the unit tests, the database will get erased losing all the work done manually.

Apart from that, the current pytest fixture to yield the DB session is removing all the rows from the tables after the test gets executed instead of rolling back.

I would like a different database to be used to run the unit tests and to roll back the transactions instead of removing all the rows after the test gets executed.

gabrielmbmb:
# Description

I've updated the unit test setup to use a different database than the one used by the local deployment of the Argilla server. In addition, I've updated the pytest fixtures that provided a connection to the DB, and now before executing every test a new transaction is created that gets rollbacked when the execution of the test has finished, leaving no traces in the DB.

Also, I've aligned the SQLAlchemy models metadata with the metadata in the current alembic migrations scripts.

Closes #2987 

**Type of change**

**How Has This Been Tested**

All the unit tests are passing using this new DB for the tests.

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,10 @@ These are the section headers that we use:
 - Added boolean `use_markdown` property to `TextFieldSettings` model.
 - Added boolean `use_markdown` property to `TextQuestionSettings` model.
 
+## Changed
+
+- Database setup for unit tests. Now the unit tests use a different database than the one used by the local Argilla server ([#2987]).
+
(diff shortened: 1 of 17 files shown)


frascuchon (approved):
super nice!

PR #2994, opened 2023-05-26
fix : wrong init flag value for showing toast

keithCuniah:
# Description

The initial value of the flag to show or to not show the toast to prevent user loosing his current modifications was wrongly initiate. 

**Type of change**

**How Has This Been Tested**

Only feedback task

### DIFF
--- a/frontend/pages/dataset/_id/annotation-mode/index.vue
+++ b/frontend/pages/dataset/_id/annotation-mode/index.vue
@@ -53,7 +53,7 @@ export default {
   },
   data: () => {
     return {
-      areResponsesUntouched: false,
+      areResponsesUntouched: true, // NOTE - this flag is used to show or to not show a toast when questionnaire is touched (to prevent loosing current modification)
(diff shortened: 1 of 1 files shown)


frontend/pages/dataset/_id/annotation-mode/index.vue
```diff
@@ -53,7 +53,7 @@ export default {
   },
   data: () => {
     return {
-      areResponsesUntouched: false,
+      areResponsesUntouched: true, // NOTE - this flag is used to show or to not show a toast when questionnaire is touched (to prevent loosing current modification)
```
frascuchon, line 56: We need to set the same initialization value for other components using this logic (PaginationFeedback..., )

keithCuniah: I add "TODO" comment to tackle this in another moment

frascuchon (approved):

PR #3001, opened 2023-05-29
refactor: split `FeedbackDataset` and `pydantic.BaseModel`s

alvarobartt:
# Description

This PR basically restructures the former `argilla/client/feedback.py` to be more maintainable, by creating a separate module for it at `argilla/client/feedback/` where the `pydantic.BaseModel` schemas and the `FeedbackDataset` itself have been split.

So now the contents of `argilla/client/feedback` are:
* `schemas.py` with all the relevant `pydantic.BaseModel` schemas defined
* `dataset.py` with the `FeedbackDataset` class itself

Additionally, this PR also includes the `use_markdown` attribute in the `TextField` and `TextQuestion` schemas.

**Type of change**

**How Has This Been Tested**

- [X] Re-run the same unit tests but placed under a new module, and using the updated imports. No need to test anything else since from the user perspective nothing changed, just internally for maintenance/scalability/etc. reasons.

**Checklist**


src/argilla/client/feedback/schemas.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+class ValueSchema(BaseModel):
+    value: Union[StrictStr, StrictInt]
+
+
+class ResponseSchema(BaseModel):
+    user_id: Optional[UUID] = None
+    values: Dict[str, ValueSchema]
+    status: Literal["submitted", "discarded"] = "submitted"
```
gabrielmbmb, line 44: In #3033, which has been already merged to `develop`, we added a new possible status called `draft`. It's not being used yet by the UI, but I think that we can add it here already.

src/argilla/client/feedback/schemas.py
```diff
@@ -0,0 +1,157 @@
...
+        return v
+
+
+class FeedbackRecord(BaseModel):
+    fields: Dict[str, str]
+    responses: Optional[Union[ResponseSchema, List[ResponseSchema]]] = None
+    external_id: Optional[str] = None
+
+    @validator("responses", always=True)
+    def responses_must_be_a_list(cls, v):
```
gabrielmbmb:
```suggestion
    def responses_must_be_a_list(cls, v: Optional[Union[ResponseSchema, List[ResponseSchema]]]) -> List[ResponseSchema]:
```

src/argilla/client/feedback/schemas.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+class FieldSchema(BaseModel):
+    name: str
+    title: Optional[str] = None
+    required: Optional[bool] = True
+    settings: Dict[str, Any]
+
+    @validator("title", always=True)
+    def title_must_have_value(cls, v, values):
```
gabrielmbmb:
```suggestion
    def title_must_have_value(cls, v: Optional[str], values: Dict[str, Any]) -> str:
```

src/argilla/client/feedback/schemas.py
```diff
@@ -0,0 +1,157 @@
...
+            return values["name"].capitalize()
+        return v
+
+
+class TextField(FieldSchema):
+    settings: Dict[str, Any] = Field({"type": "text"})
+    use_markdown: bool = False
+
+    @validator("use_markdown", always=True)
+    def update_settings_with_use_markdown(cls, v, values):
```
gabrielmbmb:
```suggestion
    def update_settings_with_use_markdown(cls, v: bool, values: Dict[str, Any]) -> bool:
```

src/argilla/client/feedback/schemas.py
```diff
@@ -0,0 +1,157 @@
...
+
+class QuestionSchema(BaseModel):
+    name: str
+    title: Optional[str] = None
+    description: Optional[str] = None
+    required: Optional[bool] = True
+    settings: Dict[str, Any]
+
+    @validator("title", always=True)
+    def title_must_have_value(cls, v, values):
```
gabrielmbmb:
```suggestion
    def title_must_have_value(cls, v: Optional[str], values: Dict[str, Any]) ->str:
```

src/argilla/client/feedback/schemas.py
```diff
@@ -0,0 +1,157 @@
...
+        return v
+
+
+# TODO(alvarobartt): add `TextResponse` and `RatingResponse` classes
+class TextQuestion(QuestionSchema):
+    settings: Dict[str, Any] = Field({"type": "text", "use_markdown": False})
+    use_markdown: bool = False
+
+    @validator("use_markdown", always=True)
+    def update_settings_with_use_markdown(cls, v, values):
```
gabrielmbmb:
```suggestion
    def update_settings_with_use_markdown(cls, v: bool, values: Dict[str, Any]) -> bool:
```

src/argilla/client/feedback/schemas.py
```diff
@@ -0,0 +1,157 @@
...
+        validate_assignment = True
+        extra = Extra.forbid
+
+
+class RatingQuestion(QuestionSchema):
+    settings: Dict[str, Any] = Field({"type": "rating"})
+    values: List[int] = Field(unique_items=True)
+
+    @validator("values", always=True)
+    def update_settings_with_values(cls, v, values):
```
gabrielmbmb:
```suggestion
    def update_settings_with_values(cls, v: List[int], values: Dict[str, Any]) -> List[int]:
```

gabrielmbmb: not sure if in this case would be better to use a `pydantic.root_validator`, or use `@validator("settings")` and return the updated settings.

src/argilla/client/feedback/dataset.py
```diff
@@ -234,7 +140,7 @@ def __init__(
         self,
         *,
         fields: List[FieldSchema],
-        questions: List[AllowedQuestionTypes],
+        questions: List[Union[TextQuestion, RatingQuestion]],
```
gabrielmbmb, line 143: Maybe we should use here `argilla.client.feedback.schemas.AllowedQuestionTypes`?

src/argilla/client/feedback/utils.py
```diff
@@ -0,0 +1,136 @@
...
+        >>> from argilla.client.feedback.dataset import feedback_dataset_in_argilla
+        >>> fds_exists, fds_cls = feedback_dataset_in_argilla(name="my-dataset")
+    """
+    assert name or (name and workspace) or id, (
+        "You must provide either the `name` and `workspace` (the latter just if"
+        " applicable, if not the default `workspace` will be used) or the `id`, which"
+        " is the Argilla ID of the `rg.FeedbackDataset`."
+    )
+
+    httpx_client: "httpx.Client" = rg.active_client()._client.httpx
```
gabrielmbmb: why we don't use the `http_client` property here?

src/argilla/client/feedback/utils.py
```diff
@@ -0,0 +1,136 @@
...
+        ValueError: if the `workspace` is not a `rg.Workspace` instance or a string.
+        Exception: if the `FeedbackDataset` could not be listed from Argilla.
+
+    Examples:
+        >>> import argilla as rg
+        >>> rg.init(api_url="...", api_key="...")
+        >>> from argilla.client.feedback.dataset import feedback_dataset_in_argilla
+        >>> fds_exists, fds_cls = feedback_dataset_in_argilla(name="my-dataset")
+    """
+    assert name or (name and workspace) or id, (
```
gabrielmbmb:
```suggestion
    assert (name and workspace) or name or id, (
```

gabrielmbmb: Otherwise, if `name` is provided, the condition `(name or workspace)` will never be evaluated

src/argilla/client/feedback/utils.py
```diff
@@ -0,0 +1,136 @@
...
+        "You must provide either the `name` and `workspace` (the latter just if"
+        " applicable, if not the default `workspace` will be used) or the `id`, which"
+        " is the Argilla ID of the `rg.FeedbackDataset`."
+    )
+
+    httpx_client: "httpx.Client" = rg.active_client()._client.httpx
+
+    if name or (name and workspace):
```
gabrielmbmb:
```suggestion
    if (name and workspace) or name:
```

src/argilla/client/feedback/utils.py
```diff
@@ -0,0 +1,136 @@
...
+            raise ValueError(f"Workspace must be a `rg.Workspace` instance or a string, got {type(workspace)}")
+
+        try:
+            datasets = datasets_api_v1.list_datasets(client=httpx_client).parsed
+        except Exception as e:
+            raise Exception(f"Failed while listing the `FeedbackTask` datasets from Argilla with exception: {e}")
+
+        for dataset in datasets:
+            if dataset.name == name and dataset.workspace_id == workspace.id:
+                return (True, dataset)
```
gabrielmbmb:
```suggestion
                return True, dataset
```

src/argilla/client/feedback/utils.py
```diff
@@ -0,0 +1,136 @@
...
+        return (False, None)
```
gabrielmbmb:
```suggestion
        return False, None
```

src/argilla/client/feedback/utils.py
```diff
@@ -0,0 +1,136 @@
...
+    else:
+        try:
+            return (True, datasets_api_v1.get_dataset(client=httpx_client, id=id).parsed)
```
gabrielmbmb:
```suggestion
            return True, datasets_api_v1.get_dataset(client=httpx_client, id=id).parsed
```

src/argilla/client/feedback/utils.py
```diff
@@ -0,0 +1,136 @@
...
+        except:
+            return (False, None)
```
gabrielmbmb:
```suggestion
            return False, None
```

gabrielmbmb (commented):
Hi Alvaro, LGTM! I left some suggestions for some missing type hints and for some conditions.

Also, it would be good to add more unit tests for covering all the branches in `feedback_dataset_in_argilla` function.

src/argilla/client/feedback/dataset.py
alvarobartt: Yes, it's like that in the PR I created on top of this to add also the new question types 👍🏻

src/argilla/client/feedback/schemas.py
alvarobartt: My take here is that we shouldn't confuse the user with "useless" stuff, so maybe we can leave it as is until we decide what to do with the "draft" status?

src/argilla/client/feedback/schemas.py
alvarobartt: Indeed I also changed this in the PR for the new question-types, will forward it to you once published! 😄

src/argilla/client/feedback/utils.py
alvarobartt: True, we should use it instead of the protected attribute `_client`, nice catch!

src/argilla/client/feedback/schemas.py
gabrielmbmb: okay! we can add it later in another PR

gabrielmbmb (approved):

PR #3005, opened 2023-05-29
feat: add single label selection question to the API

### LINKED ISSUE #2995: Create single-choice API setting class
Add the required setting class that will allow the creation of single-choice questions in the API using the `POST /api/v1/datasets/:dataset_id/questions`

gabrielmbmb:
# Description

I've added the required class `LabelSelectionQuestionSettings` to create the new type of question label selection question (also known as a single choice question) and to validate the responses submitted for this kind of question. As discussed with @frascuchon, I've just added the classes to `src/argilla/server/models/questions.py` (not in `src/argilla/server/schemas/v1/datasets.py` this time) to test if that's something that we can do also for the rest of the questions and avoid the code duplication. 

In addition, I've refactored the `RatingQuestionSettings` class to extract the methods `option_values` and `check_response` to a new class that can be used both by `RatingQuestionSettings` and `LabelSelectionQuestionSettings`, as they perform the same kind of validation.

Also, I've added some extra unit tests to test the responses submitted for all the kinds of questions that are currently available.

Closes #2995 

**Type of change**

**How Has This Been Tested**

I've updated the unit tests to check all the possible cases in which a label selection question is created and responses for this kind of question are created.

**Checklist**


src/argilla/server/models/questions.py
```diff
@@ -52,17 +55,41 @@ class RatingQuestionSettingsOption(BaseModel):
...
-QuestionSettings = Annotated[Union[TextQuestionSettings, RatingQuestionSettings], Field(..., discriminator="type")]
+class RatingQuestionSettings(ValidOptionCheckerMixin[int], BaseQuestionSettings):
+    type: Literal[QuestionType.rating]
+    options: List[RatingQuestionSettingsOption]
+
+
+class LabelSelectionQuestionSettingsOption(BaseModel):
+    value: str
+    text: str
+    description: Optional[str] = None
```
frascuchon, line 75: Maybe make sense having some length limits here

src/argilla/server/schemas/v1/datasets.py
```diff
@@ -148,13 +154,19 @@ class RatingQuestionSettings(BaseModel):
     )
 
 
+QuestionSettings = Annotated[
```
alvarobartt, line 188: What if we import the `QuestionSettings` from `argilla/server/models/questions.py` instead to avoid defining this twice?

src/argilla/server/schemas/v1/datasets.py
```diff
@@ -18,14 +18,20 @@
 from uuid import UUID
 
 from pydantic import BaseModel, conlist, constr, validator
-from pydantic import Field as ModelField
+from pydantic import Field as PydanticField
```
alvarobartt, line 21: Do we actually need the `as PydanticField`, since this is the only `Field` import we're doing here, right?

src/argilla/server/search_engine.py
```diff
@@ -102,7 +102,7 @@ def _field_mapping_for_question(self, question: Question):
         if settings.type == QuestionType.rating:
             # See https://www.elastic.co/guide/en/elasticsearch/reference/current/number.html
             return {"type": "integer"}
-        elif settings.type == QuestionType.text:
+        elif settings.type in [QuestionType.text, QuestionType.label_selection]:
             # See https://www.elastic.co/guide/en/elasticsearch/reference/current/text.html
             return {"type": "text", "index": False}
```
frascuchon:
```suggestion
        elif settings.type in [QuestionType.text, QuestionType.label_selection]:
            # TODO: Review mapping for label selection. Could make sense to use `keyword` mapping instead. See https://www.elastic.co/guide/en/elasticsearch/reference/current/keyword.html
            # See https://www.elastic.co/guide/en/elasticsearch/reference/current/text.html
            return {"type": "text", "index": False}
```

src/argilla/server/models/questions.py
```diff
@@ -52,17 +55,41 @@ class RatingQuestionSettingsOption(BaseModel):
...
+
+
+class LabelSelectionQuestionSettings(ValidOptionCheckerMixin[str], BaseQuestionSettings):
+    type: Literal[QuestionType.label_selection]
+    options: conlist(
+        LabelSelectionQuestionSettingsOption,
+        min_items=LABEL_SELECTION_OPTIONS_MIN_ITEMS,
+        max_items=LABEL_SELECTION_OPTIONS_MAX_ITEMS,
+    )
+    visible_options: Optional[PositiveInt] = None
```
frascuchon: I remember why we separate schemas and models for the settings: the field validations make sense only for input schemas models. Maybe we should keep 2 models, for validation/non-validation scenarios.

src/argilla/server/schemas/v1/datasets.py
gabrielmbmb: Hi @alvarobartt, we cannot do that as the `pydantic.BaseModel` in `argilla/server/models/questions.py` are a bit different from the ones in `argilla/server/schemas/v1/datasets.py`.

src/argilla/server/models/questions.py
gabrielmbmb: I see, so I will update the PR to add the `pydantic.BaseModel` with the relaxed validation for when reading from the DB.

src/argilla/server/models/questions.py
gabrielmbmb: Yes, it makes sense. How about a maximum of 100 characters?

src/argilla/server/schemas/v1/datasets.py
frascuchon: I would align values with what we're doing for [question and fields names, titles, and descriptions](https://github.com/argilla-io/argilla/pull/3005/files#diff-154a4fe702792a783f557b4e95846e0f9f58dc1c430df9afd0e9d2206f2ef918R39): 200 for name, 500 for titles (can correspond to the text) and 1000 for description. 

We can better adjust those numbers later. My concern about lower values is to be too much restrictive.

alvarobartt: True @gabrielmbmb didn't see the code above, and thought you were just importing those `pydantic.BaseModel`s from the `questions.py` module 👍🏻

alvarobartt: True @gabrielmbmb didn't see the code above, and thought you were just importing those pydantic.BaseModels from the questions.py module 👍🏻

src/argilla/server/models/questions.py
frascuchon: I've put my comment but I didn't push it. We could align to the [questions and fields limits](https://github.com/argilla-io/argilla/pull/3005/files#diff-154a4fe702792a783f557b4e95846e0f9f58dc1c430df9afd0e9d2206f2ef918R35):

* `name`/`value`: 200
* `title`/`text`: 500
* `description`: 1000

My concern about lower values is to be too much restrictive.

src/argilla/server/schemas/v1/datasets.py
```diff
@@ -148,13 +158,46 @@ class RatingQuestionSettings(BaseModel):
...
+    text: constr(
+        min_length=LABEL_SELECTION_TEXT_MIN_LENGTH,
+        max_length=LABEL_SELECTION_TEXT_MAX_LENGTH,
+    )
+    description: Optional[
+        constr(
+            min_length=LABEL_SELECTION_DESCRIPTION_MIN_LENGTH,
+            max_length=LABEL_SELECTION_DESCRIPTION_MAX_LENGTH,
+        )
+    ] = PydanticField(None, max_length=100)
```
frascuchon: We can remove the `PydanticField` definition since we're using `constr`, right?
```suggestion
    ]
```

gabrielmbmb: yes! I forgot to push that. I've added some unit tests also to validate the min/max length for the label selection question option attributes.

frascuchon (approved):

PR #3012, opened 2023-05-29
chore: upgrade `flair` to 0.12

alvarobartt:
# Description

As older versions of `flair` (< 0.12) were having problems during the installation, I've decided to upgrade `flair` to 0.12 and pin it. Besides the package upgrade, some things had to be updated in the `FlairMonitor` for the unit tests to pass when installing the latest version of `flair`.

Some more information about the installation issue at https://github.com/flairNLP/flair/issues/3129

**Type of change**

**How Has This Been Tested**

- [x] Updated the `FlairMonitoring` unit tests

**Checklist**

### DIFF
--- a/src/argilla/monitoring/_flair.py
+++ b/src/argilla/monitoring/_flair.py
@@ -13,7 +13,7 @@
 #  limitations under the License.
 
 from datetime import datetime
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 from argilla import TokenClassificationRecord
 from argilla.client.api import Api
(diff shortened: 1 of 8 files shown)


src/argilla/monitoring/_flair.py
```diff
@@ -63,7 +65,7 @@ def predict(self, sentences: Union[List[Sentence], Sentence], *args, **kwargs):
         metadata = kwargs.pop("metadata", None)
         result = self.__model__.predict(sentences, *args, **kwargs)
 
-        if isinstance(sentences, Sentence):
+        if not isinstance(sentences, list):
```
frascuchon, line 68: Be aware of using this condition only will match for `list`. If you're using a generator or other kinds of sequences to provide a sentence batch, the `predict` wrapper will fail. Even if the original method (`model.predict`) is working

```python
instance((i for i in [1,2,3]), list) # False
```

alvarobartt: Even though you're right, I was following the type-hint for the `sentences` arg, that's why I kept that check as `isinstance(..., list)`, as I guess we're only expecting either a list of `Sentence` or a single `Sentence`.

gabrielmbmb (approved):
LGTM!

davidberenstein1957 (approved):
LGTM, but maybe the SetFit updates were supposed to go to another PR?

alvarobartt:
> LGTM, but maybe the SetFit updates were supposed to go to another PR?

Those were indeed part of another PR, see https://github.com/argilla-io/argilla/pull/3015, those are the commits from the merge 😄 (not sure why those were rebased anyway)

PR #3015, opened 2023-05-30
Fix function-args retrieval and file-deletion after train

alvarobartt:
# Description

Some `setfit` unit tests were not passing as it can be seen at https://github.com/argilla-io/argilla/actions/runs/5113755403/jobs/9193313124, with the exception `FileNotFound` which was due to the function defined in `tests/training/helpers.py` named `train_with_cleanup` was not properly cleaning the file, since it was trying to `shutil.rmtree` over a file that could or could not exist. I assume that issue was not appearing on other runs since probably the file was cached at some point.

Then after solving that issue, another issue appeared `TypeError: got multiple values for argument 'model_id'`, and `TypeError: got multiple values for argument 'revision'`, which was also leading the `setfit` tests to fail.

That was due to the function defined at `argilla/training/utils.py` named `get_default_args` was not working as expected, since it was having issues with the inheritance of the classmethod `from_pretrained`, and adding some args that were not from the original function such as `model_id` and `revision` which are args from the `_from_pretrained` method, so on, since `from_pretrained` keeps the `kwargs` and then propagates those to `_from_pretrained` (called as part of the return statement), it was receiving the `model_id` and `revision` args set during `from_pretrained`, as well as the ones propagated via `kwargs` since those are not valid args in `from_pretrained`. So on, I've fixed this by using `inspect.getfullargspec` instead, and that seems to work without any issue when inheriting from multiple classes or even inheriting from a MixIn.

**Type of change**

**How Has This Been Tested**

- [X] Re-run the `setfit` unit tests on Python 3.8, 3.9, and 3.10 and all those are passing properly now

**Checklist**

### DIFF
--- a/docs/_source/community/developer_docs.md
+++ b/docs/_source/community/developer_docs.md
@@ -38,7 +38,7 @@ git remote add upstream https://github.com/argilla-io/argilla.git
 
 ### Install dependencies
 
-To be able to use Argilla via either the Python client or the Python Command Line Interface (CLI), you'll need to 
+To be able to use Argilla via either the Python client or the Python Command Line Interface (CLI), you'll need to
 install `argilla` and its extra dependencies, if applicable. To do so, you can either install it via `pip` or setup
 a Conda environment with all the required dependencies.
 
@@ -75,7 +75,7 @@ conda env create -f environment_dev.yml
 conda activate argilla
 ```
 
-In the new Conda environment, *Argilla* will already be installed in [editable mode](https://pip.pypa.io/en/stable/cli/pip_install/#install-editable) 
+In the new Conda environment, *Argilla* will already be installed in [editable mode](https://pip.pypa.io/en/stable/cli/pip_install/#install-editable)
 with all the server dependencies. But if you're willing to install any other dependency you can do so via `pip` to install your own, or just
 see the available extras besides the `server` extras, which are: `listeners`, `postgresql`, and `tests`; all those installable as `pip install -e ".[<EXTRA_NAME>]"`.
 
--- a/src/argilla/training/setfit.py
+++ b/src/argilla/training/setfit.py
@@ -72,18 +72,14 @@ class ArgillaSetFitTrainer(ArgillaTransformersTrainer):
 
     def update_config(
         self,
-        **setfit_kwargs,
+        **kwargs,
     ):
         """
         Updates the `setfit_model_kwargs` and `setfit_trainer_kwargs` dictionaries with the keyword
         arguments passed to the `update_config` function.
         """
-        from setfit import SetFitTrainer
-
-        self.setfit_model_kwargs.update(setfit_kwargs)
-
-        self.setfit_trainer_kwargs.update(setfit_kwargs)
-        self.setfit_trainer_kwargs = filter_allowed_args(SetFitTrainer.__init__, **self.setfit_trainer_kwargs)
+        self.setfit_model_kwargs.update((k, v) for k, v in kwargs.items() if k in self.setfit_model_kwargs)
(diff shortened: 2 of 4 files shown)


tomaarsen (commented):
I don't love the readability of the new `default_args` definition. It'd already be better if we turn the in-line conditional into a normal if-statement.
Other than that, I trust that this produces the correct and desired default arguments, as much as possible anyways: `kwargs` will always be an issue.

alvarobartt:
> I don't love the readability of the new `default_args` definition. It'd already be better if we turn the in-line conditional into a normal if-statement. Other than that, I trust that this produces the correct and desired default arguments, as much as possible anyways: `kwargs` will always be an issue.

Sure, I'll update the code to be more readable/understandable, and we can also add the `kwargs`, feel free to check https://docs.python.org/3/library/inspect.html#inspect.getfullargspec

tomaarsen:
I've just tried it out, to see if it would capture the `kwargs`, and I get this:
```python
>>> inspect.getfullargspec(setfit.SetFitModel.from_pretrained)
FullArgSpec(args=[], varargs='args', varkw='kwargs', defaults=None, kwonlyargs=[], kwonlydefaults=None, annotations={'return': ~T})
>>> # For reference, the old behaviour:
>>> inspect.signature(setfit.SetFitModel.from_pretrained)
<Signature (pretrained_model_name_or_path: str, force_download: bool = False, resume_download: bool = False, proxies: Optional[Dict] = None, token: Union[str, bool, NoneType] = None, cache_dir: Optional[str] = None, local_files_only: bool = False, **model_kwargs) -> ~T>
```
This doesn't seem like what we want. What do you get, here?

alvarobartt:
See that tests are passing below:

image

docs/_source/community/developer_docs.md
```diff
@@ -38,7 +38,7 @@ git remote add upstream https://github.com/argilla-io/argilla.git
 
 ### Install dependencies
 
-To be able to use Argilla via either the Python client or the Python Command Line Interface (CLI), you'll need to 
+To be able to use Argilla via either the Python client or the Python Command Line Interface (CLI), you'll need to
 install `argilla` and its extra dependencies, if applicable. To do so, you can either install it via `pip` or setup
```
davidberenstein1957, line 42: should we mention the extras specifically?

src/argilla/training/utils.py
```diff
@@ -18,7 +18,7 @@
 args_to_remove = ["self", "cls", "model_args", "args", "kwargs", "model_kwargs"]
 
 
-def get_default_args(func):
+def get_default_args(fn):
```
davidberenstein1957:
```suggestion
def get_default_args(fn) -> dict:
```

src/argilla/training/setfit.py
```diff
@@ -62,18 +72,14 @@ def init_training_args(self):
 
     def update_config(
         self,
-        **setfit_kwargs,
+        **kwargs,
     ):
```
davidberenstein1957:
```suggestion
    ) -> None: 
```

src/argilla/training/setfit.py
```diff
@@ -47,7 +47,17 @@ def __init__(self, *args, **kwargs):
     def init_training_args(self):
```
davidberenstein1957:
```suggestion
    def init_training_args(self) -> None:
```

davidberenstein1957 (approved):
some hint and comment

docs/_source/community/developer_docs.md
alvarobartt: What do you mean with that? Anyway the change on this file should be tackled in the parent PR not in this once, as that was automatically fixed by the pre-commit bot without me noticing :( It was just the last space removal BTW

PR #3020, opened 2023-05-30
Add example in `add_records` docstring & fix issue in `FeedbackDataset` docstring

alvarobartt:
# Description

This PR just adds an example in the `add_records` docstring in `FeedbackDataset`, as well as fixing an issue in the `FeedbackDataset` docstring as `responses` were not logged as they should.

**Type of change**

- [X] Documentation update

**Checklist**

### DIFF
--- a/src/argilla/client/feedback.py
+++ b/src/argilla/client/feedback.py
@@ -446,6 +446,43 @@ class FeedbackDataset:
 
         Raises:
             ValueError: if the given records do not match the expected schema.
+
+        Examples:
+            >>> import argilla as rg
+            >>> rg.init(api_url="...", api_key="...")
+            >>> dataset = rg.FeedbackDataset(
+            ...     fields=[
+            ...         rg.TextField(name="text", required=True),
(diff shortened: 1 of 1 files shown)


src/argilla/client/feedback.py
```diff
@@ -446,6 +446,43 @@ def add_records(
...
         Raises:
             ValueError: if the given records do not match the expected schema.
+
+        Examples:
+            >>> import argilla as rg
+            >>> rg.init(api_url="...", api_key="...")
+            >>> dataset = rg.FeedbackDataset(
+            ...     fields=[
+            ...         rg.TextField(name="text", required=True),
+            ...         rg.TextField(name="label", required=True),
```
frascuchon, line 456: This is a comment for the future. Maybe we could separate free-text fields from term-like or class-like ones. Apart from semantics differences, for the second one you can improve the metrics or searches, since the search space can be reduced to a bunch of terms, instead of a full-text search.

frascuchon (approved):

PR #3024, opened 2023-05-30
Release v1.8.0

frascuchon:


src/argilla/server/schemas/v1/datasets.py
```diff
@@ -131,6 +132,7 @@ class FieldCreate(BaseModel):
 
 class TextQuestionSettings(BaseModel):
     type: Literal[QuestionType.text]
+    use_markdown: bool = False
```
alvarobartt, line 135: @frascuchon @davidberenstein1957 was this intended? Should we release 1.8.0 with `use_markdown`? Wasn't this intended for the 1.9.0?

frascuchon: Yes, some merge from develop somewhere

PR #3036, opened 2023-05-31
docs: refactor code-blocks that were looping over `records` as a `dict` instead of `pydantic.BaseModel`

alvarobartt:
# Description

As recently discussed with @frascuchon we should return the `FeedbackDataset.records` as `pydantic.BaseModel`s instead of Python dictionaries, since those are more user friendly and we should keep the validation layer for future releases.

**Type of change**

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] Refactor (change restructuring the codebase without changing functionality)
- [X] Documentation update

**How Has This Been Tested**

- [X] Re-run all the Jupyter Notebooks and scripts to make sure those work, even fixed some that were not working before (neither with the previous approach)

**Checklist**

- [X] I have merged the original branch into my forked branch
- [X] I made corresponding changes to the documentation

**Missing**

In `docs/_source/guides/llms/conceptual_guides/rm.md` there's a code-block that mentions triplets, and that's assuming that some fields are created in the `FeedbackDataset` while in the definition on a code-block above those are not created, is that an extra piece of code, or should we create the fields `response 1`, `response 2`, and `response 3` mentioned there?

PR #3038, opened 2023-05-31
refactor: `FeedbackDataset.records` are now `FeedbackRecord` instead of Python `dict`

alvarobartt:
# Description

As of a recent discussion with @frascuchon we should return the `FeedbackRecord` `pydantic.BaseModel` itself, instead of the Python dict conversion of it. So this PR basically removes all the `.dict()` conversion towards the Python client, while it's kept in the API. Besides that we also use `pydantic.BaseModel` validation instead of using `.construct` when there are nested `pydantic.BaseModels`.

Docs have been updated accordingly at #3036 

**Type of change**

- [X] Documentation update

**How Has This Been Tested**

- [X] Updated unit tests accordingly

**Checklist**

### DIFF
--- a/src/argilla/client/feedback.py
+++ b/src/argilla/client/feedback.py
@@ -218,13 +218,13 @@ class FeedbackDataset:
         ...     ]
         ... )
         >>> dataset.records
-        [FeedbackRecord(fields={"text": "This is the first record", "label": "positive"}, responses=[{"user_id": None, "values": {"question-1": {"value": "This is the first answer"}, "question-2": {"value": 5}}}])]
+        [FeedbackRecord(fields={"text": "This is the first record", "label": "positive"}, responses=[ResponseSchema(user_id=None, values={"question-1": ValueSchema(value="This is the first answer"), "question-2": ValueSchema(value=5)})])]
         >>> dataset.push_to_argilla(name="my-dataset", workspace="my-workspace")
         >>> dataset.argilla_id
         "..."
         >>> dataset = rg.FeedbackDataset.from_argilla(argilla_id="...")
         >>> dataset.records
-        [FeedbackRecord(fields={"text": "This is the first record", "label": "positive"}, responses=[{"user_id": None, "values": {"question-1": {"value": "This is the first answer"}, "question-2": {"value": 5}}}])]
+        [FeedbackRecord(fields={"text": "This is the first record", "label": "positive"}, responses=[ResponseSchema(user_id=None, values={"question-1": ValueSchema(value="This is the first answer"), "question-2": ValueSchema(value=5)})])]
     """
 
     argilla_id: Optional[str] = None
@@ -474,7 +474,7 @@ class FeedbackDataset:
             ...     ]
             ... )
             >>> dataset.records
-            [FeedbackRecord(fields={"text": "This is the first record", "label": "positive"}, responses=[{"user_id": None, "values": {"question-1": {"value": "This is the first answer"}, "question-2": {"value": 5}}}])]
+            [FeedbackRecord(fields={"text": "This is the first record", "label": "positive"}, responses=[ResponseSchema(user_id=None, values={"question-1": ValueSchema(value="This is the first answer"), "question-2": ValueSchema(value=5)})])]
         """
         if isinstance(records, list):
             records = [FeedbackRecord(**record) if isinstance(record, dict) else record for record in records]
--- a/src/argilla/client/feedback.py
+++ b/src/argilla/client/feedback.py
@@ -437,6 +437,7 @@ class FeedbackDataset:
                 with the fields of the record.
 
         Raises:
+            ValueError: if the given records are an empty list.
             ValueError: if the given records are neither: `FeedbackRecord`, list of `FeedbackRecord`,
                 list of dictionaries as a record or dictionary as a record.
             ValueError: if the given records do not match the expected schema.
@@ -479,6 +480,8 @@ class FeedbackDataset:
             [FeedbackRecord(fields={"text": "This is the first record", "label": "positive"}, responses=[ResponseSchema(user_id=None, values={"question-1": ValueSchema(value="This is the first answer"), "question-2": ValueSchema(value=5)})])]
         """
         if isinstance(records, list):
+            if len(records) == 0:
+                raise ValueError("Expected `records` to be a non-empty list of `dict` or `rg.FeedbackRecord`.")
             new_records = []
             for record in records:
                 if isinstance(record, dict):
--- a/src/argilla/client/feedback.py
+++ b/src/argilla/client/feedback.py
@@ -437,6 +437,8 @@ class FeedbackDataset:
                 with the fields of the record.
 
         Raises:
+            ValueError: if the given records are neither: `FeedbackRecord`, list of `FeedbackRecord`,
+                list of dictionaries as a record or dictionary as a record.
             ValueError: if the given records do not match the expected schema.
 
         Examples:
@@ -477,11 +479,23 @@ class FeedbackDataset:
             [FeedbackRecord(fields={"text": "This is the first record", "label": "positive"}, responses=[ResponseSchema(user_id=None, values={"question-1": ValueSchema(value="This is the first answer"), "question-2": ValueSchema(value=5)})])]
         """
         if isinstance(records, list):
-            records = [FeedbackRecord(**record) if isinstance(record, dict) else record for record in records]
-        if isinstance(records, dict):
+            for record in records:
+                if isinstance(record, dict):
+                    self.add_records(record)
+                elif isinstance(record, FeedbackRecord):
+                    self.add_records(record)
+                else:
+                    raise ValueError(
+                        f"Expected `records` to be a list of `dict` or `rg.FeedbackRecord`, got type {type(record)} instead."
+                    )
+        elif isinstance(records, dict):
             records = [FeedbackRecord(**records)]
-        if isinstance(records, FeedbackRecord):
+        elif isinstance(records, FeedbackRecord):
             records = [records]
+        else:
+            raise ValueError(
+                f"Expected `records` to be a `dict` or `rg.FeedbackRecord`, got type {type(records)} instead."
+            )
 
         if self.__fields_schema is None:
             self.__fields_schema = generate_pydantic_schema(self.fields)
--- a/src/argilla/client/feedback.py
+++ b/src/argilla/client/feedback.py
@@ -970,7 +970,7 @@ class FeedbackDataset:
                     fields={field.name: hfds[index][field.name] for field in cls.fields},
                     responses=list(responses.values()) or None,
                     external_id=hfds[index]["external_id"],
-                ).dict()
+                )
             )
         del hfds
         return cls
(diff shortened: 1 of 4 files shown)


alvarobartt:
Just as a side-note @frascuchon, the current script in `scripts/load_data.py` should remain as-is, since now we're using the `pydantic.BaseModel` so it will work fine, anyway, let me know 👍🏻  I've run it on my side and it works fine, but let me know whether that works for you or not

src/argilla/client/feedback.py
```diff
@@ -91,16 +90,7 @@ def responses_must_be_a_list(cls, v):
...
-
-
-class OfflineFeedbackRecord(BaseModel):
-    id: Optional[str] = None
-    fields: Dict[str, str]
-    responses: List[ResponseSchema] = []
-    external_id: Optional[str] = None
-    inserted_at: Optional[str] = None
-    updated_at: Optional[str] = None
+        extra = Extra.ignore
```
frascuchon, line 94: Can we show a warning if more properties than expected are passed?  Just to let it know  to the user

src/argilla/client/feedback.py
```diff
@@ -228,13 +218,13 @@ class FeedbackDataset:
         ...     ]
         ... )
         >>> dataset.records
-        [FeedbackRecord(fields={"text": "This is the first record", "label": "positive"}, responses=[{"user_id": None, "values": {"question-1": {"value": "This is the first answer"}, "question-2": {"value": 5}}}])]
+        [FeedbackRecord(fields={"text": "This is the first record", "label": "positive"}, responses=[ResponseSchema(user_id=None, values={"question-1": ValueSchema(value="This is the first answer"), "question-2": ValueSchema(value=5)})])]
```
frascuchon, line 221:
```suggestion
        [FeedbackRecord(fields={"text": "This is the first record", "label": "positive"}, responses=[ResponseSchema(values={"question-1": ValueSchema(value="This is the first answer"), "question-2": ValueSchema(value=5)})])]
```

src/argilla/client/feedback.py
```diff
@@ -228,13 +218,13 @@ class FeedbackDataset:
...
         >>> dataset.push_to_argilla(name="my-dataset", workspace="my-workspace")
         >>> dataset.argilla_id
         "..."
         >>> dataset = rg.FeedbackDataset.from_argilla(argilla_id="...")
         >>> dataset.records
-        [FeedbackRecord(fields={"text": "This is the first record", "label": "positive"}, responses=[{"user_id": None, "values": {"question-1": {"value": "This is the first answer"}, "question-2": {"value": 5}}}])]
+        [FeedbackRecord(fields={"text": "This is the first record", "label": "positive"}, responses=[ResponseSchema(user_id=None, values={"question-1": ValueSchema(value="This is the first answer"), "question-2": ValueSchema(value=5)})])]
```
frascuchon, line 227:
```suggestion
        [FeedbackRecord(fields={"text": "This is the first record", "label": "positive"}, responses=[ResponseSchema(values={"question-1": ValueSchema(value="This is the first answer"), "question-2": ValueSchema(value=5)})])]
```

src/argilla/client/feedback.py
```diff
@@ -412,7 +402,7 @@ def fetch_records(self) -> None:
             first_batch = datasets_api_v1.get_records(
                 client=httpx_client, id=self.argilla_id, offset=0, limit=FETCHING_BATCH_SIZE
             ).parsed
-            self.__records = first_batch.items
+            self.__records = [FeedbackRecord(**item.dict()) for item in first_batch.items]
```
frascuchon: Should we use `FeedbackRecord.parse_obj` instead? Not sure what's the difference

src/argilla/client/feedback.py
```diff
@@ -426,7 +416,7 @@ def fetch_records(self) -> None:
                         offset=FETCHING_BATCH_SIZE * current_batch,
                         limit=FETCHING_BATCH_SIZE,
                     ).parsed
-                    records = batch.items
+                    records = [FeedbackRecord(**item.dict()) for item in batch.items]
```
frascuchon: ditto

src/argilla/client/feedback.py
```diff
@@ -484,7 +474,7 @@ def add_records(
             ...     ]
             ... )
             >>> dataset.records
-            [FeedbackRecord(fields={"text": "This is the first record", "label": "positive"}, responses=[{"user_id": None, "values": {"question-1": {"value": "This is the first answer"}, "question-2": {"value": 5}}}])]
+            [FeedbackRecord(fields={"text": "This is the first record", "label": "positive"}, responses=[ResponseSchema(user_id=None, values={"question-1": ValueSchema(value="This is the first answer"), "question-2": ValueSchema(value=5)})])]
```
frascuchon, line 481: same change here, since the `user_id` is an optional value and we're setting it to `None` in the example, makes sense to remove it from the response initialization.

src/argilla/client/feedback.py
alvarobartt: I don't think we can combine `Extra.ignore` with a `root_validator` to make sure that just the existing fields are provided, maybe @gabrielmbmb knows a cleaner way to do so?

src/argilla/client/feedback.py
```diff
@@ -498,20 +488,18 @@ def add_records(
 
         for record in records:
             try:
-                record.fields = self.__fields_schema.parse_obj(record.fields)
+                self.__fields_schema.parse_obj(record.fields)
```
frascuchon, line 511: This line, is just to apply the fields_schema validation, right? I understand that fields will be accessed as dicts instead of class instances:

```python
records.fields["text"]
```
instead of 
```python
records.fields.text
```

src/argilla/client/feedback.py
```diff
@@ -91,16 +90,7 @@ def responses_must_be_a_list(cls, v):
         return v
 
     class Config:
-        extra = Extra.forbid
-
-
-class OfflineFeedbackRecord(BaseModel):
```
davidberenstein1957, line 97: Do all Pydantic models doo validation during attribute assignment?

src/argilla/client/feedback.py
alvarobartt: Yes, since before we were mapping the `FeedbackRecord.fields` to `FieldsSchema`, but we were calling `construct` over `FeedbackRecord` instead of creating it as `FeedbackRecord(...)`, so it was still a Python dict, adding that doesn't make too much sense for the moment, so we're just validating the `fields` based on `FeedbackDataset.fields` auto-generated `pydantic` schema

src/argilla/client/feedback.py
alvarobartt: So we can either `FeedbackRecord(**item.dict())` or `FeedbackRecord.parse_obj(item.dict())`, it's the same thing in this case

src/argilla/client/feedback.py
```diff
@@ -484,7 +474,7 @@ def add_records(
             ...     ]
             ... )
             >>> dataset.records
-            [FeedbackRecord(fields={"text": "This is the first record", "label": "positive"}, responses=[{"user_id": None, "values": {"question-1": {"value": "This is the first answer"}, "question-2": {"value": 5}}}])]
+            [FeedbackRecord(fields={"text": "This is the first record", "label": "positive"}, responses=[ResponseSchema(user_id=None, values={"question-1": ValueSchema(value="This is the first answer"), "question-2": ValueSchema(value=5)})])]
         """
         if isinstance(records, list):
```
davidberenstein1957, line 483: I think `elif` statements would be better suited for the other cases

src/argilla/client/feedback.py
alvarobartt: Also note that we want to ignore the extras, also the ones that come from Argilla for the moment i.e. `updated_at`, `created_at`, etc. fields that are not required/needed by the user, also to keep consistency for the records. So when we log records to Argilla and then we want to pull them, just the `FeedbackRecord` fields are kept from the API response

src/argilla/client/sdk/v1/datasets/api.py
```diff
@@ -207,7 +207,7 @@ def get_records(
     )
 
     if response.status_code == 200:
-        parsed_response = FeedbackRecordsModel.construct(**response.json())
+        parsed_response = FeedbackRecordsModel(**response.json())
```
frascuchon, line 210: `parse_obj` like?

src/argilla/client/sdk/v1/datasets/api.py
```diff
@@ -286,7 +286,7 @@ def get_fields(
     response = client.get(url=url)
 
     if response.status_code == 200:
-        parsed_response = [FeedbackFieldModel.construct(**item) for item in response.json()["items"]]
+        parsed_response = [FeedbackFieldModel(**item) for item in response.json()["items"]]
```
frascuchon, line 289: ditto

src/argilla/client/feedback.py
alvarobartt: So the thing is that since it's optional it's set to `None` by default, so when we print the `FeedbackRecord` we see the `responses` as a `ResponseSchema` with all its values, so I think it's nice that we keep it as-is in the docstrings, as that's what you get when you `print(self.records)`, and also to give the users visibility about the `user_id` field in `ResponseSchema`

src/argilla/client/feedback.py
alvarobartt: Yes they do, unless you call `.construct()`, that just builds it as a `pydantic.BaseModel` ignoring the nested `pydantic.BaseModel`s and assuming those are Python dicts, which is not what we want here. So instantiating or calling `parse_obj` over a `pydantic.BaseModel` does validate it 😄

src/argilla/client/feedback.py
frascuchon: I realize that `item` is an object. Should we use [`from_orm`](https://docs.pydantic.dev/latest/usage/models/#orm-mode-aka-arbitrary-class-instances)?  So, we don't need to convert them to dict.

src/argilla/client/sdk/v1/datasets/api.py
alvarobartt: Same thing as above, we removed the usage of `.construct` as that's not doing the validation and the nested `pydantic.BaseModel`s are lost and transformed into Python dictionaries, but here we could either do the current or `FeedbackRecordsModel.parse_obj(response.json())`, the outcome will be the same

src/argilla/client/feedback.py
alvarobartt: Those are non-colliding if statements, but sure, I can make those `elif` instead, even though there's not a clear `else` condition, should we raise `ValueError`?

src/argilla/client/feedback.py
alvarobartt: Sure we can also do that, but we need to update the `Config` then, right?

alvarobartt: But I like this to avoid converting to `dict()` in between whenever we receive `pydantic.BaseModel`s already

frascuchon (approved):

frascuchon (approved):

src/argilla/client/feedback.py
alvarobartt: I've used [`parse_obj_as`](https://docs.pydantic.dev/latest/usage/models/#parsing-data-into-a-specified-type), as we're moving from a `pydantic.BaseModel` to another `pydantic.BaseModel`, not from a generic class, so I think this may be more aligned to what we want 😄

src/argilla/client/feedback.py
alvarobartt: Same here, `parse_obj_as` has been used instead!

PR #3042, opened 2023-05-31
♻️ Front : updates to prepare single label

keithCuniah:
# Description
Two updates have to be done to prepare  `SINGLE_LABEL` component development
1/ in the backend it's not anymore `single_label` but `label_selection`
2/ in the backend development for the `SINGLE_LABEL`, since the `value` (value of the response) and the `text` (text to show in the response) are dissociated, updates in common in logics between form components have been made (when reparing questions with responses) 
3/ instead of `value_in_api`, the `value` and the `is_selected` attibute have been implemented in the options from the form to know which value is selected or not

For a future refactoring, the states of component (`is_selected`) will be removed from the orm

**Type of change**

**How Has This Been Tested**

- Only feedback task


frontend/components/commons/forms/questions-form/QuestionsForm.component.vue
```diff
@@ -27,7 +27,7 @@
           v-if="input.component_type === COMPONENT_TYPE.FREE_TEXT"
           :title="input.question"
           :placeholder="input.placeholder"
-          v-model="input.options[0].text"
+          v-model="input.options[0].value_in_api"
```
leiyre: Just a comment, I feel a bit confused with value_in_api name in a dynamic frontend v-model

frontend/components/commons/forms/questions-form/QuestionsForm.component.vue
```diff
@@ -443,12 +443,39 @@ export default {
 
               switch (componentType) {
                 case COMPONENT_TYPE.SINGLE_LABEL:
+                  formattedOptions = formattedOptions.map((option) => {
+                    if (option.value_in_api === newResponse.value) {
```
leiyre: Two objects are almost the same, is it possible to make the condition at object value level?

frontend/components/commons/forms/questions-form/QuestionsForm.component.vue
keithCuniah: I replace it by `value`

leiyre (approved):

frascuchon (commented):
We need to move all the options logic to the proper components and keep at ORM level only the question `settings` info.

frascuchon (approved):

PR #3055, opened 2023-06-01
feat: check duplicated values for rating, label selection and multi label selection questions

gabrielmbmb:
# Description

This PR adds a new class called `UniqueValuesCheckerMixin` which includes a `pydantic.root_validator` to check that the items in the `options` attribute of the `RatingQuestionSettings`, `LabelSelectionQuestionSettings` and `MultiLabelSelectionQuestionSettings` classes are not duplicated based on the `value` attribute of each option.

**Type of change**

**How Has This Been Tested**

Manually calling the API and I've added new test cases to the `test_create_dataset_question_with_invalid_settings` unit test to check that when creating a rating or label selection question with duplicated values a `422` status code is returned by the API.

**Checklist**


src/argilla/server/schemas/v1/datasets.py
```diff
@@ -145,11 +145,27 @@ class TextQuestionSettings(BaseModel):
     use_markdown: bool = False
 
 
+class UniqueValuesCheckerMixin(BaseModel):
+    @root_validator
+    def check_unique_values(cls, values: Dict[str, Any]) -> Dict[str, Any]:
+        options = values.get("options", [])
+        seen = set()
+        duplicates = set()
+        for option in options:
```
davidberenstein1957, line 154:
```python
options = [opt.value for opt in values.get("options", [])]
if len(options) != len(set(options)):
    raise ValueError(f"Option values must be unique, found duplicates: {duplicates}")
```

davidberenstein1957 (approved):
maybe we can write the check a bit shorter?

src/argilla/server/schemas/v1/datasets.py
gabrielmbmb: yes, I was thinking to do it that way, but then we cannot provide the list of duplicate values in the exception message (it might be not important).

davidberenstein1957: Does not matter much. Whatever you prefer:)

frascuchon: Yes, I think is useful to return the duplicated values. We could reduce some lines with:

```python
seen = set()
duplicates = [option.value for option in options if option.value in seen or seen.add(option.value)]
```

but I think is not clear to read.

gabrielmbmb: For me it is clear enough to read, but yes, it might be better to use the "long" syntax to make it clearer.

frascuchon (approved):

PR #3056, opened 2023-06-01
fix: wrong `FeedbackDatasetConfig` parsing in `from_huggingface`

alvarobartt:
# Description

When loading back a dataset previously pushed to the HuggingFace Hub via `push_to_huggingface`, the config was not being properly parsed when reading it from the HuggingFace Hub repository, as there was an `Union` in the `pydantic.BaseModel` that was ignored. So when the first value was parsed e.g. `TextQuestion`, then that model was kept and the remaining values were parsed using the same model, even though the type-hint was e.g. `Union[TextQuestion, RatingQuestion, ...]`.

Using `Config.smart_union = True` solves it, as the `Union` is properly identified and the parsing is "smart".

Besides that, I also had to restrict the `pydantic.BaseModel`s to be unique, since both could parse any question, so adding `Config.extra = forbid` solves it.

**Type of change**

**How Has This Been Tested**

- [x] Add unit tests to make sure that the types of the loaded `fields` and `questions` from the `FeedbackDatasetConfig` match the ones originally dumped

**Checklist**

### DIFF
--- a/src/argilla/client/feedback/schemas.py
+++ b/src/argilla/client/feedback/schemas.py
@@ -147,3 +147,6 @@ class FeedbackDatasetConfig(BaseModel):
     fields: List[AllowedFieldTypes]
     questions: List[AllowedQuestionTypes]
     guidelines: Optional[str] = None
+
+    class Config:
+        smart_union = True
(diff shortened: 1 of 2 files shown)


gabrielmbmb (approved):
LGTM!

PR #3064, opened 2023-06-02
:sparkles: Feat : Multi choice component

### LINKED ISSUE #2976: Create multi-choice question component

keithCuniah:
# Description
Front component for multi-choice component

Closes #2976 

**Type of change**

**How Has This Been Tested**

- Only feedback task


frontend/components/commons/forms/questions-form/fields/multi-label/MultiLabel.component.vue
```diff
@@ -0,0 +1,125 @@
...
+    this.uniqueOptions = this.options.reduce((accumulator, current) => {
+      if (!accumulator.find((item) => item.id === current.id)) {
+        accumulator.push(current);
+      }
+      return accumulator;
+    }, []);
+  },
+  computed: {
+    showSearch() {
+      return this.uniqueOptions.length >= 12;
```
frascuchon: Create a constant instead of a magic number. I saw this number also for SingleLabel... You should use the constant in both places
```suggestion
      return this.uniqueOptions.length >= constans.OPTIONS_THRESHOLD_TO_ENABLE_SEARCH;
```

frontend/components/commons/forms/questions-form/fields/multi-label/MultiLabelSelection.component.vue
```diff
@@ -0,0 +1,215 @@
+<template>
+  <div class="container">
+    <div class="component-header" v-if="showSearch || showCollapseButton">
```
frascuchon: What happens if the `div` section is always rendered? Will affect how the search bar, the collapse button, and the labels will be shown?

frontend/components/commons/forms/questions-form/fields/multi-label/MultiLabel.component.vue
```diff
@@ -0,0 +1,125 @@
...
+        badge-border-color="white"
+        v-tooltip="{ content: tooltipMessage, backgroundColor: '#FFF' }"
+      />
+    </div>
+
+    <MultiLabelSelectionComponent
+      :options="uniqueOptions"
+      :componentId="questionId"
+      :showSearch="showSearch"
+      :multiple="true"
```
frascuchon: I don't see this property defined in `MultiLabelSelectionoComponent`. Are we using int for something?

frontend/components/commons/forms/questions-form/fields/multi-label/MultiLabelSelection.component.vue
frascuchon, line 1: What's the difference between the `MultiLabelSelection` and `SingleLabelSelection` components?  I would expect to reuse the same component for both use cases with some kind of  multi=true|false property.

keithCuniah: I did a refactoring to use the same component for both case called `LabelSelection` with a props `multiple` which is true or false depending if we want a single or multi selection

frontend/components/commons/forms/questions-form/fields/multi-label/MultiLabelSelection.component.vue
keithCuniah: The purpose of the conditional rendering is to not have empty `div` in the dom if the question does not have a `search` or a `collapse button` . 

Without the div it will be like this : 
image

frontend/components/commons/forms/questions-form/QuestionsForm.component.vue
```diff
@@ -42,7 +42,22 @@
           :isRequired="input.is_required"
           :tooltipMessage="input.description"
           :visibleOptions="input.settings.visible_options"
-          @on-error="onError"
+          :optionsThresholdToEnableSeach="
```
frascuchon: Maybe this property is not necessary. You can keep this logic internally for the Single/MultiLabel components. We can expose it later if we find a question settings attribute to enable/disable the search bar on labels.

frontend/components/commons/forms/questions-form/fields/multi-label/MultiLabel.component.vue
```diff
@@ -0,0 +1,129 @@
...
+    this.uniqueOptions = this.options.reduce((accumulator, current) => {
+      if (!accumulator.find((item) => item.id === current.id)) {
+        accumulator.push(current);
+      }
+      return accumulator;
+    }, []);
+  },
+  computed: {
+    showSearch() {
+      return this.uniqueOptions.length >= this.optionsThresholdToEnableSeach;
```
frascuchon: I think we can use `OPTIONS_THRESHOLD_TO_ENABLE_SEARCH` constant here directly
```suggestion
      return this.uniqueOptions.length >= OPTIONS_THRESHOLD_TO_ENABLE_SEARCH;
```

frontend/components/commons/forms/questions-form/fields/shared-components/label-selection/LabelSelection.component.vue
```diff
@@ -128,6 +140,17 @@ export default {
...
+        this.tranformOptionsForMonoSelectionAndEmitNewOptions({
+          id,
+          is_selected,
+        });
+    },
+    tranformOptionsForMonoSelectionAndEmitNewOptions({ id, is_selected }) {
+      this.factoryOptionsForMonoSelection({ id, is_selected });
+      this.$emit("on-change", this.options);
+    },
+    factoryOptionsForMonoSelection({ id, is_selected }) {
```
frascuchon: Do we need all this refactor?  Maybe the change is easier to understand if we just add a condition related to the `multiple` prop:

```suggestion
      if (this.multiple) {
        return;
      }
```

```javascript
   if (this.multiple) return;
   else {
      this.options.map((option) => {
        if (option.id === id) {
          option.is_selected = is_selected;
        } else {
          option.is_selected = false;
        }
        return option;
      });
      this.$emit("on-change", this.options);
   }
```

frontend/components/commons/forms/questions-form/fields/shared-components/label-selection/LabelSelection.component.vue
```diff
@@ -128,6 +140,17 @@ export default {
...
       this.options.map((option) => {
```
frascuchon, line 154:
```suggestion
```

frontend/components/commons/forms/questions-form/fields/single-label/SingleLabel.component.vue
```diff
@@ -55,6 +56,10 @@ export default {
       type: Number | null,
       required: false,
     },
+    optionsThresholdToEnableSeach: {
+      type: Number,
+      required: true,
+    },
```
frascuchon: I think this prop is not needed yet. We can use directly the constant

frontend/components/commons/forms/questions-form/fields/single-label/SingleLabel.component.vue
```diff
@@ -74,7 +79,7 @@ export default {
   },
   computed: {
     showSearch() {
-      return this.uniqueOptions.length >= 12;
+      return this.uniqueOptions.length >= this.optionsThresholdToEnableSeach;
```
frascuchon:
```suggestion
      return this.uniqueOptions.length >= OPTION_THRESHOLD_TO_ENABLE_SEARCH;
```

frontend/components/commons/forms/questions-form/fields/multi-label/MultiLabel.component.vue
```diff
@@ -0,0 +1,129 @@
...
+      />
+    </div>
+
+    <LabelSelectionComponent
+      :options="uniqueOptions"
+      :multiple="true"
+      :componentId="questionId"
+      :showSearch="showSearch"
+      :maxOptionsToShowBeforeCollapse="maxOptionsToShowBeforeCollapse"
+      @change="$emit('update:options', options)"
```
frascuchon: Maybe if we're using the event instead of the v-model we can change the `LabelSelectionComponent` definition

frontend/components/commons/forms/questions-form/fields/multi-label/MultiLabel.component.vue
```diff
@@ -0,0 +1,129 @@
...
+      default: () => "",
+    },
+    visibleOptions: {
+      type: Number | null,
+      required: false,
+    },
+    optionsThresholdToEnableSeach: {
+      type: Number,
+      required: true,
+    },
```
frascuchon:
```suggestion
```

frontend/components/commons/forms/questions-form/fields/shared-components/label-selection/LabelSelection.component.vue
```diff
@@ -77,6 +81,10 @@ export default {
       type: Boolean,
       default: () => false,
     },
+    multiple: {
+      type: Boolean,
+      default: () => false,
+    },
   },
   model: {
```
frascuchon, line 89: We're not using the `v-model` directive. Maybe we can remove this.

frascuchon: The linked event to the model is `on-change` which is emitted only for `multiple === false`. Correct me if I'm wrong but, should not emit the event also when `multiple === true` to do the model binding work in that case?

frontend/components/commons/forms/questions-form/fields/shared-components/label-selection/LabelSelection.component.vue
```diff
@@ -93,11 +101,13 @@ export default {
...
-          .toLowerCase()
-          .includes(this.searchInput.toLowerCase())
-      );
+      return this.options
+        .filter((option) =>
+          String(option.text)
+            .toLowerCase()
+            .includes(this.searchInput.toLowerCase())
+        )
+        .sort((a, b) => b.is_selected - a.is_selected);
```
frascuchon: This will put first all selected options, which means label movements when they are visible. If we want to keep labels static when they are visible, we should change the approach. 

Maybe we can revisit the `visibleOptions` computed methods to add the selected non-visible options at the end of the visible ones (I think this is the current behavior for text classification)

leiyre (approved):

frascuchon (approved):

PR #3068, opened 2023-06-02
feat: add `FeedbackDataset` search endpoint

### LINKED ISSUE #3067: Feedback Dataset search endpoint
Create a new search endpoint `POST /api/v1/me/datasets/{dataset_id}/records/search` allowing to search records using the `search` function introduced in #3037.

## Input

The input payload of the endpoint will be for searching in all the fields of the dataset:

```json
{
  "query": {
    "text": { "q": "bad" }
  }
}
```

Input payload for searching in a specific field:

```json
{
  "query": {
    "text": { "q": "bad", "field": "text" }
  }
}
```

Additionally, the following query parameters could be provided:

- `include`: to specify which additional information has to be included in the output (like the `responses`)
- `response_status`: to filter the results based on the response status

## Output

```json
{
  "items": [
    {
      "record": {
        "id": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
        "fields": {},
        "external_id": "string",
        "responses": [
          {
            "id": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
            "values": {
              "question-1": {
                "value": "string"
              },
              "question-2": {
                "value": "string"
              },
              "question-3": {
                "value": "string"
              }
            },
            "status": "draft",
            "user_id": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
            "inserted_at": "2023-06-02T14:42:30.335Z",
            "updated_at": "2023-06-02T14:42:30.335Z"
          }
        ],
        "inserted_at": "2023-06-02T14:42:30.335Z",
        "updated_at": "2023-06-02T14:42:30.335Z"
      },
      "query_score": 14.234
    }
  ]
}
```

gabrielmbmb:
# Description

This PR introduces a new endpoint `POST /api/v1/me/datasets/{dataset_id}/records/search` that will allow the user to search records using some basic queries over the fields.

Closes #3067

**Type of change**

**How Has This Been Tested**

New unit tests have been added to test the functionality of this new endpoint. In addition, I manually tested the endpoint using the Argilla dolly dataset.

**Checklist**


src/argilla/server/apis/v1/handlers/datasets.py
```diff
@@ -59,6 +69,16 @@ def _get_dataset(db: Session, dataset_id: UUID):
...
 
 
+def _merge_search_records(search_responses: "SearchResponses", records: List[Records]) -> List[SearchRecord]:
+    search_records = []
+    for response in search_responses.items:
+        record = next((r for r in records if r.id == response.record_id), None)
+        if record:
+            search_records.append(SearchRecord(record=record, query_score=response.score))
+            records.remove(record)
+    return search_records
```
gabrielmbmb: maybe this function would be better off in another file, but I'm not sure where to put it, since it is an auxiliary function for the endpoint

src/argilla/server/apis/v1/handlers/datasets.py
```diff
@@ -285,6 +295,61 @@ async def create_dataset_records(
...
+    records = datasets.get_records_by_ids(
+        db=db, dataset_id=dataset_id, record_ids=list(record_id_score_map.keys()), include=include
+    )
+
+    return SearchRecordsResult(
+        items=sorted(
+            [SearchRecord(record=record, query_score=record_id_score_map[record.id]) for record in records],
+            key=lambda x: x.query_score,
+            reverse=True,
+        )
```
frascuchon: As default, the search engine will sort results by score relevance. In any case, I think this could be a search engine's responsibility. So I think we can remove the sort from here:
```suggestion
        items=[SearchRecord(record=record, query_score=record_id_score_map[record.id]) for record in records]
```

src/argilla/server/schemas/v1/datasets.py
```diff
@@ -347,3 +347,12 @@ def check_user_id_is_unique(cls, values):
 
 class RecordsCreate(BaseModel):
     items: conlist(item_type=RecordCreate, min_items=RECORDS_CREATE_MIN_ITEMS, max_items=RECORDS_CREATE_MAX_ITEMS)
+
+
+class SearchRecord(BaseModel):
+    record: Record
+    query_score: float
```
frascuchon: It makes sense to be more flexible on score restrictions and let the possibility of no scores at all.

```suggestion
    query_score: Optional[float]
```

src/argilla/server/contexts/datasets.py
```diff
@@ -189,6 +189,17 @@ def get_record_by_id(db: Session, record_id: UUID):
     return db.get(Record, record_id)
 
 
+def get_records_by_ids(
+    db: Session, dataset_id: UUID, record_ids: List[UUID], include: List[RecordInclude] = []
+) -> List[Record]:
+    query = db.query(Record).filter(Record.dataset_id == dataset_id, Record.id.in_(record_ids))
+    if RecordInclude.responses in include:
+        query = query.options(joinedload(Record.responses))
```
frascuchon: We should return only connected user responses (if any) instead of all responses (this will be enabled in a separate endpoint for admins). Take a look [here](https://calendar.google.com/calendar/u/0/r?cid=c_09f749125621f991d957be9dd672970956530dc9e4e2c647a9bdfa7c35a78ba7@group.calendar.google.com&pli=1)

src/argilla/server/apis/v1/handlers/datasets.py
gabrielmbmb: Yes, that is true, but when the record data is retrieved from the SQL DB and we include the responses, the records will be sorted based on the absence of responses, so records with no responses will be before records with responses, and as we would like to return the records sorted by the search score, that's why we need this additional sort.

src/argilla/server/contexts/datasets.py
gabrielmbmb: True! I will update this and add some unit tests.

src/argilla/server/apis/v1/handlers/datasets.py
```diff
@@ -285,6 +295,67 @@ async def create_dataset_records(
...
+    response_model=SearchRecordsResult,
+    response_model_exclude_unset=True,
+)
+async def search_dataset_records(
+    *,
+    db: Session = Depends(get_db),
+    search_engine: SearchEngine = Depends(get_search_engine),
+    telemetry_client: TelemetryClient = Depends(get_telemetry_client),
+    dataset_id: UUID,
+    query: SearchEngineQuery,
```
frascuchon: Maybe this could be changed and having a specific schema model for the API call. Also, once we start introducing complex filters (metadata combinations, with other things) hard to specify them as query param, they should be included as part of the body. So, even if can be redundant no, maybe makes sense to define the endpoint body as:

```json
{
   "query": { "text": ...}
}
```
later, adding filters in the body can be just:
```
{ 
   "query": { "text": ...}, # This part will be relevant to the score computation
   "filter": {   # This part won't be used for the score computation
      "text": { 
            "field": "category",
            "q": "animals"
      }
    }
}
```

gabrielmbmb: agree!

src/argilla/server/enums.py
frascuchon, line 1: We can keep this definition here, but maybe in the future, we should consider moving it into the search module.

frascuchon (approved):

frascuchon (approved):

PR #3088, opened 2023-06-06
feat: add `SearchEngine` pagination offset

### LINKED ISSUE #3074: Add pagination `offset` to search endpoint

gabrielmbmb:
# Description

This PR updates the `SearchEngine.search` method to add an `offset` parameter (Qdrant, Weaviate, Milvus, etc use `offset` and `limit` combination too) that will allow to skip results. Also, the `offset` parameter has been added to the search endpoint as a query parameter.

In addition, this PR fixes an issue for the search endpoint which was returning an empty list of `responses` for each record even if the `responses` were not in the `includes` query parameter.

Closes #3074

**Type of change**

**How Has This Been Tested**

Manual testing and I've added unit tests.

**Checklist**


src/argilla/server/apis/v1/handlers/datasets.py
```diff
@@ -347,7 +349,7 @@ async def search_dataset_records(
 
     for record in records:
         record_id_score_map[record.id]["search_record"] = SearchRecord(
-            record=record, query_score=record_id_score_map[record.id]["query_score"]
+            record=record.__dict__, query_score=record_id_score_map[record.id]["query_score"]
```
frascuchon, line 352: I think this is needed regarding the empty `responses` error, right?

src/argilla/server/search_engine.py
```diff
@@ -191,21 +192,15 @@ async def search(
             "_source": False,
             "query": {"bool": bool_query},
             "sort": ["_score", {"id": "asc"}],
+            "from": offset,
+            "size": limit,
```
frascuchon: I see we have all the extra body params as `from`, `size`, `sort`, and `source` defined in the `client.search`. I think we should pass them as method args instead of a raw body.
```suggestion
```

src/argilla/server/search_engine.py
```diff
@@ -191,21 +192,15 @@ async def search(
...
         }
-        # TODO: Work on search pagination after endpoint integration
-        next_page_token = None
-        if next_page_token:
-            # See https://www.elastic.co/guide/en/elasticsearch/reference/current/paginate-search-results.html
-            body["search_after"] = next_page_token
 
-        response = await self.client.search(index=self._index_name_for_dataset(dataset), size=limit, body=body)
+        response = await self.client.search(index=self._index_name_for_dataset(dataset), body=body)
```
frascuchon:
```suggestion
        response = await self.client.search(index=self._index_name_for_dataset(dataset),body=body, from_=offset, size=size, _source=False, sort=["_score", {"id": "asc"}])
```

frascuchon (approved):

src/argilla/server/apis/v1/handlers/datasets.py
gabrielmbmb: It won't give an error, but if we pass `record` instead of `record.__dict__`, `pydantic` tries to get the value of `record.responses`, and in this case `sqlalchemy` will return an empty list.

`record.__dict__` doesn't contain the `responses` key, and this way the value in the `pydantic.BaseModel` will be `None` (because it's an `Optional` field), which in combination with the `response_model_exclude_unset` endpoint flag makes the endpoint to not include the `responses` key in the output payload, which is the desired behaviour (instead of receiving `"responses": []`.

frascuchon (approved):

PR #3010, opened 2023-05-29
feat: add multi label selection question to the API

gabrielmbmb:
# Description

I've added the required class `MultiLabelSelectionQuestionSettings` to create the new type of question multi-label selection question (also known as a multi-choice question) and to validate the responses submitted for this kind of question.

Also, I've removed the parametrization of the `test_create_record_response_with_extra_question_responses` unit test that I'd added in #3005 accidentally because I thought the validation of this depended on the question itself.

Closes #2996

**Type of change**

- [x] New feature (non-breaking change which adds functionality)

**How Has This Been Tested**

I've created some unit tests to test the creation of this new type of question and also to validate the responses submitted to it.

**Checklist**

- [x] I have merged the original branch into my forked branch
- [ ] I added relevant documentation
- [x] follows the style guidelines of this project
- [x] I did a self-review of my code
- [ ] I made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added relevant notes to the CHANGELOG.md file (See https://keepachangelog.com/)

PR #3100, opened 2023-06-06
feat: add `LabelQuestion` and `MultiLabelQuestion` in Python client

### LINKED ISSUE #3099: Support for single and multi-label selection question types in Python client

alvarobartt:
# Description

This PR adds support to the recently included `LabelSelectionQuestion` and `MultiLabelSelectionQuestion` new types of questions as of #3005 and #3010, respectively, in the Python client side.

Closes #3099 

**Type of change**

**How Has This Been Tested**

- [x] Add integration tests for the recently included questions

**Checklist**

### DIFF
--- a/src/argilla/client/feedback/dataset.py
+++ b/src/argilla/client/feedback/dataset.py
@@ -40,9 +40,9 @@ from argilla.client.feedback.schemas import (
     FeedbackDatasetConfig,
     FeedbackRecord,
     FieldSchema,
+    LabelQuestion,
     MultiLabelQuestion,
     RatingQuestion,
-    SingleLabelQuestion,
     TextField,
     TextQuestion,
 )
@@ -86,7 +86,7 @@ class FeedbackDataset:
         TypeError: if `fields` is not a list of `FieldSchema`.
         ValueError: if `fields` does not contain at least one required field.
         TypeError: if `questions` is not a list of `TextQuestion`, `RatingQuestion`,
-            `SingleLabelQuestion`, and/or `MultiLabelQuestion`.
+            `LabelQuestion`, and/or `MultiLabelQuestion`.
         ValueError: if `questions` does not contain at least one required question.
 
     Examples:
@@ -109,7 +109,7 @@ class FeedbackDataset:
         ...             required=True,
         ...             values=[1, 2, 3, 4, 5],
         ...         ),
-        ...         rg.SingleLabelQuestion(
+        ...         rg.LabelQuestion(
         ...             name="question-3",
         ...             description="This is the third question",
         ...             required=True,
@@ -163,7 +163,7 @@ class FeedbackDataset:
             TypeError: if `fields` is not a list of `FieldSchema`.
             ValueError: if `fields` does not contain at least one required field.
             TypeError: if `questions` is not a list of `TextQuestion`, `RatingQuestion`,
-                `SingleLabelQuestion`, and/or `MultiLabelQuestion`.
+                `LabelQuestion`, and/or `MultiLabelQuestion`.
             ValueError: if `questions` does not contain at least one required question.
             TypeError: if `guidelines` is not None and not a string.
             ValueError: if `guidelines` is an empty string.
@@ -188,7 +188,7 @@ class FeedbackDataset:
             ...             required=True,
             ...             values=[1, 2, 3, 4, 5],
             ...         ),
-            ...         rg.SingleLabelQuestion(
+            ...         rg.LabelQuestion(
             ...             name="question-3",
             ...             description="This is the third question",
             ...             required=True,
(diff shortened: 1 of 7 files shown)


src/argilla/client/feedback/dataset.py
```diff
@@ -199,9 +221,10 @@ def __init__(
             raise TypeError(f"Expected `questions` to be a list, got {type(questions)} instead.")
         any_required = False
         for question in questions:
-            if not isinstance(question, (TextQuestion, RatingQuestion)):
+            if not isinstance(question, (TextQuestion, RatingQuestion, SingleLabelQuestion, MultiLabelQuestion)):
```
frascuchon: could we use the `AllowedQuestionTypes ` typing somehow?

src/argilla/client/feedback/dataset.py
```diff
@@ -527,7 +560,19 @@ def delete_and_raise_exception(dataset_id: str, exception: Exception) -> None:
...
+                        id=argilla_id,
+                        question=question.dict(
+                            include={
+                                "name": True,
+                                "title": True,
+                                "description": True,
+                                "required": True,
+                                "settings": True,
+                            }
+                        ),
```
frascuchon: I think you can use directly a set in this case:

```suggestion
                        question=question.dict(
                            include={ "name", "title", "description", "required", "settings" }
                        ),
```

src/argilla/client/feedback/schemas.py
```diff
@@ -112,40 +117,54 @@ def title_must_have_value(cls, v: Optional[str], values: Dict[str, Any]) -> str:
...
+        return values
 
-    class Config:
-        validate_assignment = True
-        extra = Extra.forbid
+
+class _LabelQuestion(QuestionSchema):
+    settings: Dict[str, Any] = Field({})
+    labels: List[str] = Field(unique_items=True)
+    visible_labels: Optional[PositiveInt] = None
```
frascuchon: We talked about a default value of `20` from the Python client.

src/argilla/client/feedback/dataset.py
alvarobartt: We cannot call `isintance` over an `Union`, neither `issubclass` 😢

src/argilla/client/feedback/schemas.py
alvarobartt: Oh ok I missed that, I just followed the API spec, but 20 makes sense 😄

src/argilla/client/feedback/dataset.py
```diff
@@ -199,9 +221,10 @@ def __init__(
             raise TypeError(f"Expected `questions` to be a list, got {type(questions)} instead.")
         any_required = False
         for question in questions:
-            if not isinstance(question, (TextQuestion, RatingQuestion)):
+            if not isinstance(question, (TextQuestion, RatingQuestion, LabelQuestion, MultiLabelQuestion)):
```
gabrielmbmb, line 224: Maybe we could create a base class for the questions and just do `isinstance(question, BaseQuestion)`

src/argilla/client/feedback/schemas.py
```diff
@@ -103,49 +108,80 @@ class QuestionSchema(BaseModel):
...
+        return values
 
-    class Config:
-        validate_assignment = True
-        extra = Extra.forbid
+
+class _LabelQuestion(QuestionSchema):
+    settings: Dict[str, Any] = Field(default_factory=dict, allow_mutation=False)
+    labels: Union[List[str], Dict[str, str]] = Field(unique_items=True, min_items=2)
+    visible_labels: Optional[conint(ge=3)] = 20
```
gabrielmbmb: in the backend we're using `options` instead of `labels`, maybe it's better to align to that here

src/argilla/client/feedback/schemas.py
```diff
@@ -103,49 +108,80 @@ class QuestionSchema(BaseModel):
...
+    def labels_dict_must_be_valid(cls, v: Union[List[str], Dict[str, str]]) -> Union[List[str], Dict[str, str]]:
+        if isinstance(v, list):
+            return v
+        if isinstance(v, dict):
+            if len(v.keys()) < 2:
+                raise ValueError("ensure this dict has at least 2 items")
+            if len(set(v.values())) != len(v.values()):
+                raise ValueError("ensure this dict has unique values")
+            return v
+        raise TypeError("ensure this value is a list or a dict")
```
frascuchon: I think this is handled by pydantic using the field definition, isn't it?

src/argilla/client/feedback/dataset.py
gabrielmbmb: I've seen that we could use `QuestionSchema`, but if for some reason a user creates a class inheriting from it, it would pass this validation and we would have some errors later.

src/argilla/client/feedback/schemas.py
alvarobartt: It's not, the `Field(unique_items, min_items)` just works for lists, not for dicts; so we need to add this validation layer besides the `Field`

alvarobartt: Or you mean the `TypeError`?

src/argilla/client/feedback/dataset.py
alvarobartt: We wanted to specifically constraint this to the supported questions in the Python client, that's why :(

src/argilla/client/feedback/schemas.py
alvarobartt: Yes I know, but we decided that user-wise it's better to use `labels` as doesn't make much sense to define a `LabelQuestion` with `options` instead of `labels`, but if you both agree @frascuchon, I can revert that to be `options` too 😄

frascuchon (approved):
I think we could setup the `settings` attribute as private for the Questions and Field types. Even more, since we're providing the question configuration  through specific properties, these `settings` field  could be generate later when needed (only for the API calls)

PR #2940, opened 2023-05-18
Feat/2664 add argillatraining module for autotrain

davidberenstein1957:
# Description

added support for AutoTrain via direct Argilla dataset integrations.

Closes #2664

**Type of change**

(Please delete options that are not relevant. Remember to title the PR according to the type of change)

- [X] New feature (non-breaking change which adds functionality)
- [X] Refactor (change restructuring the codebase without changing functionality)
- [X] Improvement (change adding some improvement to an existing functionality)

**How Has This Been Tested**

pytest tests/training/test_autotrain.py

**Checklist**

- [ ] I added relevant documentation
- [X] follows the style guidelines of this project
- [X] I did a self-review of my code
- [X] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added relevant notes to the CHANGELOG.md file (See https://keepachangelog.com/)

PR #3105, opened 2023-06-07
chore: upgrade `flair` to latest version (0.12.2)

alvarobartt:
# Description

As of #3012, we decided to upgrade `flair` to 0.12.0 to avoid some known installation issues with both dependencies: `tokenizers` and `sentencepiece`. It was solved, but we recently included the `ArgillaAutoTrainTrainer` at #2940 (kudos to @davidberenstein1957), and it seems that `autotrain-advanced` dependency expects a greater `huggingface_hub` version, while `flair` 0.12.0 had it pinned to `==0.10.0` which was pretty restrictive and, so on, running into conflicts with `autotrain-advanced`.

So on, it seems that `flair` 0.12.2 already defines `huggingface_hub` dependency as `>=0.10.0` which is more relaxed, and doesn't run into any conflict with the rest of the dependencies.

See the previous `huggingface_hub` dependency pinned to ==0.10.0 at https://github.com/flairNLP/flair/blob/d13f8ea36e60c5ab1d7468893498db96d91fd4be/requirements.txt#L22, and now being pinned to >=0.10.0 at https://github.com/flairNLP/flair/blob/621e7dd58be8e3c5985176d9a7f54eb5c0cd8c88/requirements.txt#L22, as of 0.12.2, since in 0.12.1 it's still pinned to ==0.10.0.

**Type of change**

- [X] Upgrade/Add dependency

**How Has This Been Tested**

- [X] Installed via `pip` in Python 3.8, 3.9, 3.10, and 3.11; and with `conda` in Python 3.8

**Checklist**

### DIFF
--- a/environment_dev.yml
+++ b/environment_dev.yml
@@ -37,7 +37,7 @@ dependencies:
       - cleanlab~=2.0.0 # With this version, tests are failing
       - datasets>1.17.0,!= 2.3.2 # TODO: push_to_hub fails up to 2.3.2, check patches when they come out eventually
       - huggingface_hub
-      - flair==0.12
+      - flair>=0.12.2
       - faiss-cpu
       - flyingsquid
       - pgmpy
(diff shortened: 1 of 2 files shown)


alvarobartt:
P.S. It also works in Python 3.7, in case we continue giving support to it at least for the v1.9.0 👍🏻

PR #3108, opened 2023-06-07
made issues templates a bit friendlier

davidberenstein1957:
- added a call to action about contacting me or starting a PR
- added some default tags to bug and features
- added a separate issue for docs

### DIFF
--- a/.github/ISSUE_TEMPLATE/🆕-feature-request.md
+++ b/.github/ISSUE_TEMPLATE/🆕-feature-request.md
@@ -1,11 +1,13 @@
 ---
-name: Feature request
-about: Suggest an idea for this project
-title: ''
+name: "\U0001F195 Feature request"
+about: Cool new ideas for the project
+title: "[FEATURE]"
 labels: enhancement
 assignees: ''
 
 ---
+
+<!--  Ask David for help you to contribute https://calendly.com/argilla-office-hours/30min or feel free to submit a pull request straight away: https://github.com/argilla-io/argilla/pulls or  -->
 
 **Is your feature request related to a problem? Please describe.**
 A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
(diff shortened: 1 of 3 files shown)


alvarobartt:
LGTM! But will the emojis in the filename work? Besides that, can we also add a file related to issues with the integrations? e.g. `langchain` as in the `ArgillaCallbackHandler` we're redirecting the users to report issues with it as https://github.com/hwchase17/langchain/blob/b3ae6bcd3f42ec85ee65eb29c922ab22a17a0210/langchain/callbacks/argilla_callback.py#L119-L124 (since that was the standard practice as of the rest of the callbacks), but we don't actually let the users create tags, so maybe an integrations template would solve it 😄

davidberenstein1957:
> LGTM! But will the emojis in the filename work?
> 
> Besides that, can we also add a file related to issues with the integrations? e.g. `langchain` as in the `ArgillaCallbackHandler` we're redirecting the users to report issues with it as https://github.com/hwchase17/langchain/blob/b3ae6bcd3f42ec85ee65eb29c922ab22a17a0210/langchain/callbacks/argilla_callback.py#L119-L124 (since that was the standard practice as of the rest of the callbacks), but we don't actually let the users create tags, so maybe an integrations template would solve it 😄

Yes emojis will work.

@alvarobartt I will create another PR for the integrations template.

alvarobartt (commented):
LGTM! But will the emojis in the filename work?

Besides that, can we also add a file related to issues with the integrations? e.g. `langchain` as in the `ArgillaCallbackHandler` we're redirecting the users to report issues with it as https://github.com/hwchase17/langchain/blob/b3ae6bcd3f42ec85ee65eb29c922ab22a17a0210/langchain/callbacks/argilla_callback.py#L119-L124 (since that was the standard practice as of the rest of the callbacks), but we don't actually let the users create tags, so maybe an integrations template would solve it 😄

alvarobartt (approved):

PR #3110, opened 2023-06-07
style: curate auto-generated `DatasetCard` for `FeedbackDataset`

### LINKED ISSUE #3018: Curate generated `DatasetCard` for `FeedbackDataset`s
**Describe the bug**

The generated `DatasetCard` for `FeedbackDataset` when calling `push_to_huggingface` with `generate_card=True` the generated `DatasetCard` seems to have some issues when not all the information is there e.g. `description` for questions.

**Screenshots**

image

### LINKED ISSUE #3066: Enhance Dataset Card Structure and Content
I'd like to enhance the structure and content of the generated dataset card. It might be a good idea to start using a template where we can update the structure rather than doing it programmatically (in the feedback.py)

There are a few changes I'd like to make and it's open to suggestions from @alvarobartt @davidberenstein1957  @tomaarsen @gabrielmbmb:

Include details in the  metadata YML content at the beginning of the file, at least the following:

```yaml
---
tags:
- argilla
- human-feedback
size_categories:
- 1K<n<10K (I assume we can compute this easily?)
---
```

Start with an H1 where we include the dataset name. If we think this is an issue, let's use the provided path/dataset name from the Hub but let's add something:

`# Dataset card for {dataset_name}`

Then this section is fine:

`## Load with Argilla`

To load this dataset with Argilla, you'll just need to install Argilla as `pip install argilla --upgrade` and then use the following code:

```python
import argilla as rg

ds = rg.FeedbackDataset.from_huggingface('argilla/comparison-dataset-dolly-curated-falcon')
```

`## Load with Datasets`

To load this dataset with Datasets, you'll just need to install Datasets as `pip install datasets --upgrade` and then use the following code:

```python

ds = load_dataset('argilla/comparison-dataset-dolly-curated-falcon')
```

Then the `questions` and `fields`. Can we maybe make this more human-readable, provide the full info about the attributes and have them in a markdown table?

Then the Guidelines, but I'd recommend including Annotation Guidelines ( I know we use guidelines internally but for people outside it's not so clear what these are.

Finally what would be awesome (and good for the community) is to use the [recommended template](https://github.com/huggingface/datasets/blob/main/templates/README.md) (based on the Dataset cards paper).

Here we could try to align and fit our content as much as possible (the order of the content is not so relevant and we could push the things we know or can compute at card creation time).

Let me know what you think!

alvarobartt:
# Description

Curate the auto-generated `DatasetCard` when calling `FeedbackDataset.push_to_huggingface(..., generate_card=True)` since there were some things that were not looking fine in terms of style and readability.

So now we're using the `DatasetCard` template provided by the HuggingFace Hub team, but with some minor tweaks to create `ArgillaDatasetCard`. We're using the template defined at https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/templates/datasetcard_template.md.

Closes #3018 and #3066

**Type of change**

**How Has This Been Tested**

- [X] Uploaded various datasets to the HuggingFace Hub via `FeedbackDataset.push_to_huggingface` with questions without description, questions with description, rating questions with different values, and different combination of those.

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ These are the section headers that we use:
 - Added `MultiLabelSelectionQuestionSettings` class allowing to create multi-label selection (multi-choice) questions in the API.
 - Database setup for unit tests. Now the unit tests use a different database than the one used by the local Argilla server (Closes [#2987]).
 - Updated `alembic` setup to be able to autogenerate revision/migration scripts using SQLAlchemy metadata from Argilla server models ([#3044])
+- Improved `DatasetCard` generation on `FeedbackDataset.push_to_huggingface` when `generate_card=True`, following the official HuggingFace Hub template, but suited to `FeedbackDataset`s from Argilla ([#3110])
 
 ## [1.8.0](https://github.com/argilla-io/argilla/compare/v1.7.0...v1.8.0)
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -91,8 +91,14 @@ homepage = "https://www.argilla.io"
 documentation = "https://docs.argilla.io"
 repository = "https://github.com/argilla-io/argilla"
 
+[tool.setuptools.packages.find]
+where = ["src"]
+
 [tool.setuptools.dynamic]
 version = {attr = "argilla.__version__"}
+
+[tool.setuptools.package-data]
+"argilla.client.feedback.card" = ["argilla_template.md"]
 
 [tool.pytest.ini_options]
 log_format = "%(asctime)s %(name)s %(levelname)s %(message)s"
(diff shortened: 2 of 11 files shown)


dvsrepo:
@alvarobartt to close #3066, I'd recommend starting with [the recommended template](https://github.com/huggingface/datasets/blob/main/templates/README.md):

- [x] Use the recommended template
- [x] Add yml metadata as suggested in #3066. If it's an overkill don't compute the dataset size but include at least the fixed suggested properties

I can help fit our content into the template if it helps


A very good example of using a template approach is [skops](https://github.com/skops-dev/skops/tree/main/skops/card)

dvsrepo (approved):
💯

PR #3133, opened 2023-06-08
Review docker and scripts

frascuchon:
# Description

The docker and script have been reviewed:

- The docker build context will be placed in the docker folder, so we don't need to change anything inside the dockerfiles
- The unneeded scripts have been removed
- The specific docker script has been moved into docker/scripts
- The scripts used outside docker have been placed in the scripts folder, otherwise, script execution may change. And not much sense to put them inside the docker folder since they are used outside docker images.
- Docs have been updated
- The `release.Dockerfile` has been renamed to `Dockerfile` (more standard way for the principal docker image)


CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-# Changelog
+it add# Changelog
```
```suggestion
# Changelog
```

davidberenstein1957 (approved):

davidberenstein1957 (approved):

PR #3138, opened 2023-06-09
:sparkles: FEAT: Search for feedbacktask

### LINKED ISSUE #3087: Search for feedback task ui

keithCuniah:
# Description

Search for feedback task

What have been done :
- New component for the search in feedback task
- If user try to apply search and if the `QuestionForm` have been touched => a toast will appears
- If the user apply a search with

Closes #3087 

**Type of change**

**How Has This Been Tested**

- Only feedback task


frontend/components/feedback-task/SearchBar.base.vue
```diff
@@ -16,87 +16,99 @@
...
 
 <script>
 export default {
+  name: "SearchBarComponent",
   props: {
-    currentSearchText: {
+    value: {
       type: String,
-      default: "",
+      default: () => "",
```
leiyre: The prop works without arrow function for String, there is a purpose?

leiyre (approved):

frontend/components/feedback-task/DatasetFilters.component.vue
```diff
@@ -62,35 +71,35 @@ export default {
...
     },
+    searchFromRoute() {
+      return this.$route.query?._search;
+    },
   },
   watch: {
     selectedStatus(newValue) {
       this.$root.$emit("status-filter-changed", newValue);
     },
+    async searchInput(searchInput) {
```
frascuchon:
```suggestion
    searchInput(searchInput) {
```

frontend/components/feedback-task/SearchBar.base.vue
frascuchon: We should be consistent with the props definition. We should apply the same way of default props definition (I mean if we decide to use arrow functions, we should define all props defaults as arrow functions for primitive types, if we choose to use the value directly, we should apply this to all default prop definitions.

frontend/components/page-contents/feedback-task-content/RecordFeedbackTaskAndQuestionnaire.content.vue
```diff
@@ -307,9 +330,46 @@ export default {
 
       this.reRenderQuestionForm++;
     },
+    async applySearchFilter(searchFilter) {
+      console.log("apply: ", searchFilter);
```
frascuchon: remove this line.

frontend/components/page-contents/feedback-task-content/RecordFeedbackTaskAndQuestionnaire.content.vue
```diff
@@ -409,6 +482,38 @@ export default {
...
+          JSON.stringify({
+            query: {
+              text: {
+                q: searchText,
+              },
+            },
+          })
+        );
+
+        const { data } = await this.$axios.post(url, body);
```
frascuchon: I think axios supports query params to be passed as an object. This can simplify the URL definition

```suggestion
        const url = `/v1/me/datasets/${datasetId}/records/search`;

        const body = JSON.parse(
          JSON.stringify({
            query: {
              text: {
                q: searchText,
              },
            },
          })
        );
        const params = { 
           include: "responses",
           response_status: responseStatus,
           limit: numberOfRecordsToFetch,
           offset
        };
           
        const { data } = await this.$axios.post(url, body, { params } );
```

PR #3162, opened 2023-06-13
feat: migrate to async SQLAlchemy engine

### LINKED ISSUE #3035: Create a SQLAlchemy async engine and update the API to use it
SQLAlchemy allows to create an async engine that could be used in the `async` endpoints of the API to interact with the database. This could improve the performance of the API because it could perform non-blocking I/O operations while waiting for database queries.

We should create a new async engine to be used by the API, but we would need to keep the sync engine to be used in other parts like the CLI.

gabrielmbmb:
# Description

This PR updates the API to use an async SQLAlchemy engine to connect to the DB: 

- Added a new `AsyncEngine` instance, `sessionmaker` that will create `AsyncSession`s and a new async generator function `get_async_db` that will be used in the API. The sync engine and `sessionmaker` creating `Session`s are still there as we need them for the CLI.
-  Update the `Base` (`DeclarativeBase`) SQLAlchemy ORM class to also inherit from `AsyncAttrs` class. This will allow to load/get relationships using `relationship = await model.awaitable_attrs.relationship_attribute_name`. When using the sync engine, this was not required because of the lazy loading, but using the async engine [this is not possible](https://docs.sqlalchemy.org/en/20/orm/extensions/asyncio.html#asyncio-orm-avoid-lazyloads).
- Update all the `PolicyAction`s methods/function to be async (required because they are used from the API and the API will use the async connection to the DB)
- Added a new `AsyncSQLAlchemyModelFactory` factory_boy factory that allows to create SQLAlchemy ORM models using an async connection. All the factories classes have been updated to use this class as parent class. In some cases, like for the `UserFactory` which is used in some tests for the CLI which uses the sync DB connection, a sync version of the factory has been created (`UserSyncFactory`).
- Update parts of the code in which we were adding, updating or deleting records for a Feedback Dataset to rollback the DB transaction (i.e. not doing any change in the DB) if an exception is raised while inserting the record info in ElasticSearch. 

Closes #3035 

**Type of change**

**How Has This Been Tested**

All the tests using the async connection to the DB have been updated to be async. The previous unit tests that not require to connect to the DB (at least directly) are passing.

**Checklist**


src/argilla/server/contexts/datasets.py
```diff
@@ -43,82 +42,100 @@
...
 
-    dataset.status = DatasetStatus.ready
-    db.commit()
-    db.expire(dataset)
-
     try:
+        dataset.status = DatasetStatus.ready
         await search_engine.create_index(dataset)
+        await db.commit()
+        if is_unit_test():
```
gabrielmbmb: I'll try to get rid of this

src/argilla/server/contexts/datasets.py
```diff
@@ -306,44 +324,53 @@ async def create_records(
...
-
-        db.commit()
-        db.expire_all()
-
+        await db.flush(records)
+        for record in records:
+            await record.awaitable_attrs.responses
         await search_engine.add_records(dataset, records)
+        await db.commit()
+        if is_unit_test():
```
gabrielmbmb: I'll try to get rid of this

alvarobartt:
image

alvarobartt:
cc @tomaarsen any idea on how we can avoid the HuggingFace unit tests from failing due to HTTP issues? Is there any retry-method you're currently using in SetFit to avoid that from making the CI/CD fail? Thanks 🤗

gabrielmbmb:
We could use https://github.com/pytest-dev/pytest-rerunfailures to try to rerun tests failing because HTTP exceptions when connecting to HF

tomaarsen:
> cc @tomaarsen any idea on how we can avoid the HuggingFace unit tests from failing due to HTTP issues? Is there any retry-method you're currently using in SetFit to avoid that from making the CI/CD fail? Thanks 🤗

We don't do any retrying. It tends to occur really infrequently, as we cache all models that we load. We also purposefully only load tiny models and datasets. We haven't really had these issues.

tomaarsen:
One potential fix is to wrap the tests in a decorator that catches `requests.exceptions.ReadTimeout` errors and calls `pytest.skip()` if they occur.

src/argilla/server/database.py
```diff
@@ -50,5 +60,16 @@ def get_db() -> Generator["Session", None, None]:
...
+    try:
+        db: "AsyncSession" = AsyncSessionLocal()
+        yield db
+    finally:
+        await db.close()
+
+
+class Base(AsyncAttrs, DeclarativeBase):
+    def __eq__(self, value: object) -> bool:
+        if is_unit_test():
```
frascuchon: what it this? 👀

src/argilla/server/schemas/v1/questions.py
```diff
@@ -35,13 +40,40 @@ class RatingQuestionSettings(BaseModel):
...
+QuestionSettings = Annotated[
+    Union[
+        TextQuestionSettings,
+        RatingQuestionSettings,
+        LabelSelectionQuestionSettings,
+        MultiLabelSelectionQuestionSettings,
+    ],
+    Field(..., discriminator="type"),
+]
+
```
frascuchon: This is nothing about the async refactor. Just notice this schema was outdated, right?

src/argilla/server/settings.py
```diff
@@ -65,6 +65,7 @@ class Settings(BaseSettings):
     home_path: Optional[str] = Field(description="The home path where argilla related files will be stored")
     base_url: Optional[str] = Field(description="The default base url where server will be deployed")
     database_url: Optional[str] = Field(description="The database url that argilla will use as data store")
+    database_url_async: Optional[str] = Field(description="The async database url that argilla will use as data store")
```
frascuchon: For others db than SQLite, do we need a specific async URL? If not I think we can just compute the async URL connection based on the database_url settings property.
```python
@property
def database_url_async(self):
   return database_url.replace("sqlite:///", "sqlite+aiosqlite:///")

```

src/argilla/utils/utils.py
```diff
@@ -165,3 +166,8 @@ def start_background_loop(loop: asyncio.AbstractEventLoop) -> None:
         thread.start()
         __LOOP__, __THREAD__ = loop, thread
     return __LOOP__, __THREAD__
+
+
+def is_unit_test() -> bool:
```
frascuchon: Why the `Base` class __eq__ method is different in a test environment?

frascuchon (commented):
I have some questions about the `is_unit_test` function and another code block that looks like another topic (just to confirm).

Also, in the future, I think we can define TestSuite classes with the @pytest.asyncio decorator instead of having every test function decorated. See the TestSuite for the Search engine https://github.com/argilla-io/argilla/pull/3162/files#diff-6188cca586c2d3876c05d742e761b86b921be53fad6a72436a5678b02e084847L142

src/argilla/server/settings.py
gabrielmbmb: yes, we do. For example, for Postgres the URL would be `postgresql+asyncpg://...`

Anyways, I think that we can compute the async URL from the sync one. We could have a `dict[str, str]` mapping sync protocols to async ones.

src/argilla/server/database.py
gabrielmbmb: I had to add this so basically the `assert_called_with` works:

https://github.com/argilla-io/argilla/blob/df19a5a6168948325e34b67c197202d03125e8bf/tests/server/api/v1/test_responses.py#L70

We have a few mocked functions in which we're passing an SQLAlchemy ORM model. `assert_called_once_with` uses the `==` operator (and therefore the `__eq__` magic method) to check if the arguments are the same. The problem is that we're usually passing objects to this methods that have been generated with a factory (from `factory_boy`), and this happens:

image

They contain the same information, but as you can see in the image above they are not the same object and this is because they are attached to a different SQLAlchemy session.

For the `tests`, we were using an [`scoped_session`](https://docs.sqlalchemy.org/en/20/orm/contextual.html), which basically is a registry of sessions. This class maintains a session per thread, so if we do this:

```
from sqlalchemy.orm import scoped_session

Session = scoped_session(...)

session1 = Session()
session2 = Session()
```

they will be basically the same session. So the session in the unit tests and the factories is the same because it's getting executed in the same thread. With async this is different. We have to use [`async_scoped_session`](https://docs.sqlalchemy.org/en/20/orm/extensions/asyncio.html#using-asyncio-scoped-session), which do not use the threads to maintain the registry and instead it maintains a session per `asyncio.Task`. So basically, the factories and the unit tests are not running in the same `asyncio.Task` and they are getting a different session. I'll try to find a fix for this.

gabrielmbmb: Okay, I managed to get rid of the `is_unit_test` function :)

alvarobartt: @gabrielmbmb 🐐

src/argilla/server/schemas/v1/questions.py
gabrielmbmb: yes, it was outdated. In fact, I forgot also to include the `RankingQuestionSettings` here too in #3232. I just added it.

frascuchon (approved):

PR #3166, opened 2023-06-13
feat: return `total` number of search hits

gabrielmbmb:
# Description

I've added the `total` attribute to the `SearchResponses` and `SearchRecordResult` (schema) classes, so the `POST /api/v1/me/datasets/{dataset_id}/records/search` will return a number of the total records that the ElasticSearch query returned.

**Type of change**

**How Has This Been Tested**

Manually doing several requests to a local server with a dataset containing IMDB reviews. Also, I've updated the unit tests regarding the `SearchEngine` and `search` endpoint.

**Checklist**

### DIFF
--- a/src/argilla/server/search_engine.py
+++ b/src/argilla/server/search_engine.py
@@ -202,6 +202,7 @@ class SearchEngine:
             size=limit,
             _source=False,
             sort="_score:desc,id:asc",
+            track_total_hits=True,
         )
 
         items = [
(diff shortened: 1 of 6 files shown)


src/argilla/server/search_engine.py
```diff
@@ -206,8 +207,9 @@ async def search(
         items = [
             SearchResponseItem(record_id=UUID(hit["_id"]), score=hit["_score"]) for hit in response["hits"]["hits"]
         ]
+        total = response["hits"]["total"]["value"]
```
frascuchon, line 211: Elasticsearch computes the total number of hits with optimization by default (see here:https://www.elastic.co/guide/en/elasticsearch/reference/8.8/search-your-data.html#track-total-hits)

If we want to return always the proper number, we should add the `track_total_hits=True` argument on the search method to compute it.

frascuchon (approved):

Issue #3079, opened 2023-06-05
feature: extend `CLI` support for user management


davidberenstein1957:
**Is your feature request related to a problem? Please describe.**
We want to add extended CLI support for `src/argilla/tasks/users` because it is limited now.

**Describe the solution you'd like**
Our [docs](https://docs.argilla.io/en/latest/getting_started/installation/configurations/user_management.html) mention a lot of operations which are not supported by our CLI. 

**Describe alternatives you've considered**
Using the examples from the docs directly.

**Additional context**
https://docs.argilla.io/en/latest/getting_started/installation/configurations/user_management.html

https://github.com/argilla-io/argilla/issues/2968

khishh:
Hi @davidberenstein1957, can I pick this up? I might need some extra context but am interested in contributing this!

davidberenstein1957:
Hi @khishh, that would be awesome. Do you need help getting started with this?

davidberenstein1957:
@alvarobartt, perhaps it is good to align this with the new user management layer PR.

alvarobartt:
Sure, @khishh feel free to contact me and/or the community via Slack at https://rubrixworkspace.slack.com/archives/C05BCGM277B, and I can follow up with you there to help if needed 😄

davidberenstein1957:
https://github.com/argilla-io/argilla/pull/3169

khishh:
> Sure, @khishh feel free to contact me and/or the community via Slack at https://rubrixworkspace.slack.com/archives/C05BCGM277B, and I can follow up with you there to help if needed 😄

That would be great!! I set the dev environment up locally using docker-compose but I will probably need some context before working on this ticket. I will reach out to you on Slack👋🏼

alvarobartt:
Closed as completed in the following PRs:
* https://github.com/argilla-io/argilla/issues/3590
* https://github.com/argilla-io/argilla/issues/3589
* https://github.com/argilla-io/argilla/issues/3590

PR #3169, opened 2023-06-13
feat: add `User` class to manage and create users within Argilla

### LINKED ISSUE #2968: Add a user-management layer from the Python client

Even though the users can already be managed through the API, I think it would be much clearer if we also included that within the Python client, to avoid having to define requests to the API via Python to get information about the users in Argilla.

This would prevent users from having to do the following to list Argilla users:

```python
import httpx

rg_client= rg.active_client().client
auth_headers = {"X-Argilla-API-Key": rg_client.token}
http=httpx.Client(base_url=rg_client.base_url, headers=auth_headers)
users = http.get("/api/users").json()
```

The class to manage the users should allow to create new users, change roles, see their information, delete/remove users, list users, filter users, etc. and any other feature that the current user-management allows via the API.

alvarobartt:
# Description

This PR addresses the addition of a new user-management layer in the Python client via the `User` class, wrapping up the `/api/users` and `/api/me` endpoints in the API v0.

Closes #2968

**Type of change**

**How Has This Been Tested**

- [x] Add unit tests for the new `User` class
- [X] Add unit tests for the `user_api` endpoints
- [X] Revisit existing tests to pass with updated `UserModel` (formerly named `User`)

**Checklist**


davidberenstein1957:
shouldn't we also be able to update users? or will we then need to delete and re-create?

src/argilla/client/users.py
```diff
@@ -0,0 +1,118 @@
...
+    def __user_exists(username: Optional[str] = None, id: Optional[str] = None) -> Optional["UserModel"]:
+        # loop through `list_users` and look for matches on either `username` or `id`, if
+        # the user exists return the reponse, otherwise return None
+        for user in list_users():
+            if user.username == username or user.id == id:
+                return user
+        return None
+
+    def __str__(self) -> str:
+        return f"User(id={self.id}, username={self.username}, role={self.role}, workspaces={self.workspaces}, api_key={self.api_key}, first_name={self.first_name}, last_name={self.last_name}, role={self.role}, inserted_at={self.inserted_at}, updated_at={self.updated_at})"
```
davidberenstein1957: Perhaps this can formatted a bit better to make it more readable within newlines??

src/argilla/client/sdk/users/models.py
```diff
@@ -12,17 +12,35 @@
...
 
-class User(BaseModel):
-    """User data model"""
 
+class UserCreateModel(BaseModel):
+    first_name: str = Field(min_length=1)
+    last_name: Optional[str] = Field(min_length=1)
+    username: str = Field(min_length=1, regex=r"^(?!-|_)[a-z0-9-_]+$")
+    role: Roles = "annotator"
+    password: str = Field(min_length=8, max_length=100)
```
davidberenstein1957, line 34: isn't 100 a bit much for as password?

alvarobartt:
> shouldn't we also be able to update users? or will we then need to delete and re-create?

That's a nice question indeed, maybe @frascuchon can answer this.

But for the moment, the only available endpoints for user-management are the following, so I guess we're not able to, but it's a nice addition for sure.

image

src/argilla/client/sdk/users/api.py
```diff
@@ -12,10 +12,147 @@
...
+
+    Returns:
+        A `Response` object containing a `parsed` attribute with the parsed response if
+        the request was successful, which is an instance of `UserModel`.
+    """
+    url = f"/api/users/{user_id}"
+
+    response = client.delete(url=url)
+
+    if response.status_code == 200:
```
davidberenstein1957, line 143: maybe we can simplify this a bit because it is duplicated a couple of time.

src/argilla/client/users.py
```diff
@@ -0,0 +1,118 @@
...
+    def create(
+        cls,
+        username: str,
+        password: Union[str, UUID],
+        first_name: Optional[str] = None,
+        last_name: Optional[str] = None,
+        role: Optional[Roles] = None,
+    ) -> "User":
+        if not first_name:
+            warnings.warn(
```
davidberenstein1957, line 168: does `warn.warn` include rich formatting?

src/argilla/client/users.py
```diff
@@ -0,0 +1,118 @@
...
+                "Since the `first_name` hasn't been provided, it will be set to the same value as the `username`."
+            )
+            first_name = username
+        if not role:
+            warnings.warn("Since the `role` hasn't been provided, it will be set to `annotator`.")
+            role = "annotator"
+
+        if isinstance(password, UUID):
+            warnings.warn("Since the `password` is a UUID, it will be converted to a string.")
```
davidberenstein1957: Maybe this is irrelevant since the str(UUID) looks exactly the same right?

davidberenstein1957:
> But for the moment, the only available endpoints for user management are the following, so I guess we're not able to, but it's a nice addition for sure.

Yes exactly, because I would expect to be able to update passwords, roles etc.

src/argilla/client/sdk/users/models.py
alvarobartt: That's the API default, so we should be compliant with it. Anyway feel free to open a discussion with @frascuchon and/or @gabrielmbmb 😄

src/argilla/client/users.py
alvarobartt: Sure, I'll do so whenever I come back to this PR 👍🏻

src/argilla/client/sdk/users/models.py
gabrielmbmb: I think 100 is OK! 1Password generates passwords up to 100 characters.

frascuchon:
> > shouldn't we also be able to update users? or will we then need to delete and re-create?
> 
> That's a nice question indeed, maybe @frascuchon can answer this.
> 
> But for the moment, the only available endpoints for user-management are the following, so I guess we're not able to, but it's a nice addition for sure.
> 
> image

The CLI commands for users management are not using the API for the actions. I would prefer to provide basic support from python cli (by using available endpoints) and add more features later iteratively (which maybe means creating a new endpoint to update users, or reset passwords,...)

davidberenstein1957:
> 

Perfect. @frascuchon could you potentially add some context [here](https://github.com/argilla-io/argilla/issues/3079) for our contributors?

src/argilla/client/users.py
alvarobartt: No it does not, but we decided to use both `logger.warning` and `warnings.warn`, as those are intended to different purposes, see https://docs.python.org/3/howto/logging.html#when-to-use-logging

image

alvarobartt: In this case, defining a `first_name` makes the warning disappear, so this is why here we're using `warnings.warn` instead of `logging.warning`, anyway we can have a more in-detail discussing for this matter! 👍🏻

src/argilla/client/users.py
alvarobartt: Remove it! Thanks 👍🏻

src/argilla/client/sdk/users/api.py
alvarobartt: We will refactor this at some point if applicable yes, but closing this for now!

src/argilla/client/sdk/users/api.py
```diff
@@ -12,10 +12,155 @@
...
+    """Sends a GET request to `/api/me` endpoint to get the current user information.
+
+    Args:
+        client: the authenticated Argilla client to be used to send the request to the API.
+
+    Returns:
+        A `UserModel` instance with the current user information.
+    # Returns:
+    #     A `Response` object containing a `parsed` attribute with the parsed response if
+    #     the request was successful, which is an instance of `UserModel`.
```
gabrielmbmb: I guess that we will use this once we update the function to return `Response` instead of `UserModel`. In the meantime, can we remove or move this comment to inside of the function? It will get rendered in the user docstring.

src/argilla/client/sdk/users/models.py
```diff
@@ -12,17 +12,35 @@
...
 #  limitations under the License.
 
+from datetime import datetime
 from typing import List, Optional
+from uuid import UUID
 
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 
+from argilla.server.models import UserRole as Roles
```
frascuchon: I think you should have your own UserRole class instead of using it from the server module

src/argilla/client/sdk/users/models.py
```diff
@@ -12,17 +12,35 @@
...
 
-class User(BaseModel):
-    """User data model"""
 
+class UserCreateModel(BaseModel):
+    first_name: str = Field(min_length=1)
+    last_name: Optional[str] = Field(min_length=1)
+    username: str = Field(min_length=1, regex=r"^(?!-|_)[a-z0-9-_]+$")
+    role: Roles = "annotator"
```
frascuchon:
```suggestion
    role: Roles = Roles.annotator
```

src/argilla/client/users.py
```diff
@@ -0,0 +1,310 @@
...
+    @staticmethod
+    def __active_client(httpx: bool = True) -> Union["httpx.Client", "AuthenticatedClient"]:
+        """Returns the active Argilla `httpx.Client` instance."""
+        try:
+            client = active_client().http_client
+            if httpx:
+                client = client.httpx
+            return client
+        except Exception as e:
+            raise RuntimeError(f"The `rg.active_client()` is not available or not respoding.") from e
```
gabrielmbmb, line 126: I think that we should create a base class that could be used in client classes that have to use the `httpx` client.

src/argilla/client/users.py
```diff
@@ -0,0 +1,310 @@
...
+
+from argilla.client import active_client
+from argilla.client.sdk.commons.errors import (
+    AlreadyExistsApiError,
+    BaseClientError,
+    NotFoundApiError,
+)
+from argilla.client.sdk.users import api as users_api
+from argilla.client.sdk.users.models import UserCreateModel, UserModel
+from argilla.server.models import UserRole
```
frascuchon: Same here, this class definition should be in the `sdk.users.models` module

src/argilla/client/users.py
```diff
@@ -0,0 +1,310 @@
...
+        except NotFoundApiError as e:
+            raise ValueError(
+                f"User with username=`{self.username}` doesn't exist in Argilla, so please"
+                " make sure that the name you provided is a valid one."
+            ) from e
+        except BaseClientError as e:
+            raise RuntimeError(f"Error while deleting user with username=`{self.username}` from Argilla.") from e
+
+    @classmethod
+    def __new_instance(cls, client: Optional["httpx.Client"] = None, ws: Optional["UserModel"] = None) -> "User":
```
gabrielmbmb: does `ws` stand for workspace user?

src/argilla/client/users.py
```diff
@@ -0,0 +1,310 @@
...
+        """Returns a new `User` instance."""
+        instance = cls.__new__(cls)
+        instance.__client = client or cls.__active_client()
+        if isinstance(ws, UserModel):
```
frascuchon:
```suggestion
        if isinstance(user, UserModel):
```

src/argilla/client/users.py
frascuchon:
```suggestion
    def __new_instance(cls, client: Optional["httpx.Client"] = None, user: Optional["UserModel"] = None) -> "User":
```

gabrielmbmb (approved):

src/argilla/client/sdk/users/api.py
alvarobartt: True I'll remove everything in the same commit to easily revert it later!

src/argilla/client/users.py
alvarobartt: Nope that was a copy-paste issue from the `__new_instance` method defined in `Workspace` 😄

src/argilla/client/users.py
alvarobartt: Sure, we can do that on another PR, to also make sure the most recent classes are updated and aligned i.e. `FeedbackDataset`, `User` and `Workspace`

src/argilla/client/users.py
alvarobartt: I agree, it was something temporary that I forgot to update!

src/argilla/client/sdk/users/models.py
alvarobartt: Regarding how `Enum`s are accessed in Python 3.11, do you know if this may eventually lead to any issue? cc @gabrielmbmb

src/argilla/client/sdk/users/models.py
alvarobartt: I agree! Thanks

tests/client/sdk/users/test_api.py
```diff
@@ -12,25 +12,103 @@
...
 
-def test_whoami_with_connection_error():
-    with pytest.raises(BaseClientError):
-        whoami(AuthenticatedClient(base_url="http://localhost:6900", token="wrong-apikey"))
+def test_list_users(owner: User) -> None:
+    UserFactory.create(username="user_1")
+    UserFactory.create(username="user_2")
+    httpx_client = ArgillaSingleton.init(api_key=owner.api_key).http_client.httpx
+
+    response = list_users(client=httpx_client)
```
frascuchon, line 52: For this kind of tests I think you can use directly the `client: TestClient`, fixture
```suggestion
def test_list_users(client: TestClient, owner: User, owner_auth_header: dict) -> None:
    UserFactory.create(username="user_1")
    UserFactory.create(username="user_2")
    client.headers.update(owner_auth_header)

    response = list_users(client=client)
```

tests/client/test_users.py
```diff
@@ -0,0 +1,92 @@
...
+        User(name="test_user")
+
+    with pytest.raises(
+        Exception,
+        match=r"`User` cannot be initialized via the `__init__` method | you should use `User.from_id\('00000000-0000-0000-0000-000000000000'\)`",
+    ):
+        User(id="00000000-0000-0000-0000-000000000000")
+
+
+def test_user_from_name(owner: UserModel):
```
frascuchon: Maybe is more clear as ServerUser
```suggestion
def test_user_from_name(owner: ServerUser):
```

tests/client/test_users.py
```diff
@@ -0,0 +1,92 @@
...
+    assert user.username == new_user.username
+    assert isinstance(user.id, UUID)
+
+    with pytest.raises(ValueError, match="User with username="):
+        User.from_name("non-existing-user")
+
+
+def test_user_from_id(owner: UserModel):
+    new_user = UserFactory.create(username="test_user")
+    ArgillaSingleton.init(api_key=owner.api_key)
```
frascuchon, line 56: It should work with the general `rg.init` function. Just to be more familiar with inits

alvarobartt: Yes, it was basically to avoid using `rg` in the unit tests and use the actual class instead, which seems more clear when debugging to have the direct import instead of the reference to `rg`

tests/client/sdk/users/test_api.py
alvarobartt: But won't the `ArgillaSingleton` use the `TestClient` as configured in the `conftest.py`? Is this change worth it? Asking because if so, we should change this is more than one place, LMKWYT!

src/argilla/client/sdk/users/models.py
```diff
@@ -12,17 +12,40 @@
...
 from typing import List, Optional
+from uuid import UUID
 
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 
 
-class User(BaseModel):
-    """User data model"""
+class UserRoles(str, Enum):
```
frascuchon: Normally this enun class definition is named in singular since attributes and properties will have one single value of this class so one `UserRole`

src/argilla/client/sdk/users/models.py
```diff
@@ -12,17 +12,40 @@
...
+    owner = "owner"
+    admin = "admin"
+    annotator = "annotator"
 
+
+class UserCreateModel(BaseModel):
+    first_name: str = Field(min_length=1)
+    last_name: Optional[str] = Field(min_length=1)
+    username: str = Field(min_length=1, regex=r"^(?!-|_)[a-z0-9-_]+$")
+    role: UserRoles = UserRoles.annotator
```
frascuchon: I think is better as `role: UserRole`

tests/client/test_users.py
frascuchon: we should align then the rest of the tests ( we can tackle in a separate PR)

frascuchon (approved):

tests/client/sdk/users/test_api.py
frascuchon: Yes, just to be more explicit here in tests if we're using the HTTP client directly. The conftest is a fixture to use the rg.init functions with an already mocked client

PR #3173, opened 2023-06-13
:sparkles: FEAT : add `total_records` when search is apply

keithCuniah:
# Description

Add `total_records` when the search is applied.
The info is passed to the search component by bus event.

Note that we update the `total_records` in specific cases: 
- when the search is applied with a value
- when the search is removed
- when the filter status change and a search is applied

**Type of change**

**How Has This Been Tested**

- Only feedback task


frontend/components/feedback-task/DatasetFilters.component.vue
```diff
@@ -58,12 +61,22 @@ export default {
...
 
     this.sortedFiltersValue = Object.values(this.filters).sort((a, b) =>
       a.order > b.order ? -1 : 1
     );
   },
   computed: {
+    additionalInfoForSearchComponent() {
+      if (isNil(this.totalRecords)) return null;
+
+      if ([0, 1].includes(this.totalRecords))
```
frascuchon: Since we have already a page when a search returns no data, maybe we can skip the message when 0 records were found?

cc @nataliaElv @Amelie-V

frontend/components/page-contents/feedback-task-content/RecordFeedbackTaskAndQuestionnaire.content.vue
```diff
@@ -269,6 +243,52 @@ export default {
...
+          _status: this.recordStatusToFilterWith,
+        },
+      });
+    },
+    async recordStatusToFilterWith(newValue) {
+      await this.$router.push({
+        path: this.$route.path,
+        query: {
+          ...this.$route.query,
+          _search: "",
```
frascuchon: do we need to reset the search query param when the record status filter changes?

frascuchon (approved):

PR #3180, opened 2023-06-14
feat: extend functionality of `Workspace` class

alvarobartt:
# Description

This PR extends the functionality of the `Workspace` class created as of Argilla 1.8.0, to not just do name-to-id, but also to be able to manage the workspaces and the users assigned to those, as the API supports those features via the `api/workspaces` endpoints.

**Type of change**

**How Has This Been Tested**

- [x] Add unit tests

**Checklist**


alvarobartt:
Fails in CI/CD are un-related cc @frascuchon @gabrielmbmb

src/argilla/client/feedback/utils.py
```diff
@@ -111,10 +111,10 @@ def feedback_dataset_in_argilla(
 
     if (name and workspace) or name:
         if workspace is None:
-            workspace = rg.Workspace.from_name(rg.active_client().get_workspace())
+            workspace = rg.Workspace(rg.active_client().get_workspace())
```
frascuchon: That means that `Workspace.__init___` will call internally the argilla API, right? I would prefer to have this operation outside of an object constructor. I would say to separate the `Workspace` class for the data model and `Workspaces` class to interact with workspaces in Argilla.
```python
import argilla as rg 

workspace= rg.Workspaces.from_name("my-workspace")
workspace.id # UUID(......)
workspace.name # my-workspace

new_workspace= rg.Workspaces.create(Workspace(name="new-workspace")
new_workspace.id # UUID (...)
new_workspace.name # new-workspace
```

This approach allows us to decouple the internal API integration (will be in Workspaces) from the workspace model itself.

src/argilla/client/sdk/v1/workspaces/api.py
```diff
@@ -0,0 +1,54 @@
...
+    ErrorMessage,
+    HTTPValidationError,
+    Response,
+)
+from argilla.client.sdk.v1.workspaces.models import WorkspaceModel
+
+
+def get_workspace(
+    client: httpx.Client,
+    id: str,
```
frascuchon, line 30: We can pass an UUID since workspace ids will be always an UUID

frascuchon: We can pass a UUID since workspace ids will be always an UUID 
```suggestion
    id: UUID,
```

src/argilla/client/sdk/workspaces/api.py
```diff
@@ -42,7 +40,121 @@ def list_workspaces(client: httpx.Client) -> Response[Union[List[WorkspaceModel]
     response = client.get(url=url)
 
     if response.status_code == 200:
-        parsed_response = [WorkspaceModel(**workspace) for workspace in response.json()]
+        parsed_response = [WorkspaceModel.construct(**workspace) for workspace in response.json()]
```
frascuchon: By using the `construct` method, all automatic casting will be applied. For example `str`-> `UUID` or `str`-> `datetime` will be applied?

src/argilla/client/workspaces.py
```diff
@@ -14,67 +13,238 @@
...
-    def __init__(self, name: str, *, id: Optional[str] = None) -> None:
-        """Initializes a new `Workspace` instance using the provided name. If no name
-        is provided, the current workspace will be used.
+    id: UUID
+    name: str
+    users: Optional[List["WorkspaceUserModel"]] = None
+    inserted_at: datetime
+    updated_at: datetime
+
+    def __init__(self, name: Optional[str] = None, *, id: Optional[str] = None) -> None:
```
frascuchon:
```suggestion
    def __init__(self, name: Optional[str] = None, *, id: Optional[Union[str,UUID]] = None) -> None:
```

but in any case we should check that id contains a valid UUID

src/argilla/client/workspaces.py
```diff
@@ -14,67 +13,238 @@
...
+                    existing_workspace = workspace
+                    break
+
+            if existing_workspace is None:
+                raise ValueError(
+                    f"Workspace with name=`{name}` doesn't exist in Argilla, so please"
+                    "create it via the `Workspace.create` method as follows:"
+                    f" `Workspace(name=`{name}`)`."
+                )
+        self.__dict__.update(existing_workspace.dict())
```
frascuchon: What this line exactly does?

src/argilla/client/workspaces.py
```diff
@@ -14,67 +13,238 @@
...
+
+    @property
+    def users(self) -> List["WorkspaceUserModel"]:
+        """Returns the list of users linked to the workspace.
+
+        Returns:
+            A list of `WorkspaceUserModel` instances.
+        """
+        if not hasattr(self, "__users") or self.__users is None:
+            self.__users = workspaces_api.list_workspace_users(active_client().http_client.httpx, self.id).parsed
```
frascuchon: We should reduce the couple with the active_client logic. I think the httpx client should be part of the workspace on object initialization. Then we can do:
```suggestion
            self.__users = workspaces_api.list_workspace_users(self.__httpx, self.id).parsed
```
otherwise, we can start having problems in multi-thread multi client environments (a process connecting to different argilla server to sync some datasets/workspaces)

src/argilla/client/workspaces.py
```diff
@@ -14,67 +13,238 @@
...
+        Examples:
+            >>> from argilla import rg
+            >>> workspace = rg.Workspace("my-workspace")
+            >>> workspace.add_user("my-user")
         """
-        self.name = name
-        self.id = id or None
+        try:
+            created_user = workspaces_api.create_workspace_user(
+                client=active_client().http_client.httpx,
```
frascuchon: same here
```suggestion
                client=self.__httpx,
```

src/argilla/client/workspaces.py
```diff
@@ -14,67 +13,238 @@
...
+            A list of `WorkspaceUserModel` instances.
+        """
+        if not hasattr(self, "__users") or self.__users is None:
+            self.__users = workspaces_api.list_workspace_users(active_client().http_client.httpx, self.id).parsed
+        return self.__users
+
+    def __str__(self) -> str:
+        return f"Workspace(id={self.id}, name={self.name}, users={self.users}, insterted_at={self.inserted_at}, updated_at={self.updated_at}"
+
+    def add_user(self, id: str) -> None:
```
frascuchon: put UUID instead of str

src/argilla/client/workspaces.py
```diff
@@ -14,67 +13,238 @@
...
+
+        Examples:
+            >>> from argilla import rg
+            >>> workspace = rg.Workspace("my-workspace")
+            >>> workspace.add_user("my-user")
         """
-        self.name = name
-        self.id = id or None
+        try:
+            created_user = workspaces_api.create_workspace_user(
```
frascuchon: you're creating a workspace-user relationship, right?

tests/client/test_workspaces.py
```diff
@@ -17,16 +17,121 @@
 from argilla.client import api
 from argilla.client.workspaces import Workspace
 
+from tests.helpers import SecuredClient
+
+
+def test_workspace_from_name(mocked_client: SecuredClient) -> None:
+    the_api = api.active_api()
+    workspace = the_api.http_client.post("/api/workspaces", json={"name": "test_workspace"})
+    assert workspace["name"] == "test_workspace"
```
frascuchon: You can setup your test by using the DB factories.
```suggestion
    workspace = WorkspaceFactory.create(name="test_workspace") # this will create a workspace in DB that is accessible through the API
```

tests/client/test_workspaces.py
```diff
@@ -17,16 +17,121 @@
...
+    with pytest.warns(DeprecationWarning):
+        workspace = Workspace.from_name("test_workspace")
+        assert workspace.name == "test_workspace"
+        assert isinstance(workspace.id, str)
+
+
+def test_workspace_init(mocked_client: SecuredClient) -> None:
+    the_api = api.active_api()
+    workspace = the_api.http_client.post("/api/workspaces", json={"name": "test_workspace"})
+    assert workspace["name"] == "test_workspace"
```
frascuchon: same here, use test factories instead, will simplify your test

tests/client/test_workspaces.py
```diff
@@ -17,16 +17,121 @@
...
+        json={
+            "first_name": "string",
+            "last_name": "string",
+            "username": "test_user",
+            "role": "admin",
+            "password": "stringst",
+        },
+    )
+    assert user["username"] == "test_user"
+    assert isinstance(user["id"], str)
```
frascuchon: Again you can use the UserFactory to create a user fo the test

src/argilla/client/workspaces.py
alvarobartt: This line is just mapping the `WorkspaceModel` to the current model, so all the attributes are mapped, instead of having to manually set all the attributes one by one e.g. `self.name = existing_workspace.name`

alvarobartt: That's also why I've defined the attributes on the top of the class before the `__init__`

src/argilla/client/workspaces.py
alvarobartt: Sure, then I'll also add the client on top to be reused in the whole class :smile:

src/argilla/client/workspaces.py
alvarobartt: Yes, I just renamed the method to be `add_user` in the `Workspace` class just to avoid using the `create_workspace_user` naming as I found it a little bit confusing, while `add_user` seems more straight forward

tests/client/test_workspaces.py
alvarobartt: Right, I'll refactor the unit tests then!

alvarobartt:
Thanks for the review @frascuchon I'll try to address all the comments now! 👍🏻

src/argilla/client/sdk/workspaces/api.py
alvarobartt: No it wont', so maybe we just prefer to avoid `construct` usage? Even though it will just affect the `id` and the `inserted/created_at` attributes (and the last two ones are not to be modified, and a `str` formatted as an UUID can be parsed as UUID, so I see no issue. But if we want to be more strict with typing sure, we can just just avoid the `construct` to validate the fields (I was just using the latter when there were nested `pydantic.BaseModel`s)

src/argilla/client/feedback/utils.py
alvarobartt: This is what we had before, but it's a little bit weird because users will be able to register workspaces that don't exist in Argilla, and this functionality is just intended to be used within Argilla, but if you think it's better to keep the init simpler then we can just do nothing on init, and keep the `from_name` which is still there

src/argilla/client/sdk/workspaces/models.py
```diff
@@ -14,12 +13,33 @@
 #  limitations under the License.
 
 from datetime import datetime
+from typing import List, Optional
+from uuid import UUID
+
+try:
+    from typing import Literal
+except ImportError:
+    from typing_extensions import Literal
```
gabrielmbmb, line 22: I think that we can just import `from typing import Literal` without trying to catch the exception. `Literal` was added in Python 3.8, which is the oldest version that we support, right?

I think that I've seen this in some other places, we should update it everywhere.

alvarobartt: So we've already removed support for Python 3.7? I can create a separate PR to remove the unsupported stuff from now on 👍🏻

src/argilla/client/workspaces.py
```diff
@@ -13,68 +12,281 @@
...
+            >>> from argilla import rg
+            >>> workspace = rg.Workspace(
+            ...     client=rg.active_client().http_client.httpx,
+            ...     name="my-workspace",
+            ...     id="...",
+            ...     inserted_at=datetime.datetime(2021, 8, 31, 10, 0, 0),
+            ...     updated_at=datetime.datetime(2021, 8, 31, 10, 0, 0),
+            ... )
         """
+        self.client = client
```
frascuchon: I think we can keep this attribute private, since won't be used outside this class
```suggestion
        self._client = client
```

src/argilla/client/workspaces.py
```diff
@@ -13,68 +12,281 @@
...
-                f"The `rg.Workspace` doesn't exist yet, you need to create it in advance. Otherwise, use `rg.Workspace.from_name` to make sure that the workspace exists."
+    @property
+    def users(self) -> List["WorkspaceUserModel"]:
+        """Returns the list of users linked to the workspace.
+
+        Returns:
+            A list of `WorkspaceUserModel` instances.
+        """
+        if not hasattr(self, "__users") or self.__users is None:
+            self.__users = workspaces_api.list_workspace_users(self.client, self.id).parsed
```
frascuchon: Should be better to create the __users attribute on init, just to know what attributes (public and privates) contain the instances by reading the init method?

src/argilla/client/workspaces.py
```diff
@@ -13,68 +12,281 @@
...
+            ValueError: if the user with the provided ID already exists in the workspace.
+            RuntimeError: if there was an error while adding the user to the workspace.
+
+        Examples:
+            >>> from argilla import rg
+            >>> workspace = rg.Workspace.from_name("my-workspace")
+            >>> workspace.add_user("my-user-id")
+        """
+        try:
+            created_user = workspaces_api.create_workspace_user(
```
frascuchon: should we check if the user is found in the local store before?

src/argilla/client/workspaces.py
```diff
@@ -13,68 +12,281 @@
...
+            ValueError: if the user with the provided ID doesn't exist in the workspace.
+            RuntimeError: if there was an error while deleting the user from the workspace.
+
+        Examples:
+            >>> from argilla import rg
+            >>> workspace = rg.Workspace.from_name("my-workspace")
+            >>> workspace.delete_user("my-user-id")
+        """
+        try:
+            workspaces_api.delete_workspace_user(
```
frascuchon, line 167: same here, maybe we need to find it in list before, just for early validations

src/argilla/client/workspaces.py
alvarobartt: Since this is just intended to be used internally, we're just defining `users` as that's the property the users will use to list the linked users to a given workspace, there's no need to create `self.__users`, as it won't be accessed/used until `self.users` is accessed.

alvarobartt: Since this is just intended to be used internally, we're just defining `users` as that's the property the users will use to list the linked users to a given workspace, there's no need to create `self.__users`, as it won't be accessed/used until `self.users` is accessed.

src/argilla/client/workspaces.py
alvarobartt: This is what we had before, a validation layer against Argilla in the `__init__`, but since we decided to remove it, I think we can just rely on the API validation included as part of those endpoints

alvarobartt: Either calling the API to check whether the user exists or not and then to add it to the workspace (2 API calls) or just add it to the workspace (1 API call) and if it fails we handle the exception to show either that the user doesn't exist and/or the user is already there, in case it fails.

src/argilla/client/workspaces.py
alvarobartt: So same response as above, is it worth to have another API call to repeat a validation that is already included as part of the endpoints we're calling? Isn't just better to catch the exceptions and show the relevant error messages based on those?

src/argilla/client/workspaces.py
```diff
@@ -13,68 +12,303 @@
...
+            raise ValueError(f"User with id=`{id}` already exists in workspace with id=`{self.id}`.") from e
+        except BaseClientError as e:
+            raise RuntimeError(f"Error while adding user with id=`{id}` to workspace with id=`{self.id}`.") from e
+
+        if len(self.users) == 0:
+            self.__users = [created_user]
+        else:
+            self.__users.append(created_user)
+
+    # TODO(alvarobartt): also allow user addition via name (which is more user-friendly)
```
frascuchon: Regarding this, once we have the `User` classes we can pass User instances instead of IDs or names and let the responsibility of find users by name to the `User.from_name` class method

alvarobartt: Yes, indeed I'm already preparing the user-management layer so that `users` contains a list of validated users instead, but I agree that we can either allow id, name, and or `User` instance

src/argilla/client/workspaces.py
alvarobartt: Ok I'll make it private then! 👍🏻

src/argilla/client/workspaces.py
```diff
@@ -13,68 +12,303 @@
...
+        return self.__users
+
+    def __str__(self) -> str:
+        return (
+            f"Workspace(id={self.id}, name={self.name}, users={self.users}, insterted_at={self.inserted_at},"
+            f" updated_at={self.updated_at}"
+        )
+
+    # TODO(alvarobartt): also allow user addition via name (which is more user-friendly)
+    def add_user(self, id: str) -> None:
```
frascuchon: I'm missing a way to refresh o resync the workspace with Argilla server. If an external process is adding or deleting new users, it would be nice to refresh data locally to fetch latest changes without a new `Workspace.from_id(ws.id)`

A workflow could be:
```python
workspace = Workspace.from_name("ws")
workspace.users # []
... # here another process added 2 users (user_id=1 and user_id=2) to this workspace
workspace.add_user(user_id=1) # this will fail because the user already exists in this workspace
workspace.refresh()
workspace.users # will return the updated user list
```

frascuchon: we can explicit the id is for a user
```suggestion
    def add_user(self, user_id: str) -> None:
```

src/argilla/client/workspaces.py
```diff
@@ -13,68 +12,303 @@
...
+        except BaseClientError as e:
+            raise RuntimeError(f"Error while adding user with id=`{id}` to workspace with id=`{self.id}`.") from e
+
+        if len(self.users) == 0:
+            self.__users = [created_user]
+        else:
+            self.__users.append(created_user)
+
+    # TODO(alvarobartt): also allow user addition via name (which is more user-friendly)
+    def delete_user(self, id: str) -> None:
```
frascuchon: same here
```suggestion
    def delete_user(self, user_id: str) -> None:
```

src/argilla/client/workspaces.py
alvarobartt: Sure I can easily included, I just didn't think about it as a possible workflow, but you never know 😆

src/argilla/client/workspaces.py
```diff
@@ -13,68 +12,305 @@
...
+        Examples:
+            >>> from argilla import rg
+            >>> workspace = rg.Workspace.create("my-workspace")
+        """
+        instance = cls.__new__(cls)
+        instance.__client = cls.__active_client()
+        try:
+            ws = workspaces_api.create_workspace(client=instance.__client, name=name).parsed
+            instance.__dict__.update(ws.dict())
+            return instance
```
frascuchon: Maybe we can create a new instance after we're sure everything went well. 
```suggestion
        try:
            ws = workspaces_api.create_workspace(client=instance.__client, name=name).parsed
            # This code block is a potential WorspaceFactory method
            instance = cls.__new__(cls)
            instance.__client = cls.__active_client()
            instance.__dict__.update(ws.dict())
            return instance
```

src/argilla/client/workspaces.py
```diff
@@ -13,68 +12,305 @@
...
+        Examples:
+            >>> from argilla import rg
+            >>> workspace = rg.Workspace.from_id("my-workspace-id")
+        """
+        instance = cls.__new__(cls)
+        instance.__client = cls.__active_client()
+        try:
+            ws = workspaces_api_v1.get_workspace(instance.__client, id).parsed
+            instance.__dict__.update(ws.dict())
+            return instance
```
frascuchon: same here, and I think we can reuse the new-instance code block

src/argilla/client/workspaces.py
```diff
@@ -13,68 +12,305 @@
...
+        instance.__client = cls.__active_client()
+        try:
+            workspaces = workspaces_api.list_workspaces(instance.__client).parsed
+        except Exception as e:
+            raise RuntimeError("Error while retrieving the list of workspaces from Argilla.") from e
+
+        for ws in workspaces:
+            if ws.name == name:
+                instance.__dict__.update(ws.dict())
+                return instance
```
frascuchon: same here, I think we can first find the proper workspace and then create the instance with workspace + client

src/argilla/client/workspaces.py
```diff
@@ -13,68 +12,305 @@
...
-                return cls(name, id=workspace.id)
-        raise ValueError(f"Workspace {name} not found in the current Argilla account.")
+        client = cls.__active_client()
+        try:
+            workspaces = workspaces_api.list_workspaces(client).parsed
+            for ws in workspaces:
+                instance = cls.__new__(cls)
+                instance.__client = client
+                instance.__dict__.update(ws.dict())
+                yield instance
```
frascuchon: code duplication

frascuchon (approved):

PR #3190, opened 2023-06-14
:bug: FIX : correction in search logic

keithCuniah:
# Description

Correction in search for feedback task.
1/ be able to always delete an active search
2/ render the close button 

**Type of change**

- Bug fix

**How Has This Been Tested**

- only feedback task

### DIFF
--- a/frontend/components/feedback-task/SearchBar.base.vue
+++ b/frontend/components/feedback-task/SearchBar.base.vue
@@ -55,7 +55,7 @@
     />
 
     <BaseIconWithBadge
-      v-if="isSearchActive"
+      v-if="showDelete"
       class="searchbar__icon"
       icon="close"
       :show-badge="false"
@@ -69,6 +69,8 @@
 </template>
 
 <script>
+import { isNil } from "lodash";
+
 export default {
   name: "SearchBarComponent",
   props: {
@@ -97,14 +99,24 @@ export default {
   },
   computed: {
     isSearchActive() {
-      return this.value?.length;
+      return this.value?.length ?? false;
+    },
+    isActiveSearchEmpty() {
+      return isNil(this.value) || this.value.length === 0;
+    },
+    showDelete() {
+      if (isNil(this.searchValue)) return false;
+      if (this.isActiveSearchEmpty && this.searchValue.length === 0)
+        return false;
+
+      return true;
     },
   },
   watch: {
     value: {
       immediate: true,
       handler(newValue) {
-        this.searchValue = newValue ?? "";
+        this.searchValue = newValue;
       },
     },
   },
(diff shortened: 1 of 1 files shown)


leiyre (approved):

frontend/components/feedback-task/SearchBar.base.vue
```diff
@@ -93,14 +99,27 @@ export default {
   },
   computed: {
     isSearchActive() {
-      return this.value?.length;
+      return this.value?.length ?? false;
```
frascuchon: could we rewrite this in terms of `isActiveSearchEmpty`? It seems the opposite thing. Could you do:
```javascript
isSearchActive() {
   return !this.isActiveSearchEmpty;
}
```

or 

```javascript
showDelete() {
      if (isNil(this.searchValue) && this.searchValue.length === 0)) return false;
      return this.isSearchActive;
    },
```

frontend/components/feedback-task/SearchBar.base.vue
```diff
@@ -92,15 +99,28 @@ export default {
...
+    },
+    searchValueIsEmpty() {
+      return isNil(this.searchValue) || this.searchValue.length === 0;
+    },
+    showDelete() {
+      if (isNil(this.searchValue)) return false;
+      if (this.isActiveSearchEmpty && this.searchValue.length === 0)
+        return false;
+
+      return true;
```
frascuchon: is this equivalent?

```suggestion
      return !(this.isActiveSearchEmpty && this.searchValue.length === 0);
```

keithCuniah: !(this.isActiveSearchEmpty && this.searchValue`?`.length === 0); to add also the isNil test.
I make it in several lines instead of one to make it more readable. What is it better in this case ?

frontend/components/feedback-task/SearchBar.base.vue
```diff
@@ -48,10 +48,14 @@
       @keydown.enter.exact.stop=""
     />
 
-    <span class="search-area__additional-info" v-text="additionalInfo" />
+    <span
+      class="search-area__additional-info"
+      v-if="!searchValueIsEmpty"
```
frascuchon: I think we should show the additional info only when the query search prop is the same as the query search in the search bar

frascuchon:
```suggestion
      v-if="value === searchValue"
```

keithCuniah: value === searchValue && value.length => to not show additional info when value and searchValue are empty

frascuchon: A question here, additional info is passed as a prop and I figure out that an empty or null value is passed when the value is not set, right?

In that case, we should add the condition of having some additional info to show

Issue #3191, opened 2023-06-14
Review  the `sdk.v1.datasets` data model API


frascuchon:
The low-level layer is using direct feedback schemas or raw dictionaries to communicate with. We should improve this layer definition to:

1. Improve data serialization/deserialization using pydantic features
2. Decouple low-level schemas and feedback client schemas

alvarobartt:
Closed as completed upon https://github.com/argilla-io/argilla/pull/3784 completion

PR #3192, opened 2023-06-14
Apply proper serializer when pushing feedback dataset in Argilla

### LINKED ISSUE #3189: [BUG] Cannot copy Feedback dataset to be published as another name/workspace
**Describe the bug**

When I load a feedback dataset using `from_argilla` method and I try to push into argilla with another name/workspace, I fetch the error:
Screenshot 2023-06-14 at 16 45 13

**To Reproduce**
1. Load a Feedback dataset from argilla using `.from_argilla`
2. Push the same dataset in Argilla again with another name `.push_to_argilla("new_name")`

**Expected behavior**
This is a way to copy/update/reindex this kind of dataset, so it should be allowed.

 - OS [e.g. iOS]:
 - Argilla Version [e.g. 1.0.0]: 1.9.0

frascuchon:
# Description

By default, the `model.dict()` method won't apply custom serialization to the `UUID` or `datetatime` fields. Since pydantic provides a `.json` method which applies serialization properly, we should use them in order to support pushing data in Argilla.

The current solution may be improved since we're serializing more times than needed. But for that, we should change the sdk.v1.api layer. See issue #3191

Closes #3189 

**Type of change**

**How Has This Been Tested**

New tests have been added

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,7 @@ These are the section headers that we use:
 
 - Resolve breaking issue with `ArgillaSpanMarkerTrainer` for Named Entity Recognition with `span_marker` v1.1.x onwards.
 - Move `ArgillaDatasetCard` import under `@requires_version` decorator, so that the `ImportError` on `huggingface_hub` is handled properly ([#3174](https://github.com/argilla-io/argilla/pull/3174))
-- Allow load feedback datasets from Argilla and push them back in Argilla with different name/workspace (Closes [#3189](https://github.com/argilla-io/argilla/issues/3189))
+- Allow flow `FeedbackDataset.from_argilla` -> `FeedbackDataset.push_to_argilla` under different dataset names and/or workspaces ([#3192](https://github.com/argilla-io/argilla/issues/3192))
 
 ## [1.9.0](https://github.com/argilla-io/argilla/compare/v1.8.0...v1.9.0)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ These are the section headers that we use:
 
 - Resolve breaking issue with `ArgillaSpanMarkerTrainer` for Named Entity Recognition with `span_marker` v1.1.x onwards.
 - Move `ArgillaDatasetCard` import under `@requires_version` decorator, so that the `ImportError` on `huggingface_hub` is handled properly ([#3174](https://github.com/argilla-io/argilla/pull/3174))
+- Allow load feedback datasets from Argilla and push them back in Argilla with different name/workspace (Closes [#3189](https://github.com/argilla-io/argilla/issues/3189))
 
 ## [1.9.0](https://github.com/argilla-io/argilla/compare/v1.8.0...v1.9.0)
 
--- a/src/argilla/client/feedback/dataset.py
+++ b/src/argilla/client/feedback/dataset.py
@@ -11,10 +11,11 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
-
+import json
 import logging
 import tempfile
 from typing import TYPE_CHECKING, Any, Dict, Iterator, List, Optional, Union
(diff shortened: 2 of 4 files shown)


src/argilla/client/feedback/dataset.py
```diff
@@ -555,7 +556,7 @@ def delete_and_raise_exception(dataset_id: str, exception: Exception) -> None:
 
             for field in self.fields:
                 try:
-                    datasets_api_v1.add_field(client=httpx_client, id=argilla_id, field=field.dict())
+                    datasets_api_v1.add_field(client=httpx_client, id=argilla_id, field=json.loads(field.json()))
```
alvarobartt, line 559: Can't we just provide the `.json()` instead?

src/argilla/client/feedback/dataset.py
```diff
@@ -568,9 +569,7 @@ def delete_and_raise_exception(dataset_id: str, exception: Exception) -> None:
             for question in self.questions:
                 try:
                     datasets_api_v1.add_question(
-                        client=httpx_client,
-                        id=argilla_id,
-                        question=question.dict(include={"name", "title", "description", "required", "settings"}),
+                        client=httpx_client, id=argilla_id, question=json.loads(question.json())
```
alvarobartt, line 572: We shouldn't include extra fields, that why there was an `include=set()` before, IMO we should keep it

alvarobartt (commented):
LGTM! Just some minor comments regarding the `json.loads(schema.json())`, and about the removal of `.dict(includes=set())` 🤗

src/argilla/client/feedback/dataset.py
frascuchon: The .json returns a string representation of the field. I didn't do it to avoid changing the sdk.api logic in this PR. We should have a look into that layer later.

src/argilla/client/feedback/dataset.py
frascuchon: Yes, but currently the API will discard them.

Ideally, we should have an SDK schema for each operation and create them from our feedback.schema instances:

```python
class QuestionCreate(BaseModel):
    title: str
    name: str
    settings: dict


api.add_question(client, QuestionCreate.from_raw(feedback_question)
```

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ These are the section headers that we use:
 
 - Resolve breaking issue with `ArgillaSpanMarkerTrainer` for Named Entity Recognition with `span_marker` v1.1.x onwards.
 - Move `ArgillaDatasetCard` import under `@requires_version` decorator, so that the `ImportError` on `huggingface_hub` is handled properly ([#3174](https://github.com/argilla-io/argilla/pull/3174))
+- Allow load feedback datasets from Argilla and push them back in Argilla with different name/workspace (Closes [#3189](https://github.com/argilla-io/argilla/issues/3189))
```
```suggestion
- Allow flow `FeedbackDataset.from_argilla` -> `FeedbackDataset.push_to_argilla` under different dataset names and/or workspaces ([#3192](https://github.com/argilla-io/argilla/issues/3192))
```

src/argilla/client/feedback/dataset.py
```diff
@@ -11,10 +11,11 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
-
+import json
```
alvarobartt, line 14: I guess we could maybe use `orjson` instead as it's faster? https://github.com/ijl/orjson

alvarobartt: Unless this is just temporary, in that case I guess we're good

alvarobartt (approved):

src/argilla/client/feedback/dataset.py
frascuchon: Yes, this would be a temporary solution. Ideally, we should to setup orjson as the default serializer/deserializer used by the `httpx.Client`, but still an open discussion here: https://github.com/encode/httpx/issues/717

PR #3194, opened 2023-06-14
feat: add `metadata` to `Record`

### LINKED ISSUE #3155: [FEATURE] Add metadata field to the FeedbackRecord
Add support for adding metadata to the FeedbackRecord. 

Currently, when pushing the records to the dataset, I can only push the fields that are later used as a `TextField`. In some scenarios, I would like to push not only the values but also the metadata, e.g., the hash values associated with different fields. They come in handy when analysing the results after the labelling is completed. I'm aware of the `external_id` but in many use cases it is not sufficient.

The solution would be to either:
- not require all the used fields to be present in `TextField`.
- add a `metadatafield` param that would capture such fields.

The code: 

```py
fields = [
    rg.TextField(name="instruction", required=True),
    rg.TextField(name="input", required=True),
    rg.TextField(name="completion_a", required=True),
    rg.TextField(name="completion_b", required=True),
]

dataset = rg.FeedbackDataset(
    guidelines="Please, read the prompt carefully and...",
    questions=questions,
    fields=fields,
)


def build_records(path: Path) -> list[rg.FeedbackRecord]:
    with open(path) as f:
        data = json.load(f)

    records = []

    for point in data:
        record = rg.FeedbackRecord(
            fields={
                    "instruction_hash": point["instruction_hash"],
                    "instruction": point["instruction"],
                    "input": point["input"],
                    "completion_a": point["completion_a"],
                    "completion_b": point["completion_b"],
                    "completion_hash_a": point["completion_hash_a"],
                    "completion_hash_b": point["completion_hash_b"],
                    "pair_hash": point["pair_hash"],
                    }
        )
        records.append(record)

        return records


records = build_records(path=Path(__file__).parent / "data.json")
dataset.add_records(records)
dataset.push_to_argilla(name="my-dataset")
```

The exception

```
Exception: Failed while adding the records to the `FeedbackTask` dataset in 
Argilla with exception: Argilla server returned an error with http status: 422
Error details: [{'response': "Error: found fields values for non configured 
fields: ['instruction_hash', 'completion_hash_a', 'completion_hash_b', 
'pair_hash']", 'params': {}}]
```

gabrielmbmb:
# Description

This PR adds a new attribute called `metadata` to the `Record` of the `FeedbackDataset`.

- A new `metadata` column/attribute has been added to the `Record` ORM class (a new column `metadata` will be added by the migration script generated by `alembic`)
- All the endpoints in the API v1 listing records has been updated to return this new `metadata` column
- The SDK has been updated to parse the `metadata` key returned by the API
- The Python client has been updated so new records can be created including `metadata`

Additionally, I've refactor some `if/else` conditions in some methods from the `FeedbackDataset` class.

Closes #3155

**Type of change**

- [x] Refactor

**How Has This Been Tested**

Unit tests has been updated and I've done some manual tests.

**Checklist**


src/argilla/client/feedback/dataset.py
```diff
@@ -611,12 +611,10 @@ def delete_and_raise_exception(dataset_id: UUID, exception: Exception) -> None:
...
                 _LOGGER.warning(
-                    "Since the current object is already a `FeedbackDataset` pushed to"
-                    " Argilla, you'll keep on interacting with the same dataset in"
-                    " Argilla, even though the one you just pushed holds a different"
-                    f" ID ({argilla_id}). So on, if you want to switch to the newly"
-                    " pushed `FeedbackDataset` instead, please use"
-                    f" `FeedbackDataset.from_argilla(id='{argilla_id}')`."
+                    "Since the current object is already a `FeedbackDataset` pushed to Argilla, you'll keep on"
+                    " interacting with the same dataset in Argilla, even though the one you just pushed holds a"
+                    f" different ID ({argilla_id}). So on, if you want to switch to the newly  pushed `FeedbackDataset`"
```
alvarobartt:
```suggestion
                    f" different ID ({argilla_id}). So on, if you want to switch to the newly pushed `FeedbackDataset`"
```

src/argilla/client/feedback/dataset.py
```diff
@@ -489,7 +489,7 @@ def push_to_argilla(self, name: Optional[str] = None, workspace: Optional[Union[
         """
         httpx_client: "httpx.Client" = rg.active_client().http_client.httpx
 
-        if not name or (not name and not workspace):
+        if name is None:
```
frascuchon, line 492: With this change, the warning message looks outdated. Do we still need to check the workspace argument?

src/argilla/client/feedback/dataset.py
```diff
@@ -549,46 +549,38 @@ def delete_and_raise_exception(dataset_id: UUID, exception: Exception) -> None:
...
                 datasets_api_v1.publish_dataset(client=httpx_client, id=argilla_id)
             except Exception as e:
-                delete_and_raise_exception(
-                    dataset_id=argilla_id,
-                    exception=Exception(
-                        f"Failed while publishing the `FeedbackTask` dataset in Argilla with exception: {e}"
-                    ),
-                )
+                exc = Exception(f"Failed while publishing the `FeedbackTask` dataset in Argilla with exception: {e}")
+                exc.__cause__ = e
```
frascuchon: maybe we should use the conventional way to do this? https://docs.python.org/3/library/exceptions.html#exception-context
```python
delete_dataset(dataset_id=argilla_id)
raise exc from e
```

src/argilla/client/feedback/utils.py
```diff
@@ -78,7 +78,7 @@ def feedback_dataset_in_argilla(
     *,
     workspace: Optional[Union[str, rg.Workspace]] = None,
     id: Optional[str] = None,
-) -> Tuple[bool, Optional["FeedbackDatasetModel"]]:
+) -> Union["FeedbackDatasetModel", None]:
```
frascuchon, line 81: should this be the same?
```suggestion
) -> Optional["FeedbackDatasetModel"]:
```

src/argilla/server/database.py
```diff
@@ -27,7 +27,7 @@
     from sqlalchemy.orm import Session
 
 ALEMBIC_CONFIG_FILE = os.path.normpath(os.path.join(os.path.dirname(argilla.__file__), "alembic.ini"))
-TAGGED_REVISIONS = OrderedDict({"1.7": "1769ee58fbb4", "1.8": "ae5522b4c674"})
+TAGGED_REVISIONS = OrderedDict({"1.7": "1769ee58fbb4", "1.8": "ae5522b4c674", "1.10": "3ff6484f8b37"})
```
frascuchon: Probably this will change to 1.11

frascuchon: Maybe we can have this track in the CHANGELOG as a change instead of supporting this revision tagging. Let's discuss it.

src/argilla/server/models/models.py
```diff
@@ -114,6 +114,7 @@ class Record(Base):
 
     id: Mapped[UUID] = mapped_column(primary_key=True, default=uuid4)
     fields: Mapped[dict] = mapped_column(JSON, default={})
+    metadata_: Mapped[Optional[dict]] = mapped_column("metadata", JSON, nullable=True)
```
frascuchon, line 117: `metadata` is a reserved attribute in sqlalchemy orm models, right?

src/argilla/server/schemas/v1/datasets.py
```diff
@@ -287,9 +288,19 @@ class RecordInclude(str, Enum):
     responses = "responses"
 
 
+class RecordGetterDict(GetterDict):
+    def get(self, key: Any, default: Any = None) -> Any:
+        if key == "metadata":
+            return getattr(self._obj, "metadata_", {})
+        if key == "responses" and "responses" not in self._obj.__dict__:
```
frascuchon, line 295: Now I understand the change in handlers. Much better like this 👍

src/argilla/server/schemas/v1/datasets.py
```diff
@@ -299,11 +310,21 @@ class Record(BaseModel):
 
     class Config:
         orm_mode = True
+        getter_dict = RecordGetterDict
+
+    @classmethod
+    def from_orm(cls: Type["Record"], obj: Any) -> "Record":
+        dict_copy = obj.__dict__.copy()
+        return cls(**{"metadata": dict_copy.pop("metadata_", None), **dict_copy})
```
frascuchon: why do we need this?

src/argilla/server/schemas/v1/datasets.py
```diff
@@ -299,11 +310,21 @@ class Record(BaseModel):
...
 
 
 class Records(BaseModel):
     items: List[Record]
 
+    @classmethod
+    def from_orm(cls: Type["Records"], obj: List["Record"]) -> "Records":
+        return cls(items=[Record.from_orm(record) for record in obj])
+
```
frascuchon: same here. Why do we need this?

src/argilla/server/models/models.py
gabrielmbmb: it is, yes

src/argilla/client/feedback/utils.py
gabrielmbmb: I prefer to use `Union["FeedbackDatasetModel", None]` because the syntax is close to what we will have in Python 3.11 `"FeedbackDatasetModel" | None`, and I think that for return type hints the intent is more clear.

https://fastapi.tiangolo.com/python-types/#using-union-or-optional

src/argilla/server/schemas/v1/datasets.py
gabrielmbmb: In the listing records endpoints we were using `record.__dict__` to build the `Record` schema. This is because if we use `record` directly, then `pydantic` when creating the `Record` class will try to access `record.responses` and `sqlalchemy` will return an empty list no matter what, so in the output payload we would have `"responses": []` even if the client has not asked for the responses.

To fix this, we use `record.__dict__` which does not include the `responses` if they were not included in the query with a `joinedload` or other load option. Now that we've added the `metadata_` attribute, we cannot use `record.__dict__` directly to build the `Record` class, because the field in this class is called `metadata` and we have to map this manually.

As we had to do this in several endpoints, I've decided that it was a good idea to put this logic in this method.

src/argilla/client/feedback/dataset.py
gabrielmbmb: Yes, I think that's better than doing weird stuff like this :) So `delete_and_raise_exception` -> `delete_dataset`, right?

src/argilla/server/schemas/v1/datasets.py
gabrielmbmb: I just realise that we don't need this custom `GetterDict`. I added this to try to make it work with the default `from_orm` method, but if we return `None` in this condition `if key == "responses" and "responses" not in self._obj.__dict__:`, `pydantic` will detect that we have set a value for the `Record.responses` attribute, and then in the listing records endpoints `response_model_exclude_unset=True` won't affect this field, and we will have `"responses: null"` in the output payload.

src/argilla/client/feedback/dataset.py
frascuchon: exactly!

src/argilla/client/feedback/utils.py
frascuchon: Okay, so we should review all Optional defs in the server and client to keep them aligned

src/argilla/server/schemas/v1/datasets.py
frascuchon: Okay, so we keep using record.__dict___?

gabrielmbmb: yes

src/argilla/client/feedback/dataset.py
gabrielmbmb: no, we don't. I'll update the warning message

src/argilla/server/schemas/v1/datasets.py
frascuchon: Can we use an alias then for the record.metadata field? Or use a RecordGetter class to tackle this? I feel this from_orm override is a bit tricky

gabrielmbmb: I managed to do the same with the  `GetterDict` and keep the `response` as unset if the query didn't have the `joinedload` option. It was as easy as returning the `default` argument instead of `None`. 

```python
class RecordGetterDict(GetterDict):
    def get(self, key: str, default: Any) -> Any:
        if key == "metadata":
            return getattr(self._obj, "metadata_", None)
        if key == "responses" and "responses" not in self._obj.__dict__:
            return default
        return super().get(key, default)
```

This way `pydantic` detects that the field has not been set and in combination with the decorator parameter `response_model_exclude_unset=True` it works.

frascuchon (approved):

PR #3221, opened 2023-06-19
Improvement: Harden dockerfile

osintalex:
# Hardening Dockerfile to run without root user

First pass on hardening the dockerfile. Wasn't sure how you are generating the `dist` file and have been getting some errors about accessing the sql database so this isn't ready yet. Wanted to get the ball rolling though.

Can you let me know how you're making the dist bundle and then I can test it more properly 😇 
**Type of Change**


release.Dockerfile
```diff
@@ -1,33 +1,42 @@
+FROM python:3.9-slim AS builder
+
+# Copying argilla distribution files
+COPY dist/*.whl /packages/
+RUN python -m venv /my-virtual-env
+ENV PATH="/my-virtual-env/bin:$PATH"
+RUN apt-get update && \
+    apt-get install -y python-dev-is-python3 libpq-dev gcc && \
```
osintalex: I couldn't get `python-dev` to install so changed to `python-dev-is-python3`

release.Dockerfile
```diff
@@ -1,33 +1,42 @@
...
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/* && \
+    rm -rf /packages
+
+
 FROM python:3.9-slim
 
 # Environment Variables
-ENV ARGILLA_HOME_PATH=/var/lib/argilla
+ENV ARGILLA_HOME_PATH=/home/worker/argilla
```
osintalex: `/var/lib` doesn't allow non-root ownership afaik

gabrielmbmb:
Hi @osintalex, thanks for the PR! For building the dist bundle we're using https://github.com/pypa/build which by default generates the dist build in `.whl` and `.tar.gz` formats.

davidberenstein1957:
Hi @osintalex thanks for picking this up!💪🏽🚀

frascuchon:
Thanks @osintalex for this improvement. As @gabrielmbmb pointed out, the distribution packages are generated using the `build` command. We skip the build in the dockerfile build since all web static files must be generated, which can increase the final docker image.

davidberenstein1957:
@frascuchon @gabrielmbmb can this be merged?

osintalex:
Got time to have another look at this, the errors are happening because of the argilla users database requires root permission to access it. 

Do you know how that database gets created? Bit confused since I can see that you're install postgresql but looks like the database is sqlite. Haven't had time to poke around in the source code properly to see where it comes from.

davidberenstein1957:
Normally we default to sqlite but if we were to have a configuration of postgresql availble, we would turn to that. This is also not really clearly documented. 

@frascuchon @gabrielmbmb

release.Dockerfile
frascuchon: Yes, the latest python base image upgraded the OS, and the `python-dev` is not working anymore. Thanks

release.Dockerfile
frascuchon: 👍 

May this change affect old containers updating to this image?

frascuchon:
> Got time to have another look at this, the errors are happening because of the argilla users database requires root permission to access it.
> 
> Do you know how that database gets created? Bit confused since I can see that you're install postgresql but looks like the database is sqlite. Haven't had time to poke around in the source code properly to see where it comes from.

Hi @osintalex, sorry for the late response. The Postgresql dependencies are installed just to allow the use of this image to connect with a Postgresql database. As you mention, the default dataset is SQLite and the `argilla.db` file is created under the `ARGILLA_HOME` path. 

Then, the script [`start_argilla_server.sh` ](https://github.com/osintalex/argilla/blob/harden-dockerfile/scripts/start_argilla_server.sh#L5) will apply the database migration to properly setup the database.

docker/Dockerfile
```diff
@@ -1,33 +1,42 @@
...
+ENV USERS_DB=/config/.users.yml
+ENV UVICORN_PORT=6900
+
+RUN useradd -ms /bin/bash worker
+COPY scripts/start_argilla_server.sh /home/worker
+COPY --chown=worker:worker --from=builder /my-virtual-env /home/worker/my-virtual-env
+ENV PATH="/home/worker/my-virtual-env/bin:$PATH"
+
 # Create argilla volume
 RUN mkdir -p "$ARGILLA_HOME_PATH"
```
frascuchon: I've been doing some tests but I found problems applying the database migration. Argilla's home folder is created by `root` user. So, when Argilla is running, the `worker` user has no permissions to execute database migrations (and all write-db-related actions) in `/home/worker/argilla`

frascuchon: After fixing permission problems, the db migration task can be applied, but I found an error pointing to a missing file:

```bash
docker-argilla-1        | start_argilla_server.sh: line 19: /home/worker/my-virtual-env/bin/uvicorn: cannot execute: required file not found
```

osintalex: Hey! Thanks for doing all that and explaining the sqlite thing.

I'm not sure why that has happened tbh, I did notice that before. I got round it when testing by just doing `python -m uvicorn` rather than `uvicorn`. 

Probably need to do something clever when copying the virtual environment over in the multistage build so uvicorn is on path or tbh it might be simpler to just install uvicorn as part of the second stage docker build.

Please feel free to push commits/makes changes to my code if that's faster/easier. Otherwise will take a look when I next get time, hopefully tomorrow or next week.

release.Dockerfile
osintalex: tl;dr don't think so but I reckon you would know better!

Not sure I totally understand the question so forgive me if this is wrong... What other containers have this image as a dependency? You mean like if someone else is running their own say production webapp that then makes api calls to an argilla container for something?

In that case yes it would be a breaking change if they expect some argilla stuff to be in `/var/lib/argilla` rather than `/home/worker/argilla`. 

I don't know the code as well as you but that doesn't seem to me like the kind of place another container should be looking. In that example I gave there where say my webapp container is making api requests to argilla I would imagine I'd just be hitting api routes like `http://argilla:6900/some_route` rather than poking around in the argilla container file system?

frascuchon (approved):

PR #3224, opened 2023-06-19
fix: `format_as("datasets")` when no responses

alvarobartt:
# Description

As just reported by @dvsrepo, the following flow: `FeedbackDataset.from_argilla() -> .format_as("datasets")` was failing under some scenarios where either the optional responses had no value or there were no responses at all, as either we were trying to retrieve those when any, or we were setting it's value to `None` if not found, instead of `{"user_id": None, "values": None, "status": None}` which is what 🤗 Datasets is producing based on the features.

So on, this PR solves that bug and now the flow mentioned above is possible and works as expected!

**Type of change**

**How Has This Been Tested**

- [X] To also cover the scenario when there are no responses, the fixture `feedback_dataset_records` has been modified to contain a record with and another without responses

**Checklist**

### DIFF
--- a/src/argilla/client/feedback/dataset.py
+++ b/src/argilla/client/feedback/dataset.py
@@ -757,14 +757,16 @@ class FeedbackDataset:
                     dataset[field.name].append(record.fields[field.name])
                 for question in self.questions:
                     dataset[question.name].append(
-                        [
-                            {
-                                "user_id": r.user_id or None,
-                                "value": r.values[question.name].value or None,
-                                "status": r.status or None,
-                            }
-                            for r in record.responses
-                        ]
+                        {
+                            "user_id": [r.user_id for r in record.responses],
+                            "value": [
+                                r.values[question.name].value if question.name in r.values else None
+                                for r in record.responses
+                            ],
+                            "status": [r.status for r in record.responses],
+                        }
+                        if record.responses
+                        else None
                     )
                 dataset["metadata"].append(json.dumps(record.metadata) if record.metadata else None)
                 dataset["external_id"].append(record.external_id or None)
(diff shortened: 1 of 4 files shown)


dvsrepo (commented):
cool @alvarobartt ! please ask someone to review this as I might just approve it blindly 😄

src/argilla/client/feedback/dataset.py
```diff
@@ -759,13 +759,12 @@ def format_as(self, format: Literal["datasets"]) -> "Dataset":
                     dataset[question.name].append(
                         [
                             {
-                                "user_id": r.user_id,
-                                "value": r.values[question.name].value,
-                                "status": r.status,
+                                "user_id": r.user_id or None,
+                                "value": r.values[question.name].value or None,
```
frascuchon: this may fail when the value is a `0`. In those cases, the returning value will be `None, ' which is unexpected. Can you be more explicit in setting the default value for user_id, value, and status?

tests/client/conftest.py
```diff
@@ -428,17 +428,6 @@ def feedback_dataset_records() -> List[FeedbackRecord]:
...
-                {
-                    "values": {
-                        "question-1": {"value": "This is a response to question 1"},
-                        "question-2": {"value": 1},
-                        "question-3": {"value": "a"},
-                        "question-4": {"value": ["a", "b"]},
-                    },
-                    "status": "submitted",
-                }
-            ],
```
frascuchon, line 441: I think we should include a specific test case to check the bugfix. Not sure this change is covering it.

alvarobartt: This covers it, as well as ensuring robustness since the responses are optional, so we're mocking a `FeedbackRecord` with and without responses 👍🏻

src/argilla/client/feedback/dataset.py
alvarobartt: Nice catch, didn't think about the `0 or None` case, but you're right!

frascuchon: Since some questions can be optional, I think we should prevent this.
```suggestion
                                "value": r.values.get(question.name) or None, # TODO: we should fetch first the response and then the value
```

gabrielmbmb (approved):

PR #3227, opened 2023-06-20
chore: move `license_header` to root dir

alvarobartt:
# Description

Just a minor chore change to place the `license_header.txt` in the root dir instead of in `src` to ease the navigation via GitHub, and place it as a static file in the root directory rather than under `src/`, unless there was a specific reason for it TBC with @frascuchon 

**How Has This Been Tested**

- [x] Added a file with no header, and run `pre-commit run` to make sure the header was injected


frascuchon (approved):
Thanks, @alvarobartt 

We don't need this file under `src` folder

PR #3231, opened 2023-06-20
fix: `push_to_huggingface` parsing of UUIDs

alvarobartt:
# Description

Due to the recent constraint for the `FeedbackRecord.user_id`s to be `UUID`s instead of `str`s, the conversion to `dict()` was failing, as the `UUID`s are not JSON-serializable, which was leading to some issues when trying to add the example record to the `DatasetCard` via the `dict()` conversion.

**Type of change**

**How Has This Been Tested**

- [X] Re-run unit tests to pass when generating the `DatasetCard` for `FeedbackDataset`s

**Checklist**

### DIFF
--- a/src/argilla/client/feedback/dataset.py
+++ b/src/argilla/client/feedback/dataset.py
@@ -891,7 +891,7 @@ class FeedbackDataset:
             repo_type="dataset",
             **hub_auth,
         )
-        with open(config_path, "rb") as f:
+        with open(config_path, "r") as f:
             config = FeedbackDatasetConfig.parse_raw(f.read())
 
         cls = cls(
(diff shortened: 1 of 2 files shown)


gabrielmbmb (approved):
Hey Alvaro, LGTM for now :) maybe we could think to add a custom type like `StrUUID` https://docs.pydantic.dev/latest/usage/types/#classes-with-__get_validators__, so when calling `.dict()` we get an `str` instead of a `UUID`.

Using `json.loads(model.json())` makes the trick, but it's a bit inefficient as we're serializing/deserializing the whole model.

alvarobartt:
> Hey Alvaro, LGTM for now :) maybe we could think to add a custom type like `StrUUID` https://docs.pydantic.dev/latest/usage/types/#classes-with-__get_validators__, so when calling `.dict()` we get an `str` instead of a `UUID`.
> 
> Using `json.loads(model.json())` makes the trick, but it's a bit inefficient as we're serializing/deserializing the whole model.

Yes, I agree with this, I'll tackle next release, this was a patch proposed by @frascuchon but since it's just the `UUID`s and the `datetime`s, and we're not pushing `datetime`s the solution above or a custom encoder would do the work! Thanks

PR #3234, opened 2023-06-21
fix: `push_to_argilla` and `from_argilla` consistency issues

alvarobartt:
# Description

This PR removes the unnecessary `json.loads(field.json())` and `json.loads(question.json())` as all the fields in those `pydantic.BaseModel`s are JSON-serializable, so we can just provide the `dict` to the endpoint.

Besides that, we're also using the `dict` for `FeedbackRecord`s, and then ensuring that the `user_id` in each response, which is an `UUID` and so on, not JSON-serializable, is properly serialized as a string in the `add_records` SDK function.

This also aligns both the `FeedbackRecord` logging in Argilla under the scenarios of pushing a new dataset and pushing records to an existing dataset.

Finally, also the `from_argilla` field and question re-construction is patched, as we were still using `construct` which was not respecting the `Config` set on those `pydantic.BaseModel`s as `construct` implies `Config.extra = 'allow'`. So on, now that's patched too, and the `FeedbackDataset` retrieved via `from_argilla` is exactly the same one as the one logged via `push_to_argilla` and/or `push_to_huggingface`; as well as the `argilla_id` assignment when retrieving a `FeedbackDataset.from_argilla` once the `argilla_id` has already been set as `cls.argilla_id = ...`.

**Type of change**

**How Has This Been Tested**

- [X] Add unit tests for `push_to_argilla` to make sure the API doesn't throw any HTTP 422 when the payload contains UUIDs
- [X] Add unit tests for `from_argilla` to make sure the retrieve `FeedbackDataset` is consistent to what we should expect

**Checklist**


gabrielmbmb (approved):
LGTM!

PR #3237, opened 2023-06-21
docs: update assignments code in tutorials and llm guides

nataliaElv:
# Description

Changes code snippets in the LLM practical guides and examples to apply the changes in user and workspace management from version 1.11.0

**Type of change**

- [x] Documentation update

**Checklist**

### DIFF
--- a/docs/_source/guides/llms/examples/curating-feedback-instructiondataset.ipynb
+++ b/docs/_source/guides/llms/examples/curating-feedback-instructiondataset.ipynb
@@ -45,7 +45,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "%pip install argilla==1.11.0 datasets pandas httpx plotly -qqq"
+    "%pip install argilla==1.11.0 datasets pandas plotly -qqq"
    ]
   },
   {
@@ -66,7 +66,7 @@
     "from datasets import Dataset, load_dataset\n",
     "import pandas as pd\n",
     "import random\n",
-    "from collections import defaultdict, Counter\n",
+    "from collections import defaultdict\n",
     "import plotly.express as px"
    ]
   },
--- a/docs/_source/guides/llms/examples/curating-feedback-instructiondataset.ipynb
+++ b/docs/_source/guides/llms/examples/curating-feedback-instructiondataset.ipynb
@@ -45,7 +45,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "%pip install argilla==1.11.0 datasets pandas plotly -qqq"
+    "%pip install --upgrade argilla datasets pandas plotly -qqq"
    ]
   },
   {
--- a/docs/_source/guides/llms/examples/curating-feedback-instructiondataset.ipynb
+++ b/docs/_source/guides/llms/examples/curating-feedback-instructiondataset.ipynb
@@ -45,7 +45,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "%pip install argilla datasets pandas httpx plotly -qqq"
+    "%pip install argilla==1.11.0 datasets pandas httpx plotly -qqq"
    ]
   },
   {
@@ -65,7 +65,6 @@
     "import argilla as rg\n",
     "from datasets import Dataset, load_dataset\n",
     "import pandas as pd\n",
-    "import httpx\n",
     "import random\n",
     "from collections import defaultdict, Counter\n",
     "import plotly.express as px"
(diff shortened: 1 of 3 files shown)


docs/_source/guides/llms/practical_guides/set_up_annotation_team.md
```diff
@@ -20,22 +20,29 @@ Here's how you can do this:
...
 # make a request using your Argilla Client to get the list of users
 rg_client= rg.active_client().client
 auth_headers = {"X-Argilla-API-Key": rg_client.token}
 http=httpx.Client(base_url=rg_client.base_url, headers=auth_headers)
 users = http.get("/api/users").json()
 
-# optional: filter users to get only those with annotator role
-users = [u for u in users if u["role"] == "annotator"]
+# filter users to get only those with annotator role
+users = [user for user in users if user['role']=='annotator']
```
alvarobartt: Here I think we should be consistent with the double-quotes usage, as we're doing in most of the cases, correct me if I'm wrong @davidberenstein1957 but we should advocate for " > ', right?

nataliaElv: Totally my bad! I have the bad habit of not being consistent with the quotes. Changing it now!

alvarobartt (approved):
LGTM!

docs/_source/guides/llms/examples/curating-feedback-instructiondataset.ipynb
```diff
@@ -45,7 +45,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "%pip install argilla datasets pandas httpx plotly -qqq"
+    "%pip install argilla==1.11.0 datasets pandas plotly -qqq"
```
frascuchon: This version will be outdated quickly. Maybe just an `-U` to keep packages updated?

docs/_source/guides/llms/practical_guides/set_up_annotation_team.md
```diff
@@ -20,22 +20,29 @@ Here's how you can do this:
...
 import argilla as rg
 
 rg.init(
     api_url="...",
     api_key="..."
 )
 
+# get the list of users
+# optional: filter users to get only those with annotator role
+users = [u for u in rg.User.list() if u.role == "annotator"]
```
frascuchon, line 32: Here we can use `UserRole.annotator`, right @alvarobartt ?

alvarobartt: Yes, but `UserRole` is not included as an init import so if that's the case we should `from argilla.client.sdk.users.models import UserRole`, so I guess we're good like this.

alvarobartt: Anyway, I guess that in Python 3.11 this may fail as we need to access Enum value https://docs.python.org/3/library/enum.html#enum.Enum.value, i.e. `users = [user for user in rg.User.list() if user.role.value == "annotator"]`

Note that the above just applies to Python 3.11 or greater

docs/_source/guides/llms/examples/curating-feedback-instructiondataset.ipynb
nataliaElv: I'm just worried that the tutorial could also get outdated and if we don't keep updating the code often tutorials might not work with future Argilla versions. This one we know it works for 1.11.0 and that's why I wanted to mention it like that, but if you think it's preferable to keep updating tutorials to the last Argilla version then I'm happy changing it.

frascuchon: For me make sense to align with the rest of the tutorials. If this is the way to install argilla in other tutorials, let's keep it as is.

docs/_source/guides/llms/practical_guides/set_up_annotation_team.md
frascuchon: So in that case I prefer to expose `UserRole` definition in the root module and use it.
```python
from argilla import UserRole

...
if user.role == UserRole.annotator:

```

 Also, [python refs](https://docs.argilla.io/en/docs-1.11.0-docs/reference/python/python_users.html#argilla.client.users.User.role) make reference to this `UserRole` enum. And maybe we should align also the User Model [attribute](https://docs.argilla.io/en/docs-1.11.0-docs/getting_started/installation/configurations/user_management.html#user-model)

docs/_source/guides/llms/examples/curating-feedback-instructiondataset.ipynb
alvarobartt: It was internally agreed that we should use `--upgrade` instead, and we'll review the rest of the tutorials iteratively in upcoming releases.

PR #3249, opened 2023-06-22
feat: create `/data` directory in quickstart Docker image

gabrielmbmb:
# Description

I've updated the quickstart image to create the `/data` directory which will be used to store Argilla Server data and Elasticsearch data.

This change is mainly needed so the quickstart is compatible with the new HF Spaces feature, persistent storage. HF Spaces persistent storage mounts a volume in the `/data` directory of the container, so `ARGILLA_HOME_PATH` should be set to `/data/argilla` (or other directory but within `/data`). Additionally, we need to also change the `path.data` config parameter of Elasticsearch to `/data/elasticsearch`, so elastic data is also persisted.

If the HF persistent storage is not used (volume is not mounted), then the `/data` directory won't exist and the Linux user used in the quickstart Docker image (`elasticsearch`) won't have the rights needed to create it. To avoid this from happening, we also create the `/data` directory in the Dockerimage and we give permissions to the `elasticsearch` user to manage this directory.

I've also moved the `ENV` in the top of the Dockerfile to the bottom. This is to take advantage of the Dockerimage layering system, so if we change one of these `ENV`s the cached layers are not invalidated and we don't have to reinstall everything again.

**Type of change**

**How Has This Been Tested**

Running the new `quickstart` image locally.

**Checklist**

### DIFF
--- a/quickstart.Dockerfile
+++ b/quickstart.Dockerfile
@@ -1,24 +1,13 @@
 FROM docker.elastic.co/elasticsearch/elasticsearch:8.5.3
 
-ENV ADMIN_USERNAME=admin
-ENV ADMIN_PASSWORD=12345678
-ENV ADMIN_API_KEY=admin.apikey
-
-ENV ANNOTATOR_USERNAME=argilla
-ENV ANNOTATOR_PASSWORD=12345678
-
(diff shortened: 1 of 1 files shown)


alvarobartt (approved):
LGTM!

PR #3253, opened 2023-06-22
🚀 Login users to demo env without credentials

damianpumar:
This feature allow users to access without enter the username and password to login.

See #3252 

Screen Recording 2023-06-23 at 08 58 38

### DIFF
--- a/frontend/pages/login.spec.js
+++ b/frontend/pages/login.spec.js
@@ -3,9 +3,13 @@ import Login from "./login.vue";
 
 const stubs = ["BaseLoading", "brand-logo", "geometric-shape-a", "base-button"];
 
+const validAuthToken = btoa(
+  JSON.stringify({ username: "USERNAME", password: "PASSWORD" })
+);
+
 describe("Login page should", () => {
   it("still in the same page if the auth token is not valid", () => {
-    const userLoginSpy = jest.spyOn(Login.methods, "userLogin");
+    const loginUserSpy = jest.spyOn(Login.methods, "loginUser");
 
     shallowMount(Login, {
       mocks: {
@@ -17,11 +21,11 @@ describe("Login page should", () => {
       },
     });
 
-    expect(userLoginSpy).toHaveBeenCalledTimes(0);
+    expect(loginUserSpy).toHaveBeenCalledTimes(0);
   });
 
   it("still in the same page if the auth token query params is empty", () => {
-    const userLoginSpy = jest.spyOn(Login.methods, "userLogin");
+    const loginUserSpy = jest.spyOn(Login.methods, "loginUser");
 
     shallowMount(Login, {
       mocks: {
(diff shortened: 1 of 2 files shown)


tomaarsen:
🙏 Awesome! I've wanted this feature for a while!

frontend/pages/login.vue
```diff
@@ -102,6 +110,20 @@ export default {
           : this.error;
       }
     },
+    authToken() {
```
leiyre: In the case that we expect the url to be non-reactive we can not use a computed property and instead use `created`

frontend/pages/login.vue
```diff
@@ -102,6 +110,20 @@ export default {
...
+      const rawAuthToken = this.$route.query.auth;
+
+      if (!rawAuthToken) return;
+
+      try {
+        const auth = JSON.parse(atob(rawAuthToken));
+
+        if (!!auth.username && !!auth.password) return auth;
+      } catch {}
```
leiyre: Please fix `Empty block statement` error

damianpumar: Hi Leire, The error in this case is not important, as the token can have any value.

leiyre: Is because the python package build is failing

tomaarsen: @damianpumar https://github.com/argilla-io/argilla/actions/runs/5350353696/jobs/9702843904?pr=3253

frontend/pages/login.spec.js
```diff
@@ -0,0 +1,182 @@
+import { shallowMount } from "@vue/test-utils";
+import Login from "./login.vue";
+
+const stubs = ["BaseLoading", "brand-logo", "geometric-shape-a", "base-button"];
+
+const validAuthToken = btoa(
+  JSON.stringify({ username: "USERNAME", password: "PASSWORD" })
```
frascuchon: A question here, not sure if make sense to align the expected format with the [Basic auth format](https://datatracker.ietf.org/doc/html/rfc7617#autoid-3):

> To receive authorization, the client
> 
>    1.  obtains the user-id and password from the user,
> 
>    2.  constructs the user-pass by concatenating the user-id, a single
>        colon (":") character, and the password,
> 
>    3.  encodes the user-pass into an octet sequence (see below for a
>        discussion of character encoding schemes),
> 
>    4.  and obtains the basic-credentials by encoding this octet sequence
>        using Base64 ([[RFC4648], Section 4](https://datatracker.ietf.org/doc/html/rfc4648#section-4)) into a sequence of US-ASCII
>        characters ([[RFC0020](https://datatracker.ietf.org/doc/html/rfc0020)]).

 instead of an encoded custom json one. Just to simplify the usability

damianpumar: Yes, why not. I was thinking in a token a bit more "safe" although your proposal is simpler. I don't have any problem to update it.

frascuchon (approved):

leiyre (approved):

PR #3256, opened 2023-06-23
feat: add `ArgillaSpaCyTransformersTrainer` & improve `ArgillaSpaCyTrainer`

alvarobartt:
# Description

This PR adds support for `spacy-transformers` via the new `ArgillaSpaCyTransformersTrainer` class allowing the user to lock the `transformer` model not to be updated, and the `ArgillaSpaCyTrainer` is improved to allow re-using the `tok2vec` or freeze it, if available.

Besides that, this PR also includes a new arg in `ArgillaTrainer` named `framework_kwargs` which is a Python dict to contain the framework-specific kwargs, in this case intended to be created as `{"update_transformer": False}` and `{"freeze_tok2vec": False}` for both `ArgillaSpaCyTransformersTrainer` and `ArgillaSpaCyTrainer`, respectively. Ideally, we should also move the `spaCy` specific-args already included as `ArgillaTrainer` args.

**Type of change**

**How Has This Been Tested**

- [X] Add unit-tests for `ArgillaSpaCyTransformersTrainer`

**Checklist**

- [X] I filled out [the contributor form](https://tally.so/r/n9XrxK) (see text above)


src/argilla/training/base.py
```diff
@@ -46,6 +46,7 @@ def __init__(
         train_size: Optional[float] = None,
         seed: Optional[int] = None,
         gpu_id: Optional[int] = -1,
+        framework_kwargs: Optional[dict] = {},
```
davidberenstein1957, line 58: @alvarobartt For me it makes more sense to add this to the `update_config()`.

src/argilla/training/spacy.py
```diff
@@ -35,18 +37,17 @@ def __init__(
         language: Optional[str] = None,
         gpu_id: Optional[int] = -1,
         model: Optional[str] = None,
+        optimize: Literal["efficiency", "accuracy"] = "efficiency",
```
davidberenstein1957, line 40: perhaps for me this makes sense to add to `update_config`too

tests/training/test_spacy.py
```diff
@@ -13,89 +13,98 @@
...
-        assert trainer._trainer.use_gpu == spacy.prefer_gpu(0)
-
-
-def test_predict_wo_training(dataset_text_classification):
-    trainer = ArgillaTrainer(name=dataset_text_classification, framework=FRAMEWORK, model=MODEL)
-    trainer._trainer.predict("This is a text")
+from tests.training.helpers import cleanup_spacy_config, train_with_cleanup
+
+
+@pytest.mark.parametrize(
```
davidberenstein1957, line 22: You are not really passing the new variables in the tests right?

davidberenstein1957 (commented):
LGTM and thanks for the fix. Perhaps it makes more sense to me to pass the variable via the `update_config` to align with the usage of the rest of the frameworks. Also, I don't see the new config options being test right?

tests/training/test_spacy.py
alvarobartt: Yes we are, as we included the args in the `test_` methods and then created the `TestSpaCyTrainer` class, so those values are passed to every function, and we're "looping" twice, once for `spacy` and another time for `spacy-transformers` 😄

src/argilla/training/spacy.py
alvarobartt: It's not so straight forward to modify that from the configuration file itself, so we should keep it as is, since we're using the `optimize` value to generate one config or the other initially, more things differ in those configuration files, not just a string as it may seem according to the `optimize` arg

src/argilla/training/base.py
alvarobartt: As we can also define the `optimize` value there I'd keep it as is, plus in an upcoming iteration moving the `spaCy`-specific stuff we have there to the `framework_kwargs` arg, to be provided in the `__init__`

src/argilla/training/spacy.py
```diff
@@ -95,9 +100,10 @@ def __init__(
             self._dataset if isinstance(self._dataset, tuple) and len(self._dataset) > 1 else (self._dataset, None)
         )
         self._train_dataset_path = "./train.spacy"
-        self._eval_dataset_path = "./dev.spacy" if self._eval_dataset else None
+        self._eval_dataset_path = "./dev.spacy" if self._eval_dataset else "./train.spacy"
```
davidberenstein1957, line 103: do we need to provide an eval dataset by default? For me it feels weird to use the training dataset to do this.

davidberenstein1957:
Hi @alvarobartt, could you also include the implementation and tests for the FeedbackDataset?

src/argilla/training/spacy.py
alvarobartt: Yes, it is mandatory, the solution/workaround here is to use the train dataset, but note that here I just simplified the code, before we had the exact same thing just that we were setting it to None first and then setting it to `train` if it was None

But as it's mandatory I also agree with raising an error if not provided

alvarobartt:
> Hi @alvarobartt, could you also include the implementation and tests for the FeedbackDataset?

Not sure I'll have time for this release, need to focus now on the `RankingQuestion`, but we can create a ticket for the next one

davidberenstein1957:
I can have a look tomorrow afternoon. I implemented most stuff so it should only require some small changes.

src/argilla/training/spacy.py
davidberenstein1957: No, this seems like a better solution. Thanks

alvarobartt:
Hi @davidberenstein1957 can you have a look into the unit test that is failing? Is there anything that you changed w.r.t. `ArgillaSpaCyTrainer` in the `ArgillaTrainer` for the `FeedbackDataset`s? I guess we're good to merge as the failing unit test is unrelated?

davidberenstein1957:
@alvarobartt, I added the integration and some test. Can you have a final look tomorrow before merging?

davidberenstein1957 (approved):

PR #3263, opened 2023-06-25
Bugfixes/align availables workspaces and set workspace operation

### LINKED ISSUE #3262: Cannot set workspace after create a new one
Once one owner user creates a new workspace, the `rg.set_workspace` function does not accept the new workspace.

```python
import argilla as rg

ws = rg.Workspace.create(name)


rg.set_workspace(ws.name)
# Exception: Wrong provided workspace <name>

```

frascuchon:
# Description

This PR adapts the client workflow to check available workspaces before the `rg.set_workspace` function.  Since the `set_workspace` is based on `list_workspaces` method, a code refactor has been applied to return only available workspaces for the connected user.

@alvarobartt we need to align the `User.workspaces`, `list_workspace`, and `Workspace.create`

Closes #3262 

**Type of change**

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@ These are the section headers that we use:
 
 ## [Unreleased]
 
+### Fixed
+
+- Check available workspaces on Argilla on `rg.set_workspace` (Closes [#3262](https://github.com/argilla-io/argilla/issues/3262))
+
 ## [1.11.0](https://github.com/argilla-io/argilla/compare/v1.10.0...v1.11.0)
 
 ### Fixed
--- a/src/argilla/client/client.py
+++ b/src/argilla/client/client.py
@@ -28,6 +28,7 @@ from rich.progress import Progress
 
 from argilla._constants import (
     DEFAULT_API_KEY,
+    DEFAULT_USERNAME,
     ES_INDEX_REGEX_PATTERN,
     WORKSPACE_HEADER_NAME,
 )
@@ -136,7 +137,24 @@ class Argilla:
         )
 
         self._user = users_api.whoami(client=self.http_client)  # .parsed
-        self.set_workspace(workspace or self._user.username)
+
+        if not workspace and self._user.username == DEFAULT_USERNAME:
+            warnings.warn(
+                "Default user was detected and no workspace configuration was provided, so "
+                f"the default {DEFAULT_USERNAME!r} workspace will be used. "
+                "If you want to setup another workspace, use the `rg.set_workspace` function "
+                "or provide a different one on `rg.init`",
+                category=UserWarning,
+            )
+            workspace = DEFAULT_USERNAME
+        if workspace:
+            self.set_workspace(workspace or self._user.username)
+        else:
+            warnings.warn(
+                "No workspace configuration was detected. To work with Argilla datasets, "
+                "specify a valid workspace name on `rg.init` or set it up through the `rg.set_workspace` function. ",
+                category=UserWarning,
+            )
 
         self._check_argilla_versions()
 
@@ -214,10 +232,12 @@ class Argilla:
             )
 
         if workspace != self.get_workspace():
-            if workspace == self.user.username or (self.user.workspaces and workspace in self.user.workspaces):
-                self.http_client.update_headers({WORKSPACE_HEADER_NAME: workspace})
-            else:
-                raise Exception(f"Wrong provided workspace {workspace}")
+            # TODO(frascuchon): user should list ONLY their accessible workspaces
+            for ws in self.list_workspaces():
+                if ws.name == workspace:
+                    self.http_client.update_headers({WORKSPACE_HEADER_NAME: ws.name})
+                    return
+            raise ValueError(f"Wrong provided workspace {workspace!r}")
 
     def get_workspace(self) -> str:
         """Returns the name of the active workspace.
--- a/src/argilla/client/client.py
+++ b/src/argilla/client/client.py
@@ -140,10 +140,10 @@ class Argilla:
 
         if not workspace and self._user.username == DEFAULT_USERNAME:
             warnings.warn(
-                "Default user was detected and no workspace configuration was provided, so "
-                f"the default {DEFAULT_USERNAME!r} workspace will be used. "
-                "If you want to setup another workspace, use the `rg.set_workspace` function "
-                "or provide a different one on `rg.init`",
+                "Default user was detected and no workspace configuration was provided,"
+                f" so the default {DEFAULT_USERNAME!r} workspace will be used. If you"
+                " want to setup another workspace, use the `rg.set_workspace` function"
+                " or provide a different one on `rg.init`",
                 category=UserWarning,
             )
             workspace = DEFAULT_USERNAME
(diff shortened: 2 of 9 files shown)


src/argilla/client/client.py
```diff
@@ -136,7 +137,24 @@ def __init__(
...
+                "or provide a different one on `rg.init`",
+                category=UserWarning,
+            )
+            workspace = DEFAULT_USERNAME
+        if workspace:
+            self.set_workspace(workspace or self._user.username)
+        else:
+            warnings.warn(
+                "No workspace configuration was detected. To work with Argilla datasets, "
+                "specify a valid workspace name on `rg.init` or set it up through the `rg.set_workspace` function. ",
```
alvarobartt:
```suggestion
                "specify a valid workspace name on `rg.init` or set it up through the `rg.set_workspace` function.",
```

src/argilla/client/client.py
```diff
@@ -136,7 +137,24 @@ def __init__(
...
+                category=UserWarning,
+            )
```
alvarobartt:
```suggestion
        if not workspace and self._user.username == DEFAULT_USERNAME:
    warnings.warn(
        "Default user was detected and no workspace configuration was provided,"
        f" so the default {DEFAULT_USERNAME!r} workspace will be used. If you"
        " want to setup another workspace, use the `rg.set_workspace` function"
        " or provide a different one on `rg.init`",
        category=UserWarning,
    )
    workspace = DEFAULT_USERNAME

if not workspace:
    warnings.warn(
        "No workspace configuration was detected. To work with Argilla"
        " datasets, specify a valid workspace name on `rg.init` or set it"
        " up through the `rg.set_workspace` function.",
        category=UserWarning,
    )

self.set_workspace(workspace or self._user.username)
```

src/argilla/client/client.py
```diff
@@ -228,13 +246,15 @@ def get_workspace(self) -> str:
...
     def list_workspaces(self) -> List[WorkspaceModel]:
-        """Lists all the availble workspaces for the current user.
+        """Lists all the available workspaces for the current user.
 
         Returns:
             A list of `WorkspaceModel` objects, containing the workspace
             attributes: name, id, created_at, and updated_at.
         """
-        return workspaces_api.list_workspaces(client=self.http_client.httpx).parsed
+        user_workspaces = users_api.whoami(self.http_client).workspaces
```
alvarobartt, line 256: TODO: at some point we should also refactor the `whoami` to use the `httpx.Client` instead and return the response, instead of directly the parsed response, plus we should add proper error handling

src/argilla/client/sdk/users/api.py
```diff
@@ -42,7 +41,20 @@ def whoami(
...
+            return Response(
+                status_code=response.status_code,
+                content=response.content,
+                headers=response.headers,
+                parsed=UserModel(**response.json()),
+            )
+        else:
+            handle_response_error(response)
+    else:
+        return UserModel(**response)
```
alvarobartt: I agree with this TODO, but I guess we can revert this for the moment, until we create an actual PR to refactor it, taking both scenarios into consideration is not a good thing IMO, we already had this but decided to revert in favour of the current approach until we fix the mocks

```suggestion
    return UserModel(**response)
```

src/argilla/client/users.py
alvarobartt, line 1: Even if `workspaces` is a property, we can have it defined in the attributes, as the only diff with the rest of the attributes is that it may raise an `AttributeError` if we try to set it, as we didn't define a setter for the property

src/argilla/client/users.py
```diff
@@ -115,16 +114,21 @@ def __str__(self) -> str:
         )
 
     @staticmethod
-    def __active_client(httpx: bool = True) -> Union["httpx.Client", "AuthenticatedClient"]:
+    def __active_client() -> Union["httpx.Client", "AuthenticatedClient"]:
```
alvarobartt: As mentioned before, I wouldn't revert this until the mock is fixed in a separate PR

src/argilla/client/users.py
```diff
@@ -115,16 +114,21 @@ def __str__(self) -> str:
...
-                client = client.httpx
-            return client
+            return active_client().http_client.httpx
         except Exception as e:
             raise RuntimeError(f"The `rg.active_client()` is not available or not respoding.") from e
 
+    @property
+    def workspaces(self) -> List[str]:
+        try:
+            user = users_api.whoami(self.__client).parsed
```
alvarobartt: What if the user we want to list the workspaces is not the active one, because we're the owner and we want to list the workspaces for a given user?

We shouldn't use `whoami` here, but a request to the API to get the workspaces for the current user based on an ID or name match, but not using `whoami`, as the activated user may be owner or admin

src/argilla/client/users.py
```diff
@@ -278,10 +282,10 @@ def me(cls) -> "User":
             >>> from argilla import rg
             >>> user = rg.User.me()
         """
-        client = cls.__active_client(httpx=False)
+        client = cls.__active_client()
         try:
-            user = users_api.whoami(client)  # .parsed
-            return cls.__new_instance(client.httpx, user)
+            user = users_api.whoami(client).parsed
+            return cls.__new_instance(client, user)
```
alvarobartt: Same here, I would revert this until fixed

alvarobartt (changes requested):
LGTM, but in `User` there are some inconsistencies as we're not always the current `User` but e.g. the owner listing all the users, so `workspaces` listing would fail, and I wouldn't update the SDK until the mock is fixed in a separate PR. But I agree with the missing type-hints, extended unit tests, and `Argilla.__init__` fixes

src/argilla/client/users.py
frascuchon: Totally true. I will revert all changes related to the `workspaces` attribute and we'll plan a PR to improve this (maybe we should create a new endpoint to fetch info for an specific user)

src/argilla/client/sdk/users/api.py
```diff
@@ -26,11 +27,9 @@
 from argilla.client.sdk.users.models import UserCreateModel, UserModel, UserRole
 
 
-# TODO(alvarobartt,frascuchon): use `httpx.Client` instead of `AuthenticatedClient` and
+# TODO(alvarobartt,frascuchon): use ONLY `httpx.Client` instead of `AuthenticatedClient` and
 # fix mock in `tests/conftest.py` to use `httpx.Client` instead of `AuthenticatedClient`
-def whoami(
-    client: AuthenticatedClient,
-) -> UserModel:
+def whoami(client: AuthenticatedClient) -> Union[UserModel, Response[UserModel]]:
```
frascuchon:
```suggestion
def whoami(client: AuthenticatedClient) -> UserModel:
```

src/argilla/client/users.py
frascuchon: rolled back

src/argilla/client/users.py
frascuchon: rollback

src/argilla/client/users.py
frascuchon: all this is rolled back

CHANGELOG.md
```diff
@@ -18,6 +18,10 @@ These are the section headers that we use:
 
 ## [Unreleased]
 
+### Fixed
```
alvarobartt, line 21: Shouldn't this go under the 1.11.1 if we're planning to release it as a bug-fix?

frascuchon: This fix will be released in v1.12.0

alvarobartt (approved):

PR #3278, opened 2023-06-27
feat: Use `__repr__`   for new user and workspace client models

### LINKED ISSUE #3261: Implement __str__ and __repr__ for new User and Workspace client models
Python consoles should show user info properly instead of the generic one: 

```python
In [4]: rg.User.me()
Out[4]: <argilla.client.users.User at 0x7fdf9247e130>
```

frascuchon:
Based on what [documentation says](https://docs.python.org/3.9/reference/datamodel.html#object.__repr__) about *`__repr__`* and *`__str__`* methods, this PR implement the `__repr__` method for both `rg.User` and `rg.Workspace` classes, so the `str` will use the same implement.

Closes #3261


tests/client/test_workspaces.py
```diff
@@ -122,3 +122,16 @@ def test_workspace_delete_user(owner: "ServerUser") -> None:
...
     with pytest.raises(ValueError, match="Either the user with id="):
         workspace.delete_user(owner.id)
+
+
+def test_print_workspace(owner: "ServerUser"):
+    workspace = WorkspaceFactory.create(name="test_workspace")
+
+    import argilla as rg
+
+    rg.init(api_key=owner.api_key)
```
alvarobartt:
```suggestion
    ArgillaSingleton.init(api_key=owner.api_key)
```

tests/client/test_users.py
```diff
@@ -93,3 +93,15 @@ def test_user_delete_user(owner: "ServerUser") -> None:
     user.delete()
     with pytest.raises(ValueError, match="doesn't exist in Argilla"):
         user.delete()
+
+
+def test_print_user(owner: "ServerUser"):
+    import argilla as rg
+
+    rg.init(api_key=owner.api_key)
```
alvarobartt:
```suggestion
    ArgillaSingleton.init(api_key=owner.api_key)
```

src/argilla/client/workspaces.py
```diff
@@ -110,12 +110,13 @@ def users(self) -> List["WorkspaceUserModel"]:
...
         """
+        # TODO(@alvarobartt): Maybe we should return a list of rg.User instead.
         return workspaces_api.list_workspace_users(self.__client, self.id).parsed
 
-    def __str__(self) -> str:
+    def __repr__(self) -> str:
         return (
-            f"Workspace(id={self.id}, name={self.name}, users={self.users},"
-            f" insterted_at={self.inserted_at}, updated_at={self.updated_at}"
+            f"Workspace(id={self.id}, name={self.name},"
```
alvarobartt, line 118: Why did we decide to remove the users from the `print`-statement? Should we keep it there for more visibility?

alvarobartt (approved):
LGTM, just some minor nits

src/argilla/client/workspaces.py
frascuchon: For the sake of readability, I think is better to exclude them. Otherwise, the info can be hard to read. Also, we have the `workspace.users` property which will show the users correctly.

tests/client/test_workspaces.py
frascuchon: Not sure about this change. If we're coding a test for the client, we should use the way we're using normally the client, so `import argilla as rg; rg.....`

tests/client/test_users.py
frascuchon: same here

tests/client/test_workspaces.py
alvarobartt: I guess we should try to avoid to specifically `import argilla as rg` inside our code base, as that's just for the users and examples/tutorials, internally we should make everything more explicit to have some traceability IMO

tests/client/test_users.py
frascuchon: In that case, we should add a test to validate that the main Argilla functions are accessible from the main client module. Imagine a new function `rg.pentola_me` mapping to an internal `argilla.client.init.Argilla.pentola_me` method. If we use the direct method, we never know if the exposed `rg.pentola_me` function is really working.

PR #3289, opened 2023-06-28
feat: warn user when trying to load `FeedbackDataset` using `rg.load` 

### LINKED ISSUE #3175: [BUG] Issue loading FeedbackDataset
**Describe the bug**
After creating a feedback data set when I try to use rg.load(name = feedbackDatasetName) I get an error. 

```python

questions = [
    rg.RatingQuestion(
        name="response_ranking",
        title="Rank the responses\n1: first response is better,\n 2: second response is better,\n 3: both are equal ",
        required=True,
        values=[1, 2, 3]
    ),
    rg.TextQuestion(
        name="correct_response",
        title="If none of the responses are helpful and correct, provide the response",
        required=False
    ),
]
fields = [
    rg.TextField(name="response-1", required=True),
    rg.TextField(name="response-2", required=True)
]

dataset = rg.FeedbackDataset(
    guidelines="Please, read the prompt carefully and...",
    questions=questions,
    fields=fields
)

record = rg.FeedbackRecord(fields={"prompt": "test", "response-1": "response 1", "response-2": "response_2"})
records.append(record)

# Add records to the dataset
dataset.add_records(records)
# This publishes the dataset and pushes the records into Argilla
dataset.push_to_argilla(name="feedback_testing ")

# load predictions for review
dataset_rg = rg.load(
    name="feedback_testing"
)

```
Error is: 

NotFoundApiError: Argilla server returned an error with http status: 404
Error details: [{'code': 'argilla.api.errors::EntityNotFoundError', 'params': 
{'name': 'safety_testing', 'type': 'ServiceDataset'}}]

**Expected behavior**
I expect to read a dataset like I do any other dataset 

**Environment:**
 - Argilla Version 1.9.0
 - Argilla server 1.9.0

gabrielmbmb:
# Description

A few users have already reported an "error" (one on GitHub and a few others on Slack) when trying to load a `FeedbackDataset` using `rg.load`. `rg.load` cannot be used to load `FeedbackDataset`s and `rg.FeedbackDataset.from_argilla` has to be used instead.

This PR updated the `rg.load` function to try to check if there's a `FeedbackDataset` with the provided `name` and `workspace` if `NotFoundApiError` is raised when trying to get the V0 dataset. If the `FeedbackDataset` is found, then it will raise an error telling the user how to load the `FeedbackDataset`. Otherwise, the original exception will be re-raised.

Closes #3175

**Type of change**

**How Has This Been Tested**

Some manual tests and I've added a new unit test covering this case.

**Checklist**

### DIFF
--- a/src/argilla/client/api.py
+++ b/src/argilla/client/api.py
@@ -24,6 +24,9 @@ from argilla.client.models import (  # TODO Remove TextGenerationRecord
     BulkResponse,
     Record,
 )
+from argilla.client.sdk.commons import errors
+from argilla.client.sdk.v1.datasets.api import list_datasets as list_datasets_api_v1
+from argilla.client.sdk.workspaces.api import list_workspaces as list_workspaces_api_v0
 
 _LOGGER = logging.getLogger(__name__)
 
@@ -235,6 +238,22 @@ async def log_async(
     return await asyncio.wrap_future(future)
 
 
+def _check_if_feedback_dataset_exists(client: Argilla, name: str, workspace: str) -> bool:
+    response = list_workspaces_api_v0(client.http_client.httpx)
+    workspace_id = None
+    for ws in response.parsed:
+        if ws.name == workspace:
+            workspace_id = ws.id
+            break
+
+    response = list_datasets_api_v1(client.http_client.httpx)
+    for dataset in response.parsed:
+        if dataset.name == name and dataset.workspace_id == workspace_id:
+            return True
+
+    return False
+
+
 def load(
     name: str,
     workspace: Optional[str] = None,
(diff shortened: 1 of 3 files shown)


src/argilla/client/api.py
```diff
@@ -235,6 +238,22 @@ async def log_async(
...
+        if ws.name == workspace:
+            workspace_id = ws.id
+            break
+
+    response = list_datasets_api_v1(client.http_client.httpx)
+    for dataset in response.parsed:
+        if dataset.name == name and dataset.workspace_id == workspace_id:
+            return True
+
+    return False
```
gabrielmbmb, line 254: There is a similar function `argilla.client.feedback.utils.feedback_dataset_in_argilla` but it was impossible to use it in this module because it generates cyclic imports errors. I tried to fix them, but that would have required moving a lot of stuff... so I decided that it's not worth it and created this function instead.

src/argilla/client/api.py
```diff
@@ -295,20 +314,30 @@ def load(
...
+            include_vectors=include_vectors,
+            as_pandas=as_pandas,
+        )
+    except errors.NotFoundApiError as e:
+        workspace = workspace or argilla.get_workspace()
+        if _check_if_feedback_dataset_exists(client=argilla, name=name, workspace=workspace):
+            raise ValueError(
+                f"The dataset '{name}' exists but it is a `FeedbackDataset`. Use `rg.FeedbackDataset.from_argilla`"
+                " instead to load it."
+            )
```
frascuchon, line 339: I'm not sure if a warning  + return `None` would be better than an exception only for this.

gabrielmbmb: I think it is better to stop the execution of the program as soon as possible. Maybe some users decide to suppress the `argilla` log messages and do not see the warning message, and get a `None` instead and make them even more confused later when their program starts raising `AttributeError: 'NoneType' object has no attribute ...`

alvarobartt: Agree with @gabrielmbmb here, it's better to just raise the exception straight away

alvarobartt (approved):

frascuchon (approved):

Issue #3291, opened 2023-06-28
[REFACTOR] Split unit and integration tests and use different CI/CD conditions


alvarobartt:
**Describe the problem**

I think we should carefully split and revisit the whole test suite to split the unit tests with the integration tests, and be less strict for the integration ones, otherwise the CI/CD always (or mostly always) ends up failing due to HTTP errors, or integration related issues.

This way the unit and integration tests would run faster, and we would have more control, rather than waiting up to 25 minutes for every CI/CD run to report the errors.

Open for suggestions, but WDYT @frascuchon @gabrielmbmb? I'm not saying we need to tackle this right now, but ideally we should start cleaning the test-suite a bit

PR #3300, opened 2023-06-30
ci: Optimize build + test pipeline

frascuchon:
# Description

This PR reviews and organizes the CI pipeline. With the changes introduced here, all tests are running as expected.  

Also, the cache action has been updated in order to fix a cache resolution problem which allows to reuse the cache for the rest of the jobs.

## Two blocks of tests are launched separately 

**base tests** (most of the server tests and some basic client tests) and **extra tests** (training, monitor, listeners...). The pipeline falls into errors if resource all tests are launched in the same job. 
In the next step, those groups should be reviewed and improved.  

Refs #3291 

## Reusable jobs extraction

The job running the tests has been extracted as a separate workflow, which helps with reusability and workflow code maintenance. The same has been applied to the build Python package.

## Improve conditional executions

The `IS_DEPLOYABLE` step conditions have been removed. Now, this condition is executed just once at the related job level.


.github/workflows/build-python-package.yml
```diff
@@ -0,0 +1,33 @@
...
+      - name: Cache pip 👜
+        uses: actions/cache@v3
+        env:
+          # Increase this value to reset cache if pyproject.toml has not changed
+          CACHE_NUMBER: 0
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-pip-${{ env.CACHE_NUMBER }}-${{ hashFiles('pyproject.toml') }}
+      - name: Setup Node.js
+        uses: actions/setup-node@v2
```
gabrielmbmb: should we update this to `actions/setup-node@v3`?

.github/workflows/build-python-package.yml
```diff
@@ -0,0 +1,33 @@
...
+        with:
+          node-version: "14"
+      - name: Build Package 🍟
+        run: |
+          pip install -U build
+          scripts/build_distribution.sh
+      - name: Upload package artifact
+        uses: actions/upload-artifact@v2
```
gabrielmbmb: should we update this to `actions/upload-artifact@v3`?

.github/workflows/run-python-tests.yml
```diff
@@ -0,0 +1,92 @@
...
+      - name: Setup ElasticSearch 🔎
+        if: inputs.searchEngine == 'elasticsearch'
+        uses: ankane/setup-elasticsearch@v1
+        with:
+          elasticsearch-version: ${{ inputs.searchEngineVersion }}
+      - name: Setup OpenSearch 🔎
+        if: inputs.searchEngine == 'opensearch'
+        uses: ankane/setup-opensearch@v1
+        with:
+          opensearch-version: ${{ inputs.searchEngineVersion }}
```
gabrielmbmb, line 82: maybe we could replace these actions with this https://docs.github.com/en/actions/using-containerized-services/about-service-containers#creating-service-containers?

gabrielmbmb (approved):
LGTM!

.github/workflows/run-python-tests.yml
frascuchon: Yes, for sure. I will tackle this in a separate PR. Thanks!

.github/workflows/build-python-package.yml
frascuchon: Done

.github/workflows/build-python-package.yml
frascuchon: Done

.github/workflows/run-python-tests.yml
```diff
@@ -0,0 +1,92 @@
...
+          pip install -e ".[server,listeners]"
+          alembic upgrade head
+          pytest --cov=argilla --cov-report=xml ${{ inputs.pytestArgs }}
+      - name: Upload Coverage to Codecov 📦
+        uses: codecov/codecov-action@v1
+        if: inputs.coverageReport
+        with:
+          files: ./coverage.xml
+          flags: pytest
+          fail_ci_if_error: false
```
gabrielmbmb: I think that we should remove this from here, and have an step in `package.yml` that [combines](https://coverage.readthedocs.io/en/stable/cmd.html#combining-data-files-coverage-combine) the `coverage` files from both jobs (`run_tests` and `run_tests_extra`) and then upload the combined file to codecov.

pyproject.toml
```diff
@@ -131,7 +131,7 @@ log_format = "%(asctime)s %(name)s %(levelname)s %(message)s"
 log_date_format = "%Y-%m-%d %H:%M:%S"
 log_cli = "True"
 testpaths = ["tests"]
-addopts = ["--ignore", "tests/training"]
```
tomaarsen, line 134: Nice, it seems good to run these tests locally too.

tomaarsen (approved):

.github/workflows/run-python-tests.yml
frascuchon: I finally could work on this with success. Now all coverage reports are combined and published automatically. So final coverage increased substantially

PR #3306, opened 2023-07-03
refactor: simplify old bulk endpoints to avoid create datasets if does not exists

### LINKED ISSUE #3244: Simplify old bulk endpoints to avoid create datasets if does not exists
Change old bulk API endpoints to avoid creating the dataset if does not exist. This will break old workflows, so keep in mind this t reflect in the changelog.

Refs #3162

frascuchon:
# Description

This PR changes the `POST /api/datasets/:dataset-id/:task/bulk` endpoints behavior to not create the dataset if does not exist. The dataset must be created first using the `POST /api/datasets` endpoint.

Since version 1.11, the Python client is already adapted, but the `rg.log` may fail for older clients.

Closes #3244

**Type of change**

**Checklist**

### DIFF
--- a/src/argilla/server/apis/v0/handlers/text2text.py
+++ b/src/argilla/server/apis/v0/handlers/text2text.py
@@ -73,32 +73,17 @@ def configure_router():
     ) -> BulkResponse:
         task = task_type
         workspace = common_params.workspace
-        try:
-            dataset = await datasets.find_by_name(
-                current_user,
-                name=name,
-                task=task,
-                workspace=workspace,
-            )
-            await datasets.update(
-                user=current_user,
-                dataset=dataset,
-                tags=bulk.tags,
-                metadata=bulk.metadata,
-            )
-        except EntityNotFoundError:
-            dataset = CreateDatasetRequest(name=name, workspace=workspace, task=task, **bulk.dict())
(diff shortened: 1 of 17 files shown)


src/argilla/server/apis/v0/handlers/text2text.py
```diff
@@ -73,32 +73,17 @@ async def bulk_records(
...
-                tags=bulk.tags,
-                metadata=bulk.metadata,
-            )
-        except EntityNotFoundError:
-            dataset = CreateDatasetRequest(name=name, workspace=workspace, task=task, **bulk.dict())
-            dataset = await datasets.create_dataset(user=current_user, dataset=dataset)
+
+        dataset = await datasets.find_by_name(current_user, name=name, task=task, workspace=workspace)
+
+        await datasets.update(user=current_user, dataset=dataset, tags=bulk.tags, metadata=bulk.metadata)
```
gabrielmbmb, line 79: maybe we should move updating the dataset also out of the bulk endpoints and call `PATCH /api/datasets/{name}` one time in `rg.log` before adding the records

gabrielmbmb (approved):

src/argilla/server/apis/v0/handlers/text2text.py
frascuchon: We can tackle this in a separate PR.

PR #3307, opened 2023-07-03
fix: unit test failing if not local db

gabrielmbmb:
# Description

The API startup event was using the `argilla.server.database.get_async_db` function to get a DB session to check if the default argilla user exists. We were overriding the `get_async_db` dependency for the API endpoints, but we were not mocking the function when being used from the API startup, and therefore the function was returning a session to the local sqlite DB (`$HOME/.argilla/argilla.db`). If the sqlite local db file didn't exist and the migrations hadn't been applied, then a `sqlalchemy.errors. OperationalError` was being raised as the `User` table didn't exist.

**Type of change**

- [ ] Documentation update

**How Has This Been Tested**

Running the tests without the `$HOME/.argilla/argilla.db` file doesn't raise an exception anymore.

**Checklist**


src/argilla/server/server.py
```diff
@@ -212,9 +212,10 @@ async def check_telemetry():
             print(message, flush=True)
 
 
-def configure_database(app: FastAPI):
-    get_db_wrapper = contextlib.asynccontextmanager(get_async_db)
+_get_db_wrapper = contextlib.asynccontextmanager(get_async_db)
```
gabrielmbmb, line 215: This has to be here, otherwise it cannot be mocked. I tried to mock `get_async_db`, but I couldn't mock it directly using `argilla.server.server.get_async_db`. I was able to mock it using `argilla.server.database.get_async_db` but that caused the rest of the unit tests to fail :)

frascuchon: Make sense to move this function to `argilla.server.database` and use it here?

frascuchon (commented):
Once we have this working, we should remove alembic migrations in CI pipeline: [Here](https://github.com/argilla-io/argilla/blob/fix/unit_test_failing_if_not_local_db/.github/workflows/package.yml#L30) and [here](https://github.com/argilla-io/argilla/blob/fix/unit_test_failing_if_not_local_db/.github/workflows/package.yml#L98)

src/argilla/server/server.py
gabrielmbmb: right now is only being used here and I think it won't be needed in any other part of the code. I would keep it here to avoid confusions with `get_async_db` function.

frascuchon (approved):

PR #3328, opened 2023-07-04
feat: warn users when using integer record ids

### LINKED ISSUE #3265: [BUG] Validating multi-label text classification records makes new copies of them instead of updating
**Describe the bug**
In multi-label text classification, whenever I annotate records, they get copied to a validated version, but the original record stays the same. So I have a dataset with 1000 records, and after annotating 20 of them, I now have 1020 records in my dataset, the 1000 records with status "default" and the 20 annotated records. Is this expected behaviour? Because like this I can't filter for records I haven't annotated yet, when I filter for status:default I still get all 1000 records.

I run argilla and elasticsearch with the supplied docker-compose file without any modifications.

**To Reproduce**
```python
DATASET_NAME = '...'
labels = [...]
settings = rg.TextClassificationSettings(label_schema=labels)
rg.configure_dataset_settings(name=DATASET_NAME, settings=settings)

# sample is a pandas dataframe containing 1000 records
records = sample.apply(lambda x: rg.TextClassificationRecord(
    inputs={"user": x.username, "body": x.rawContent},
    metadata={"hashtags": x.hashtags},
    id=x.id,
    multi_label=True
), axis=1).values.tolist()

rg.log(records, DATASET_NAME)
```

**Expected behavior**
I expect the original records to change status, so after annotating 20 records I should still have 1000 records in my dataset, now 980 with status "default" and 20 with status "validated"

 - OS: macOS 11.7.5
 - Browser: Firefox
 - Argilla Version: 1.10.0
 - Docker Image: docker-compose 3.9 (argilla-server:latest, elasticsearch:8.5.3)

Edit: Update on some more behaviour:
- when I query for a word that results in more than 1 result, I get the unlabeled and the validated records
- when I query for a word/combination of words that exist(s) only in 1 record, I also get both records (via .load and in the ui), but in the ui I can only see the validated record, but there is blank space where the unlabeled record would be (there is no element for it in the dom though, the validated record is offset via `translateY`)

frascuchon:
# Description

This PR deprecates the integer support for record id. Also, warn if big integers are used as record id

Refs #3265

**Type of change**

**How Has This Been Tested**

After the changes the, issue cannot be reproduced anymore. Tested in dev environments

**Checklist**


src/argilla/server/daos/models/records.py
```diff
@@ -56,7 +56,7 @@ class BaseEmbeddingVectorDB(BaseModel):
 
 
 class BaseRecordInDB(GenericModel, Generic[AnnotationDB]):
```
davidberenstein1957, line 59: Don't we want to validate, warn, and potentially convert an integer to string?

tests/server/functional_tests/datasets/test_get_record.py
```diff
@@ -112,7 +112,7 @@ def test_get_record_by_id(mocked_client, task, expected_record_class):
 
     assert response.status_code == 200
     record = expected_record_class.parse_obj(response.json())
-    assert record.id == record_id
+    assert record.id == str(record_id)
```
davidberenstein1957: We do convert here. Is it not cleaner to do this in the record definition?

davidberenstein1957 (approved):

tests/server/functional_tests/datasets/test_get_record.py
frascuchon: Here in the test, the provided record id is defined as an integer. I kept it like this just to validate that the integer valuer is still supported, even if the found record has the string representation of the id.

src/argilla/server/daos/models/records.py
frascuchon: Pydantic already casts integers to strings for free. What you mean is to include a custom validator to warn this, right? It can be helpful, but may also fill the server logs with this kind of warnings

src/argilla/client/models.py
```diff
@@ -107,7 +116,25 @@ def _none_to_empty_dict(cls, v):
...
+                f"Integer ids won't be supported in future versions. We recommend start using strings instead. "
+                "For dataset already containing integer values, you can take a look into the docs section "
+                "to migrate them. See https://docs.argilla.io/en/latest/getting_started/installation/configurations"
+                "/database_migrations.html#elasticsearch"
+            )
+            warnings.warn(message, DeprecationWarning)
+            # See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER
+            if v > _JS_MAX_SAFE_INTEGER:
+                message = (
+                    "You've provided a big integer value. Use a string instead, otherwise may experiment with some "
```
davidberenstein1957:
```suggestion
                    "You've provided a big integer value. Use a string instead, otherwise you may experience some "
```

src/argilla/client/models.py
```diff
@@ -107,7 +116,25 @@ def _none_to_empty_dict(cls, v):
         return v
 
     @validator("id", check_fields=False, always=True)
-    def _none_to_generated_uid64(cls, v):
+    def _normalize_id(cls, v):
+        if isinstance(v, int):
+            message = (
+                f"Integer ids won't be supported in future versions. We recommend start using strings instead. "
```
davidberenstein1957:
```suggestion
                f"Integer ids won't be supported in future versions. We recommend to start using strings instead. "
```

src/argilla/client/models.py
```diff
@@ -107,7 +116,25 @@ def _none_to_empty_dict(cls, v):
...
+                "For dataset already containing integer values, you can take a look into the docs section "
```
davidberenstein1957:
```suggestion
                "For datasets already containing integer values we recommend migrating them to avoid deprecation issues."
```

src/argilla/client/models.py
```diff
@@ -107,7 +116,25 @@ def _none_to_empty_dict(cls, v):
...
+                "to migrate them. See https://docs.argilla.io/en/latest/getting_started/installation/configurations"
```
davidberenstein1957:
```suggestion
                "See https://docs.argilla.io/en/latest/getting_started/installation/configurations"
```

src/argilla/server/daos/models/records.py
```diff
@@ -115,11 +115,28 @@ def prepare_record_for_db(cls, values):
...
     @validator("id", always=True, pre=True)
-    def default_id_if_none_provided(cls, id: Optional[str]) -> str:
-        """Validates id info and sets a random uuid if not provided"""
-        if id is None:
-            return str(uuid4())
-        return id
+    def _normalize_id(cls, v):
+        if isinstance(v, int):
+            message = (
+                f"Integer ids won't be supported in future versions. We recommend start using strings instead. "
```
davidberenstein1957:
```suggestion
                f"Integer ids won't be supported in future versions. We recommend to start using strings instead. "
```

src/argilla/server/daos/models/records.py
```diff
@@ -115,11 +115,28 @@ def prepare_record_for_db(cls, values):
...
+                "For dataset already containing integer values, you can take a look into the docs section "
+                "to migrate them. See https://docs.argilla.io/en/latest/getting_started/installation/configurations"
```
davidberenstein1957:
```suggestion
                "See https://docs.argilla.io/en/latest/getting_started/installation/configurations"
```

src/argilla/server/daos/models/records.py
```diff
@@ -115,11 +115,28 @@ def prepare_record_for_db(cls, values):
...
-    def default_id_if_none_provided(cls, id: Optional[str]) -> str:
-        """Validates id info and sets a random uuid if not provided"""
-        if id is None:
-            return str(uuid4())
-        return id
+    def _normalize_id(cls, v):
+        if isinstance(v, int):
+            message = (
+                f"Integer ids won't be supported in future versions. We recommend start using strings instead. "
+                "For dataset already containing integer values, you can take a look into the docs section "
```
davidberenstein1957:
```suggestion
                "For datasets already containing integer values we recommend migrating them to avoid deprecation issues."
```

src/argilla/server/daos/models/records.py
```diff
@@ -115,11 +115,28 @@ def prepare_record_for_db(cls, values):
...
+                "to migrate them. See https://docs.argilla.io/en/latest/getting_started/installation/configurations"
+                "/database_migrations.html#elasticsearch"
+            )
+            warnings.warn(message, DeprecationWarning)
+            # See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER
+            if v > _JS_MAX_SAFE_INTEGER:
+                message = (
+                    "You've provided a big integer value. Use a string instead, otherwise may experiment with some "
```
davidberenstein1957:
```suggestion
                    "You've provided a big integer value. Use a string instead, otherwise you may experience some "
```

tests/client/test_models.py
```diff
@@ -292,3 +292,37 @@ def test_record_validation_on_assignment(record):
...
+
+def test_big_integer_record_id():
+    import random
+
+    record_id = random.getrandbits(64)
+    with pytest.warns(UserWarning) as warn_record:
+        record = TextClassificationRecord(text="This is a text", id=record_id)
+        assert record.id == record_id
+        assert (
+            "You've provided a big integer value. Use a string instead, otherwise may experiment with some "
```
davidberenstein1957:
```suggestion
            "You've provided a big integer value. Use a string instead, otherwise you may experience some "
```

tests/client/test_models.py
```diff
@@ -292,3 +292,37 @@ def test_record_validation_on_assignment(record):
...
+
+
+def test_cast_record_id():
+    record_id = 1000
+
+    with pytest.warns(DeprecationWarning) as warn_record:
+        record = TextClassificationRecord(text="This is a text", id=record_id)
+        assert record.id == record_id
+        assert (
+            "Integer ids won't be supported in future versions. We recommend start using "
```
davidberenstein1957:
```suggestion
            "Integer ids won't be supported in future versions. We recommend to start using strings instead. "
```

tests/client/test_models.py
```diff
@@ -292,3 +292,37 @@ def test_record_validation_on_assignment(record):
...
+            "strings instead. For dataset already containing integer values, you can take "
```
davidberenstein1957:
```suggestion
            "For datasets already containing integer values we recommend migrating them to avoid deprecation issues."
```

tests/client/test_models.py
```diff
@@ -292,3 +292,37 @@ def test_record_validation_on_assignment(record):
...
+            "a look into the docs section to migrate them. See "
```
davidberenstein1957:
```suggestion
            "See https://docs.argilla.io/en/latest/getting_started/installation/configurations"
```

tests/client/test_models.py
```diff
@@ -292,3 +292,37 @@ def test_record_validation_on_assignment(record):
...
+            "https://docs.argilla.io/en/latest/getting_started/installation/configurations/database_migrations.html"
+            "#elasticsearch"
```
davidberenstein1957:
```suggestion
            "/database_migrations.html#elasticsearch"
```

davidberenstein1957 (approved):

src/argilla/client/models.py
```diff
@@ -106,10 +115,29 @@ def _none_to_empty_dict(cls, v):
...
+            # See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER
+            if v > _JS_MAX_SAFE_INTEGER:
+                message = (
+                    "You've provided a big integer value. Use a string instead, otherwise you may experience some "
+                    "problems using the UI. See "
+                    "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number"
+                    "/MAX_SAFE_INTEGER"
+                )
+                warnings.warn(message, UserWarning)
+        elif not isinstance(v, str):
```
frascuchon, line 139: This manual type-condition is needed. I tried using `conint` and `constr` in the field definition but I found some weird behavior and unexpected int->str automatic conversions.

PR #3308, opened 2023-07-03
feat: add list user workspaces endpoint

gabrielmbmb:
# Description

This PR adds a new endpoint to API v1 to list the workspaces to which a user belongs.

Closes #3273

**Type of change**

- [x] New feature (non-breaking change which adds functionality)

**How Has This Been Tested**

Manually and unit tests have been added to test this new endpoint.

**Checklist**

- [ ] I added relevant documentation
- [x] follows the style guidelines of this project
- [x] I did a self-review of my code
- [ ] I made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I filled out [the contributor form](https://tally.so/r/n9XrxK) (see text above)
- [x] I have added relevant notes to the CHANGELOG.md file (See https://keepachangelog.com/)

PR #3334, opened 2023-07-04
feat: add `list_user_workspaces` and `User.workspaces` property

alvarobartt:
# Description

This PR adds the `list_user_workspaces` function in the SDK and the Python client, as of the recently included API endpoint at https://github.com/argilla-io/argilla/pull/3308, so this PR needs to wait upon its parent one to be merged first.

This PR then, solves the issue we had when listing workspaces from an user, since we were listing all the workspaces, and just available for certain users.

**Type of change**

**How Has This Been Tested**

- [X] Add unit tests for both the SDK and the Python client

**Checklist**


src/argilla/client/users.py
```diff
@@ -68,7 +70,7 @@ class User:
     last_name: Optional[str]
     full_name: Optional[str]
     role: UserRole
-    workspaces: Optional[List[str]]
+    workspaces: Optional[List[WorkspaceModel]]
```
gabrielmbmb: This should be removed, right?

gabrielmbmb (approved):

src/argilla/client/users.py
alvarobartt: I guess we can keep it as an attribute even though it's a property, but we can also remove it, do you think it's better to remove it?

PR #3336, opened 2023-07-04
chore: added telemetry to ArgillaTrainer

### LINKED ISSUE #3325: [FEATURE] add `telemetry` to the `ArgillaTrainer`
We don't have direct insight into the usage of the `ArgillaTrainer` and frameworks.

N.A.

N.A.

**Additional context**
N.A.

davidberenstein1957:
# Description

- moved `argilla.server.commons.telemetry` to `argilla.utils.telemetry`
- added `telemetry` to `ArgillaTrainer` for Framework and NLP-task

Closes #3325 

**Type of change**

**How Has This Been Tested**

N.A.

**Checklist**

N.A.


alvarobartt:
Before tackling this I guess we should talk about whether adding telemetry in the Python client side is something we want to do @frascuchon, as for now we just have that in the API

davidberenstein1957:
I briefly discussed this with @frascuchon and @dvsrepo before. For me it makes sense to do so in order to keep track of integrations like the trainer and the callbacks/plugins.

davidberenstein1957:
@frascuchon I opted for moving telemetry to `argilla.utils` instead of creating another `commons` directory just for telemetry.

src/argilla/training/base.py
```diff
@@ -234,6 +236,12 @@ def __init__(
             )
 
         self._logger.info(self)
+        self._track_trainer_usage(framework=framework, task=self._rg_dataset_type._RECORD_TYPE.__name__)
+
+    def _track_trainer_usage(self, framework: str, task: str):
+        if not self._CLIENT:
```
frascuchon: I think this is not necessary since the telemetry client will be always initialized: See https://github.com/argilla-io/argilla/blob/6fb45b3a70ccd511b5e4b8305af212efb6cc6df3/src/argilla/utils/telemetry.py#L89C1-L89C8

src/argilla/training/base.py
```diff
@@ -234,6 +236,12 @@ def __init__(
...
+            return
+        self._CLIENT.track_data(action="ArgillaTrainerUsage", data={"framework": framework, "task": task})
```
frascuchon, line 242: We should update the telemetry [section in the docs](https://docs.argilla.io/en/latest/reference/telemetry.html#telemetry) to include this new tracking.

frascuchon (approved):
Cool!!

I've included some comments 

Also, maybe the changes referring to the [training import issue](https://github.com/argilla-io/argilla/pull/3346) should be  removed, or wait to merge first #3346 ?

davidberenstein1957:
@frascuchon I assumed #3346 would be resolved during the patch.

PR #3337, opened 2023-07-04
feat: delete record endpoint

### LINKED ISSUE #3310: [FEATURE] Add API v1 endpoints to remove records
Add two endpoints that allows to remove records from a `FeedbackDataset` in the API v1:

- `DELETE /api/v1/records/{record_id}` to remove one record
- `POST /api/v1/datasets/{dataset_id}/records/bulk-delete` to remove several records.

gabrielmbmb:
# Description

This PR adds one new endpoint for deleting records:

- `DELETE /api/v1/records/{record_id}`: for removing a specific record given its ID.

It's been needed to add a new method `delete_records` to the `SearchEngine` that will remove the records also from the index created for the dataset in Elastic Search.

Closes #3310

**Type of change**

**How Has This Been Tested**

Manually in a local environment and I've added unit tests.

**Checklist**


src/argilla/server/apis/v1/handlers/datasets.py
```diff
@@ -300,6 +301,25 @@ async def create_dataset_records(
...
+):
+    dataset = await _get_dataset(db, dataset_id)
+
+    await authorize(current_user, DatasetPolicyV1.bulk_delete_records(dataset))
+
+    deleted_records_ids = await datasets.bulk_delete_records(db, dataset=dataset, records_ids=bulk_delete.ids)
+
+    return BulkDeleteResponse(
+        deleted=deleted_records_ids, not_deleted=list(set(bulk_delete.ids) - set(deleted_records_ids))
+    )
```
gabrielmbmb: I though that it might be useful to return which records were deleted and which not (the user can provide IDs that don't exist), but maybe we can return an empty response and a 204.

frascuchon: For me it depends. Why some of the records have not been deleted, Did any error occur? In that case, I would say to return a 400 or 422 error with failed record ids. If everything goes well, I think the 204 is enough as a response.

src/argilla/server/policies.py
```diff
@@ -307,6 +316,20 @@ async def is_allowed(actor: User) -> bool:
...
+            return actor.is_owner or (
+                actor.is_admin
+                and await _exists_workspace_user_by_user_and_workspace_id(actor, record.dataset.workspace_id)
+            )
+
+        return is_allowed
+
+    @classmethod
+    def bulk_delete(cls, actor: User) -> bool:
+        pass
```
frascuchon, line 331: I think this stayed here after some refactoring...
```suggestion
```

frascuchon: I think this stayed here after some refactoring...
```suggestion
```

src/argilla/server/schemas/v1/common.py
```diff
@@ -0,0 +1,27 @@
...
+from uuid import UUID
+
+from pydantic import BaseModel
+
+
+class BulkDelete(BaseModel):
+    ids: List[UUID]
+
+
+class BulkDeleteResponse(BaseModel):
```
frascuchon: Related to my previous comment. The happy path for this endpoint can be a 204 response. In that case, this schema is not needed anymore.

src/argilla/server/schemas/v1/common.py
frascuchon, line 1: I prefer to create this common module once we have more endpoints sharing schemas. Since the only endpoint using this is the bulk-delete, I would say to move the schemas to the `v1.schemas.datasets` module

tests/server/api/v1/test_records.py
```diff
@@ -905,3 +905,46 @@ async def test_create_record_suggestion_as_annotator(client: TestClient):
     )
 
     assert response.status_code == 403
+
+
+@pytest.mark.parametrize("role", [UserRole.owner, UserRole.admin])
+@pytest.mark.asyncio
+async def test_delete_record(client: TestClient, db: "AsyncSession", role: UserRole):
```
frascuchon: maybe adding a test covering the deletion of a non-existing record?

frascuchon (approved):
I think makes sense to return in the future a 422 when `bulk_delete` with non-existing records.

alvarobartt:
> I think makes sense to return in the future a 422 when `bulk_delete` with non-existing records.

Shouldn't we return 404 in that case? As we want to state that the introduced record IDs were not found, right?

src/argilla/server/contexts/datasets.py
```diff
@@ -217,6 +217,10 @@ async def get_record_by_id(
     return result.scalar_one_or_none()
 
 
+async def delete_record(db: "AsyncSession", record: Record) -> Record:
+    return await record.delete()
```
frascuchon: We should remove also the indexed document from the elasticsearch index.

frascuchon (approved):
💯

PR #3357, opened 2023-07-06
ci: skip running tests when no python changes

frascuchon:
This PR fixes conditions to skip running tests when no python changes were detected.

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -37,7 +37,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: check_repo_files
     env:
-      IS_DEPLOYABLE: ${{ secrets.AR_DOCKER_USERNAME != '' && needs.check_repo_files.outputs.buildChanges }}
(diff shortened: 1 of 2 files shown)


tomaarsen:
Seems to have worked. We can test the other case by pushing some arbitrary change in Python, but I'm fine with merging this and figuring out whether the `pythonChanges == 'true'` case works on the fly.

GitHub actions also knows about booleans ([docs](https://docs.github.com/en/actions/learn-github-actions/expressions#literals)), but I reckon that `pythonChanges` is not a boolean, as it was written as a (string) environment variable.

tomaarsen (approved):

frascuchon:
Thanks @tomaarsen.

I also checked with forcing some changes in python (see [this execution](https://github.com/argilla-io/argilla/actions/runs/5483834101)).

Regarding the boolean condition, I think is not a new topic. I found some comments and what people do is parsing using `fromJSON` https://github.com/actions/runner/issues/1483

PR #3366, opened 2023-07-09
fix: `ArgillaDatasetCard` to handle all question `options`

### LINKED ISSUE #3365: [BUG] `ArgillaDatasetCard` is not handling the possible values of question responses
**Describe the bug**

Currently, when pushing a `FeedbackDataset` to the HuggingFace Hub via `push_to_huggingface` with the default args (assuming `generate_card=True`), the `ArgillaDatasetCard` generated is not properly handling the question values/labels for the newly introduced questions i.e. `RankingQuestion`, `LabelQuestion` and `MultiLabelQuestion`.

image

image

**Expected behavior**

The expected should be the same as in the screenshot above, but including the list of possible values/labels of those as for the `RatingQuestion`, instead of displaying `N/A`.

**Environment:**
 - Argilla Version [e.g. 1.0.0]: latest but happening since 1.11.0

alvarobartt:
# Description

This PR addresses the issue mentioned below, to extend the current support from the `argilla_template.md` used by the `ArgillaDatasetCard` class to properly generate the card for all the questions to include their options (aka values/labels).

image

image

Closes #3365

**Type of change**

**How Has This Been Tested**

- [X] Generated `ArgillaDatasetCard` for a dataset including all the question types supported as of 1.13.0 (unreleased)

**Checklist**

### DIFF
--- a/src/argilla/client/feedback/integrations/huggingface/card/argilla_template.md
+++ b/src/argilla/client/feedback/integrations/huggingface/card/argilla_template.md
@@ -75,7 +75,7 @@ The **questions** are the questions that will be asked to the annotators. They c
 
 | Question Name | Title | Type | Required | Description | Values/Labels |
 | ------------- | ----- | ---- | -------- | ----------- | ------------- |
-{% for question in argilla_questions %}| {{ question.name }} | {{ question.title }} | {{ question.__class__.__name__  }} | {{ question.required | default(true, true) }} | {{ question.description | default("N/A", true) }} | {% if question.settings.type == 'rating' %}{{ question.settings.options | map(attribute="value") | list }}{% else %} N/A {% endif %} |
+{% for question in argilla_questions %}| {{ question.name }} | {{ question.title }} | {{ question.__class__.__name__  }} | {{ question.required | default(true, true) }} | {{ question.description | default("N/A", true) }} | {% if question.settings.type in ["rating", "label_selection", "multi_label_selection", "ranking"] %}{{ question.settings.options | map(attribute="value") | list }}{% else %}N/A{% endif %} |
 {% endfor %}
 
 Finally, the **guidelines** are just a plain string that can be used to provide instructions to the annotators. Find those in the [annotation guidelines](#annotation-guidelines) section.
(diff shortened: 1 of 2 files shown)


src/argilla/client/feedback/integrations/huggingface/card/argilla_template.md
```diff
@@ -75,7 +75,7 @@ The **questions** are the questions that will be asked to the annotators. They c
 
 | Question Name | Title | Type | Required | Description | Values/Labels |
 | ------------- | ----- | ---- | -------- | ----------- | ------------- |
-{% for question in argilla_questions %}| {{ question.name }} | {{ question.title }} | {{ question.__class__.__name__  }} | {{ question.required | default(true, true) }} | {{ question.description | default("N/A", true) }} | {% if question.settings.type == 'rating' %}{{ question.settings.options | map(attribute="value") | list }}{% else %} N/A {% endif %} |
+{% for question in argilla_questions %}| {{ question.name }} | {{ question.title }} | {{ question.__class__.__name__  }} | {{ question.required | default(true, true) }} | {{ question.description | default("N/A", true) }} | {% if question.settings.type in ["rating", "label_selection", "multi_label_selection", "ranking"] %}{{ question.settings.options | map(attribute="value") | list }}{% else %}N/A{% endif %} |
```
gabrielmbmb, line 78: I think it would be good to have a constant in the client with all the available kind of questions and use that constant everywhere, so when a new question is added we just need to add a string in one place.

gabrielmbmb (approved):
LGTM! Just a comment, but we can tackle that in a separate PR.

src/argilla/client/feedback/integrations/huggingface/card/argilla_template.md
alvarobartt: Yes I agree, for most of the stuff we're already re-using the pre-defined types and just looping over those to avoid re-writing those, but in this case we're using Jinja2 to generate the card, and the behavior here needs to be manually checked, as not all the questions are the same under this scenario.

Issue #3277, opened 2023-06-27
[FEATURE] Add suggestions endpoints


gabrielmbmb:
**Description**

Add the required API v1 endpoints to manage suggestions.

PR #3370, opened 2023-07-10
feat: add `suggestions` in `FeedbackRecord`

alvarobartt:
# Description

As of the recent addition of the `suggestions` in the Argilla API at https://github.com/argilla-io/argilla/issues/3277, we make sure that's covered in both the SDK and the Python client.

So on, this PR adds the following:

* `suggestions` included in `FeedbackRecord`
* Update `FeedbackRecord` to "listen" for updates (for the moment just applies to `suggestions`, but may eventually be extended to `external_id`, `metadata`, and `responses`)
* Add `add_suggestions` function in the SDK
* Add `include=responses` and `include=suggestions` as default in `get_records`
* `add_question` and `add_field` now return the schema, `QuestionSchema` and `FieldSchema`, respectively (to get the Argilla ID of those, required to log `suggestions` as we use the Argilla Question ID)
* Extend `pydantic.BaseModel`s to add the `id` attribute for anything Argilla-related
* Constrain `pydantic.BaseModel`s to disallow extra fields and have more control over those
* Handle `suggestions` on both pull and push from Argilla
* Handle `suggestions` on both pull and push from HuggingFace

Additionally, I've also included some missing type-hints, improved `question_by_name` and `field_by_name` methods, fix some return type-hints, rename "`FeedbackTask` dataset" occurrences to "`FeedbackDataset`", and some other minor fixes.

Closes https://github.com/argilla-io/Team/issues/115

**Type of change**

**How Has This Been Tested**

- [X] Add unit tests for the newly included `suggestions` attribute in `FeedbackRecord`
- [X] Add unit tests for the `add_suggestions` function in the SDK
- [X] Add unit tests for `FeedbackDataset.records` related functions/methods

**Checklist**


src/argilla/client/feedback/dataset.py
```diff
@@ -311,26 +311,26 @@ def guidelines(self, guidelines: str) -> None:
...
         return self.__fields
 
-    def field_by_name(self, name: str) -> Dict[str, FieldSchema]:
-        for item in self.fields:
-            if item.name == name:
-                return item
-        raise KeyError(f"Field with name '{name}' not found in self.fields")
+    def field_by_name(self, name: str) -> AllowedFieldTypes:
+        try:
+            return next(filter(lambda f: f.name == name, self.fields))
```
gabrielmbmb: I would just use a for loop, it's more readable for everyone I think

alvarobartt:
This PR should be merged right after https://github.com/argilla-io/argilla/pull/3391

src/argilla/client/feedback/integrations/huggingface/dataset.py
```diff
@@ -320,7 +356,7 @@ def from_huggingface(cls: Type["FeedbackDataset"], repo_id: str, *args: Any, **k
                                 "status": status,
                                 "values": {},
                             }
-                        user_without_id_response = True
+                            user_without_id_response = True
```
alvarobartt, line 359:
```suggestion
                        user_without_id_response = True
```

src/argilla/client/feedback/integrations/huggingface/card/argilla_template.md
```diff
@@ -78,6 +78,8 @@ The **questions** are the questions that will be asked to the annotators. They c
 {% for question in argilla_questions %}| {{ question.name }} | {{ question.title }} | {{ question.__class__.__name__  }} | {{ question.required | default(true, true) }} | {{ question.description | default("N/A", true) }} | {% if question.settings.type in ["rating", "label_selection", "multi_label_selection", "ranking"] %}{{ question.settings.options | map(attribute="value") | list }}{% else %}N/A{% endif %} |
 {% endfor %}
 
+**✨ NEW** Additionally, we also have **suggestions**, which are linked to the existing questions, and so on, named appending "-suggestion" and "-suggestion-metadata" to those, containing the value/s of the suggestion and its metadata, respectively. So on, the possible values are the same as in the table above.
+
```
gabrielmbmb, line 82: Maybe we can combine this paragraph with the one that you added below? What do you think?

src/argilla/client/feedback/dataset.py
```diff
@@ -521,32 +543,84 @@ def push_to_argilla(
...
+                        0,
+                        len(updated_records),
+                        PUSHING_BATCH_SIZE,
+                        desc="Updating records in Argilla...",
+                        disable=show_progress,
+                    ):
+                        for record in updated_records[i : i + PUSHING_BATCH_SIZE]:
+                            for suggestion in record.suggestions:
+                                suggestion.question_id = question_name2id[suggestion.question_name]
+                                datasets_api_v1.add_suggestion(
```
frascuchon: since this is adding/updating a suggestion, maybe should be more descriptive calling `set_suggestion` instead of `add_suggestion`

alvarobartt: Since that's the SDK function, should we name it `put_suggestion` instead?

alvarobartt: Also, should we rename the `FeedbackRecord` method from `add_suggestions` to something else?

src/argilla/client/feedback/integrations/huggingface/dataset.py
```diff
@@ -330,7 +366,18 @@ def from_huggingface(cls: Type["FeedbackDataset"], repo_id: str, *args: Any, **k
                     if value is not None:
                         if question.settings["type"] == "ranking":
                             value = [{"rank": r, "value": v} for r, v in zip(value["rank"], value["value"])]
-                        responses[user_id or "user_without_id"]["values"].update({question.name: {"value": value}})
+                        if value is not None:
+                            responses[user_id or "user_without_id"]["values"].update({question.name: {"value": value}})
+
+                if f"{question.name}-suggestion" in hfds[index]:  # Here for backwards compatibility
+                    if hfds[index][f"{question.name}-suggestion"] is not None:
```
gabrielmbmb: maybe these two ifs could be combined with an `and`

src/argilla/client/feedback/schemas.py
```diff
@@ -93,51 +93,165 @@ def user_id_must_have_value(cls, v):
...
-        if not v:
-            return []
-        if isinstance(v, ResponseSchema):
-            return [v]
-        return v
+    _unified_responses: Optional[Dict[str, List["UnifiedValueSchema"]]] = PrivateAttr(default_factory=dict)
+    _updated: bool = PrivateAttr(default=False)
+
+    def __init__(self, **data: Dict[str, Any]) -> None:
+        super().__init__(**data)
```
gabrielmbmb: we can remove this `__init__`, right?

src/argilla/client/feedback/dataset.py
```diff
@@ -568,66 +642,79 @@ def push_to_argilla(
...
+                    records = []
+                    for record in self.records[i : i + PUSHING_BATCH_SIZE]:
+                        if record.suggestions:
+                            for suggestion in record.suggestions:
+                                suggestion.question_id = question_name2id[suggestion.question_name]
+                        records.append(
+                            record.dict(
+                                exclude={"id": ..., "suggestions": {"__all__": {"question_name"}}}, exclude_none=True
+                            )
+                        )
```
frascuchon, line 699: this is duplicated code, maybe we can refactor a bit more to avoid some code duplication

src/argilla/client/feedback/schemas.py
```diff
@@ -93,51 +93,165 @@ def user_id_must_have_value(cls, v):
...
         ...     external_id="entry-1",
         ... )
 
     """
 
+    id: Optional[UUID] = None
     fields: Dict[str, str]
-    metadata: Optional[Dict[str, Any]] = None
-    responses: Optional[Union[ResponseSchema, List[ResponseSchema]]] = None
+    metadata: Optional[Dict[str, Any]] = Field(default_factory=dict)
```
gabrielmbmb: if we have `{}` as default, then we don't need the `Optional` anymore. Just one comment, in the backend the `metadata` can be `None`, but I guess in the client we want to default to `{}` so it's more user friendly, right?

src/argilla/client/feedback/schemas.py
```diff
@@ -93,51 +93,165 @@ def user_id_must_have_value(cls, v):
...
+    responses: Optional[List[ResponseSchema]] = Field(default_factory=list)
```
gabrielmbmb: `Optional` can be removed I guess

src/argilla/client/feedback/schemas.py
```diff
@@ -93,51 +93,165 @@ def user_id_must_have_value(cls, v):
...
+    suggestions: Optional[List[SuggestionSchema]] = Field(default_factory=list, allow_mutation=True)
```
gabrielmbmb: ditto

src/argilla/client/feedback/schemas.py
```diff
@@ -93,51 +93,165 @@ def user_id_must_have_value(cls, v):
...
-        if isinstance(v, ResponseSchema):
-            return [v]
-        return v
+    _unified_responses: Optional[Dict[str, List["UnifiedValueSchema"]]] = PrivateAttr(default_factory=dict)
+    _updated: bool = PrivateAttr(default=False)
+
+    def __init__(self, **data: Dict[str, Any]) -> None:
+        super().__init__(**data)
+
+    def add_suggestions(self, suggestions: Union[SuggestionSchema, List[SuggestionSchema]]) -> None:
```
gabrielmbmb:
```suggestion
    def add_suggestions(self, suggestions: Union[SuggestionSchema, List[SuggestionSchema], Dict[str, Any], List[Dict[str, Any]]]) -> None:
```

src/argilla/client/feedback/dataset.py
frascuchon: For me, it makes sense to call `record.set_suggestion` since will set one single suggestion for a question. For the API call function, I would say not showing implementation details in the name (put_suggestions, or post_suggestions) we can adopt more general naming (set_suggestion, or update_suggestion, or replace_suggestion)

src/argilla/client/feedback/schemas.py
```diff
@@ -93,51 +93,165 @@ def user_id_must_have_value(cls, v):
...
+                " you can simply call `FeedbackDataset.fetch_records()` to fetch them and"
+                " automatically set the `id`, to call `add_suggestions` on top of that."
+            )
+        for suggestion in parsed_suggestions:
+            existing_suggestion = next(
+                filter(
+                    lambda s: (s.question_id == suggestion.question_id)
+                    or (s.question_name == suggestion.question_name),
+                    self.suggestions,
+                ),
```
gabrielmbmb: I would use a list comprehension as I think it's more readable

src/argilla/client/feedback/schemas.py
```diff
@@ -93,51 +93,165 @@ def user_id_must_have_value(cls, v):
...
+                warnings.warn(
+                    "You are trying to update the existing `suggestions` with a new value,"
+                    " which is not allowed. You should use the `add_suggestions` method"
+                    " instead, and the existing `suggestions` will be overwritten based"
+                    " on the provided `question_id`s/`question_name`/s for those `suggestions`."
+                )
+        super().__setattr__(name, value)
+
+    def _reset_updated(self) -> None:
+        self._updated = False
```
gabrielmbmb, line 245: I think we're not using this method anywhere yet, right?

frascuchon (approved):
Just 2 comment related to a minimal code refactoring

src/argilla/client/sdk/v1/datasets/api.py
```diff
@@ -201,12 +201,9 @@ def get_records(
         A `Response` object containing a `parsed` attribute with the parsed response if the
         request was successful, which is an instance of `FeedbackRecordsModel`.
     """
-    url = f"/api/v1/datasets/{id}/records"
+    url = f"/api/v1/datasets/{id}/records?offset={offset}&limit={limit}&include=responses&include=suggestions"
```
gabrielmbmb: why this change?

src/argilla/client/feedback/dataset.py
alvarobartt: Sure then I'll rename that to `set_suggestions` in the client and `set_suggestion` in the SDK, thanks!

tests/client/conftest.py
```diff
@@ -456,6 +456,28 @@ def feedback_dataset_records() -> List[FeedbackRecord]:
                     "status": "submitted",
                 }
             ],
+            suggestions=[
+                {
+                    "question_name": "question-1",
```
gabrielmbmb, line 460: can we add here one suggestion with all `type`, `score` and `agent`?

gabrielmbmb (approved):
LGTM! I left some minor comments

src/argilla/client/sdk/v1/datasets/api.py
alvarobartt: Because otherwise from the Python client we cannot define a Python dict with two keys named equally i.e. `{"include": "responses", "include": "suggestions"}`

src/argilla/client/feedback/schemas.py
alvarobartt: Yes, in `push_to_argilla`

src/argilla/client/feedback/schemas.py
alvarobartt: You mean a for-loop?

frascuchon: just as an idea. could we just define record suggestions as a dictionary where the key is the question name and the value is the suggestion itself?

frascuchon: I also feel a bit hard to read this block of code

src/argilla/client/sdk/v1/datasets/api.py
frascuchon: You can pass a list as value. See https://www.python-httpx.org/quickstart/

```python
>>> params = {'key1': 'value1', 'key2': ['value2', 'value3']}
>>> r = httpx.get('https://httpbin.org/get', params=params)
>>> r.url
URL('https://httpbin.org/get?key1=value1&key2=value2&key2=value3')
```

src/argilla/client/feedback/schemas.py
alvarobartt: Note that for the suggestions we cannot do that as we may also want to add `type`, `score` and `agent`, so I included the `question_name` and the mapping to `question_id` before pushing to Argilla to make it easier for the user

alvarobartt: That block of code is an iterator that checks for each entry that condition until is matched, otherwise returns None, but if you both feel like a for-loop is better I'll revert

src/argilla/client/feedback/schemas.py
alvarobartt: Yes it's just for that and so that anyone can just `record.metadata.update({...: ...})` but we can make it None if you want (also we're excluding the unset values with `exclude_unset` when calling `dict()` so if that's an empty dict we'll just push `None`)

src/argilla/client/feedback/dataset.py
alvarobartt: I'll refactor everything in an upcoming PR to create `ArgillaMixin` to add all those things there instead 👍🏻

src/argilla/client/feedback/integrations/huggingface/card/argilla_template.md
alvarobartt: I think it's better to keep it like this as we're presenting different information on each section, we can confirm it later, but I think it's nice to add the information in the sections as the card recommends.

PR #3381, opened 2023-07-11
fix: `HuggingFaceDatasetMixIn` dependencies and usage

alvarobartt:
# Description

This PR fixes some issues recently found in the `HuggingFaceDatasetMixIn` as described below:

* Records with more than one response with an empty `user_id` i.e. `user_id=None` could be pushed to the HuggingFace Hub, and when pulling it we were assuming that all the users had a valid `user_id`, so now we are just keeping the first `user_id=None` for each question, joining them and discarding the rest
* `HuggingFaceDatasetMixIn` dependencies were being imported in the top of the file, which was leading to conflicts, as `datasets` and `huggingface_hub` are optional dependencies, and as the `HuggingFaceDatasetMixIn` was being used as part of `FeedbackDataset`, an `ImportError` was being automatically raised if `datasets` and/or `huggingface_hub` were not installed, even though those were not used
* `HuggingFaceDatasetMixIn` is now properly used, without defining wrapping methods, and directly using the methods defined in the MixIn itself
* Add `Config` to manage the configuration for the available integrations (for the moment just HuggingFace), and check the dependencies once, not on every function call (similar approach to what 🤗`datasets` does)

**Type of change**

**How Has This Been Tested**

- [X] Add unit tests to catch the warnings raised, and to make sure everything's working as expected

**Checklist**


alvarobartt:
@gabrielmbmb FYI I've not modified the `CHANGELOG.md` as this fix is related to https://github.com/argilla-io/argilla/pull/3326, and included in this release, so there's no need to include this fix into the `CHANGELOG.md` 🤗

src/argilla/client/feedback/integrations/config.py
```diff
@@ -0,0 +1,34 @@
...
+
+from argilla.utils.dependency import require_version
+
+
+def _check_if_installed(dependency: str) -> None:
+    try:
+        require_version(dependency)
+        return True
+    except Exception as e:
+        return False
```
gabrielmbmb: I would move this function to `argilla.utils.dependency` module and avoid using `require_version`, and use `importlib.util.find_spec(dependency) is not None` instead.

Why you didn't use `require_version` function or `requires_version` decorator directly? I would advocate for doing this kind of checks in a standard way across all the package.

src/argilla/client/feedback/integrations/config.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+@dataclass
+class Config:
+    datasets_installed: bool = _check_if_installed("datasets")
+    huggingface_hub_installed: bool = _check_if_installed("datasets")
+
+
+CONFIG = Config()
```
gabrielmbmb: I think this class is not needed

src/argilla/client/feedback/integrations/huggingface/dataset.py
```diff
@@ -281,23 +319,44 @@ def _from_huggingface(cls: Type["FeedbackDataset"], repo_id: str, *args: Any, **
...
+                        if not user_without_id:
+                            user_without_id = True
+                            responses["user_without_id"] = {
+                                "user_id": user_id,
+                                "status": status,
+                                "values": {},
+                            }
+                        user_without_id_response = True
+                    if user_id is not None and user_id not in responses:
+                        if user_id not in responses:
```
gabrielmbmb: the second if condition is not needed as the same expression is already being evaluated in the previous if condition

gabrielmbmb (commented):
Hey @alvarobartt, looks good for me. Just some comments about how it's checked that the needed libs are installed.

Also, when using the `from_huggingface` method if the response don't have an `user_id` we will set `"user_without_id"`, right? I think that if that's the case, then we will have a problem if the user tries to push the dataset to argilla, because the `user_id` won't pass the UUID validation.

src/argilla/client/feedback/integrations/config.py
alvarobartt: Because using the decorators implies that the dependency check is ran every time the function is called, while keeping everything in the config can give us more information on the required dependencies for the existing integrations, and that's just ran once

alvarobartt:
> Hey @alvarobartt, looks good for me. Just some comments about how it's checked that the needed libs are installed.
> 
> Also, when using the `from_huggingface` method if the response don't have an `user_id` we will set `"user_without_id"`, right? I think that if that's the case, then we will have a problem if the user tries to push the dataset to argilla, because the `user_id` won't pass the UUID validation.

Thanks! And no, we're using the keys internally to differentiate users, but we're just keeping the `.values()` of the dict not the keys, so the key will be `user_without_id` but the values `{"user_id": None, ...}`

src/argilla/client/feedback/integrations/huggingface/dataset.py
alvarobartt: True, I'll remove that now!

src/argilla/client/feedback/integrations/config.py
alvarobartt: But yes, I agree that's better to keep using `requires_version` for the moment!

PR #3383, opened 2023-07-11
feat: add `allowed_for_roles` decorator in `User` and `Workspace`

### LINKED ISSUE #3367: [FEATURE] Add Python decorator to check `role`

All the functions/methods related to Argilla are exposed equally to any user no matter what role they have, and we need some sort of control to avoid sending requests to Argilla if we know that an user with that role has no access in advance. Kudos to @nataliaElv for mentioning the issue that led to the proposal of this idea.

Ideally, we should have a Python decorator that checks for the current user's role (`admin`, `owner` or `annotator` as of Argilla 1.13.0), and raise an exception if an user is trying to perform an operation that it's not allowed to via the Python client, e.g. an annotator wants to list all the users in Argilla.

**Additional context**

This feature is also nice to internally know which users have access to what, as otherwise the Python client ends up being a little bit messy and we lose control of the permission levels.

alvarobartt:
# Description

This PR includes the `allowed_for_roles` Python decorator that checks the role for the current user and checks whether that role is allowed to use the decorated function/method, otherwise, a `ValueError` is raised stating that the function/method cannot be used with that role and mentioning the permitted roles.

Kudos to @nataliaElv for mentioning the issue that lead to the definition of this PR.

Closes #3367

**Type of change**

**How Has This Been Tested**

- [X] Add unit tests with different unsupported roles to show that while `owner` can do everything, `admin` and `annotator` are limited

**Checklist**


src/argilla/client/utils.py
```diff
@@ -0,0 +1,57 @@
...
+    from typing_extensions import ParamSpec
+
+from argilla.client.api import ArgillaSingleton
+
+_P = ParamSpec("_P")
+_R = TypeVar("_R")
+
+
+def allowed_for_roles(
+    roles: List[Literal["owner", "admin", "annotator"]]
```
gabrielmbmb: this type hint could use `UserRole` enum instead, right?

src/argilla/client/utils.py
```diff
@@ -0,0 +1,57 @@
...
+        >>> def delete_user(user_id: UUID) -> None:
+        >>>     ...
+    """
+
+    def decorator(func: Callable[_P, _R]) -> Callable[_P, _R]:
+        @functools.wraps(func)
+        def wrapper(*args: _P.args, **kwargs: _P.kwargs) -> _R:
+            role = ArgillaSingleton.get().user.role
+            if role not in roles:
+                raise ValueError(
```
gabrielmbmb: maybe we can raise built-in `PermissionError` (although it's usually used with file-system permission errors) or create a custom exception

gabrielmbmb (approved):
LGTM! Just some comment related to the `allowed_for_roles` decorator

src/argilla/client/utils.py
frascuchon: Yes, should be better using the enum definition

frascuchon (approved):

PR #3389, opened 2023-07-12
:sparkles: FEAT: show required question instead of optional one

keithCuniah:
# Description

Instead of showing `(Optional)` after optional question, there is `*` at the end of a `required` question

**How Has This Been Tested**

- Only feedback task (all question types are concerned by this modification)


Amelie-V:
* in red color @keithCuniah

PR #3390, opened 2023-07-12
feat: add GET `/api/v1/me/workspaces` & update `Workspace` class

alvarobartt:
# Description

This PR adds the `/api/v1/me/workspaces` endpoint in the API V1, and also updates the `Workspace` accordingly so that both users with role annotator and/or admin, can list the workspaces they are part of.

So on, the bug related to the access to the `User.workspaces` property is now solved, as before just the `owners` were able to access it and print the model (as `workspace` was part of the repr).

Besides that, some minor things have also been tackled as part of this PR:
* Remove unused imports
* Restrict `Workspace.delete_user` to users with role `owner` only

**Type of change**

**How Has This Been Tested**

- [X] Add unit tests for GET `/api/v1/me/workspaces`
- [X] Add unit tests for `list_workspaces_me` in Python SDK
- [X] Add unit tests for `User.workspaces` in Python client

All the above tested for all the roles: owner, admin, and annotator.

**Checklist**

### DIFF
--- a/src/argilla/client/users.py
+++ b/src/argilla/client/users.py
@@ -26,13 +26,14 @@ from argilla.client.sdk.commons.errors import (
 from argilla.client.sdk.users import api as users_api
 from argilla.client.sdk.users.models import UserCreateModel, UserModel, UserRole
 from argilla.client.sdk.v1.users import api as users_api_v1
-from argilla.client.sdk.v1.workspaces.models import WorkspaceModel
+from argilla.client.sdk.v1.workspaces import api as workspaces_api_v1
 from argilla.client.utils import allowed_for_roles
 
 if TYPE_CHECKING:
     import httpx
 
     from argilla.client.sdk.client import AuthenticatedClient
+    from argilla.client.sdk.v1.workspaces.models import WorkspaceModel
 
 
 class User:
@@ -108,13 +109,16 @@ class User:
         raise Exception(error_msg)
 
     @property
-    def workspaces(self) -> Optional[List[WorkspaceModel]]:
+    def workspaces(self) -> Optional[List["WorkspaceModel"]]:
         """Returns the workspace names the current user is linked to.
 
         Returns:
             A list of `WorkspaceModel` the current user is linked to.
         """
-        return users_api_v1.list_user_workspaces(self.__client, self.id).parsed
+        if self.role == UserRole.owner:
+            return users_api_v1.list_user_workspaces(self.__client, self.id).parsed
+        else:
+            return workspaces_api_v1.list_workspaces_me(self.__client).parsed
 
     def __repr__(self) -> str:
         return (
--- a/src/argilla/client/users.py
+++ b/src/argilla/client/users.py
@@ -108,7 +108,6 @@ class User:
         raise Exception(error_msg)
 
     @property
-    @allowed_for_roles(roles=[UserRole.owner])
     def workspaces(self) -> Optional[List[WorkspaceModel]]:
         """Returns the workspace names the current user is linked to.
 
--- a/src/argilla/client/users.py
+++ b/src/argilla/client/users.py
@@ -115,10 +115,10 @@ class User:
         Returns:
             A list of `WorkspaceModel` the current user is linked to.
         """
-        if active_client().user.role == UserRole.owner:
+        connected_user = users_api.whoami_httpx(self.__client).parsed
+        if connected_user.role == UserRole.owner:
             return users_api_v1.list_user_workspaces(self.__client, self.id).parsed
-        else:
-            return workspaces_api_v1.list_workspaces_me(self.__client).parsed
+        return workspaces_api_v1.list_workspaces_me(self.__client).parsed
 
     def __repr__(self) -> str:
         return (
--- a/src/argilla/client/users.py
+++ b/src/argilla/client/users.py
@@ -61,7 +61,7 @@ class User:
         >>> from argilla import rg
         >>> user = rg.User.from_name("my-user") # or `User.from_id("...")`
         >>> print(user)
-        User(id='...', username='my-user', first_name='Luke', last_name="Skywalker', full_name='Luke Skywalker', role='annotator', workspaces=[WorkspaceModel(...), ...], api_key='...', inserted_at=datetime.datetime(2021, 8, 31, 10, 0, 0), updated_at=datetime.datetime(2021, 8, 31, 10, 0, 0))
+        User(id='...', username='my-user', role='annotator', first_name='Luke', last_name="Skywalker', full_name='Luke Skywalker', role='annotator', api_key='...', inserted_at=datetime.datetime(2021, 8, 31, 10, 0, 0), updated_at=datetime.datetime(2021, 8, 31, 10, 0, 0))
     """
 
     __client: "httpx.Client"
--- a/src/argilla/client/users.py
+++ b/src/argilla/client/users.py
@@ -123,9 +123,9 @@ class User:
     def __repr__(self) -> str:
         return (
             f"User(id={self.id}, username={self.username}, role={self.role},"
-            f" workspaces={self.workspaces}, api_key={self.api_key},"
-            f" first_name={self.first_name}, last_name={self.last_name}, role={self.role},"
-            f" inserted_at={self.inserted_at}, updated_at={self.updated_at})"
+            f" api_key={self.api_key}, first_name={self.first_name},"
+            f" last_name={self.last_name}, inserted_at={self.inserted_at},"
+            f" updated_at={self.updated_at})"
         )
 
     @staticmethod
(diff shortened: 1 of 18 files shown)


src/argilla/client/users.py
```diff
@@ -108,14 +109,16 @@ def __init__(
...
         """Returns the workspace names the current user is linked to.
 
         Returns:
             A list of `WorkspaceModel` the current user is linked to.
         """
-        return users_api_v1.list_user_workspaces(self.__client, self.id).parsed
+        if self.role == UserRole.owner:
+            return users_api_v1.list_user_workspaces(self.__client, self.id).parsed
+        else:
+            return workspaces_api_v1.list_workspaces_me(self.__client).parsed
```
frascuchon: this is working because the admin and annotator role cannot create `User` without using `User.me`, right?

frascuchon: I mean, ideally, to launch the `list_workspaces_me` the condition should verify that `self` is the connected user. Anyway, this should work

alvarobartt: No, you're right, indeed if we list users from the `owner` we won't get the real results, I'll add the corrected statement now! Nice catch 😄

frascuchon: This change will have a problem if the client is reinitialized after creating the user instance. We should find another approach

src/argilla/client/users.py
```diff
@@ -108,14 +109,16 @@ def __init__(
...
         """Returns the workspace names the current user is linked to.
 
         Returns:
             A list of `WorkspaceModel` the current user is linked to.
         """
-        return users_api_v1.list_user_workspaces(self.__client, self.id).parsed
+        if active_client().user.role == UserRole.owner:
+            return users_api_v1.list_user_workspaces(self.__client, self.id).parsed
+        else:
+            return workspaces_api_v1.list_workspaces_me(self.__client).parsed
```
frascuchon:
```suggestion
        # TODO: improve this
        connected_user = users_api.whoami_httpx(self.__client).parsed

        if connected_user.role == UserRole.owner:
            return users_api_v1.list_user_workspaces(self.__client, self.id).parsed

        elif connected_user.id == self.id:
            return workspaces_api_v1.list_workspaces_me(self.__client).parsed
        else:
            raise PermissionError(
                f"User with role={connected_user.role} is not allowed to workspaces for user `{self.id}`."
            )
```

frascuchon: Or just skip the id check:
```suggestion
        # TODO: improve this
        connected_user = users_api.whoami_httpx(self.__client).parsed

        if connected_user.role == UserRole.owner:
            return users_api_v1.list_user_workspaces(self.__client, self.id).parsed
        return workspaces_api_v1.list_workspaces_me(self.__client).parsed
```

alvarobartt: We can skip indeed the `user_id` check because any other user than `owner` is not able to use any other method besides `User.me()` initially

alvarobartt: And regarding the proposal, shouldn't we use the current active user instead of an HTTP request? This way we'll avoid sending 2 requests to Argilla on every call to `User.workspaces`

alvarobartt: I assume the use-case is changes in the role of the current user, which would be problematic if we're using an `admin` or `annotator`, as we can just use the `User.me` classmethod, and changing the role of the active user will still make the class invalid, as the `role` attribute won't be updated, so IMO there's no need to send that extra HTTP request

frascuchon: Yes, it's why I put the TODO to improve this. My concern about your proposal is regarding the use of `active_client` inside this method. This client can be misaligned with the `__client` when the user instance was created, so we can use the wrong info to check the workspaces access

src/argilla/server/apis/v1/handlers/workspaces.py
```diff
@@ -44,3 +44,15 @@ async def get_workspace(
...
+@router.get("/me/workspaces", response_model=Workspaces)
+async def list_workspaces_me(
+    *,
+    db: AsyncSession = Depends(get_async_db),
+    current_user: User = Security(auth.get_current_user),
+) -> Workspaces:
+    await authorize(current_user, WorkspacePolicyV1.list_workspaces_me)
+
+    workspaces = await accounts.list_workspaces_by_user_id(db, current_user.id)
+    return Workspaces(items=workspaces)
```
frascuchon:
```suggestion
    await authorize(current_user, WorkspacePolicyV1.list_workspaces_me)

    if current_user.is_owner:
        workspaces = await accounts.list_workspaces(db)
    else:
        workspaces = await accounts.list_workspaces_by_user_id(db, current_user.id)
    return Workspaces(items=workspaces)
```

tests/server/api/v1/test_workspaces.py
```diff
@@ -75,3 +75,40 @@ async def test_get_workspace_with_nonexistent_workspace_id(client: TestClient, o
...
+
+    assert response.status_code == 200
+    assert len(response.json()["items"]) == len(workspaces)
+    for workspace in workspaces:
+        assert {
+            "id": str(workspace.id),
+            "name": workspace.name,
+            "inserted_at": workspace.inserted_at.isoformat(),
+            "updated_at": workspace.updated_at.isoformat(),
+        } in response.json()["items"]
```
frascuchon:
```suggestion
@pytest.mark.asyncio
async def test_list_workspaces_me_as_owner(client: TestClient, owner_auth_header: dict) -> None:
    workspaces = await WorkspaceFactory.create_batch(size=5)

    response = client.get("/api/v1/me/workspaces", headers=owner_auth_header)

    assert response.status_code == 200
    assert len(response.json()["items"]) == len(workspaces)
    for workspace in workspaces:
        assert {
            "id": str(workspace.id),
            "name": workspace.name,
            "inserted_at": workspace.inserted_at.isoformat(),
            "updated_at": workspace.updated_at.isoformat(),
        } in response.json()["items"]
 
               
@pytest.mark.asyncio
@pytest.mark.parametrize("role", [UserRole.annotator, UserRole.admin])
async def test_list_workspaces_me(client: TestClient, role: UserRole) -> None:
    workspaces = await WorkspaceFactory.create_batch(size=5)
    user = await UserFactory.create(role=role, workspaces=workspaces)

    response = client.get("/api/v1/me/workspaces", headers={API_KEY_HEADER_NAME: user.api_key})

    assert response.status_code == 200
    assert len(response.json()["items"]) == len(workspaces)
    for workspace in workspaces:
        assert {
            "id": str(workspace.id),
            "name": workspace.name,
            "inserted_at": workspace.inserted_at.isoformat(),
            "updated_at": workspace.updated_at.isoformat(),
        } in response.json()["items"]
```

tests/client/sdk/v1/workspaces/test_api.py
```diff
@@ -0,0 +1,51 @@
...
+    user = await UserFactory.create(role=role, workspaces=workspaces)
+
+    httpx_client = ArgillaSingleton.init(api_key=user.api_key).http_client.httpx
+
+    response = list_workspaces_me(client=httpx_client)
+    assert response.status_code == 200
+    assert isinstance(response.parsed, list)
+    assert len(response.parsed) > 0
+    assert isinstance(response.parsed[0], WorkspaceModel)
+    assert len(response.parsed) == len(workspaces)
```
frascuchon:
```suggestion
@pytest.mark.asyncio
async def test_list_workspaces_me_as_owner(owner: User) -> None:
    workspaces = await WorkspaceFactory.create_batch(size=5)

    httpx_client = ArgillaSingleton.init(api_key=user.api_key).http_client.httpx

    response = list_workspaces_me(client=httpx_client)
    assert response.status_code == 200
    assert isinstance(response.parsed, list)
    assert len(response.parsed) > 0
    assert isinstance(response.parsed[0], WorkspaceModel)
    assert len(response.parsed) == len(workspaces)

@pytest.mark.asyncio
@pytest.mark.parametrize("role", [UserRole.admin, UserRole.annotator])
async def test_list_workspaces_me(role: UserRole) -> None:
    workspaces = await WorkspaceFactory.create_batch(size=5)
    user = await UserFactory.create(role=role, workspaces=workspaces)

    httpx_client = ArgillaSingleton.init(api_key=user.api_key).http_client.httpx

    response = list_workspaces_me(client=httpx_client)
    assert response.status_code == 200
    assert isinstance(response.parsed, list)
    assert len(response.parsed) > 0
    assert isinstance(response.parsed[0], WorkspaceModel)
    assert len(response.parsed) == len(workspaces)
```

src/argilla/server/apis/v1/handlers/workspaces.py
alvarobartt: Thanks for the suggestion (sorry I forgot to include you as co-author 😞)

frascuchon (approved):

PR #3304, opened 2023-07-03
feat: add `Suggestion` endpoints

gabrielmbmb:
# Description

This PR adds the following endpoints:

- `GET /api/v1/records/{record_id}/suggestions`
- `POST /api/v1/records/{record_id}/suggestions`

and updates the following ones:

- `GET /api/v1/datasets/{dataset_id}/records`: output payload updated to add suggestions depending on the value of the query parameter `include`
- `GET /api/v1/me/datasets/{dataset_id}/records`: output payload updated to add suggestions depending on the value of the query parameter `include`
- `POST /api/v1/me/datasets/{dataset_id}/records/search`: output payload updated to add suggestions depending on the value of the query parameter `include`
- `POST /api/v1/datasets/{dataset_id}/records`: input payload updated to add suggestions

Additionally and not directly related to the suggestions, this PR has updated the `GET /api/v1/me/datasets/{dataset_id}/records` endpoint to only include the responses of the record in the output payload when the value `responses` has been provided via the `include` query parameter. Old behaviour always returned the responses.

Finally, I've added a `TimestampMixin` class that can be used with a SQLAlchemy ORM model class to add the `inserted_at` and `updated_at` attributes.

Some notes about the suggestions:

- I've decided that it is a good idea to leave the `Suggestion.score` value unrestricted. At first I thought it would be good to force the user to provide a value between `0` and `1`, but users may want to assign a suggestion based on a distance based metric whose value is not bounded between `0` and `1`.
- `Suggestion.type` has only two possible values for now: `human` and `model` (machine learning, rule-based, heuristic, etc). The idea is to complement with `Suggestion.agent`.

Closes #3277

**Type of change**

- [x] Refactor
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

**How Has This Been Tested**

Calling the endpoints in a local environment manually and adding unit tests covering all the possible cases.

**Checklist**

- [ ] I added relevant documentation
- [x] follows the style guidelines of this project
- [x] I did a self-review of my code
- [ ] I made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I filled out [the contributor form](https://tally.so/r/n9XrxK) (see text above)
- [x] I have added relevant notes to the CHANGELOG.md file (See https://keepachangelog.com/)

PR #3391, opened 2023-07-12
feat: update create suggestion endpoint to `PUT`

gabrielmbmb:
# Description

This PR updates the `POST` create suggestion endpoint added in #3304 to be a `PUT` endpoint instead that allows creating a suggestion for a given combination of question + record (if it doesn't exist) or update an existing one:

- The endpoint will return a `201` when the suggestion is created
- The endpoint will return a `200` when the suggestion existed and it has been updated.

Also, this PR updates the `Suggestion` SQLAlchemy ORM model adding a new unique constraint for record id and question id. As the migration file added in #3304 to create the `suggestions` table has not been applied yet in any prod environment, just in `dev.argilla.io`, I modified the migration file to include this unique constraint. Before deploying these changes in `dev.argilla.io` we will need to:

1. Revert the migration
2. Apply the latest migration

**Type of change**

- [x] Breaking change

**How Has This Been Tested**

Manually in a local environment and I've added unit tests.

**Checklist**


src/argilla/server/schemas/v1/suggestions.py
```diff
@@ -32,6 +32,10 @@ class SuggestionCreate(BaseSuggestion):
     pass
 
 
+class SuggestionCreateWithRecordId(BaseSuggestion):
+    record_id: UUID
```
frascuchon: If this schema is not used in any endpoint, we should move from here. Not sure if as part of the `contexts`, or related to the `models` modules, as we have for questions. Otherwise, could create confusion regarding the layer separation.

frascuchon (approved):
Just a comment regarding the new schema

alvarobartt (approved):
LGTM! Feel free to merge 👍🏻

PR #3400, opened 2023-07-13
ci: using extended-runner for tests

frascuchon:
# Description

Forcing the use of extended-runner for running tests. This will help to random errors in ci pipelines.

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -34,7 +34,7 @@ jobs:
 
   deployable_check:
     name: Check if must build & deploy from environment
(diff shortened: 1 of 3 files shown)


alvarobartt:
May I ask what's the `extended-runner`? is that something custom?

frascuchon:
> May I ask what's the `extended-runner`? is that something custom?

Yes. A [runner](https://github.com/organizations/argilla-io/settings/actions/github-hosted-runners/7?viewing_from_runner_group=false) with more resources than default one

PR #3421, opened 2023-07-17
feat: field `PATCH` endpoint

### LINKED ISSUE #3397: [FEATURE] Add `PATCH` endpoint for `FeedbackDataset` fields
Some attributes of a field of a `FeedbackDataset` cannot be updated once created.

Add a `PATCH /api/v1/fields/:field_id` endpoint that allows to update the following attributes:
- `title`
- `settings.use_markdown`

gabrielmbmb:
# Description

This PRs adds a new `PATCH /api/v1/fields/{field_id}` endpoint that allows to partially update a field of a `FeedbackDataset` in the API. The attributes that can be updated using this new endpoint are:

- `title`
- `settings.use_markdown`

To be able to update the `dict` column, it was needed to update the `CRUDMixin.fill` method to detect if the column to be filled is a `dict` column. In that case, it will iterate the keys and values of the received value, and it will set the key and value of the `dict` col one by one to avoid overriding the whole dict.

Closes #3397

**Type of change**

**How Has This Been Tested**

I've tested it manually in a local environment and I've added unit tests.

**Checklist**


src/argilla/server/models/mixins.py
```diff
@@ -50,6 +50,15 @@ def fill(self, **kwargs: Any) -> Self:
...
                 raise AttributeError(f"Model `{self.__class__.__name__}` has no attribute `{key}`")
+            # If the value is a dict, set value for each key one by one, as we want to update only the keys that are in
+            # `value` and not override the whole dict.
+            if isinstance(value, dict):
+                dict_col = getattr(self, key)
+                if not dict_col:
+                    dict_col = {}
+                for k, v in value.items():
+                    dict_col[k] = v
+                value = dict_col
```
frascuchon: I think we can simplify this code block
```suggestion
                dict_col = getattr(self, key, {})
                dict_col.update(value)
                value = dict_col
```

src/argilla/server/models/models.py
```diff
@@ -60,7 +61,7 @@ class Field(DatabaseModel):
     name: Mapped[str] = mapped_column(Text, index=True)
     title: Mapped[str] = mapped_column(Text)
     required: Mapped[bool] = mapped_column(default=False)
-    settings: Mapped[dict] = mapped_column(JSON, default={})
+    settings: Mapped[dict] = mapped_column(MutableDict.as_mutable(JSON), default={})
```
frascuchon, line 64: Do we need this line? Or just prevent a prover config for future event handlers listening to the `onChanged` event?

src/argilla/server/schemas/v1/fields.py
```diff
@@ -38,3 +38,12 @@ class Field(BaseModel):
...
         orm_mode = True
+
+
+class FieldSettingsUpdate(BaseModel):
+    use_markdown: Optional[bool]
+
+
+class FieldUpdate(BaseModel):
+    title: Optional[str]
+    settings: Optional[FieldSettingsUpdate]
```
frascuchon: Just to align with other definitions
```suggestion
class TextFieldSettingsUpdate(BaseModel):
    use_markdown: Optional[bool]


class FieldUpdate(BaseModel):
    title: Optional[str]
    settings: Optional[TextFieldSettingsUpdate]
```

tests/factories.py
```diff
@@ -225,7 +225,7 @@ class Meta:
 
 
 class TextFieldFactory(FieldFactory):
-    settings = {"type": FieldType.text.value}
+    settings = {"type": FieldType.text.value, "use_markdown": False}
```
frascuchon, line 228: I would say to keep using the default values for `use_markdown` and use the desired values in tests

src/argilla/server/models/models.py
gabrielmbmb: Right now we don't need it, but might be useful in the future if we want to use `sqlalchemy.event.listens_for` decorator.

src/argilla/server/models/mixins.py
```diff
@@ -50,6 +50,12 @@ def fill(self, **kwargs: Any) -> Self:
         for key, value in kwargs.items():
             if not hasattr(self, key):
                 raise AttributeError(f"Model `{self.__class__.__name__}` has no attribute `{key}`")
+            # If the value is a dict, set value for each key one by one, as we want to update only the keys that are in
+            # `value` and not override the whole dict.
+            if isinstance(value, dict):
+                dict_col = getattr(self, key, {})
```
frascuchon: My bad. The previous line works when the attribute is not defined, but no when is defined with a None value
```suggestion
                dict_col = getattr(self, key) or {}
```

src/argilla/server/schemas/v1/fields.py
```diff
@@ -38,3 +38,12 @@ class Field(BaseModel):
 
     class Config:
         orm_mode = True
+
+
+class TextFieldSettingsUpdate(BaseModel):
+    use_markdown: Optional[bool]
+
+
+class FieldUpdate(BaseModel):
```
frascuchon, line 48: We should align the field validations with the `FieldCreate` class
```python
class FieldCreate(BaseModel):
    name: constr(
        regex=FIELD_CREATE_NAME_REGEX,
        min_length=FIELD_CREATE_NAME_MIN_LENGTH,
        max_length=FIELD_CREATE_NAME_MAX_LENGTH,
    )
    title: constr(
        min_length=FIELD_CREATE_TITLE_MIN_LENGTH,
        max_length=FIELD_CREATE_TITLE_MAX_LENGTH,
    )
```

gabrielmbmb: Done!

frascuchon (approved):

frascuchon (approved):

PR #3422, opened 2023-07-17
feat: update `package.yml` triggers

gabrielmbmb:
# Description

This PR updates the triggers of the `package.yml` workflow:

- Remove all the branches but `main`, `develop` and `releases/**` from the `push` trigger
- Add `workflow_dispatch` trigger to allow triggering the `package.yml` workflow manually.

This way, the unit tests will gets executed just one time per push when a PR for the branch is opened. If there isn't a PR opened for the branch and for some reason the Docker image for that branch needs to be build, then the `package.yml` workflow can get executed manually.

This PR is an option and a personal opinion, if you agree we can merge it. Otherwise, I'll close the PR.

**Type of change**

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

**How Has This Been Tested**

**Checklist**


.github/workflows/package.yml
```diff
@@ -13,10 +15,6 @@ on:
     branches:
       - "main"
       - "develop"
-      - "integration"
-      - "releases/**"
```
alvarobartt, line 17: I guess we can keep this one in case we're pushing hotfixes to `releases/**` right before releasing it, as sometimes we need to just push late time stuff there without creating a PR (which I guess shouldn't be the case, but sometimes is)

alvarobartt (approved):
LGTM, just some minor comment before merging TBC with @frascuchon ideally

.github/workflows/package.yml
gabrielmbmb: yes, makes sense

frascuchon:
I agree with you in the sense we should control some flow executions. But, since the docker images will be automatically generated just once when the PR is created, we need to keep all the team notified to understand how to launch the pipeline manually.

frascuchon (approved):

PR #3428, opened 2023-07-18
fix: ensure `suggestions` provided to `FeedbackRecord.set_suggestions` are unique

alvarobartt:
# Description

This PR makes sure that the `suggestions` argument provided to `FeedbackRecord.set_suggestions` are unique once parsed to `SuggestionSchema`, to ensure that we're not sending bad requests to the API.

As of now, even though seems like an edge case, the code below would fail:

```python
FeedbackRecord.set_suggestions(suggestions=[{"question_name": "question-1", "value": 1}, {"question_name": "question-1", "value": 2}])
```

While the following works fine:

```python
FeedbackRecord.set_suggestions(suggestions={"question_name": "question-1", "value": 1})
FeedbackRecord.set_suggestions(suggestions={"question_name": "question-1", "value": 2})
```

**Type of change**

- [X] Bug fix (non-breaking change which fixes an issue)

**How Has This Been Tested**

(Please describe the tests that you ran to verify your changes. And ideally, reference `tests`)

- [x] Unit tests included on `FeedbackRecord.set_suggestions` at #3427

**Checklist**

- [ ] I added relevant documentation
- [X] follows the style guidelines of this project
- [X] I did a self-review of my code
- [ ] I made corresponding changes to the documentation
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [ ] I filled out [the contributor form](https://tally.so/r/n9XrxK) (see text above)
- [ ] I have added relevant notes to the CHANGELOG.md file (See https://keepachangelog.com/)

PR #3430, opened 2023-07-18
refactor: improve suggestions workflow

frascuchon:
hello, 

# Description

Given the following code snippet:
```python
from argilla.feedback import FeedbackRecord
​
record = FeedbackRecord(fields={"text": "this is a text"}, suggestions=[{"question_name": "qt", "value": 3}])
​
record.set_suggestions({"question_name": "dd", "value": 234})
#  /argilla/client/feedback/schemas.py:203: UserWarning: Ignore the following
# if you are creating a new `FeedbackDataset` with `FeedbackRecord`s, or if you are just working with a
# `FeedbackRecord`. Otherwise, if the `FeedbackRecord` is already pushed to Argilla, note that `suggestions` have
# been provided, but the `id` is not set, which means that the `FeedbackRecord` has been pushed to Argilla,
# but hasn't been fetched, so the `id` is missing. To solve that, you can simply call `FeedbackDataset.fetch_records()`
# to fetch them and automatically set the `id`, to call `set_suggestions` on top of that.
#   warnings.warn(
# /argilla/client/feedback/schemas.py:219: UserWarning: A suggestion for question
# `qt` has already been provided, so the provided suggestion will overwrite it.
#   warnings.warn(
# /argilla/client/feedback/schemas.py:231: UserWarning: Ignore this warning if
# you're already using `set_suggestions` method. Otherwise, if you are trying to set `suggestions` directly,
# which is not allowed. You should use the `set_suggestions` method instead.
#   warnings.warn(
​
records.suggestions = [{"question_name": "ab", "value": 33}]
```

improve the workflow to:

- Avoid suggestions assignation and force using `set_suggestions` to update the record suggestions
- Remove extra warnings (all related to suggestions assignment) 
- Move record-id-related warnings to the `push_to_argilla` method since in that context have more sense
- Skip updating a record in argilla if there is no id for the record (related to the previous point)

This enhancement is based on using tuples for store suggestions (since a tuple is not mutable), and the `allow_mutation=False` pydantic parameter (to avoid field assignment)

Also, some tests have been included to validate the flow when [creating](https://github.com/argilla-io/argilla/blob/refactor/improve-suggesions-workflow/tests/client/feedback/test_dataset.py#L146) or [updating](https://github.com/argilla-io/argilla/blob/refactor/improve-suggesions-workflow/tests/client/feedback/test_dataset.py#L173) records with suggestions (by checking the expected content instead of the expected local state after pushing the data). [One of those tests ](https://github.com/argilla-io/argilla/blob/refactor/improve-suggesions-workflow/tests/client/feedback/test_dataset.py#L173) is still failing but, testing the same flow outside the tests, it works. So, not sure about the error (this error happened before changes if a test check the expected suggestions for a record after updating them)

Finally, the bug fixed in https://github.com/argilla-io/argilla/pull/3428 is already resolved by using a dictionary structure in `set_suggestions` method. So the fix is not needed anymore.

**Type of change**

**How Has This Been Tested**

Launching several workflows locally 

**Checklist**

### DIFF
--- a/src/argilla/client/feedback/dataset.py
+++ b/src/argilla/client/feedback/dataset.py
@@ -574,6 +574,20 @@ class FeedbackDataset(HuggingFaceDatasetMixin):
                         disable=show_progress,
                     ):
                         for record in updated_records[i : i + PUSHING_BATCH_SIZE]:
+                            if not record.id:
+                                warnings.warn(
+                                    "A record is already pushed to Argilla but no record id found, which means"
+                                    " that the `FeedbackRecord` has been pushed to  Argilla, but hasn't been fetched,"
+                                    " so the `id` is missing. Record update will be skipped for this record"
+                                )
+                                if record.suggestions:
+                                    warnings.warn(
+                                        "The `suggestions` have been provided, but the `id`"
+                                        " is not set. To solve that, you can simply call"
+                                        " `FeedbackDataset.fetch_records()` to fetch them and"
+                                        " automatically set the `id`, to call `set_suggestions` on top of that."
+                                    )
+                                continue
                             for suggestion in record.suggestions:
                                 suggestion.question_id = question_name2id[suggestion.question_name]
                                 datasets_api_v1.set_suggestion(
--- a/src/argilla/client/feedback/schemas.py
+++ b/src/argilla/client/feedback/schemas.py
@@ -192,8 +192,10 @@ class FeedbackRecord(BaseModel):
     _updated: bool = PrivateAttr(default=False)
 
     @validator("suggestions", always=True)
-    def normalize_suggestions(cls, values) -> Tuple:
-        return tuple([v for v in values])
+    def normalize_suggestions(cls, values: Any) -> Tuple:
+        if not isinstance(values, tuple):
+            return tuple([v for v in values])
+        return values
 
     def set_suggestions(
         self, suggestions: Union[SuggestionSchema, List[SuggestionSchema], Dict[str, Any], List[Dict[str, Any]]]
--- a/src/argilla/client/feedback/schemas.py
+++ b/src/argilla/client/feedback/schemas.py
@@ -13,7 +13,7 @@
 #  limitations under the License.
 
 import warnings
-from typing import TYPE_CHECKING, Any, Dict, List, Literal, Optional, Union
+from typing import TYPE_CHECKING, Any, Dict, List, Literal, Optional, Tuple, Union
 from uuid import UUID
 
 from pydantic import (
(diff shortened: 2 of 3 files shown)


src/argilla/client/feedback/dataset.py
```diff
@@ -574,6 +574,20 @@ def push_to_argilla(
                         disable=show_progress,
                     ):
                         for record in updated_records[i : i + PUSHING_BATCH_SIZE]:
+                            if not record.id:
```
alvarobartt, line 577: These were defined in `set_suggestions` to avoid the users from calling `set_suggestions` and then after setting suggestions for all the records realising that they cannot push those to Argilla and need to complete the process again. But we can also define it here if you think is better.

src/argilla/client/feedback/schemas.py
```diff
@@ -183,12 +183,18 @@ class FeedbackRecord(BaseModel):
...
+        default_factory=tuple, allow_mutation=False
+    )
     external_id: Optional[str] = None
 
     _unified_responses: Optional[Dict[str, List["UnifiedValueSchema"]]] = PrivateAttr(default_factory=dict)
     _updated: bool = PrivateAttr(default=False)
 
+    @validator("suggestions", always=True)
+    def normalize_suggestions(cls, values) -> Tuple:
+        return tuple([v for v in values])
```
alvarobartt: Here we are missing `values` type-hint and the `return` just applies if not a `Tuple` already right? So shouldn't it be something like
```suggestion
    def normalize_suggestions(cls, values: Any) -> Tuple:
        if not isinstance(values, tuple):
            return tuple([v for v in values])
        return values
```

src/argilla/client/feedback/schemas.py
```diff
@@ -199,42 +205,21 @@ def set_suggestions(
...
-                    self.suggestions.remove(existing_suggestion)
-                    break
-        self.suggestions += parsed_suggestions
+            if suggestion.question_name in suggestions_dict:
+                warnings.warn(
+                    f"A suggestion for question `{suggestion.question_name}` has already"
+                    " been provided, so the provided suggestion will overwrite it.",
+                    category=UserWarning,
+                )
+            suggestions_dict[suggestion.question_name] = suggestion
```
alvarobartt, line 219: Even though this works, isn't it better to just keep the first one we find among the ones provided in the scenario mentioned at https://github.com/argilla-io/argilla/pull/3428, also including a warning so that the use knows what's happening? To differentiate between the scenario where the `suggestion` was already there, versus the scenario where the user provided two suggestions for the same question within `set_suggestions`? LMKWYT

tests/client/feedback/test_dataset.py
```diff
@@ -554,7 +584,7 @@ async def test_update_dataset_records_in_argilla(
                 },
             ]
         )
-    with pytest.warns(UserWarning, match="if you are trying to set `suggestions` directly"):
+    with pytest.raises(TypeError, match='"suggestions" has allow_mutation set to False and cannot be assigned'):
```
alvarobartt, line 594:
```suggestion
    with pytest.raises(TypeError, match="\"suggestions\" has allow_mutation set to False and cannot be assigned"):
```

alvarobartt: Didn't try this, but I assume we should keep the standard for using double-quotes and just scape the nested ones

alvarobartt (approved):
LGTM! Thanks for the fix, makes sense, just some minor comments but we can merge and release and I can address those afterwards if applicable 👍🏻

tests/client/feedback/test_dataset.py
frascuchon: The black process changed the string to the single quotes style in this case. So I'll keep as is.

tests/client/feedback/test_dataset.py
```diff
@@ -142,6 +143,68 @@ def test_init_wrong_questions(
...
+    ds.fetch_records()
+    assert len(ds.records) == 1
+    for record in ds.records:
+        assert record.id is not None
+        assert record.suggestions == ()
+
+        record.set_suggestions([{"question_name": "text", "value": "This is a suggestion"}])
+
+    ds.push_to_argilla()
+
```
frascuchon:
```suggestion
# TODO: Review this requirement for tests and explain, try to avoid use or at least, document.
```

PR #3444, opened 2023-07-24
improvement: Better efficiency of Weak Labels when vectors exist

### LINKED ISSUE #3404: [BUG] WeakLabels take so much time to compute when vectors are present
**Describe the bug**
Applying `WeakLabels` take very long when vectors are present in the dataset.

```python
from argilla.labeling.text_classification import Rule, WeakLabels, load_rules

weak_labels = WeakLabels(dataset="descripciones_usuarios_twitter")
```

**Expected behavior**

The weak labels are computed without loading the vectors.

**Environment:**
 - Argilla Version [e.g. 1.0.0]:

**Additional context**

It seems that the problem is caused by [this line](https://github.com/argilla-io/argilla/blob/9be4b00ab42fedb31916f7f485ff13d14e22e2d1/src/argilla/labeling/text_classification/weak_labels.py#L96), which is not using the `include_vectors` parameter. I'm not sure if it should or not, though (as the `extend_matrix` method seems to use this information). 

```[tasklist]
### Tasks
```

tomaarsen:
Closes #3404

Hello!

# Description

As mentioned in #3404, weak labels are slow when the dataset contains vectors. This weak labeling support was added before we even supported vectors, so when these vectors started being included by default in every `rg.load` call, the weak labeling support became much slower. 
In short, the vectors were added in ~January 2023, while weak labeling is from ~October 2022. It seems that we can safely set `include_vectors=False` as I cannot find a place where the vectors are actually used: it seems that vectors are used via [`extend_matrix`](https://docs.argilla.io/en/latest/tutorials/notebooks/labelling-textclassification-sentencetransformers-weaksupervision.html#Extend-the-weak-labels-matrix) rather than the newer `record.vectors`.

**Type of change**

- [x] Improvement

**How Has This Been Tested**

Via `pytest .\tests\labeling\text_classification\` and by running:
```python
import time
import argilla as rg
from argilla.labeling.text_classification import Rule, WeakLabels, load_rules, add_rules

dataset_ds = load_dataset("argilla/agnews_weak_labeling", split="train")
dataset_rb = rg.read_datasets(dataset_ds, task="TextClassification")
rg.log(dataset_rb, name="agnews_weak_labeling_test")

# define queries and patterns for each category (using ES DSL)
queries = [
    (["money", "financ*", "dollar*"], "Business"),
    (["war", "gov*", "minister*", "conflict"], "World"),
    (["footbal*", "sport*", "game", "play*"], "Sports"),
    (["sci*", "techno*", "computer*", "software", "web"], "Sci/Tech"),
]

# define rules
rules = [Rule(query=term, label=label) for terms, label in queries for term in terms]

add_rules(dataset="agnews_weak_labeling_test", rules=rules)

start_t = time.time()
weak_labels = WeakLabels(dataset="agnews_weak_labeling_test")
print(f"{time.time() - start_t:.4f}s to load records and apply weak labels.")
```
This script loads a dataset with vectors, adds some rules, and then loads the weak labels with a timer. On `develop`, i.e. with vectors, this takes ~27.5 seconds. When `include_vectors=False`, it takes ~5.5 seconds.

See also these cProfile graphs:
**Before**:
Schermafbeelding_20230724_123356

**After**:
Schermafbeelding_20230724_123405

**Checklist**

(Note: I'm unsure whether this best falls under "Changed" or another category in the CHANGELOG)

---

- Tom Aarsen


frascuchon (approved):
💯

PR #3450, opened 2023-07-25
feat: update CLI to use async connection to DB

gabrielmbmb:
# Description

This PR updates the CLI, so an async connection to the DB is used in the commands instead of a sync one. As the rest of the parts of the application are using the async connection, the code for the sync connection has been removed.

The list of changes of this PR are:

- Add `AsyncTyper` class which allows to register `async` command with its method `async_command`. It executes then executes the command using `asyncio.run`.
- Add `run` function that allows to execute an `async` command.
- Remove `database_async_url` property from `Settings` class. From now on, `database_url` should have an URL with an async driver (`sqlite+aiosqlite://` or `postgres+asyncpg://`).
- Update Alembic connection to create an async engine to run the migrations.

**Type of change**

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

**How Has This Been Tested**

- [x] All the unit tests regarding the CLI are working as expected
- [x] All the commands have been executed locally and working as expected (using both `python -m argilla users create ...` and `python -m argilla.tasks.users.create ...`)
- [x] Migrations have been applied without errors in a local environment to both SQLite and PostgreSQL instances. 

**Checklist**


src/argilla/server/alembic/env.py
```diff
@@ -68,27 +70,31 @@ def run_migrations_offline() -> None:
...
-        context.configure(connection=connection, target_metadata=target_metadata)
-
-        with context.begin_transaction():
-            context.run_migrations()
+    async with connectable.connect() as connection:
+        await connection.run_sync(apply_migrations)
 
 
 if context.is_offline_mode():
     run_migrations_offline()
```
frascuchon, line 102: this line should also be wrapped in an async statement, right?

gabrielmbmb: yeah, I haven't updated it as we're not using the offline migrations, but I'll do it just in case.

gabrielmbmb: The `run_migrations_offline` function can be kept as is, as it doesn't need to connect to the DB, it just generates an SQL script.

src/argilla/server/alembic/env.py
```diff
@@ -68,27 +70,31 @@ def run_migrations_offline() -> None:
         context.run_migrations()
 
 
-def run_migrations_online() -> None:
+def apply_migrations(connection) -> None:
```
frascuchon: can you add typing for `connection` ?

src/argilla/server/alembic/env.py
frascuchon: but now `run_migrations_offline` is a coroutine so the main program can stop before finishing the coroutine completion

gabrielmbmb: no, `run_migrations_offline` is a regular function, the only one that has been updated is `run_migrations_online`

src/argilla/server/settings.py
```diff
@@ -165,19 +167,6 @@ def old_dataset_records_index_name(self) -> str:
             return index_name.replace("<NAMESPACE>", "")
         return index_name.replace("<NAMESPACE>", f".{ns}")
 
-    @property
```
frascuchon, line 168: We should support at least for the next release setting up the proper async driver when no driver was provided. Otherwise, we cannot prepare the environment for the current environments (is not possible to configure the database URL to `postgresql+asyncpg://....` because of these validations.

frascuchon: Maybe we can move this logic to the database module and use the `sqlalchemy.make_url` and the `sqlalchemy.URL` for that.

src/argilla/tasks/async_typer.py
```diff
@@ -0,0 +1,51 @@
...
+            def sync_func(*_args: P.args, **_kwargs: P.kwargs) -> R:
+                return asyncio.run(async_func(*_args, **_kwargs))
+
+            self.command(*args, **kwargs)(sync_func)
+            return async_func
+
+        return decorator
+
+
+def run(function: Callable[..., Coroutine[Any, Any, Any]]) -> None:
```
frascuchon, line 54: I think using this [alternative](https://github.com/tiangolo/typer/issues/88#issuecomment-1616045478) (already in the same thread) we don't need the `run` function and adding the extra `async_typer.run(command)` line on each command definition

gabrielmbmb: I've update the `AsyncTyper` class to rename `async_command` to `command` (both sync and async functions can be decorated with this method), but we still need the `run` function as we need to use the `AsyncTyper.command` method (which is the one able to execute both sync and async functions).

src/argilla/server/settings.py
gabrielmbmb: I've updated the server `Settings` class to validate the `database_url` provided.

- If `postgresql://...` or `postgresql+psycopg2://` URL like provided, it will be transformed to `postgresql+asyncpg://...` and a warning will be displayed to the user.
- If `sqlite://...` URL like provided, it will be transformed to `sqlite+aiosqlite://...` and a warning will be displayed to the user.

src/argilla/server/settings.py
```diff
@@ -127,7 +129,38 @@ def normalize_base_url(cls, base_url: str):
...
+        home_path = values.get("home_path")
+        sqlite_file = os.path.join(home_path, "argilla.db")
+        return database_url or f"sqlite+aiosqlite:///{sqlite_file}?check_same_thread=False"
+
+    @validator("database_url", pre=True)
+    def set_database_url(cls, database_url: str, values: dict) -> str:
+        if not database_url:
+            home_path = values.get("home_path")
+            sqlite_file = os.path.join(home_path, "argilla.db")
+            return f"sqlite+aiosqlite:///{sqlite_file}?check_same_thread=False"
```
frascuchon, line 135: Do we need this? if the `set_database_url_default` validator is executed for all cases, for an unset database url the default validator should be enough

gabrielmbmb: my bad, I forgot to remove the old validator.

PR #3459, opened 2023-07-26
chore: Fix `ruff` line length

tomaarsen:
Hello!

# Description

While the `black` line length has been correctly set to [120](https://github.com/argilla-io/argilla/blob/6424bc72eab02c03717e445359599888fc6d6146/pyproject.toml#L192), that of `ruff` was still at the (lower) default. I've set it to 120 in 14db301f3694db7bc93023e6b4fcc5a9e607fe39 using [this](https://beta.ruff.rs/docs/settings/#line-length), and fixed some issues that were caused by this in 7a4efb3297533c8c53e48fdd1dcfd5c404ba2304 by searching for `(import \(\n[^,]+),(\n\S*\))` in all source code and replacing it with `$1$2`.
This finds all cases of
```python
... import (
BlaBlaBla,
)
```
where the following might also be possible
```python
... import BlaBlaBla
```

Beyond that, I fixed an issue with a broken type hint in bd7029a28a6a3fa3b2f36b2817eaf22bd0c95067.

**Type of change**

- [x] Refactor

**How Has This Been Tested**

N/A, no real code changes.

**Checklist**

---

- Tom Aarsen

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -183,6 +183,7 @@ exclude = [
     "node_modules",
     "venv",
 ]
+line-length = 120
 
 [tool.ruff.per-file-ignores]
 # Ignore imported but unused;
--- a/src/argilla/client/api.py
+++ b/src/argilla/client/api.py
@@ -20,10 +20,7 @@ from typing import Any, Dict, Iterable, List, Optional, Tuple, Union
 
 from argilla.client.client import Argilla
 from argilla.client.datasets import Dataset
-from argilla.client.models import (  # TODO Remove TextGenerationRecord
-    BulkResponse,
-    Record,
-)
+from argilla.client.models import BulkResponse, Record  # TODO Remove TextGenerationRecord
 from argilla.client.sdk.commons import errors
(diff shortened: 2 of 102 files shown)


gabrielmbmb (approved):
LGTM!

tomaarsen:
I've gone a step further and removed all final commas before `)`, and letting ruff reformat accordingly (by replacing `,(\n\S*\))` with `$1`). This also covers a bunch of other cases like:
```python
from argilla.client.models import (  # TODO Remove TextGenerationRecord
    BulkResponse,
    Record,
)
```
becomes
```python
from argilla.client.models import BulkResponse, Record # TODO Remove TextGenerationRecord
```

or 
```python
def copy(
    dataset: str,
    name_of_copy: str,
    workspace: str = None,
):
```
becomes
```python
def copy(dataset: str, name_of_copy: str, workspace: str = None):
```

---

After going through all changes, it seems like this doesn't accidentally mess up a docstring or warning string or anything.

PR #3462, opened 2023-07-27
feat: create user with workspaces

gabrielmbmb:
# Description

This PR updates the API and the Python client so a list of workspaces names to which the user should be linked to can be provided. This is useful when creating a new user, as just one request will be needed, instead of the current behaviour, which requires one to create the user and another one to add the user to the workspace.

In addition, the code blocks in which a commit or a rollback has to be done depending on wether an exception is raised have been updated to use the async context manager `async with session.begin_nested():` which automatically will commit if the block is executed without any problem or rollback if an exception is raised.

**Type of change**

**How Has This Been Tested**

I've been able to create a user with workspaces in a local development environment. Also, the unit tests regarding the user creation have been updated.

**Checklist**


src/argilla/client/users.py
```diff
@@ -18,7 +18,12 @@
...
 from argilla.client import active_client
-from argilla.client.sdk.commons.errors import AlreadyExistsApiError, BaseClientError, NotFoundApiError
+from argilla.client.sdk.commons.errors import (
+    AlreadyExistsApiError,
+    BaseClientError,
+    NotFoundApiError,
+    ValidationApiError,
+)
 from argilla.client.sdk.users import api as users_api
 from argilla.client.sdk.users.models import UserCreateModel, UserModel, UserRole
```
alvarobartt: `UserCreateModel` is no longer used, right?

src/argilla/server/apis/v0/handlers/users.py
```diff
@@ -95,9 +95,12 @@ async def create_user(
...
         raise EntityAlreadyExistsError(name=user_create.username, type=User)
 
-    user = await accounts.create_user(db, user_create)
-    await user.awaitable_attrs.workspaces
+    try:
+        user = await accounts.create_user(db, user_create)
+    except Exception as e:
+        raise HTTPException(status_code=status.HTTP_422_UNPROCESSABLE_ENTITY, detail=str(e))
 
+    await user.awaitable_attrs.workspaces
```
alvarobartt, line 103: Just to confirm, if no `workspaces` are provided this won't add any overhead right? Couldn't we also add `if user_create.workspaces is not None` right before?

alvarobartt (approved):
LGTM! Thanks 🤗

PR #3465, opened 2023-07-27
refactor: add `FeedbackDatasetBase` and `RemoteFeedbackDataset` while keeping `FeedbackDataset` just for local

### LINKED ISSUE #3456: Python client `FeedbackDataset` refactor to simplify new user workflows implementation
```[tasklist]
### Tasks
- [ ] Create base class `FeedbackDatasetBase` and let `FeedbackDataset` as a working-locally dataset
- [ ] Implement the `ArgillaFeedbackDataset` class to work with datasets defined in argilla
- [ ] Adapt and deprecate properties and methods for `FeedbackDataset` (`fetch_records, `argilla_id`,...)
```

alvarobartt:
# Description

This PR is still in progress, and the main idea is that we move the common functionality to be used on both local and remote (Argilla) datasets to `FeedbackDatasetBase`, while we keep `FeedbackDataset` for working locally with these datasets, and add `_ArgillaFeedbackDataset` to be instantiated internally via `FeedbackDataset.from_argilla` and as part of the return statement of `FeedbackDataset.push_to_argilla`, so as to split the behaviour on some operations such as `add_records`, since locally means adding those to a local list, while remotely that means pushing those to Argilla, to avoid having to call `push_to_argilla` right after every record addition.

Some more things are tackled as part of this refactoring and will be listed down below as soon as the PR is out of draft!

Closes #3456

**Type of change**

**How Has This Been Tested**

- [x] Add unit tests for the different classes and methods, while ensuring backwards compatibility on everything related to the `FeedbackDataset`

**Checklist**


alvarobartt:
The code below works fine, the only missing piece is `fetch_records`, that won't be compatible with the upcoming approach of `_ArgillaFeedbackRecords` to detach the record-related functionality to a separate class i.e. `dataset.records.add` would be used via the shortcut `dataset.add_records`

```python
import argilla as rg

ds = rg.FeedbackDataset(...) # Initialises `FeedbackDataset` as always

records = [...]
ds.add_records(records) # Adds records locally

for record in ds.records: # Loops over the local records
    print(record)

print(ds.records[0]) # Prints the record with index 0 locally

rg.init(
    api_url="...",
    api_key="...",
)

ds.push_to_argilla("my-dataset", workspace="alvaro") # Pushes the `FeedbackDataset` to Argilla, and remaps the class instance to be now of type `_ArgillaFeedbackDataset`, also returns it, but to keep backwards compatibility we're also remapping it for the moment

ds = rg.FeedbackDataset.from_argilla("my-dataset", workspace="alvaro") # Retrieves the `FeedbackDataset` from Argilla and returns an `_ArgillaFeedbackDataset` instance

for record in ds: # Iters over the records of the `FeedbackDataset` in Argilla (yield not return)
    print(record.id)

print(ds[0]) # Prints the record with index 0 in the `FeedbackDataset` in Argilla
print(ds[1:4]) # Prints the records within the specified slice in the `FeedbackDataset` in Argilla
print(len(ds)) # Prints the total number of records in the `FeedbackDataset` in Argilla

ds.add_records(...) # Adds records directly in Argilla

ds.push_to_argilla("my-dataset", workspace="alvaro") # Raises a `DeprecationWarning` and mentions that updates are automatic
```

src/argilla/client/feedback/dataset/local.py
```diff
@@ -0,0 +1,103 @@
...
+            batch_size: the size of the batches to return. Defaults to 100.
+        """
+        for i in range(0, len(self.records), batch_size):
+            yield self.records[i : i + batch_size]
+
+    def fetch_records(self) -> None:
+        warnings.warn(
+            "As the current `FeedbackDataset` is stored locally and is pushed to Argilla,"
+            " the method `fetch_records` won't do anything. If you want to fetch the records"
+            " from Argilla, make sure you're using an `_ArgillaFeedbackDataset`, either by"
```
frascuchon: I think we can define this class as public `ArgillaFeedbackDataset`

src/argilla/client/feedback/dataset/mixins.py
```diff
@@ -0,0 +1,296 @@
...
+
+class ArgillaToFromMixin:
+    def push_to_argilla(
+        self: "FeedbackDataset",
+        name: Optional[str] = None,
+        workspace: Optional[Union[str, Workspace]] = None,
+        show_progress: bool = False,
+    ) -> _ArgillaFeedbackDataset:
+        """Pushes the `FeedbackDataset` to Argilla. If the dataset has been previously pushed to Argilla, it will be updated
+        with the new records.
```
frascuchon: The second part of the message is outdated, right?
```suggestion
        """Pushes the `FeedbackDataset` to Argilla. 
```

src/argilla/client/feedback/dataset/mixins.py
```diff
@@ -0,0 +1,296 @@
...
+
+        Note that you may need to `rg.init(...)` with your Argilla credentials before calling this function, otherwise
+        the default http://localhost:6900 will be used, which will fail if Argilla is not deployed locally.
+
+        Args:
+            name: the name of the dataset to push to Argilla. If not provided, the `argilla_id` will be used if the dataset
+                has been previously pushed to Argilla.
```
frascuchon:
```suggestion
            name: the name of the dataset to push to Argilla. 
```

src/argilla/client/feedback/dataset/mixins.py
```diff
@@ -0,0 +1,296 @@
...
+        self.__class__ = _ArgillaFeedbackDataset
+        self.__init__(
+            client=httpx_client,
+            id=argilla_id,
+            name=name,
+            workspace=workspace,
+            fields=fields,
+            questions=questions,
+            guidelines=self.guidelines,
+        )
```
frascuchon: This is for backward compatibility or is there any other reason?

Ideally, this could be removed in future releases, to allow publish the same dataset with different names/workspaces:
```python
for ws in ...:
   ds.push_to_argilla(name, ws)
```

src/argilla/client/feedback/dataset/mixins.py
```diff
@@ -0,0 +1,296 @@
...
+        return self
+
+    @classmethod
+    def from_argilla(
+        cls: Type["FeedbackDataset"],
+        name: Optional[str] = None,
+        *,
+        workspace: Optional[str] = None,
+        id: Optional[str] = None,
+        with_records: bool = True,  # TODO(alvarobartt): deprecate `with_records`
```
frascuchon: If we set to `None` as default, we can detect that a value is provided and show the warning properly.
```suggestion
        with_records: Optional[bool] = None,  # TODO(alvarobartt): deprecate `with_records`
```

src/argilla/client/feedback/dataset/mixins.py
```diff
@@ -0,0 +1,296 @@
...
+        Raises:
+            ValueError: if no `FeedbackDataset` with the provided `name` and `workspace` exists in Argilla.
+            ValueError: if no `FeedbackDataset` with the provided `id` exists in Argilla.
+
+        Examples:
+            >>> import argilla as rg
+            >>> rg.init(...)
+            >>> dataset = rg.FeedbackDataset.from_argilla(name="my_dataset")
+        """
+        if with_records:
```
frascuchon:
```suggestion
        if with_records is not None:
```

src/argilla/client/feedback/dataset/remote.py
```diff
@@ -0,0 +1,220 @@
...
+
+    from argilla.client.feedback.types import AllowedFieldTypes, AllowedQuestionTypes
+    from argilla.client.sdk.v1.datasets.models import FeedbackItemModel
+    from argilla.client.workspaces import Workspace
+
+
+warnings.simplefilter("always", DeprecationWarning)
+
+
+class _ArgillaFeedbackRecords:
```
frascuchon: I think we can define as a public class
```suggestion
class ArgillaFeedbackRecords:
```

src/argilla/client/feedback/dataset/base.py
```diff
@@ -298,19 +211,44 @@ def question_by_name(self, name: str) -> AllowedQuestionTypes:
...
+            new_records = []
+            for record in records:
+                if isinstance(record, dict):
+                    new_records.append(FeedbackRecord(**record))
+                elif isinstance(record, FeedbackRecord):
+                    new_records.append(record)
+                else:
+                    raise ValueError(
+                        f"Expected `records` to be a list of `dict` or `FeedbackRecord`, got type {type(record)} instead."
+                    )
```
gabrielmbmb: We could move this block to a function.

src/argilla/client/feedback/dataset/base.py
```diff
@@ -298,19 +211,44 @@ def question_by_name(self, name: str) -> AllowedQuestionTypes:
...
+            records = new_records
+        elif isinstance(records, dict):
+            records = [FeedbackRecord(**records)]
+        elif isinstance(records, FeedbackRecord):
+            records = [records]
+        else:
+            raise ValueError(
+                f"Expected `records` to be a `dict` or `FeedbackRecord`, got type {type(records)} instead."
```
gabrielmbmb:
```suggestion
                f"Expected `records` to be a `dict` or `FeedbackRecord`, got type `{type(records)}` instead."
```

src/argilla/client/feedback/dataset/base.py
```diff
@@ -298,19 +211,44 @@ def question_by_name(self, name: str) -> AllowedQuestionTypes:
...
-        """
-        for i in range(0, len(self.records), batch_size):
-            yield self.records[i : i + batch_size]
+        for record in records:
+            try:
+                self._fields_schema.parse_obj(record.fields)
+            except ValidationError as e:
+                raise ValueError(
+                    f"`FeedbackRecord.fields` does not match the expected schema, with exception: {e}"
+                ) from e
```
gabrielmbmb: We could move this block to a function.

src/argilla/client/feedback/dataset/local.py
gabrielmbmb: I agree, it will be used directly by the user.

src/argilla/client/feedback/dataset/mixins.py
```diff
@@ -0,0 +1,296 @@
...
+                "`push_to_argilla` will no longer be used to push changes to an existing"
+                " `FeedbackDataset` in Argilla, but to create a new one instead. So on,"
+                " please use `push_to_argilla` with the `name` argument and the `workspace`"
+                " if applicable. If you want to update an existing `FeedbackDataset` in"
+                " Argilla, you will need to either keep the returned dataset from the"
+                " `push_to_argilla` call and the functions will automatically call Argilla"
+                " or just call this function and then `from_argilla` to retrieve the"
+                " `FeedbackDataset` from Argilla.",
+                DeprecationWarning,
+            )
```
gabrielmbmb: when are we going to deprecate the current logic? `1.15.0`? once we know we should add it to this warning message

src/argilla/client/feedback/dataset/mixins.py
```diff
@@ -0,0 +1,296 @@
...
+
+        # TODO(alvarobartt): remove when `delete` is implemented
+        def delete_dataset(dataset_id: UUID) -> None:
+            try:
+                datasets_api_v1.delete_dataset(client=httpx_client, id=dataset_id)
+            except Exception as e:
+                raise Exception(
+                    f"Failed while deleting the `FeedbackDataset` with ID '{dataset_id}' from Argilla with"
+                    f" exception: {e}"
+                ) from e
```
gabrielmbmb: why we don't have this as a method already?

src/argilla/client/feedback/dataset/mixins.py
```diff
@@ -0,0 +1,296 @@
...
+                    client=httpx_client, id=argilla_id, field=field.dict(exclude={"id"})
+                ).parsed
+                field.id = new_field.id
+                fields.append(field)
+            except Exception as e:
+                delete_dataset(dataset_id=argilla_id)
+                raise Exception(
+                    f"Failed while adding the field '{field.name}' to the `FeedbackDataset` in Argilla with"
+                    f" exception: {e}"
+                ) from e
```
gabrielmbmb: We could move this block to a function.

src/argilla/client/feedback/dataset/mixins.py
```diff
@@ -0,0 +1,296 @@
...
+                ).parsed
+                question.id = new_question.id
+                questions.append(question)
+                question_name2id[new_question.name] = new_question.id
+            except Exception as e:
+                delete_dataset(dataset_id=argilla_id)
+                raise Exception(
+                    f"Failed while adding the question '{question.name}' to the `FeedbackDataset` in Argilla"
+                    f" with exception: {e}"
+                ) from e
```
gabrielmbmb: we could move this block to a function.

src/argilla/client/feedback/dataset/mixins.py
```diff
@@ -0,0 +1,296 @@
...
+                datasets_api_v1.add_records(
+                    client=httpx_client,
+                    id=argilla_id,
+                    records=records,
+                )
+            except Exception as e:
+                delete_dataset(dataset_id=argilla_id)
+                raise Exception(
+                    f"Failed while adding the records to the `FeedbackDataset` in Argilla with exception: {e}"
+                ) from e
```
gabrielmbmb: we could move this block to a function.

src/argilla/client/feedback/dataset/mixins.py
```diff
@@ -0,0 +1,296 @@
...
+        for field in datasets_api_v1.get_fields(client=httpx_client, id=existing_dataset.id).parsed:
+            base_field = field.dict(include={"id", "name", "title", "required"})
+            if field.settings["type"] == "text":
+                field = TextField(**base_field, use_markdown=field.settings["use_markdown"])
+            else:
+                raise ValueError(
+                    f"Field '{field.name}' is not a supported field in the current Python package version,"
+                    " supported field types are: `TextField`."
+                )
+            fields.append(field)
```
gabrielmbmb: This block could be a function.

src/argilla/client/feedback/dataset/mixins.py
```diff
@@ -0,0 +1,296 @@
...
+                        **question_dict, labels=labels, visible_labels=question.settings["visible_options"]
+                    )
+                elif question.settings["type"] == "ranking":
+                    question = RankingQuestion(**question_dict, values=labels)
+            else:
+                raise ValueError(
+                    f"Question '{question.name}' is not a supported question in the current Python package"
+                    f" version, supported question types are: `{'`, `'.join([arg.__name__ for arg in AllowedQuestionTypes.__args__])}`."
+                )
+            questions.append(question)
```
gabrielmbmb: This block could be a function.

src/argilla/client/feedback/dataset/remote.py
gabrielmbmb: I think it is fine as it is. The user will not use this class directly, but through the `ArgillaFeedbackDataset`, right?

src/argilla/client/feedback/dataset/local.py
alvarobartt: Hm but we don't want users to actually use it, we want to use it just internally, right?

src/argilla/client/feedback/dataset/mixins.py
alvarobartt: Yes, it's for backward compatibility, as otherwise, users would face a breaking change and would need to either capture the output or call `FeedbackDataset.from_argilla` right after the push

src/argilla/client/feedback/dataset/mixins.py
alvarobartt: Yes, docstrings and messages have not been revisited yet, just the functionality

src/argilla/client/feedback/dataset/base.py
```diff
@@ -371,14 +310,16 @@ def unify_responses(
 
         strategy.unify_responses(self.records, question)
 
+    # TODO(alvarobartt,davidberenstein1957): we should consider having something like
+    # `export(..., training=True)` to export the dataset records in any format, replacing
+    # both `format_as` and `prepare_for_training`
     def prepare_for_training(
```
frascuchon, line 306: Does make sense to have this functionality linked to local datasets? If the normal workflow is  1. download what you want from argilla locally and 2. train your model, maybe makes sense to prepare the training data only from a local version of the feedback dataset.

src/argilla/client/feedback/dataset/base.py
```diff
@@ -336,6 +274,7 @@ def format_as(self, format: Literal["datasets"]) -> "Dataset":
             return self._huggingface_format(self)
         raise ValueError(f"Unsupported format '{format}'.")
 
+    # TODO(davidberenstein1957): detatch unification into a mixin
```
frascuchon, line 267: Same here, maybe this makes sense only for local copies of the datasets.

src/argilla/client/feedback/dataset/base.py
```diff
@@ -66,63 +65,8 @@ class FeedbackDataset(ArgillaDatasetMixin, HuggingFaceDatasetMixin):
...
-        >>> dataset.push_to_argilla(name="my-dataset", workspace="my-workspace")
-        >>> dataset.argilla_id
-        "..."
-        >>> dataset = rg.FeedbackDataset.from_argilla(argilla_id="...")
-        >>> dataset.records
-        [FeedbackRecord(fields={"text": "This is the first record", "label": "positive"}, responses=[ResponseSchema(user_id=None, values={"question-1": ValueSchema(value="This is the first answer"), "question-2": ValueSchema(value=5), "question-3": ValueSchema(value="positive"), "question-4": ValueSchema(value=["category-1"])})], external_id="entry-1")]
     """
 
-    argilla_id: Optional[UUID] = None
-
```
frascuchon: If a user is using the property, this change may break the code. Maybe just a property with 2 different implementations?

```python
@property
def argilla_id() -> UUID:
   raise NotImplementedError()

# for remote dataset:
def argilla_id() -> UUID:
   warnings.warning("use self.id instead...")
   return self.id

# for local dataset
def argilla_id() -> UUID:
   warnings.warning("local datases don't track remote argilla info...")
   return None
```

src/argilla/client/feedback/dataset/base.py
```diff
@@ -298,19 +211,44 @@ def question_by_name(self, name: str) -> AllowedQuestionTypes:
...
+            records = [records]
+        else:
+            raise ValueError(
+                f"Expected `records` to be a `dict` or `FeedbackRecord`, got type {type(records)} instead."
+            )
 
-    def iter(self, batch_size: Optional[int] = FETCHING_BATCH_SIZE) -> Iterator[List[FeedbackRecord]]:
-        """Returns an iterator over the records in the dataset.
+        if self._fields_schema is None:
+            self._fields_schema = generate_pydantic_schema(self.fields)
```
frascuchon, line 224: Does it mean that once the pydantic schema is generated the field_schema is not changed after a field setting update? Maybe we can just generate each time the method is called:
```suggestion
        fields_schema  = generate_pydantic_schema(self.fields)
```
Ow, are we using the self._fields_schema outside this method?

src/argilla/client/feedback/dataset/base.py
```diff
@@ -298,19 +211,44 @@ def question_by_name(self, name: str) -> AllowedQuestionTypes:
...
 
-        Args:
-            batch_size: the size of the batches to return. Defaults to 100.
-        """
-        for i in range(0, len(self.records), batch_size):
-            yield self.records[i : i + batch_size]
+        for record in records:
+            try:
+                self._fields_schema.parse_obj(record.fields)
```
frascuchon, line 228:
```suggestion
                fields_schema.parse_obj(record.fields)
```

src/argilla/client/feedback/dataset/local.py
alvarobartt: It has been renamed to `RemoteFeedbackDataset` as internally suggested by @frascuchon

src/argilla/client/feedback/dataset/remote.py
alvarobartt: It has been renamed to `RemoteFeedbackDataset` as internally suggested by @frascuchon

src/argilla/client/feedback/dataset/base.py
alvarobartt: I've ensured that it's backwards compatible with the current approach, so that the records are automatically pulled from Argilla if we want to use a `FeedbackDataset` in Argilla i.e. `RemoteFeedbackDataset`

src/argilla/client/feedback/dataset/base.py
alvarobartt: Done, thanks!

src/argilla/client/feedback/dataset/base.py
alvarobartt: No we're not, but ideally we want to re-use it, instead of calculating it every time, and since we don't support field settings updates for the moment I'd say we can leave it as is

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ These are the section headers that we use:
 
 ### Added
 
+- Added `get_metrics` function in the SDK to get `FeedbackDataset` metrics calling `GET /api/v1/me/{dataset_id}/metrics` endpoint ([#3469](https://github.com/argilla-io/argilla/pull/3469))
```
frascuchon: I think we can remove this line since is too techie detail for a changelog 
```suggestion
```

By the way, we can add other in the changed and deprecated sections, referring the new `add_records` behaviour. or the push_to_argilla and other deprecated methods

alvarobartt: Thanks, already done!

src/argilla/client/feedback/dataset/remote.py
```diff
@@ -0,0 +1,247 @@
...
+        self.__question_id2name = {question.id: question.name for question in questions}
+        self.__question_name2id = {value: key for key, value in self.__question_id2name.items()}
+
+    def __repr__(self) -> str:
+        return (
+            "The `records` of a `FeedbackDataset` in Argilla are being lazily"
+            " fetched, and never stored locally. You can either loop over `records`"
+            " or access them by index, and those will be fetched from Argilla on the"
+            " fly."
+        )
```
gabrielmbmb: I think it's a bit weird that if you print `ArgillaFeedbackDataset.records` you get this, it's more like a warning message. This method could return something like:

`[RemoteFeedbackRecord(...), RemoteFeedbackRecord(...), ...] lazily fetched from 'https://dev.argilla.io'`

src/argilla/client/feedback/dataset/remote.py
```diff
@@ -0,0 +1,247 @@
...
+                id=self.id,
+                offset=FETCHING_BATCH_SIZE * current_batch,
+                limit=FETCHING_BATCH_SIZE,
+            ).parsed
+            for record in batch.items:
+                yield self.__parse_record(record)
+            current_batch += 1
+
+            if len(batch.items) < FETCHING_BATCH_SIZE:
+                break
```
gabrielmbmb, line 169: we can move this before the for loop

src/argilla/client/feedback/dataset/remote.py
```diff
@@ -0,0 +1,247 @@
...
+            client=self.client, id=self.id, questions=self.questions
+        )
+
+    @property
+    def argilla_id(self) -> str:
+        warnings.warn(
+            "`argilla_id` is deprected in favor of `id` and will be removed in a future"
+            " release. Please use `id` instead.",
+            DeprecationWarning,
+        )
```
gabrielmbmb: add version in which it will be deprecated once we know

src/argilla/client/feedback/dataset/remote.py
```diff
@@ -0,0 +1,247 @@
...
+    @property
+    def name(self) -> str:
+        return self._name
+
+    @property
+    def workspace(self) -> "Workspace":
+        return self._workspace
+
+    def __repr__(self) -> str:
+        return f"<ArgillaFeedbackDataset id={self.id} name={self.name} workspace={self.workspace}>"
```
gabrielmbmb: maybe it would be good to add the URL of the Argilla server

src/argilla/client/feedback/dataset/remote.py
```diff
@@ -0,0 +1,247 @@
...
+
+    def __len__(self) -> int:
+        return self.records.__len__()
+
+    def __iter__(self) -> Iterator[RemoteFeedbackRecord]:
+        return self.records.__iter__()
```
gabrielmbmb:
```suggestion
        yield from self.records
```

src/argilla/client/feedback/dataset/remote.py
```diff
@@ -0,0 +1,247 @@
...
+            DeprecationWarning,
+        )
+
+    def push_to_argilla(self, *args, **kwargs) -> None:
+        warnings.warn(
+            "`push_to_argilla` is no longer working for a `FeedbackDataset` pushed to Argilla,"
+            " as the additions, deletions and/or updates over a `FeedbackDataset` in Argilla"
+            " are being tracked automatically, so there's no need to explicitly push them.",
+            DeprecationWarning,
+        )
```
gabrielmbmb: add version in which this will be deprecated once we know

src/argilla/client/feedback/schemas/records.py
```diff
@@ -204,13 +206,48 @@ def set_suggestions(
...
-    def _reset_updated(self) -> None:
-        self._updated = False
+    def set_suggestions(
+        self, suggestions: Union[SuggestionSchema, List[SuggestionSchema], Dict[str, Any], List[Dict[str, Any]]]
+    ) -> None:
+        warnings.warn(
+            "`set_suggestions` is deprected in favor of `update` and will be removed in a future"
+            " release. Please use `update` instead.",
+            DeprecationWarning,
+        )
```
gabrielmbmb: add version in which this will get deprecated once we know

src/argilla/client/feedback/schemas/records.py
```diff
@@ -204,13 +206,48 @@ def set_suggestions(
...
+            )
+
+    def set_suggestions(
+        self, suggestions: Union[SuggestionSchema, List[SuggestionSchema], Dict[str, Any], List[Dict[str, Any]]]
+    ) -> None:
+        warnings.warn(
+            "`set_suggestions` is deprected in favor of `update` and will be removed in a future"
+            " release. Please use `update` instead.",
+            DeprecationWarning,
+        )
```
gabrielmbmb: add version in which this will get deprecated once we know

gabrielmbmb (approved):
Hey @alvarobartt, LGTM :) I have left you some comments. Also, if you could add the `stacklevel=1` to the `warnings.warn` calls, it would be good.

src/argilla/client/feedback/dataset/mixins.py
alvarobartt: I guess this was not discussed that's why I didn't specify a version, but yes, ideally we should specify the version 👍🏻

src/argilla/client/feedback/dataset/mixins.py
alvarobartt: It's on the works, PR coming later today 👍🏻

src/argilla/client/feedback/dataset/base.py
```diff
@@ -230,45 +175,14 @@ def __init__(
...
-        if len(self.records) < 1:
-            raise RuntimeError(
-                "In order to get items from `rg.FeedbackDataset` you need to either add"
-                " them first with `add_records` or fetch them from Argilla or"
-                " HuggingFace with `fetch_records`."
-            )
-        if isinstance(key, int) and len(self.records) < key:
-            raise IndexError(f"This dataset contains {len(self)} records, so index {key} is out of range.")
-        return self.records[key]
+        self.records: List[FeedbackRecord] = []
```
frascuchon: Should not we move this to the `FeedbackDataset` class? The `ArgillaFeedbackDataset` does not need a local list of records right?

Maybe the base class can just define an abstract property `records`, letting each implementation resolver in the proper way

```suggestion
   @property
   def records(self) -> Iterable[FeedbackRecord]:
      raise NotImplementedError()
```

src/argilla/client/feedback/dataset/base.py
frascuchon: 👍  In any case, I would say to warn users if the `prepare_for_training` method is done using the `RemoteArgillaDataset`. Since this will change in the future, and all the training stuff will be applied using local copies of datasets.

src/argilla/client/feedback/dataset/local.py
```diff
@@ -0,0 +1,113 @@
...
+    def argilla_id(self) -> None:
+        warnings.warn(
+            "As the current `FeedbackDataset` is stored locally, `argilla_id` won't"
+            " return anything as it's not pushed to Argilla. If you want to get the id"
+            " of a dataset in Argilla, make sure you're using an `RemoteFeedbackDataset`,"
+            " either by calling `FeedbackDataset.from_argilla` or by keeping the returned"
+            " value from `FeedbackDataset.push_to_argilla`.",
+        )
+
+    def add_records(
```
frascuchon, line 158: A question here. For local datasets, we could do something like:
```python
for data in ...:
   dataset.records.append(FeedbackRecord(**data))
``` 
to add records, right? I mean, it's possible right now, right? Maybe we should define the records as a property based on a protected attribute `__records` 


```python
@property
def records(self) -> Iterable[FeedbackRecord]:
   return (r for r in self.__records)
```

src/argilla/client/feedback/dataset/base.py
```diff
@@ -298,19 +212,44 @@ def question_by_name(self, name: str) -> AllowedQuestionTypes:
             f" {', '.join(q.name for q in self._questions)}"
         )
 
-    @property
-    def records(self) -> List[FeedbackRecord]:
-        """Returns the all the records in the dataset."""
-        return self._records + self._new_records
+    def _validate_records(
```
frascuchon: Not for this PR but, I think it make sense to validate records also based on dataset questions. Maybe records contain incompatible responses for configured dataset questions

src/argilla/client/feedback/dataset/local.py
alvarobartt: Sure, I can make `records` a property

src/argilla/client/feedback/dataset/base.py
alvarobartt: Ok makes sense to add validations for both responses and suggestions, I'll do so in an upcoming PR 😄

src/argilla/client/feedback/dataset/base.py
alvarobartt: It was defined that way because we're defining the `unify_responses` and `prepare_for_training` methods in the `FeedbackDatasetBase`, but I guess we can just define a property, but note that the property itself it's not an iterable when working with local datasets, as it's just a list, but I'll figure that out 👍🏻

src/argilla/client/feedback/dataset/remote.py
```diff
@@ -0,0 +1,249 @@
...
+
+    @property
+    def argilla_id(self) -> str:
+        warnings.warn(
+            "`argilla_id` is deprected in favor of `id` and will be removed in a future"
+            " release. Please use `id` instead.",
+            DeprecationWarning,
+        )
+        return self.id
```
frascuchon, line 267:
```suggestion
        return str(self.id)
```

src/argilla/client/feedback/dataset/remote.py
frascuchon: Yes, I would say that having `dataset.url` property  could be great

src/argilla/client/feedback/dataset/remote.py
alvarobartt: Why not return the `UUID`?

src/argilla/client/feedback/dataset/remote.py
gabrielmbmb: Also, it should be '<RemoteFeedbackDataset ...>`, right? to be aligned with the class name

src/argilla/client/feedback/dataset/remote.py
```diff
@@ -0,0 +1,257 @@
...
+    @property
+    def workspace(self) -> "Workspace":
+        return self._workspace
+
+    @property
+    def url(self) -> str:
+        return f"{self.client.base_url}/dataset/{self.id}"
+
+    def __repr__(self) -> str:
+        return f"<ArgillaFeedbackDataset id={self.id} name={self.name} workspace={self.workspace}>"
```
gabrielmbmb:
```suggestion
        return f"<FeedbackDataset id={self.id} name={self.name} workspace={self.workspace} url={self.url}>"
```

CHANGELOG.md
```diff
@@ -40,6 +40,13 @@ These are the section headers that we use:
 - Updated `GET /api/v1/me/datasets/{dataset_id}/records` endpoint to allow getting records matching one of the response statuses provided via query param.([#3359](https://github.com/argilla-io/argilla/pull/3359)).
 - Updated `POST /api/v1/me/datasets/{dataset_id}/records` endpoint to allow searching records matching one of the response statuses provided via query param.([#3359](https://github.com/argilla-io/argilla/pull/3359)).
 - Updated `SearchEngine.search` method to allow searching records matching one of the response statuses provided ([#3359](https://github.com/argilla-io/argilla/pull/3359)).
+- After calling `FeedbackDataset.push_to_argilla`, the methods `FeedbackDataset.add_records` and `FeedbackRecord.set_suggestions` will automatically call Argilla with no need of calling `push_to_argilla` explicitly ([#3465](https://github.com/argilla-io/argilla/pull/3465)).
+
+### Deprecated
```
frascuchon, line 46: does make sense also add the `pull` method to fetch a local copy of a remote Argilla dataset?

src/argilla/client/feedback/dataset/base.py
```diff
@@ -230,45 +176,18 @@ def __init__(
...
-                "In order to get items from `rg.FeedbackDataset` you need to either add"
-                " them first with `add_records` or fetch them from Argilla or"
-                " HuggingFace with `fetch_records`."
-            )
-        if isinstance(key, int) and len(self.records) < key:
-            raise IndexError(f"This dataset contains {len(self)} records, so index {key} is out of range.")
-        return self.records[key]
+    @property
+    @abstractproperty
+    def records(self) -> Any:
```
frascuchon, line 128: Could be expected an `Iterable[FeedbackRecord]` here?

alvarobartt: No because `records` is not an iterable, it's a `List[FeedbackRecord]` in `FeedbackDataset` while that's an instance of `RemoteFeedbackRecords` in `RemoteFeedbackDataset`

CHANGELOG.md
alvarobartt: Yes, adding it now!

src/argilla/client/feedback/dataset/remote.py
```diff
@@ -0,0 +1,257 @@
...
+    from argilla.client.workspaces import Workspace
+
+
+warnings.simplefilter("always", DeprecationWarning)
+
+
+class RemoteFeedbackRecords:
+    def __init__(self, client: "httpx.Client", id: "UUID", questions: List["AllowedQuestionTypes"]) -> None:
+        self.client = client
+        self.id = id
```
frascuchon: If the id refers to the dataset id, I would say to explicitly call `dataset_id`. Otherwise can be hard to know which id is referring to.

```suggestion
    def __init__(self, client: "httpx.Client", dataset_id: "UUID", questions: List["AllowedQuestionTypes"]) -> None:
        self.client = client
        self.dataset_id = dataset_id
```

src/argilla/client/feedback/dataset/remote.py
frascuchon: Yes, I agree with @gabrielmbmb. Maybe we can just print some of the records:
```suggestion
        return f"{self.records[:3]!r}..."
```

frascuchon: Even if I'm not much comfortable making remote calls in `__repr__` method. But can be useful to users.

src/argilla/client/feedback/dataset/remote.py
frascuchon: Do we need the `self.id` to be public?

src/argilla/client/feedback/dataset/remote.py
```diff
@@ -0,0 +1,257 @@
...
+                client=self.client,
+                id=self.id,
+                offset=offset,
+                limit=limit,
+            ).parsed
+            if len(fetched_records.items) == 1:
+                record = fetched_records.items[0]
+                records.append(self.__parse_record(record))
+            else:
+                records.extend([self.__parse_record(record) for record in fetched_records.items])
```
frascuchon: Why we cannot do something like:
```suggestion
            records.extend([self.__parse_record(record) for record in fetched_records.items])
```

src/argilla/client/feedback/dataset/remote.py
```diff
@@ -0,0 +1,257 @@
...
+            ) from e
+        return response.parsed.records.count
+
+    def __getitem__(self, key: Union[slice, int]) -> Union[RemoteFeedbackRecord, List[RemoteFeedbackRecord]]:
+        offsets = []
+        limit = None
+        if isinstance(key, slice):
+            start, stop, step = key.indices(len(self))
+            if step is not None and step != 1:
+                return [self[i] for i in range(start, stop, step)]
```
frascuchon: This case will call several times the API to fetch records one-by-one right? I would say start by not supporting slices with `step` != 1 for remote datasets.

alvarobartt: But what if that's what the user wants? At the end we want to expose the records so that the user can use it as a normal list but inside it's fetching from Argilla, so I think we can leave it as is, at the end it depends on the user to avoid doing that

src/argilla/client/feedback/dataset/remote.py
alvarobartt: Thanks makes sense!

src/argilla/client/feedback/dataset/remote.py
alvarobartt: Nope, making both `client` and `id` private now!

src/argilla/client/feedback/dataset/remote.py
```diff
@@ -0,0 +1,257 @@
...
+                stop += len(self)
+            if start < 0 or stop < 0:
+                raise IndexError("Index out of range")
+            limit = stop - start
+            offsets = [start] if limit < FETCHING_BATCH_SIZE else list(range(start, stop, FETCHING_BATCH_SIZE))
+        elif isinstance(key, int):
+            if key < 0:
+                key += len(self)
+            if key < 0 or key >= len(self):
+                raise IndexError("Index out of range")
```
frascuchon: Just to avoid several calls to compute the lenght
```suggestion
        offsets = []
        limit = None
        records_lenght= len(self)
        if isinstance(key, slice):
            start, stop, step = key.indices()
            if step is not None and step != 1:
                return [self[i] for i in range(start, stop, step)]
            if start < 0:
                start += records_lenght
            if stop < 0:
                stop += records_lenght
            if start < 0 or stop < 0:
                raise IndexError("Index out of range")
            limit = stop - start
            offsets = [start] if limit < FETCHING_BATCH_SIZE else list(range(start, stop, FETCHING_BATCH_SIZE))
        elif isinstance(key, int):
            if key < 0:
                key += records_lenght
            if key < 0 or key >= records_lenght:
                raise IndexError("Index out of range")
```

src/argilla/client/feedback/dataset/remote.py
alvarobartt: I agree, renamed that and made it private already!

src/argilla/client/feedback/dataset/remote.py
alvarobartt: Thanks makes sense!

src/argilla/client/feedback/dataset/remote.py
alvarobartt: I think this is not optimal and can also be confusing 😩 Maybe we can just go with @gabrielmbmb's suggestion to show something like `[RemoteFeedbackRecord(...), RemoteFeedbackRecord(...), ...] lazily fetched from 'https://dev.argilla.io'`?

alvarobartt: Or maybe just return `None` and a `warnings.warn` instead

src/argilla/client/feedback/dataset/remote.py
```diff
@@ -0,0 +1,257 @@
...
+
+    @property
+    def url(self) -> str:
+        return f"{self.client.base_url}/dataset/{self.id}"
+
+    def __repr__(self) -> str:
+        return f"<ArgillaFeedbackDataset id={self.id} name={self.name} workspace={self.workspace}>"
+
+    def __len__(self) -> int:
+        return self._records.__len__()
```
frascuchon, line 299: Is this the same as `len(self._records)`?

alvarobartt: Yes, but calling directly the magic method instead

src/argilla/client/feedback/dataset/base.py
```diff
@@ -298,19 +178,64 @@ def question_by_name(self, name: str) -> AllowedQuestionTypes:
...
+            records = new_records
+        elif isinstance(records, dict):
+            records = [FeedbackRecord(**records)]
+        elif isinstance(records, FeedbackRecord):
+            records = [records]
+        else:
+            raise ValueError(
+                "Expected `records` to be a `dict` or `FeedbackRecord`, got type" f" `{type(records)}` instead."
+            )
+        return records
```
frascuchon: I think we can simplify this method, and avoid parsing logic duplication.
```suggestion
        if isinstance(records, (dict, FeedbackRecord)):
            records = [records]

        if len(records) == 0:
            raise ValueError("Expected `records` to be a non-empty list of `dict` or `FeedbackRecord`.")
            
        new_records = []
        for record in records:
            if isinstance(record, dict):
                new_records.append(FeedbackRecord(**record))
            elif isinstance(record, FeedbackRecord):
                new_records.append(record)
            else:
                raise ValueError(
                    "Expected `record` to be a list of `dict` or `FeedbackRecord`,"
                    f" got type `{type(record)}` instead."
                )
        return new_records
```

src/argilla/client/feedback/dataset/remote.py
```diff
@@ -0,0 +1,368 @@
...
+    def __repr__(self) -> str:
+        """Doesn't return anything, but prints a warning, since the `records` of a
+        `FeedbackDataset` in Argilla are being lazily fetched, and never stored
+        locally."""
+        warnings.warn(
+            "The `records` of a `FeedbackDataset` in Argilla are being lazily"
+            " fetched, and never stored locally. You can either loop over `records`"
+            " or access them by index, and those will be fetched from Argilla on the"
+            " fly.",
+        )
```
gabrielmbmb:
```suggestion
        warnings.warn(
            "The `records` of a `FeedbackDataset` in Argilla are being lazily"
            " fetched, and never stored locally. You can either loop over `records`"
            " or access them by index, and those will be fetched from Argilla on the"
            " fly.",
            stacklevel=1
        )
```

gabrielmbmb (approved):

PR #3467, opened 2023-07-27
feat: Allow exporting data for SFT, Reward Modelling (related to RLHF), DPO, rename TrainingTaskMapping

### LINKED ISSUE #3377: [FEATURE] add `lvwerra/trl` `reward modelling` support to the `ArgillaTrainer`
I cannot do RewardModelling with the `RankingQuestion`-

We should add support for https://github.com/lvwerra/trl. 
This support should include 
- a `.for_reward_modelling()` class-method for the `TrainingTaskMapping`. 
- a top-n(=2) unification support for the `RankingQuestionStrategy`
- support for the `prepare_for_training()` method of the `FeedbackDataset`
- support for the `FrameWork(Enum)` via `trl`
- alignment with our [docs](https://docs.argilla.io/en/latest/guides/llms/examples/train-reward-model-rlhf.html) and usage of `chosen-rejected`

consider taking `carperAI/trlx` into account too. https://github.com/argilla-io/argilla/issues/3324

**Additional context**
N.A.

### LINKED ISSUE #3379: [FEATURE] add `lvwerra/trl` `SFT` support to `ArgillaTrainer`
I cannot do RewardModelling with the `TextField` and `TextQuestion`.

We should add support for https://github.com/lvwerra/trl. 
This support should include 
- a `.for_supervised_fine_tuning()` class method for the `TrainingTaskMapping`. 
- a method to support `TextField` and `TextQuestion` function for [SFTTrainer](https://github.com/lvwerra/trl#sfttrainer)
- support for the `prepare_for_training()` method of the `FeedbackDataset`
- support for the `FrameWork(Enum)` via `trl`
- alignment with our [docs](https://docs.argilla.io/en/latest/guides/llms/practical_guides/fine_tune.html)

consider taking `CarperAI/trlx` into account too. https://github.com/argilla-io/argilla/issues/3378

**Additional context**
N.A.

### LINKED ISSUE #3522: [FEATURE] add `PPOTrainer` to `trl` integration `ArgillaTrainer`
We missed the `PPOTrainer` within our first integration leap and it would be best to add this too. 
https://github.com/lvwerra/trl#ppotrainer

image

```python
task = TrainingTask.for_reward_modelling(...)
trainer = ArgillaTrainer(
   dataset=fds_dataset,
   task=task,
   framework="trl",
)
trainer.train()
trainer.save("reward_model")

# And then you can use this "reward_model" with PPO
task = TrainingTask.for_proximal_policy_optimization(...)
trainer = ArgillaTrainer(
   dataset=fds_dataset,
   task=task,
   framework="trl",
)
trainer.train(model=reward_model, generation_args)
```

N.A.

**Additional context**
N.A.

tomaarsen:
Resolves #3379, resolves #3377
Closes #3522 

Hello!

## Pull Request overview
* Prepare data for SFT, RM, DPO in TRL.
* Rename `TrainingTaskMapping` to `TrainingTask` and `task_mapping` to `task`.

# Description
## Prepare data
```python
from argilla.feedback import TrainingTask

def formatting_func(sample: Dict[str, Any]):
    ...
    yield template.format(
        prompt=sample["prompt"],
        response=sample["response"],
    )

task = TrainingTask.for_supervised_fine_tuning(formatting_func=formatting_func)
ds = fds_dataset.prepare_for_training(framework="trl", task=task)
# -> ds has "text" and "id" columns
```
Compatible with [SFTTrainer](https://huggingface.co/docs/trl/main/en/sft_trainer).

```python
task = TrainingTask.for_reward_modelling(chosen_rejected_func=chosen_rejected_func)
ds = fds_dataset.prepare_for_training(framework="trl", task=task)
# -> ds has "chosen" and "rejected" columns
```
Nearly compatible with [RewardTrainer](https://huggingface.co/docs/trl/main/en/reward_trainer).

```python
task = TrainingTask.for_direct_preference_optimization(prompt_chosen_rejected_func=prompt_chosen_rejected_func)
ds = fds_dataset.prepare_for_training(framework="trl", task=task)
# -> ds has "prompt", "chosen" and "rejected" columns
```
Compatible with [DPOTrainer](https://huggingface.co/docs/trl/main/en/dpo_trainer).

### Details
I implement this by calling `dataset.format_as("datasets")` and then passing each sample (a simple dictionary) from this dataset to the function that the user provides. This user provided function can return `None`, one sample, a list of samples, or yield samples. This allows users to export multiple training samples from a single Argilla record, e.g. when there's multiple annotators that provided useful corrections, or if the annotated record justifies 3 "chosen", "rejected" pairs because there's a ranking between 3 texts.

## Rename
`TrainingTaskMapping` is now `TrainingTask` - the "mapping" part is just unintuitive to the user. Same for `task_mapping` to `task`. **Note:** If people used `task_mapping=...` before, that will now fail. I can make this deprecation softer, but then I have to make `task` optional, which I would rather not do.

## TODO:
- [ ] Add TRL to `ArgillaTrainer`, allowing:
   ```python
   task = TrainingTask.for_supervised_fine_tuning(
       formatting_func=formatting_func
   ) # or any other task from this PR
   trainer = ArgillaTrainer(
       dataset=fds_dataset,
       task=task,
       framework="trl",
   )
   trainer.train()
   ```
- [ ] Consider renaming `FeedbackDataset.prepare_for_training` to `FeedbackDataset.export`.
- [ ] New tests
- [ ] Add documentation

**Type of change**

- [x] New feature

**How Has This Been Tested**

Not finished yet.

**Checklist**

---

- Tom Aarsen


src/argilla/client/feedback/integrations/huggingface/dataset.py
```diff
@@ -112,12 +112,17 @@ def _huggingface_format(dataset: "FeedbackDataset") -> "Dataset":
...
+            "value": [],
+            "status": [],
+        }
         for record in dataset.records:
             for field in dataset.fields:
                 hf_dataset[field.name].append(record.fields[field.name])
             for question in dataset.questions:
                 if not record.responses:
-                    hf_dataset[question.name].append(None)
+                    hf_dataset[question.name].append(empty)
```
tomaarsen: @alvarobartt Thoughts on this?

Otherwise users need None checks everywhere, and now they can just iterate over all elements in `sample["value"]`.

By the way, this is relevant for the TRL tasks. See for example the tests in this PR.

tomaarsen: Additionally, the current approach is "columnar", e.g.:
```python
{
      "user_id": [..., ..., ...],
      "value": [4, 3, 1],
      "status": ["submitted", "submitted", "discarded"],
}
```
But have we considered a row-based approach?
```python
[{"user_id": ..., "value": 4, "status": "submitted"},
{"user_id": ..., "value": 3, "status": "submitted"},
{"user_id": ..., "value": 1, "status": "discarded"}]
```
With the latter setup, it is much easier to select all values with the right status:
```python
[annotation["value"] for annotation in X if annotation["status"] == "submitted"]
```
where X is the list.

Right now, you have to do something like:
```python
[value for value, status in zip(X["value"], X["status"]) if status == "submitted"]
```

Thoughts?

tomaarsen:
The CI is failing with this message. I've seen this before I think, is it expected?
image

Perhaps you've seen it before @gabrielmbmb?

gabrielmbmb:
@tomaarsen it's a bug in the workflow... The Python package is only being built when detected changes in some directories like `src/**`. In some cases, the step for building the Python package is not executed because there are no changes in the files of these directories but the step for building the Docker image is executed and it needs the generated artifact of the build python package step.

I'll create a PR to add an additional condition for skipping the build of the Docker image if the Python package hasn't been built

davidberenstein1957:
@tomaarsen https://github.com/argilla-io/argilla/issues/3522

tomaarsen:
Updating docs:

- [x] `docs/_source/guides/llms/practical_guides/fine_tune.md`
- [x] `docs/_source/reference/python/python_training.rst`
- [x] `docs/_source/guides/llms/examples/train-reward-model-rlhf.ipynb`
- [x] `docs/_source/_common/tabs/train_prepare_for_training.md`
- [x] `docs/_source/_common/tabs/train_update_config.md`
- [ ] ~~`docs/_source/_common/tabs/unfication_strategies.md`~~
- [ ] ~~`docs/_source/_common/snippets/training/feedback-task`~~

This PR is ready for review, although the docs are not all updated yet.

- Tom

tomaarsen:
We might have to already remove `fetch_records` internally as it might throw DeprecationWarnings that can't be removed (without disabling all DeprecationWarnings).

docs/_source/_common/snippets/training/feedback-task/text-classification.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: Text classification
-description: When a RatingQuestion, LabelQuestion or MultiLabelQuestion is present in the datasets, we can define a TrainingTaskMappingForTextClassification to use our ArgillaTrainer integration for fine-tuning with openai”, “setfit”, “peft”, “spacy” and “transformers”.
+description: When a RatingQuestion, LabelQuestion or MultiLabelQuestion is present in the datasets, we can define a TrainingTaskForTextClassification to use our ArgillaTrainer integration for fine-tuning with openai”, “setfit”, “peft”, “spacy” and “transformers”.
```
alvarobartt: Replaced Windows double-quotes with standard double-quotes, and add missing one in `openai`

```suggestion
description: When a RatingQuestion, LabelQuestion or MultiLabelQuestion is present in the datasets, we can define a TrainingTaskForTextClassification to use our ArgillaTrainer integration for fine-tuning with "openai", "setfit", "peft", "spacy" and "transformers".
```

src/argilla/client/feedback/dataset/base.py
```diff
@@ -306,7 +311,12 @@ def unify_responses(
     def prepare_for_training(
         self,
         framework: Union[Framework, str],
-        task_mapping: TrainingTaskMappingForTextClassification,
+        task: Union[
+            TrainingTaskForTextClassification,
+            TrainingTaskForSupervisedFinetuning,
+            TrainingTaskForRewardModelling,
```
alvarobartt: As the user may directly interact with e.g. `TrainingTask.for_text_classification`, shouldn't we shorten these names?

e.g. `TrainingTaskForDirectPreferenceOptimization` -> `TrainingTaskForDPO`

src/argilla/client/feedback/dataset/base.py
```diff
@@ -306,7 +311,12 @@ def unify_responses(
...
+            TrainingTaskForDirectPreferenceOptimization,
+        ],
         train_size: Optional[float] = 1,
```
alvarobartt, line 317: I thought the standard was to define the `Optional` on fields that are `None` by default, as Python will mark those as optional internally if those have a predefined value as `float = 1`, so I would change this to

```suggestion
        train_size: float = 1.0,
```

src/argilla/client/feedback/dataset/base.py
```diff
@@ -376,11 +393,15 @@ def prepare_for_training(
                     lang = spacy.blank(lang)
                 else:
                     lang = spacy.load(lang)
-            return task_mapping._prepare_for_training_with_spacy(data=data, train_size=train_size, seed=seed, lang=lang)
+            return task._prepare_for_training_with_spacy(data=data, train_size=train_size, seed=seed, lang=lang)
         elif framework is Framework.SPARK_NLP:
-            return task_mapping._prepare_for_training_with_spark_nlp(data=data, train_size=train_size, seed=seed)
+            return task._prepare_for_training_with_spark_nlp(data=data, train_size=train_size, seed=seed)
```
alvarobartt, line 410: IMO we could safely drop `spaCy` and `SparkNLP` support, right? If we want to keep those for a strong reason we can discuss about it

src/argilla/client/feedback/dataset/base.py
```diff
@@ -349,19 +359,26 @@ def prepare_for_training(
...
+            task,
+            (
+                TrainingTaskForSupervisedFinetuning,
+                TrainingTaskForRewardModelling,
+                TrainingTaskForDirectPreferenceOptimization,
+            ),
+        ):
+            raise ValueError(f"Training data {type(task)} is not supported yet")
+
+        data = task._format_data(self)
```
alvarobartt, line 386: I would like to simplify the code below this line, when checking for the frameworks and the methods required to prepare the data

src/argilla/client/feedback/integrations/huggingface/dataset.py
alvarobartt: Yes I agree with a row-approach, I also think is better, we would lose consistency in case we want to add a non-matching dict, but I think that even though the `datasets.Features` may be more generic if we rely on Python dicts rather than on `datasets.Sequence`, it's probably better and easier to handle this way

src/argilla/client/feedback/training/base.py
```diff
@@ -48,7 +59,7 @@ def __init__(
 
         Args:
             dataset (FeedbackDataset): the dataset to be used for training.
-            task_mapping (TrainingTaskMappingForTextClassification): the training data to be used for training.
+            task (TrainingTask): the training data to be used for training.
```
alvarobartt: I feel we could just remove the type from the docstrings i.e. `task (TrainingTask):` as it's already defined in the `__init__` so it feels a little bit redundant IMO, thoughts? @tomaarsen @gabrielmbmb

alvarobartt: Also we're not really consistent with docstrings here and there so we should probably spend some time defining our own guidelines for docstrings 👍🏻

src/argilla/client/feedback/training/base.py
```diff
@@ -219,24 +240,31 @@ class ArgillaTrainerSkeleton(ABC):
     def __init__(
         self,
         feedback_dataset: "FeedbackDataset",
```
alvarobartt: Maybe?

```suggestion
        dataset: "FeedbackDataset",
```

src/argilla/client/feedback/training/base.py
```diff
@@ -219,24 +240,31 @@ class ArgillaTrainerSkeleton(ABC):
...
-        task_mapping: TrainingTaskMappingForTextClassification,
+        task: Union[
+            TrainingTaskForTextClassification,
+            TrainingTaskForSupervisedFinetuning,
+            TrainingTaskForRewardModelling,
+            TrainingTaskForDirectPreferenceOptimization,
```
alvarobartt: I've seen these block a couple times already, can't we just define it as `TrainingTaskTypes = Union[TrainingTaskForTextClassification, TrainingTaskForSupervisedFinetuning, TrainingTaskForRewardModelling, TrainingTaskForDirectPreferenceOptimization]`?

src/argilla/client/feedback/training/frameworks/transformers.py
```diff
@@ -57,9 +57,9 @@ def __init__(self, *args, **kwargs):
         else:
             raise NotImplementedError(f"We do not support {type(self._dataset)} yet.")
 
-        if isinstance(self._task_mapping, TrainingTaskMappingForTextClassification):
+        if isinstance(self._task, TrainingTaskForTextClassification):
             self._model_class = AutoModelForSequenceClassification
         else:
-            raise NotImplementedError(f"ArgillaTransformersTrainer does not support {type(self._task_mapping)} yet.")
+            raise NotImplementedError(f"ArgillaTransformersTrainer does not support {type(self._task)} yet.")
```
alvarobartt: Maybe?

```suggestion
            raise NotImplementedError(f"ArgillaTransformersTrainer does not support {self._task.__name__} yet.")
```

src/argilla/feedback/__init__.py
alvarobartt, line 1: We should probably refactor everything into `src/argilla/feedback/...` instead, rather than under `src/argilla/client/feedback/...`, as I guess that was initially suggested by @frascuchon and @davidberenstein1957 created this file temporarily to "ease" the imports. But maybe we can discuss the refactoring of the module where everything's placed in the upcoming release?

src/argilla/training/base.py
```diff
@@ -281,7 +281,7 @@ def predict(self, text: Union[List[str], str], as_argilla_records: bool = True,
         """
         return self._trainer.predict(text=text, as_argilla_records=as_argilla_records, **kwargs)
 
-    def train(self, output_dir: str = None):
+    def train(self, output_dir: str):
```
alvarobartt, line 284:
```suggestion
    def train(self, output_dir: str) -> None:
```

alvarobartt: Also, is the `output_dir` mandatory?

alvarobartt (commented):
LGTM overall, we can discuss about this PR in detail tomorrow! I still need to review some files, but so far so good, great work Tom!

alvarobartt:
> We might have to already remove `fetch_records` internally as it might throw DeprecationWarnings that can't be removed (without disabling all DeprecationWarnings).

We can also have a look into this tomorrow, as now we have the `pull` method for `RemoteFeedbackDataset`s

src/argilla/training/base.py
tomaarsen: Yes. It's mandatory in `transformers` and our ArgillaTrainer `transformers` integration too.

src/argilla/client/feedback/training/frameworks/transformers.py
tomaarsen: I think there's a chance that this bit is dead code, but I'll use `self._task.__class__.__name__`

src/argilla/client/feedback/training/base.py
tomaarsen: Perhaps the best play is just to have one common superclass of all of these that we can use, but I'll go with this approach for now

src/argilla/client/feedback/dataset/base.py
tomaarsen: I'm down for this. @davidberenstein1957 how do you feel about it? We'd have DPO, SFT, RM, (and later PPO).

docs/_source/_common/snippets/training/feedback-task/text-classification.md
tomaarsen: Are these Windows double quotes? I don't use them myself. Thanks for the fix.

src/argilla/client/feedback/training/base.py
tomaarsen: Done

src/argilla/client/feedback/training/base.py
tomaarsen: I'm okay with removing it. `TrainingTask` is kind of wrong anyways, as for some reason `TrainingTask` isn't a superclass of `TrainingTaskFor...`.

src/argilla/client/feedback/dataset/base.py
tomaarsen: `Optional[X]` just means `Union[None, X]`. So, this says `Union[None, float]`. When looking at the code, I think that's true. E.g. the `_prepare_for_training_with_transformers` accepts None as a valid parameter.

src/argilla/client/feedback/dataset/base.py
tomaarsen: I think it should be fine to remove once we remove/replace the old trainers. The spaCy transformer tests actually cause my pytest to abruptly terminate. Not ideal.

src/argilla/client/feedback/dataset/base.py
tomaarsen: 👍 Do you want to do that as a follow-up PR?

src/argilla/feedback/__init__.py
tomaarsen: Sounds good. I'll leave that for after this PR.

tomaarsen:
We'll still have to add docs for using DPO, but I'll leave that for another PR. This is ready for final review.

tomaarsen:
@alvarobartt Can you check out the outstanding failure tomorrow? It's relating to your changes, the `test_push_to_huggingface_and_from_huggingface` is failing. I don't have time for it anymore.

alvarobartt:
> @alvarobartt Can you check out the outstanding failure tomorrow? It's relating to your changes, the `test_push_to_huggingface_and_from_huggingface` is failing. I don't have time for it anymore.

Sure, I'll take over the PR! Thanks for the merge! Enjoy your days off 🤗

tomaarsen:
Running through the documentation myself, and my final model is returning nothing. I'll try to figure out why that is.
```
### Instruction: Is a toad a frog?

### Context: Both frogs and toads are amphibians in the order Anura, which means "without a tail." Toads are a sub-classification of frogs, meaning that all toads are frogs, but not all frogs are toads.

### Response: <|endoftext|>
```

tomaarsen:
Note: This seems unsafe when it comes to exceeding the maximum model length, you just get CUDA indexing errors. A simple-ish resolution is:
1. Setting truncation in the tokenization
```python
            def tokenize(sample):
                sample["input_ids"] = self._transformers_tokenizer.encode(sample["query"], truncation=True)
                return sample
```

2. Filter away samples that are already at the limit, as you can't generate anything beyond the limit, and these samples don't follow the correct template with "### Instruction:" at the start and "### Response:" at the end.
```python
            def remove_truncated(sample):
                return len(sample) < self._transformers_tokenizer.model_max_length

            dataset = dataset.map(tokenize, batched=False)
            size_before = len(dataset)
            dataset = dataset.filter(remove_truncated, batched=False)
            size_after = len(dataset)
            if size_after != size_before:
                self._logger.info(
                    f"Removed {size_before - size_after} samples ({1 - (size_after / size_before):%}), "
                    "as these samples were longer than the maximum model length even before the generation."
                )
```

However, it's still not amazing:
```
### Instruction: Who is the US President?

### Context:

### Response: The United States President. The United States President. The United States President. The United States President. The United States President. The United States President. The United
```
```
### Instruction: Are dogs brown?

### Context: Dogs are all kinds of colors.

### Response: Dogs are all kinds of colors. Dogs are all kinds of colors. Dogs are all kinds of colors. Dogs are all kinds of colors. Dogs are all kinds
```
```
### Instruction: Is a toad a frog?

### Context: Both frogs and toads are amphibians in the order Anura, which means "without a tail." Toads are a sub-classification of frogs, meaning that all toads are frogs, but not all frogs are toads.

### Response: Both frogs and toads are amphibians in the order Anura, which means "without a tail." Toads are a sub-classification of frogs,
```

But maybe this is just because GPT2 isn't good enough, and that I've used too little meh-quality data? And perhaps the RM isn't great?

For context, this is the SFT response:
```
### Instruction: Who is the US President?

### Context:

### Response: Barack Obama is the US President. He is the first African American president to be elected. He is the first African American president to be elected to the US Senate
```
```
### Instruction: Are dogs brown?

### Context: Dogs are all kinds of colors.

### Response: Dogs are brown. They are not brown. They are not brown. They are not brown. They are not brown. They are not brown. They are not
```
```
### Instruction: Is a toad a frog?

### Context: Both frogs and toads are amphibians in the order Anura, which means "without a tail." Toads are a sub-classification of frogs, meaning that all toads are frogs, but not all frogs are toads.

### Response: A frog is a frog if it is a frog. A frog is a frog if it is a frog if it is a frog if it is a frog if
```

docs/_source/guides/llms/practical_guides/fine_tune.md
```diff
@@ -71,54 +71,159 @@ import argilla as rg
...
+
+### Response: Virgin Australia commenced services on 31 August 2000 as Virgin Blue, with two aircraft on a single route.
+"""
 ```
 
-####  TRL
+#### TRL
 
 The [Transformer Reinforcement Learning (TRL)](https://huggingface.co/docs/trl) package provides a flexible and customizable framework for fine-tuning models. It allows users to have fine-grained control over the training process, enabling them to define their functions and to further specify the desired behavior of the model. This approach requires a deeper understanding of reinforcement learning concepts and techniques, as well as more careful experimentation. It is best suited for users who have experience in reinforcement learning and want fine-grained control over the training process. Additionally, it directly integrates with [Performance Efficient Fine Tuning](https://huggingface.co/docs/peft/index) (PEFT) decreasing the computational complexity of this step of training an LLM.
```
dvsrepo:
```suggestion
The [Transformer Reinforcement Learning (TRL)](https://huggingface.co/docs/trl) package provides a flexible and customizable framework for fine-tuning models. It allows users to have fine-grained control over the training process, enabling them to define their functions and to further specify the desired behavior of the model. This approach requires a deeper understanding of reinforcement learning concepts and techniques, as well as more careful experimentation. It is best suited for users who have experience in reinforcement learning and want fine-grained control over the training process. Additionally, it directly integrates with [Parameter-Efficient Fine-Tuning](https://huggingface.co/docs/peft/index) (PEFT) decreasing the computational complexity of this step of training an LLM.
```

src/argilla/client/feedback/training/base.py
```diff
@@ -33,7 +34,7 @@ class ArgillaTrainer(ArgillaTrainerV1):
     def __init__(
         self,
         dataset: "FeedbackDataset",
-        task_mapping: TrainingTaskMappingForTextClassification,
+        task: TrainingTaskTypes,
         framework: Framework,
         lang: Optional["spacy.Language"] = None,
         model: Optional[str] = None,
```
dvsrepo, line 40: This a cross-cutting issue, not coming from this PR, but it's specially relevant for LLMs:

Allowing users to pass only a `str` model ref is inflexible. Very often you want to configure your model with certain parameters during model instantiation (e.g., `AutoModelForCausalLM.from_pretrained`). Enabling users to pass a model (and tokenizer) object would result in a much more flexible API. 

Another potential solution would be to define a new arg like the (trainer) `framework_kwargs` but for model (and tokenizer). But I think this is less desirable in terms of developer experience and maintenance from our side.

Another alternative is to tell our users: if you want this level of flexibility just use `prepare_for_training` and do the training directly with trl, transformers, etc. But I think, if feasible, we should try to give this extra-level of flexibility even if ArgillaTrainer is a short-cut for less critical/customized training workflows.

A clear example of why users will need more flexibility is quantization (needed to train larger models with limited compute resources):

```python
compute_dtype = getattr(torch, bnb_4bit_compute_dtype)
bnb_config = BitsAndBytesConfig(
    load_in_4bit=use_4bit,
    bnb_4bit_quant_type=bnb_4bit_quant_type,
    bnb_4bit_compute_dtype=compute_dtype,
    bnb_4bit_use_double_quant=use_nested_quant,
)

model = AutoModelForCausalLM.from_pretrained(
    "meta-llama/Llama-2-7b-hf",
    quantization_config=bnb_config,
    device_map=device_map
)

```

To the best of my knowledge, using this "configured" `model` is not possible with the current `ArgillaTrainer`. I'm aware this comes from the original ArgillaTrainer API, but maybe it's a good opportunity to tackle this now?

As an additional comment, I believe passing a instantiated model would be relevant for `spacy` too (haven't checked the ArgillaSpacyTrainer so I'm guessing it's the same but might be not):

```python

def custom_tokenizer(nlp):
   ....
   return Tokenizer(....)

nlp = spacy.load('en')
nlp.tokenizer = custom_tokenizer(nlp)

```
Love to hear your thoughts @davidberenstein1957 @tomaarsen @alvarobartt

dvsrepo: Any thoughts?

tomaarsen: I intended to review this with Alvaro and David today, but I didn't get around to it.

I think it should be feasible to allow either a model string or an initialized model. We would then also expose an optional tokenizer parameter. I think this should only require minimal changes, although I think it's best to create these changes in a separate PR.

dvsrepo: Great, I agree with creating a separate PR for this.

tomaarsen:
I think this is just about ready.

dvsrepo:
> I think this is just about ready.

Great news, who could take care of the review given that @davidberenstein1957 will finish early and is off tomorrow?

tomaarsen:
I want to make one more change: if a `peft_config` is provided to any of the trainers, then I'll save only the adapters. I'll look into this right after lunch.

Edit: This should be the default behaviour, but it doesn't seem to be happening.

tests/integration/client/feedback/training/test_trainer.py
```diff
@@ -107,3 +103,33 @@ def test_prepare_for_training_text_classification(
...
+@pytest.mark.parametrize(
+    "callable",
+    (
+        lambda: TrainingTaskMapping.for_text_classification(None, None),
+        lambda: TrainingTaskMapping.for_direct_preference_optimization(None),
+        lambda: TrainingTaskMapping.for_reward_modelling(None),
+        lambda: TrainingTaskMapping.for_supervised_fine_tuning(None),
+    ),
+)
+def test_deprecations(callable) -> None:
```
alvarobartt:
```suggestion
def test_deprecations(callable: Callable[..., Any]) -> None:
```

alvarobartt: Besides that you probably also need to include `from typing import Callable`

environment_dev.yml
```diff
@@ -46,14 +46,14 @@ dependencies:
       - spacy==3.5.3
       - https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-3.5.0/en_core_web_sm-3.5.0.tar.gz
       - spacy-transformers>=1.2.5
-      - transformers[torch]>=4.19.0
+      - transformers[torch]>=4.30.0 # <- required for DPO with TRL
```
alvarobartt, line 49: Make sure to also include those dependencies in `pyproject.toml`

environment_dev.yml
```diff
@@ -46,14 +46,14 @@ dependencies:
...
       - evaluate
       - seqeval
       - setfit
       - span_marker
       - openai
       - peft
-      - autotrain-advanced==0.5.2
+      - trl>=0.5.0
```
alvarobartt, line 56: Same here, please include it in `pyproject.toml`

src/argilla/client/feedback/dataset/base.py
```diff
@@ -306,14 +313,27 @@ def unify_responses(
...
+            framework: the framework to use for training. Currently supported frameworks are: `transformers`, `peft`,
+                `setfit`, `spacy`, `spacy-transformers`, `span_marker`, `spark-nlp`, `openai`, `trl`.
+            task: the NLP task to use for training. Currently supported tasks are: `TrainingTaskForTextClassification`,
+                `TrainingTaskForSFT`, `TrainingTaskForRM`, `TrainingTaskForPPO`, `TrainingTaskForDPO`.
+            train_size: the size of the train set. If `None`, the whole dataset will be used for training.
+            test_size: the size of the test set. If `None`, the whole dataset will be used for testing.
+            seed: the seed to use for splitting the dataset into train and test sets.
+            lang: the spaCy language to use for training. If `None`, the language of the dataset will be used.
+            fetch_records: whether to fetch the records from Argilla or use the local records instead. If `None`, use local.
+        """
```
alvarobartt, line 336: `fetch_records` will be deprecated from 1.15.0 onwards, so maybe we can just remove this and iterate over `dataset.records`, since it will fetch those from Argilla if not there, or just loop over those

alvarobartt:
```suggestion
        """
```

src/argilla/client/feedback/dataset/base.py
```diff
@@ -306,14 +313,27 @@ def unify_responses(
...
         if fetch_records is not None:
             warnings.warn(
                 "`fetch_records` is deprecated and will be removed in a future version."
```
alvarobartt, line 339: I'd say here you can just go ahead and remove it, since it is scoped from 1.15.0, so there's no need to still keep this code-block

src/argilla/client/feedback/dataset/base.py
alvarobartt: Sure, we can have a look at it upon merge! 👍🏻

src/argilla/client/feedback/training/base.py
```diff
@@ -263,13 +271,13 @@ def predict(self, text: Union[List[str], str], as_argilla_records: bool = True,
         """
 
     @abstractmethod
-    def train(self, output_dir: str = None):
+    def train(self, output_dir: str = None) -> None:
```
alvarobartt:
```suggestion
    def train(self, output_dir: Optional[str] = None) -> None:
```

tests/integration/client/feedback/training/test_trainer.py
```diff
@@ -43,8 +45,8 @@
 @pytest.mark.parametrize(
     "framework",
     [
-        Framework("spacy"),
-        Framework("spacy-transformers"),
+        # Framework("spacy"),
+        # Framework("spacy-transformers"),
```
alvarobartt: Shouldn't we just remove those if not ran?

```suggestion
```

src/argilla/client/feedback/training/frameworks/trl.py
```diff
@@ -0,0 +1,383 @@
...
+        task: Union[
+            TrainingTaskForSFT,
+            TrainingTaskForRM,
+            TrainingTaskForPPO,
+            TrainingTaskForDPO,
+        ],
+        prepared_data=None,
+        model: str = None,
+        seed: int = None,
+    ):
```
alvarobartt:
```suggestion
    ) -> None:
```

src/argilla/client/feedback/training/frameworks/trl.py
```diff
@@ -0,0 +1,383 @@
...
+        self.reward_model = reward_model
+        self.length_sampler_kwargs = length_sampler_kwargs
+        self.generation_kwargs = generation_kwargs
+
+
+class ArgillaTRLTrainer(ArgillaTrainerSkeleton):
+    _logger = logging.getLogger("ArgillaTRLTrainer")
+    _logger.setLevel(logging.INFO)
+
+    require_version("trl>=0.5.0")
```
alvarobartt, line 76: Shouldn't we also include `transformers` and `torch`?

```suggestion
    require_version("transformers")
    require_version("torch")
    require_version("trl>=0.5.0")
```

alvarobartt: Should be simpler and easier to define thanks to @plaguss PR at https://github.com/argilla-io/argilla/pull/3555

```python
require_dependencies(["trl>=0.5.0", "transformers", "torch"])
```

src/argilla/client/feedback/training/frameworks/trl.py
```diff
@@ -0,0 +1,383 @@
...
+        from datasets import DatasetDict
+        from transformers import set_seed
+
+        self._transformers_model = None
+        self._transformers_tokenizer = None
+        self.device = "cpu"
+        if torch.backends.mps.is_available():
+            self.device = "mps"
+        elif torch.cuda.is_available():
+            self.device = "cuda"
```
alvarobartt, line 102: Could we include a `use_cuda` and `use_mps` boolean flags as args? Maybe we don't want to use the device even if available

src/argilla/client/feedback/training/frameworks/trl.py
```diff
@@ -0,0 +1,383 @@
...
+
+        if self._seed is None:
+            self._seed = 42
+        set_seed(self._seed)
+
+        if self._model is None:
+            self._model = "gpt2"
```
alvarobartt: IMO we can assume the default is `gpt2-medium` instead of `gpt2`

```suggestion
            self._model = "gpt2-medium"
```

src/argilla/client/feedback/training/schemas.py
```diff
@@ -39,12 +40,18 @@
...
+if TYPE_CHECKING:
+    import datasets
+    import spacy
+
+    from argilla.client.feedback.dataset import FeedbackDataset
+
 
 class TrainingData(ABC):
-    def _format_data(self, records):
+    def _format_data(self, dataset: "FeedbackDataset"):
```
alvarobartt:
```suggestion
    def _format_data(self, dataset: "FeedbackDataset") -> pd.DataFrame:
```

src/argilla/client/feedback/training/schemas.py
```diff
@@ -128,30 +141,30 @@ def for_text_classification(
...
         label_strategy: str = None,
-    ) -> "TrainingTaskMappingForTextClassification":
+    ) -> "TrainingTaskForTextClassification":
         """
         _summary_
 
         Args:
             text (TextField): The TextField to use for training.
-            label (Union[RatingQuestion, LabelQuestion, MultiLabelQuestion, RatingQuestionUnification, LabelQuestionUnification, MultiLabelQuestionUnification]): _description_
+            label (Union[RatingQuestion, LabelQuestion, RankingQuestion, MultiLabelQuestion, RatingQuestionUnification, LabelQuestionUnification, MultiLabelQuestionUnification, RankingQuestionUnification]): _description_
```
alvarobartt: As mentioned before, I think we can just remove the type-hints from the docstrings are those are automatically included when generating the docs if the type-hint is defined in the function arguments

alvarobartt: Note that this comment also applies to the rest of the methods defined in this Python file

tests/integration/client/feedback/training/test_trl.py
```diff
@@ -0,0 +1,256 @@
...
+if TYPE_CHECKING:
+    from argilla.client.feedback.types import AllowedFieldTypes, AllowedQuestionTypes
+
+
+@pytest.mark.usefixtures(
+    "feedback_dataset_guidelines",
+    "feedback_dataset_fields",
+    "feedback_dataset_questions",
+    "feedback_dataset_records",
+)
```
alvarobartt: We should refactor the tests to use the pre-defined factories rather than those fixtures, but that can be tackled in a separate PR 👍🏻

src/argilla/client/feedback/dataset/base.py
davidberenstein1957: @alvarobartt I agree, this aligns with the mention I had regarding the differentiation between local and remote datasets.

src/argilla/client/feedback/training/frameworks/trl.py
davidberenstein1957: I agree, this was to keep people from figuring out how to use `cuda` and `mps` but it can be overwritten via `update_config`. Perhaps it would be better to include it as one of the main args like `train_size`, however, this can be tackled in a different PR IMO. I believe @dvsrepo also suggested some other API changes.

src/argilla/client/feedback/training/frameworks/trl.py
davidberenstein1957: For the other frameworks, I provided a default that works okay out of the box. Not sure if that would be reasonable here so perhaps it is better to go for a smaller model that allows for easy testing out of the box.

src/argilla/client/feedback/dataset/base.py
```diff
@@ -376,11 +404,15 @@ def prepare_for_training(
                     lang = spacy.blank(lang)
```
davidberenstein1957, line 405:
```suggestion
                    lang = spacy.blank(lang) if len(lang)==2 else spacy.load(lang)
```

src/argilla/client/feedback/training/base.py
```diff
@@ -47,37 +48,32 @@ def __init__(
...
+            seed: the random seed to ensure reproducibility. Defaults to None.
+            gpu_id: the GPU ID to use when training a SpaCy model. Defaults to -1, which means that the CPU
                 will be used by default. GPU IDs start in 0, which stands for the default GPU in the system,
                 if available.
-            framework_kwargs (dict): arguments for the framework's trainer.
+            framework_kwargs: arguments for the framework's trainer.
             **load_kwargs: arguments for the rg.load() function.
         """
         self._dataset = dataset
         self._train_size = train_size
```
davidberenstein1957, line 68: Maybe we should provide a warning if someone does provide a train_size?

src/argilla/client/feedback/training/frameworks/trl.py
```diff
@@ -0,0 +1,383 @@
...
+
+        if isinstance(self._dataset, DatasetDict):
```
davidberenstein1957, line 111: I think the train_size warning might need to go here and the concatenation of datasets can also be done here.

src/argilla/client/feedback/training/frameworks/trl.py
```diff
@@ -0,0 +1,383 @@
...
+        """
+        Updates the configuration of the trainer, but the parameters depend on the trainer.subclass.
+        """
+        from transformers import TrainingArguments
+
+        self.training_args_kwargs.update(filter_allowed_args(TrainingArguments.__init__, **kwargs))
+        self.training_args_kwargs.update(filter_allowed_args(PPOArgs.__init__, **kwargs))
+        self.trainer_kwargs.update(filter_allowed_args(self.trainer_cls.__init__, **kwargs))
+
+    def predict(self, text: Union[List[str], str], as_argilla_records: bool = True, **kwargs) -> None:
```
davidberenstein1957, line 209: maybe we should allow for generative predictions?

src/argilla/client/feedback/training/frameworks/trl.py
```diff
@@ -0,0 +1,383 @@
...
+
+        self.save(output_dir)
+
+    def save(self, output_dir: str) -> None:
+        """
+        Saves the model to the specified path.
+        """
+        self._transformers_model.save_pretrained(output_dir)
+        self._transformers_tokenizer.save_pretrained(output_dir)
+
```
davidberenstein1957, line 374: should we add a `push_to_hub` method to the `transformer` based models? Similarly, we can do that [for spaCy](https://huggingface.co/docs/hub/spacy#sharing-your-models) too. IMO, we would ideally align this with a nicely generated card with a link to the dataset. @dvsrepo @alvarobartt , what do you think?

davidberenstein1957 (approved):
Thanks @tomaarsen, looks great! Some minor API-design-related comments.  I'll make a separate PR for the docs.

src/argilla/client/feedback/dataset/base.py
davidberenstein1957: FYI this suggestions does not really work ou-of-the box😅

src/argilla/client/feedback/training/frameworks/trl.py
alvarobartt: Sure, we can prepare a card once we are happy with the training interface on a separate PR if that's necessary

src/argilla/client/feedback/training/frameworks/trl.py
tomaarsen: I think it might be best to let users provide a `device` to the ArgillaTrainer init directly.

src/argilla/client/feedback/training/frameworks/trl.py
tomaarsen: Then I think a more logical next step is to add a separate generate method?

src/argilla/client/feedback/training/frameworks/trl.py
tomaarsen: Generating a card is a very nice feature, but it would have to be custom made for each of the task-framework combinations, as they all store different information (e.g. logs) and in different ways.

tests/integration/client/feedback/training/test_trainer.py
tomaarsen: e677447831dc9d7dd53abff2bdfed7934b237a40

tomaarsen:
~~TODO: Changelog~~

Edit: Done

environment_dev.yml
tomaarsen: f357c53240243555bc8818d38eb799c0c53298f6

environment_dev.yml
tomaarsen: f357c53240243555bc8818d38eb799c0c53298f6

src/argilla/client/feedback/dataset/base.py
tomaarsen: I'll delete it from the ArgillaTrainer. Should I also remove it from `prepare_for_training`? I think that might break backwards compatibility.

tomaarsen: 2b576dfa6be51eaa106c96623a56f37fbea9de06

tests/integration/client/feedback/training/test_trainer.py
tomaarsen: Oh, oops. I think these should run as they work on the CI - they just cause my pytest locally to apruptly terminate without any message, so I comment them out.

tomaarsen: c612fc4516790b17604b5a1ff3d1b77c9ebbaf09

src/argilla/client/feedback/training/frameworks/trl.py
tomaarsen: Sure. I didn't include these as `trl` requires transformers and torch already, so installing `trl>=0.5.0` should get you all set up.

tomaarsen: We can use the `require_dependencies` later, I'll go with the triple `require_version` for now

src/argilla/client/feedback/dataset/base.py
tomaarsen: As discussed privately in Slack, we'll keep this for a separate deprecation PR.

src/argilla/client/feedback/dataset/base.py
tomaarsen: As discussed privately in Slack, we'll keep this for a separate deprecation PR.

src/argilla/client/feedback/training/frameworks/trl.py
tomaarsen: Let's track this in #3632 after this PR.

src/argilla/client/feedback/training/frameworks/trl.py
tomaarsen: Sure, I'm cool with taking a stronger default here

src/argilla/client/feedback/training/schemas.py
tomaarsen: 56f3ad5d84782f189a67aa1a0676e57144574def

src/argilla/client/feedback/training/schemas.py
tomaarsen: Sure, we can do that

tomaarsen: 1eff2089fbcf0ff6e21848da0781e600acf407fb

tests/integration/client/feedback/training/test_trl.py
tomaarsen: These are normal fixtures, no idea why the weird `usefixtures` syntax is used, it's not needed at all.

tomaarsen: 7e2c28eb14688ab8898c370462a76e5783d826c7

src/argilla/client/feedback/dataset/base.py
tomaarsen: I don't think I understand, this line is already wrapped in
```python
                if len(lang) == 2:
                    lang = spacy.blank(lang)
                else:
                    lang = spacy.load(lang)
```

src/argilla/client/feedback/training/base.py
tomaarsen: Do you mean "does not"? I think it's mostly okay either way, sometimes just exclusively training with your data is fine-ish, especially for models where a loss doesn't really say much about how good a model is.

src/argilla/client/feedback/training/frameworks/trl.py
tomaarsen: I'll make an issue for the push_to_hub

tomaarsen: #3634 #3633

src/argilla/client/feedback/training/frameworks/trl.py
tomaarsen: I'll make an issue for this

davidberenstein1957:
@tomaarsen I restructured the docs. Could you add some final docs to `docs/trl-docs-revisited`?

PR #3469, opened 2023-07-27
feat: add `get_metrics` function in SDK

alvarobartt:
# Description

This PR adds a function in the SDK named `get_metrics` that calls `GET /api/v1/me/datasets/{dataset_id}/metrics` to get the metrics of a certain dataset from a certain user, meaning that each user would just see what can be seen according to its permissions.

This PR is created on top of #3465 since we will need to get the `records.count` to override the magic method `__len__` in `_ArgillaFeedbackDataset` so that the length of a dataset is directly retrieved from the record count of the dataset, since we're no longer keeping local data when working with a remote dataset.

**Type of change**

**How Has This Been Tested**

- [X] Add unit tests for the `get_metrics` function

**Checklist**

### DIFF
--- a/src/argilla/client/sdk/v1/datasets/models.py
+++ b/src/argilla/client/sdk/v1/datasets/models.py
@@ -100,3 +100,19 @@ class FeedbackSuggestionModel(BaseModel):
     score: Optional[float] = None
     value: Any
     agent: Optional[str] = None
+
+
+class FeedbackRecordsMetricsModel(BaseModel):
+    count: int
+
+
+class FeedbackResponsesMetricsModel(BaseModel):
+    count: int
+    submitted: int
+    discarded: int
+    draft: int
+
+
+class FeedbackMetricsModel(BaseModel):
+    records: FeedbackRecordsMetricsModel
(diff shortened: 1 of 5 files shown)


src/argilla/client/sdk/v1/datasets/models.py
```diff
@@ -100,3 +100,19 @@ class FeedbackSuggestionModel(BaseModel):
...
+
+class FeedbackRecordsMetricsModel(BaseModel):
+    count: int
+
+
+class FeedbackResponsesMetricsModel(BaseModel):
+    count: int
+    submitted: int
+    discarded: int
+    draft: int
```
frascuchon: I think we don't need these metrics attributes in the Python client for now. Or do we show the user-connected metrics somewhere?

alvarobartt: It was just to align both schemas, but for now I guess we'll just use `records.count` from the Python client as part of `len(dataset.records)`

gabrielmbmb (approved):
LGTM

PR #3471, opened 2023-07-27
fix: import errors when importing from `argilla.feedback`

### LINKED ISSUE #3468: [BUG] `ModuleNotFoundError` and `ImportError` when importing `argilla.feedback.*`
**Describe the bug**
When trying to use import something from `argilla.feedback` module a `ModuleNotFoundError` is raised because `fastapi` is not installed and an `ImportError` because `pyyaml` is not installed either.

```bash
pip install argilla==1.13.2
```

```python
from argilla.feedback import SuggestionSchema
```

**Expected behavior**
I can import stuff from `argilla.feedback` module without getting a `ModuleNotFoundError` and without installing `fastapi` and `PyYAML` dependencies which should be only needed for the server and extra integrations.

**Environment:**
 - Argilla Version [e.g. 1.0.0]: 1.13.2
 - ElasticSearch Version [e.g. 7.10.2]: 8.5.3
 - Docker Image (optional) [e.g. argilla:v1.0.0]: Not relevant

**Additional context**
The bug was introduced in https://github.com/argilla-io/argilla/pull/3336, when the telemetry used in the server (it imports `fastapi`) was added to the Argilla Trainer.

gabrielmbmb:
# Description

This PRs fixes the `ModuleNotFoundError` and `ImportError` that occurred when trying to import something from `argilla.feedback` module.

The first error was caused because in #3336 the telemetry was included in the `ArgillaTrainer`, but in the `argilla.utils.telemetry` module some optional dependencies used by the server were being imported.

The second one was caused because the module in which `HuggingFaceDatasetMixin` (and from which `FeedbackDataset` is inheriting) class lives was importing classes from the `argilla.client.feedback.config` module, which was importing `pyyaml` in its root causing the `ImportError`.

Closes #3468 

**Type of change**

**How Has This Been Tested**

I've created a wheel of this branch, installed in a new virtual environment and I was able to import something `argilla.feedback` module without errors.

**Checklist**

### DIFF
--- a/src/argilla/client/feedback/integrations/huggingface/dataset.py
+++ b/src/argilla/client/feedback/integrations/huggingface/dataset.py
@@ -187,6 +187,7 @@ class HuggingFaceDatasetMixin:
         import huggingface_hub
         from huggingface_hub import DatasetCardData, HfApi
 
+        # https://github.com/argilla-io/argilla/issues/3468
         from argilla.client.feedback.config import DatasetConfig
 
         if parse_version(huggingface_hub.__version__) < parse_version("0.14.0"):
@@ -262,6 +263,7 @@ class HuggingFaceDatasetMixin:
         from huggingface_hub import hf_hub_download
         from huggingface_hub.utils import EntryNotFoundError
 
+        # https://github.com/argilla-io/argilla/issues/3468
         from argilla.client.feedback.config import (
             DatasetConfig,
             DeprecatedDatasetConfig,
--- a/src/argilla/client/feedback/integrations/huggingface/dataset.py
+++ b/src/argilla/client/feedback/integrations/huggingface/dataset.py
@@ -20,7 +20,6 @@ from typing import TYPE_CHECKING, Any, Optional, Type
 
 from packaging.version import parse as parse_version
 
-from argilla.client.feedback.config import DatasetConfig, DeprecatedDatasetConfig
 from argilla.client.feedback.constants import FIELD_TYPE_TO_PYTHON_TYPE
 from argilla.client.feedback.schemas import FeedbackRecord
 from argilla.client.feedback.types import AllowedQuestionTypes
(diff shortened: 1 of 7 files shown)


src/argilla/client/feedback/integrations/huggingface/dataset.py
```diff
@@ -188,6 +187,8 @@ def push_to_huggingface(
         import huggingface_hub
         from huggingface_hub import DatasetCardData, HfApi
 
+        from argilla.client.feedback.config import DatasetConfig
```
frascuchon, line 191: Please, add a comment here about why we are importing this locally (just point out the issue link)

src/argilla/client/feedback/integrations/huggingface/dataset.py
```diff
@@ -261,6 +262,11 @@ def from_huggingface(cls: Type["FeedbackDataset"], repo_id: str, *args: Any, **k
         from huggingface_hub import hf_hub_download
         from huggingface_hub.utils import EntryNotFoundError
 
+        from argilla.client.feedback.config import (
```
frascuchon, line 267: ditto

src/argilla/server/errors/api_errors.py
```diff
@@ -41,10 +43,20 @@ def __init__(self, error: ServerError):
 
 
 class APIErrorHandler:
+    @staticmethod
+    async def track_error(error: ServerError, request: Request):
+        data = {"code": error.code}
```
frascuchon: We can setup the headers here instead of delegating a private telemetry method. This logic has more sense here.
```suggestion
        data = {
           "code": error.code,
           "user-agent": request.headers.get("user-agent"),
           "accept-language": request.headers.get("accept-language")
        }
```

src/argilla/server/errors/api_errors.py
```diff
@@ -41,10 +43,20 @@ def __init__(self, error: ServerError):
...
+        if isinstance(error, (GenericServerError, EntityNotFoundError, EntityAlreadyExistsError)):
+            data["type"] = error.type
+
+        data.update(telemetry._process_request_info(request))
```
frascuchon:
```suggestion
```

frascuchon (approved):

PR #3474, opened 2023-07-27
styles: Feedback Task UI fine tuning

### LINKED ISSUE #3475: UI Fine tuning
- [x]  The less chevron disappear
- [x]  We should remove the shadow on the search bar for all states. Never use shadow when blue stroke 
- [ ]  The rating integration is not totally thin. The outline of the component should be at first plan
    
    

### Ranking component

- [ ]  Add some space between the row and column (too small between the ranking number and placeholder area). @Leire Do you think we can make the ranking number component square? or we’ll lose too much space?
- [ ]  We need to show visually activated the row is filled
- [ ]  The shadow needs to be improved (cut at the end and too strong). Too strong the shadow and should be visible all around not only on the bottom. The icon is also too big

### Labels style:

- [x]  Add a min size to single and multi label tag : The minimum size should the one to host 3 characters.  That means if the label is composed of 1 or 2 characters like “No” the label width should be the same as for “Yes”.

- [x]  Add a min size to rating circle button : The minimum size should host 2 numbers.  That means if the circle with “1” should have the same size than the one with “10”. Maintain a regular circle until 2 numbers

### Search bar review:

Main search bar:

- [x]  No shadow effect on active in the main search bar

Form search bar:

- [x]  Reduce the height of the bar (to make if different from the label)
- [x]  No shadow
- [x]  Reduce icon
- [x]  Reduce font

leiyre:
# Description

UI small improvement and fine tuning for Feedback Task

Closes [#3475](https://github.com/argilla-io/argilla/issues/3475)

**Type of change**

- [x] Styles improvement

**Checklist**


damianpumar:
Hi @leiyre could you add some e2e to ensure the styles?

frontend/components/commons/forms/questions-form/fields/rating/RatingMonoSelection.component.vue
```diff
@@ -65,18 +65,19 @@ export default {
...
-  background: palette(purple, 800);
+  border-radius: $border-radius-rounded;
+  height: $base-space * 4;
+  min-width: $base-space * 4;
+  padding-inline: $base-space;
   outline: none;
-  padding-inline: 12px;
-  line-height: 32px;
+  background: palette(purple, 800);
+  color: palette(purple, 200);
```
keithCuniah, line 78: from more important to less important => border-radius is less importantthan height/padding, so it's better to put the corresponding css after

damianpumar (approved):

PR #3477, opened 2023-07-28
feat: question `patch` endpoint

### LINKED ISSUE #3398: [FEATURE] Add `PATCH` endpoint for FeedbackDataset question
Some of the attributes of a question of a Feedback Dataset cannot be updated.

Add a new `PATCH /api/v1/questions/:question_id` endpoint that allows to update the following attributes:

- `title`
- `description`
- For the `LabelSelection` and `MultiLabelSelection` questions:
  - `settings.visible_labels`
- For the `TextQuestion`:
  - `settings.use_markdown`

### LINKED ISSUE #3480: Add  missing `type=text`  on fields settings update
The body for the endpoint `PATCH /api/v1/fields/:field_id` should be aligned to the field creation endpoint. So, settings should include the `type: "text"`  attribute.

Image

Image

gabrielmbmb:
# Description

This PR adds a new endpoint `PATCH /api/v1/questions/{question_id}` that allows to partially update a question of a `FeedbackDataset`.

In addition, this PR adds a new base class for the the update schemas `UpdateSchema` (inherits from `pydantic.BaseModel`). This class allows to define a set of attributes which their value cannot be set explicitly to `None`.

Finally, the schemas used in the old endpoints have been moved to a new subpackage `argilla.server.schemas.v0` for better organization.

Closes #3398 
Closes #3480

**Type of change**

**How Has This Been Tested**

Manual tests in a local development environment and new unit tests have been added to cover the additions.

**Checklist**


src/argilla/server/apis/v1/handlers/questions.py
```diff
@@ -19,27 +19,60 @@
...
+    db: AsyncSession = Depends(get_async_db),
+    question_id: UUID,
+    question_update: QuestionUpdate,
+    current_user: User = Security(auth.get_current_user),
+):
+    question = await _get_question(db, question_id)
+
+    await authorize(current_user, QuestionPolicyV1.update(question))
+
+    if question_update.settings is not None:
```
frascuchon: Could we use the pydantic discriminator to avoid this logic? I think we're already using for the `QuestionCreate` schema

src/argilla/server/schemas/base.py
```diff
@@ -0,0 +1,41 @@
...
+from pydantic import BaseModel, root_validator
+
+
+class UpdateSchema(BaseModel):
+    """Base schema for update endpoints. `__non_explicit_none__` is a set of fields that cannot be set to `None`
+    explicitly. The list of fields is validated in `validate_non_explicit_none` root validator, which will raise a
+    `ValueError` if any of the fields in the set was set to `None` explicitly.
+    """
+
+    __non_explicit_none__: Union[Set[str], None] = None
```
frascuchon: As far I understood, this base class will prevent setting `None` values for the configured non-nullable fields in `__non_explicit_none__`,  right? In that case, maybe we can change a bit the ` non_explicit_none` naming to something more descriptive, like `__non_nullable_fields__`?

src/argilla/server/schemas/base.py
```diff
@@ -0,0 +1,41 @@
...
+        if cls.__non_explicit_none__ is None:
+            return values
+
+        invalid_keys = []
+        for key in cls.__non_explicit_none__:
+            if key in values and values[key] is None:
+                invalid_keys.append(key)
+
+        if invalid_keys:
+            raise ValueError(f"Fields {invalid_keys} cannot be updated to `None`.")
```
frascuchon: cannot be null? or, regarding the API behavior, "cannot remove the value for field ..." or `you must provide a value for field ...`

src/argilla/server/schemas/base.py
frascuchon, line 1: since this will be used for the `v1` schemas, maybe we can move to the module `schemas.v1.base`

src/argilla/server/schemas/v1/questions.py
```diff
@@ -86,3 +90,21 @@ class Question(BaseModel):
 
     class Config:
         orm_mode = True
+
+
+class TextQuestionSettingsUpdate(UpdateSchema):
+    use_markdown: Optional[bool]
```
frascuchon: Regarding this schema definition, since we have only one single attribute, maybe we should force provide it when settings are sent as part of the updated body:
this
````json
{ 
   "settings": {}
}
````
is a syntactically valid update body, but maybe makes no much sense, since is equivalent to 
````json
{}
````

Maybe, we should force API consumers to provide a value for `use_markdown` if settings are part of the request body:
```json
{
   "settings": {
      "use_markdown": ...
   } 
}
```
In that case, the schema definition could be:

```python
class TextQuestionSettingsUpdate(UpdateSchema):
    use_markdown: bool
    __non_explicit_none__ = {"use_markdown"} # I think this shouldn't be needed 
```

We should review the FieldsSettings update to align it with this approach.

src/argilla/server/schemas/v1/questions.py
```diff
@@ -86,3 +90,21 @@ class Question(BaseModel):
...
+
+
+class LabelSelectionSettingsUpdate(UpdateSchema):
+    visible_options: Optional[int]
+
+
+class QuestionUpdate(UpdateSchema):
+    title: Optional[str]
+    description: Optional[str]
+    settings: Optional[Union[Dict[str, Any], TextQuestionSettingsUpdate, LabelSelectionSettingsUpdate]]
```
frascuchon: why not use the same approach as the `QuestionCreate` schema? 
```python
QuestionSettingsUpdate = Annotated[
    Union[
        TextQuestionSettingsUpdate,
        LabelSelectionQuestionSettingsUpdate,
        MultiLabelSelectionQuestionSettingsUpdate,
    ],
    PydanticField(discriminator="type"),
]

```

src/argilla/server/schemas/v1/questions.py
```diff
@@ -86,3 +90,21 @@ class Question(BaseModel):
...
+
+    __non_explicit_none__ = {"title", "description", "settings"}
```
frascuchon: Description can be nullable in DB. For me, it makes sense also provide this option from the update endpoint. 
```suggestion
    __non_explicit_none__ = {"title", "settings"}
```

tests/server/api/v1/test_questions.py
```diff
@@ -61,39 +84,172 @@
         ),
         (
             LabelSelectionQuestionFactory,
+            {"settings": {"type": "ranking", "visible_options": 3}},
```
frascuchon: this should be an invalid payload. The `type` ranking does not match with provided options. We should change the endpoint to take care of this.

tests/server/api/v1/test_questions.py
```diff
@@ -38,11 +39,33 @@
 
 
 @pytest.mark.parametrize(
-    "QuestionFactory, expected_settings",
+    "QuestionFactory, payload, expected_settings",
     [
-        (TextQuestionFactory, {"type": "text", "use_markdown": False}),
+        (
+            TextQuestionFactory,
+            {"title": "New Title", "description": "New Description", "settings": {"use_markdown": True}},
```
frascuchon: I would say to be more explicit on with question setting type consumers are trying to update. I miss the `settings.type` value as a required one.

src/argilla/server/schemas/v1/datasets.py
```diff
@@ -117,10 +118,12 @@ class DatasetCreate(BaseModel):
     workspace_id: UUID
 
 
-class DatasetUpdate(BaseModel):
+class DatasetUpdate(UpdateSchema):
     name: Optional[DatasetName]
     guidelines: Optional[DatasetGuidelines]
 
+    __non_explicit_none__ = {"name", "guidelines"}
```
frascuchon: I think the guidelines are nullable in DB, so, we should let users remove the guidelines for a dataset.

frascuchon:
```suggestion
    __non_explicit_none__ = {"name"}
```

src/argilla/server/schemas/base.py
gabrielmbmb: yes, agree. We should review the rest of the errors in the API, as I think that we've a few that describes the errors using Python terminology instead of JSON terminology.

src/argilla/server/schemas/base.py
gabrielmbmb: Agree!

src/argilla/server/schemas/v1/questions.py
gabrielmbmb: I didn't do it this way, so the client didn't need to provide the `settings.type` of questions which will be updated, but it makes sense for the client to provide the `settings.type` as it will know the type of question to be updated.

src/argilla/server/schemas/base.py
gabrielmbmb: I think this class can be useful no matter the version of the API, and in the future if we've a `v2` API probably we will want to use it there. I think it would be a good addition to have a base class also for the other schemas: https://github.com/zhanymkanov/fastapi-best-practices#8-custom-base-model-from-day-0

src/argilla/server/apis/v1/handlers/fields.py
```diff
@@ -47,12 +52,18 @@ async def update_field(
 ):
     field = await _get_field(db, field_id)
 
+    if field_update.settings and field_update.settings.type != field.settings["type"]:
+        raise HTTPException(
+            status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+            detail=f"Field type cannot be changed. Expected '{field.settings['type']}' but got '{field_update.settings.type}'",
+        )
+
     await authorize(current_user, FieldPolicyV1.update(field))
```
frascuchon: I would say to check action grants for the connected user first, and then apply action validations. Otherwise, we may share the sensible info with unauthorized users
```suggestion
    await authorize(current_user, FieldPolicyV1.update(field))
    
    if field_update.settings and field_update.settings.type != field.settings["type"]:
        raise HTTPException(
            status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
            detail=f"Field type cannot be changed. Expected '{field.settings['type']}' but got '{field_update.settings.type}'",
        )
```

gabrielmbmb: wops! my bad

frascuchon (approved):

PR #2325, opened 2023-02-09
Refactor: Replace `isort` by `ruff` in `pre-commit`

tomaarsen:
Hello!

# Pull Request overview
* Replaced `isort` by `ruff` in `pre-commit`
  * This has fixed the import sorting throughout the repo.
* Manually went through `ruff` errors and fixed a bunch:
  * Unused imports
  * Added `if TYPE_CHECKING:` statements
  * Placed `# noqa: <errcode>` where the warned-about statement is the desired behaviour
* Add basic `ruff` configuration to `pyproject.toml`

## Details
This PR focuses on replacing `isort` by `ruff` in `pre-commit`. The motivation for this is that:
* `isort` frequently breaks. I have experienced 2 separate occasions in the last few months alone where the latest `isort` release has broken my CI runs in NLTK and SetFit.
* `isort` is no longer supported for Python 3.7, whereas Argilla still supports 3.7 for now.
* `ruff` is absurdly fast, I actually can't believe how quick it is.

This PR consists of 3 commits at this time, and I would advise looking at them commit-by-commit rather than at the PR as a whole. I'll also explain each commit individually.

## [Add ruff basic configuration](https://github.com/argilla-io/argilla/commit/497420e7097b7039d479df6bf431ea73c370f90b)
I've added basic configuration for [`ruff`](https://github.com/charliermarsh/ruff), a very efficient linter. I recommend the following commands:
```
# Get all [F]ailures and [E]rrors
ruff .

# See all import sort errors
ruff . --select I

# Fix all import sort errors
ruff . --select I --fix
```

## [Remove unused imports, apply some noqa's, add TYPE_CHECKING](https://github.com/argilla-io/argilla/commit/f219acbaba8f4b7fffaea70e8e1e9d1fe3a28475)
The unused imports speaks for itself.

As for the `noqa`'s, `ruff` (like most linters) respect the `# noqa` (no quality assurance) keyword. I've used the keyword in various locations where linters would warn, but the behaviour is actually correct. As a result, the output of `ruff .` now only points to questionable code.

Lastly, I added `TYPE_CHECKING` in some locations. If type hints hint at objects that do not need to be imported during run-time, then it's common to type hint like `arr: "numpy.ndarray"`. However, IDE's won't understand what `arr` is. Python has implemented `TYPE_CHECKING` which can be used to conditionally import code *only* when type checking. As a result, the code block is not actually executed in practice, but the inclusion of it allows for IDEs to better support development.
See an example here:
```python
[...]

PR #3478, opened 2023-07-28
chore: add `pyflakes`, `pycodestyle`, `isort` and `flake8` rules to `ruff`

gabrielmbmb:
# Description

This PR updates the `ruff` configuration in `pyproject.toml`, so rules from `pyflakes`, `pycodestyle` and `isort` are checked by `ruff`. These ones are useful, as `ruff` will warn us about unused imports, not used variables, etc.

The command `ruff --fix` has been executed in the files from `src` and `tests` directories, which has automatically removed unused imports and variables, as well as sorted the imports.

After running the command, the tests were not working because in `argilla/server/models/__init__.py` we were importing everything from `argilla/server/models/models.py` and `argilla/server/models/questions.py` modules using `*` wildcard. Both modules have a `Field` class/function, so when creating the factories in `tests/factories.py`, instead of using the `Field` SQLAlchemy ORM class, the `Field` function from `pydantic` was being used causing the error. Before this error was not occurring because the `Field` from `models.py` was being imported after the one from `questions.py`.

**Type of change**

**How Has This Been Tested**

N/A

**Checklist**


tomaarsen:
@gabrielmbmb This is awesome!

Perhaps we can also incorporate these ones that I once researched but never included? 
https://github.com/argilla-io/argilla/pull/2325#issuecomment-1424320210

Other than that, LGTM

tomaarsen (approved):

alvarobartt (commented):
LGTM! Thanks for this additions, much needed 👍🏻

alvarobartt (approved):

frascuchon (commented):
Nice @gabrielmbmb !!!

Maybe we can add some refs in development guides?

tomaarsen:
You might have to just rebase efb86f788f223660da4deb61297025b4d02fad08 and 8a702a27475c050bfaa54f095de3ed53e15678ee on top of develop again, and redo the `pyproject.toml` changes before re-running `pre-commit run --all`. These merge conflicts would be horrible to deal with otherwise 😅

src/argilla/client/client.py
```diff
@@ -137,6 +138,7 @@ def __init__(
                 " datasets, specify a valid workspace name on `rg.init` or set it"
                 " up through the `rg.set_workspace` function.",
                 category=UserWarning,
+                stacklevel=2,
```
alvarobartt: I'm not a huge fan of `stacklevel=2` for every warning, as in some cases we just want to display the warning itself, not any additional lines, so this tends to make `warnings` longer for not a clear reason in most of the cases, shouldn't we carefully evaluate this one before including it as a default?

alvarobartt: As of https://docs.python.org/3/library/warnings.html#warnings.warn I understand this is useful when we wrap up the `warnings` inside a function, but this is not the case right?

gabrielmbmb: `ruff` was giving the B024 (`flake8-bugbear` rule) because the `stacklevel` was not being explicitly provided. I've updated the code and now the calls to `warnings.warn` that didn't have a stacklevel now they have `stacklevel=1`

tomaarsen: All `stacklevel` does is pick which code line is used alongside the warning text, it doesn't define the size of the stack trace. At least, if I'm not mistaken. With `stacklevel=2`, you tend to see the line of code that the user wrote which caused the warning, which is often okay I think.

alvarobartt: Yes I know, but sometimes the line where the `warnings` is raised is not relevant e.g.

```python
if deprecated_arg is not None:
  warnings.warn("`deprecated_arg` is deprecated", DeprecationWarning, stacklevel=2)
```

Here the user will see the line `if deprecated_arg is not None`, which doesn't add any relevance and/or context to the `warnings` message as it's pretty self explanatory, this is why I meant we should revisit this carefully to add `stacklevel=2` just when relevant 😄

alvarobartt: Anyway I'm still open to discussion in case you guys think that setting `stacklevel=2` as default makes more sense in most of the cases

tomaarsen: I think the default is showing the `warnings.warn` line, which is kind of useless. In your example, I would rather see e.g. 

```
DeprecationWarning: `deprecated_arg` is deprecated
   output = a_func_with_warning(deprecated_arg) in demo.py#7
```
or however it formats it exactly. Regardless - that it shows where the user made the mistake.
But then we indeed need to go case-by-case to see where we need `stacklevel=2` or where we need e.g `stacklevel=3`.

Thoughts?

alvarobartt: Sure, then we can talk tomorrow and decide whether to go ahead with `stacklevel=2` or the default for the moment, and then we can start using `stacklevel=3` when needed

pyproject.toml
```diff
@@ -92,7 +93,7 @@ integrations = [
     # TODO: some backward comp. problems introduced in 0.5.0. 0.13 does not match with setfit
     "huggingface_hub >= 0.5.0,< 0.13",
     # Version 0.12 fixes a known installation issue related to `sentencepiece` and `tokenizers`, more at https://github.com/flairNLP/flair/issues/3129
-    "flair >= 0.12.2",  # Version 0.12.2 relaxes the `huggingface_hub` dependency
+    "flair >= 0.12.2",               # Version 0.12.2 relaxes the `huggingface_hub` dependency
```
tomaarsen, line 96: Accidental tab?

gabrielmbmb: my TOML formatter is adding it automatically and I added the line accidentally. I'll remove it.

gabrielmbmb:
as @tomaarsen suggested, I've added `flake8-comprehension` and `flake8-pytest-style`. `pyupgrade` could be a nice addition, but I think that it could be added in another PR, as we will need to select which rules of `pyupgrade` to apply so everything works fine with old Python versions. For example, `pyupgrade` will update the code to [PEP604](https://peps.python.org/pep-0604/) which works in `python>=3.10`:

```python
from typing import Union
from pydantic import BaseModel

class MyModel(BaseModel):
    number: Union[int, None] = None
```

will get upgraded to:

```python

from pydantic import BaseModel

class MyModel(BaseModel):
    number: int | None = None
```

and in `python<=3.10`, `pydantic` won't be able to parse it, so we need to keep the old syntax so it's compatible in all the versions supported by `argilla`.

tomaarsen:
Makes sense. With `pyupgrade` you would have to specify `--py37-plus` in the configs somewhere. We can have a look at this some other time.

alvarobartt:
I think it's a nice chance to have another look into this, and eventually include it right before the 1.15.0 release, WDYT? @gabrielmbmb I feel like if we don't tackle this now then we'll have more conflicts and will never be tackled...

PR #3481, opened 2023-07-31
chore: prepare tests with postgresql: 1 clean code and type defs

frascuchon:
## Description

This is a first step of a series of changes to support running server tests using a PostgreSQL dataset. Since a lot of changes will be applied. I will push them grouped in different PRs  to the target branch `tests/configure-server-tests-with-postgresql`

This PR removes some unused code and reviews some imports and type definitions. Also, some logging warnings have been replaced by user warnings, for those cases which make more sense.

NOTE: Probably, tests will fail in these PRs. They will be fixed later

### DIFF
--- a/tests/database.py
+++ b/tests/database.py
@@ -30,5 +30,7 @@ def get_task() -> asyncio.Task:
     return task
 
 
-TestSession = async_scoped_session(async_sessionmaker(class_=AsyncSession, expire_on_commit=False, future=True), get_task)
+TestSession = async_scoped_session(
+    async_sessionmaker(class_=AsyncSession, expire_on_commit=False, future=True), get_task
+)
 SyncTestSession = orm.scoped_session(orm.sessionmaker(class_=orm.Session, expire_on_commit=False))
--- a/tests/database.py
+++ b/tests/database.py
@@ -31,6 +31,6 @@ def get_task() -> asyncio.Task:
 
 
 TestSession = async_scoped_session(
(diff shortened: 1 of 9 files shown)


tests/database.py
```diff
@@ -29,5 +30,5 @@ def get_task() -> asyncio.Task:
     return task
 
 
-TestSession = async_scoped_session(orm.sessionmaker(class_=AsyncSession, expire_on_commit=False), get_task)
+TestSession = async_scoped_session(async_sessionmaker(class_=AsyncSession, expire_on_commit=False, future=True), get_task)
```
gabrielmbmb:
```suggestion
TestSession = async_scoped_session(async_sessionmaker(class_=AsyncSession, expire_on_commit=False), get_task)
```

`future=True` is only needed if `sqlalchemy<=2.0` but you want to use the 2.0 API style.

tests/database.py
```diff
@@ -29,5 +30,7 @@ def get_task() -> asyncio.Task:
     return task
 
 
-TestSession = async_scoped_session(orm.sessionmaker(class_=AsyncSession, expire_on_commit=False), get_task)
+TestSession = async_scoped_session(
+    async_sessionmaker(class_=AsyncSession, expire_on_commit=False, future=True), get_task
```
gabrielmbmb:
```suggestion
    async_sessionmaker(expire_on_commit=False, future=True), get_task
```

gabrielmbmb (approved):
LGTM!

PR #3482, opened 2023-07-31
refactor: Using sync database URL for alembic migrations

frascuchon:
# Description

⚠️ This PR must be merged after #3481 

This PR changes how the alembic script creates the engine connection by using a "sync" version of the database URL. This will help to launch db migrations from tests.

PR #3483, opened 2023-07-31
tests: Define tests/unit folder

frascuchon:
# Description
⚠️ This PR must be merged after #3482

This PR defines the `tests/unit` folder, which includes unit tests for the server and client part. Those tests will check single component responsibility, as and specific endpoint functionality, or model schema validation. and won't contain tests checking more complex workflows or checking client-server integrations.

The old tests locations will be removed in future PRs


tests/unit/client/feedback/training/test_trainer.py
tomaarsen, line 1: Is there a motivation for including this file in `unit` while the feedback trainer tests are in `integration`?

tomaarsen: For the rest I think the split is pretty sensible

frascuchon: Wrong place. This should be part of the integration tests. Thanks!

tomaarsen (approved):
This split seems sensible to me. I don't know if there's any documentation or something that points to any test files, if so, we'd have to update them.

frascuchon:
> This split seems sensible to me. I don't know if there's any documentation or something that points to any test files, if so, we'd have to update them.

I will review docs later in a separate PR

PR #3484, opened 2023-07-31
tests: Define tests/integration folder

frascuchon:
# Description

⚠️ This PR must be merged after #3483 

As changes covered in #3483, this PR defines a `tests/integration` folder, containing tests that cover more complex workflows, or tests running code of different components (mainly the client-server ones).

Integration tests will contain tests checking more complex workflows, affecting the client and server integration. In the future, these tests may be launched using a running instance of the Argilla server, instead of "mocking" a local connection from the client.


tomaarsen (approved):
This split seems sensible to me. I don't know if there's any documentation or something that points to any test files, if so, we'd have to update them.

PR #3485, opened 2023-07-31
tests: Remove old test files

frascuchon:
# Description
⚠️ This PR must be merged after #3484 

This PR removes old test files since they are properly moved  to the `tests/unit` and `tests/integration` folders in #3483  and #3484


tomaarsen (approved):
I think this is sensible. The diff looks a bit weird though, but that's only because the other PRs haven't been merged yet. Afterwards, this PR should be just 1 commit I think.

PR #3486, opened 2023-07-31
fix: update date in PostreSQL

frascuchon:
# Description
⚠️ This PR cannot be merged until #3485 is merged
 
This PR solves a problem with updates using PostgreSQL as a database. The `updated_at` value is not properly updated using the native DB `func.now()` directive. This PR sends a new value on each update using the `datatime.utcnow()` function.

Also, some missing error handlers for the server configuration have been included since now, the unit tests are using the `httpx.AsyncClient` instead of the `starlette.testclient.TestClient`, which can do this work [automatically](https://www.starlette.io/testclient/):

> By default the TestClient will raise any exceptions that occur in the application. Occasionally you might want to test the content of 500 error responses, rather than allowing client to raise the server exception. In this case you should use client = TestClient(app, raise_server_exceptions=False).


gabrielmbmb (approved):
LGTM!

PR #3487, opened 2023-07-31
tests:  support for running server tests with postgresql database

### LINKED ISSUE #3291: [REFACTOR] Split unit and integration tests and use different CI/CD conditions

frascuchon:
# Description

The main target of this PR is to provide support for running server tests using a PostgreSQL database. This may help to find different behavior depending on which dataset was used.

Since Argilla is using an async connection for the database and the standard `TestClient` client is not fully compatible when using other than SQLite async connections in tests (see [here](https://fastapi.tiangolo.com/advanced/async-tests/#httpx) for more details), the tests have been split into 2 groups:  `unit` and `integration` 

The first group will affect to tests checking single component functionality, and the `httpx.AsyncClient` will be used for those server-related tests.

The second group (integration) will check a more complex workflow, where several components may be affected. For now, those tests are using a mocked variant of the fastapi `TestClient` but ideally, this will change in the future to use a real connection to an up-and-running Argilla server.

Also, this new test organization allows us to optimize the build pipeline by only launching the unit tests using different search backend versions, which reduces the execution time.

Closes #3291

**Type of change**

**How Has This Been Tested**

Several tests with multiple configurations have been launched locally 

**Checklist**


tests/unit/client/feedback/conftest.py
```diff
@@ -0,0 +1,56 @@
...
+    MultiLabelQuestion,
+    RankingQuestion,
+    RatingQuestion,
+    TextField,
+    TextQuestion,
+)
+
+
+@pytest.fixture
+def rating_question_payload():
```
gabrielmbmb:
```suggestion
def rating_question_payload() -> dict:
```

tests/unit/client/feedback/conftest.py
```diff
@@ -0,0 +1,56 @@
...
+    return {
+        "name": "label",
+        "description": "label",
+        "required": True,
+        "values": ["1", "2"],
+    }
+
+
+@pytest.fixture
+def label_question_payload():
```
gabrielmbmb:
```suggestion
def label_question_payload() -> dict:
```

tests/unit/client/feedback/conftest.py
```diff
@@ -0,0 +1,56 @@
...
+    return {
+        "name": "label",
+        "description": "label",
+        "required": True,
+        "labels": ["1", "2"],
+    }
+
+
+@pytest.fixture
+def ranking_question_payload():
```
gabrielmbmb:
```suggestion
def ranking_question_payload() -> dict:
```

gabrielmbmb (approved):

PR #3497, opened 2023-08-02
fix: parse label for dataset settings for text and token classification tasks

### LINKED ISSUE #3495: [BUG] `dataclass` has no type validation so errors end up being hard to debug
**Describe the bug**

This is more an improvement than a bug, because some errors end up being cryptic and hard to debug, because the client is just throwing an HTTP 500 error code while the actual traceback is in the server.

This is due to some missing validation when `dataclass` is used with no `__post_init__` validation, instead of `pydantic.BaseModel`.

There are more situations when this can break, but this is one of those when using `TextClassificationSettings`

```python
import argilla as rg

rg.init(api_url="...", api_key="...")

rg.configure_dataset("my-dataset", workspace=rg.get_workspace(), settings=rg.TextClassificationSettings(label_schema=["A", "B"]))
```

Fails because it's a `List[str]` instead of `Set[str]`, but the error displayed in the Python client is `AttributeError: 'str' object has no attribute 'name'`, but the traceback cannot be followed from the Python client, just from the server.

**Expected behavior**

A validation layer to make sure that everything's right before calling the API, to be able to show a nice error message to the user, and better error handling, to avoid showing API exceptions straight away in the client, but a formatted version of those.

**Environment:**
 - Argilla Version [e.g. 1.0.0]: 1.13.0 (but it's N/A)
 - Docker Image (optional) [e.g. argilla:v1.0.0]: argilla:v1.13.0

**Additional context**

This ends up with users asking for help on issues they cannot easily debug, so this would also avoid people from running into these issues or just having a simple way to solve it themselves.

frascuchon:
# Description

This PR prevents errors described in #3495 by parsing labels into strings. The backend parsing is also changed to apply label normalization after the type conversion.

Closes #3495

**Type of change**

**How Has This Been Tested**

The issue code snippet has been launched locally and verified that it works with new changes.

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ These are the section headers that we use:
 
 ## [Unreleased]
 
+### Fixed
+
+- Dataset settings (text and token classification) are properly parsed to strings for clients and for the backend endpoint (Closes [3495](https://github.com/argilla-io/argilla/issues/3495)).
+
 ### Added
 
 - Added `PATCH /api/v1/fields/{field_id}` endpoint to update the field title and markdown settings ([#3421](https://github.com/argilla-io/argilla/pull/3421)).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ These are the section headers that we use:
 
 ### Fixed
 
-- Dataset settings (text and token classification) are properly parsed to strings for clients and for the backend endpoint (Closes [3495](https://github.com/argilla-io/argilla/issues/3495)).
+  - `TextClassificationSettings` and `TokenClassificationSettings` labels are properly parsed to strings both in the Python client and in the backend endpoint (Closes [#3495](https://github.com/argilla-io/argilla/issues/3495)).
 
 ### Added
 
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,10 @@ These are the section headers that we use:
 
 ## [Unreleased]
 
+### Fixed
+
+- Dataset settings (text and token classification) are properly parsed to strings for clients and for the backend endpoint (Closes [3495](https://github.com/argilla-io/argilla/issues/3495)).
```
```suggestion
  - `TextClassificationSettings` and `TokenClassificationSettings` labels are properly parsed to strings both in the Python client and in the backend endpoint (Closes [#3495](https://github.com/argilla-io/argilla/issues/3495)).
```

src/argilla/client/apis/datasets.py
```diff
@@ -59,6 +59,9 @@ class LabelsSchemaSettings(_AbstractSettings):
 
     label_schema: Set[str]
 
+    def __post_init__(self):
+        self.label_schema = set([str(label) for label in self.label_schema])
```
alvarobartt, line 67:
```suggestion
        if not isinstance(self.label_schema, (set, list, tuple)):
            raise ValueError(f"`label_schema` is of type={type(self.label_schema)}, but type=set is preferred, and also both type=list and type=tuple are allowed.")
        self.label_schema = set([str(label) for label in self.label_schema])
```

alvarobartt (approved):
LGTM thanks! Just some minor comments before merging 👍🏻

PR #3502, opened 2023-08-03
fix: `DatasetConfig.to_yaml` and `DatasetConfig.from_yaml` to ignore `UUID`

alvarobartt:
# Description

As recently reported by @nataliaElv we had some issues while calling `FeedbackDataset.from_huggingface` as the `argilla.yaml` file couldn't be read properly. This was due to an incompatibility between the UUIDs and PyYAML, which is now patched.

We will no longer be dumping the `id` fields for both `fields` and `questions` since those are useless outside Argilla and we cannot use those to recreate a `FeedbackDataset` in Argilla, and when loading those back from HuggingFace we'll remove the `id` first with a regex to avoid loading issues.

**Type of change**

**How Has This Been Tested**

- [X] Move `test_config.py` to unit tests as it's not calling the API
- [X] Add some assertions in `test_config.py` to ensure that the `id`s are neither dumped nor loaded from `argilla.yaml`

**Checklist**

### DIFF
--- a/src/argilla/client/feedback/config.py
+++ b/src/argilla/client/feedback/config.py
@@ -12,6 +12,7 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+import re
 import warnings
 from typing import List, Optional
 
@@ -43,6 +44,7 @@ class DatasetConfig(BaseModel):
 
     @classmethod
     def from_yaml(cls, yaml: str) -> "DatasetConfig":
+        yaml = re.sub(r"(\n\s*|)id: !!python/object:uuid\.UUID\s+int: \d+", "", yaml)
         return cls(**load(yaml, Loader=SafeLoader))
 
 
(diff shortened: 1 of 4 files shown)


src/argilla/client/feedback/config.py
```diff
@@ -39,10 +40,11 @@ class DatasetConfig(BaseModel):
     guidelines: Optional[str] = None
 
     def to_yaml(self) -> str:
-        return dump(self.dict())
+        return dump(self.dict({"fields": {"__all__": {"id"}}, "questions": {"__all__": {"id"}}}))
 
     @classmethod
     def from_yaml(cls, yaml: str) -> "DatasetConfig":
+        yaml = re.sub(r"(\n\s*|)id: !!python/object:uuid\.UUID\s+int: \d+", "", yaml)
```
gabrielmbmb, line 47: this is for backward compatibility with the YAML files that contains the exported IDs, right?

gabrielmbmb (approved):

src/argilla/client/feedback/config.py
alvarobartt: Yes, as we included the `id` for both fields and questions at https://github.com/argilla-io/argilla/pull/3370, it was being automatically dumped but then we couldn't load it back. So since this bug was introduced in 1.13.0, we'll leave this here at least until 1.15.0 😄

PR #3506, opened 2023-08-04
:sparkles: FEAT : focus on first question on click inside form

### LINKED ISSUE #3503: focus on first question when user click on form

keithCuniah:
# Description

When user click on form (but not inside a question), the focus goes to first question

Closes #3503 

**Type of change**


frontend/components/commons/forms/questions-form/QuestionsForm.component.vue
```diff
@@ -1,10 +1,12 @@
 <template>
   <form
     id="formId"
```
damianpumar: if you use ref, I think that id is not necessary, right ?

PR #3508, opened 2023-08-04
fix: `push_to_argilla` to return `RemoteFeedbackDataset` without re-assigning class

alvarobartt:
# Description

This PR adds a `warnings.warn` message to let the users know that the returned object from `FeedbackDataset.push_to_argilla` needs to be handled, otherwise the dataset instance will remain local, as `push_to_argilla` with no arguments won't do anything.

So on, before we had to `push_to_argilla` a new `FeedbackDataset` with name and/or workspaces, and then we could `push_to_argilla` with no arguments to push the updates, but we no longer need to do that, as now we can re-use the returned `FeedbackDataset` from `push_to_argilla` to have a `FeedbackDataset` fully integrated with Argilla.

```diff
import argilla as rg

dataset = rg.FeedbackDataset(...)
- dataset.push_to_argilla(name="my-dataset", workspace="my-workspace")
+ remote_dataset = dataset.push_to_argilla(name="my-dataset", workspace="my-workspace")

dataset.add_records(...)
- dataset.push_to_argilla()
```

**Type of change**

**How Has This Been Tested**

- [X] Catch returned object in `push_to_argilla` and handle `DeprecationWarning`

**Checklist**

### DIFF
--- a/src/argilla/client/feedback/dataset/mixins.py
+++ b/src/argilla/client/feedback/dataset/mixins.py
@@ -192,8 +192,19 @@ class ArgillaToFromMixin:
             client=httpx_client, id=argilla_id, show_progress=show_progress, question_mapping=question_name2id
         )
 
-        self.__class__ = RemoteFeedbackDataset
-        self.__init__(
+        warnings.warn(
+            "Calling `push_to_argilla` no longer implies that the `FeedbackDataset` can"
+            " be updated in Argilla. If you want to push a `FeedbackDataset` and then"
+            " update it in Argilla, you need to catch the returned object and use it"
+            " instead as `remote_ds = ds.push_to_argilla(...)`. Otherwise, you can just"
+            " call `push_to_argilla` and then `from_argilla` to retrieve the"
+            " `FeedbackDataset` from Argilla, so the current `FeedbackDataset` can be"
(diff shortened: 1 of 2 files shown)


src/argilla/client/feedback/dataset/mixins.py
```diff
@@ -192,8 +192,19 @@ def push_to_argilla(
             client=httpx_client, id=argilla_id, show_progress=show_progress, question_mapping=question_name2id
         )
 
-        self.__class__ = RemoteFeedbackDataset
-        self.__init__(
+        warnings.warn(
+            "Calling `push_to_argilla` no longer implies that the `FeedbackDataset` can"
+            " be updated in Argilla. If you want to push a `FeedbackDataset` and then"
+            " update it in Argilla, you need to catch the returned object and use it"
+            " instead as `remote_ds = ds.push_to_argilla(...)`. Otherwise, you can just"
```
gabrielmbmb:
```suggestion
            " instead: `remote_ds = ds.push_to_argilla(...)`. Otherwise, you can just"
```

I think this way is more clear, feel free to apply or discard this suggestion

gabrielmbmb (approved):
LGTM!

PR #3511, opened 2023-08-05
fix: `publish_dataset` to check `required=True` on at least one field/question

alvarobartt:
# Description

This PR fixes a bug with the `PUT /api/v1/datasets/{dataset_id}/publish` endpoint, as there was a missing check on the `required` flag for each field/question which was allowing to publish `FeedbackTask` datasets with no required fields and/or questions.

**Type of change**

**How Has This Been Tested**

- [X] Update existing unit/integration tests
- [X] Add unit/integration tests with `required=True` and `required=False`

**Checklist**


tests/unit/server/api/v1/test_datasets.py
```diff
@@ -3183,19 +3183,45 @@ async def test_publish_dataset_without_fields(
...
+        assert (await db.execute(select(func.count(Record.id)))).scalar() == 0
+
+    async def test_publish_dataset_without_required_questions(
+        self, async_client: "AsyncClient", db: "AsyncSession", owner_auth_header: dict
+    ) -> None:
+        dataset = await DatasetFactory.create()
+        await TextFieldFactory.create(dataset=dataset, required=True)
+        await RatingQuestionFactory.create(dataset=dataset)
 
         response = await async_client.put(f"/api/v1/datasets/{dataset.id}/publish", headers=owner_auth_header)
```
gabrielmbmb: We should just keep 1 of these unit tests, they are testing the same thing

gabrielmbmb (approved):
LGTM!

tests/unit/server/api/v1/test_datasets.py
alvarobartt: It was to ensure that the error is being raised on both no questions (former check) and no required questions (updated check)

alvarobartt: Maybe we can just parametrize the `required` value for both

PR #3527, opened 2023-08-08
[DOCS] Removed integration dependencies from docs

### LINKED ISSUE #3352: [DOCS] replace `autodoc` with `sphinx-autoapi`
## Which page or section is this issue related to?
`autodoc` requires to install all extra integration dependencies, but `sphinx-autoapi` does not, hence it is faster and easier to maintain.

davidberenstein1957:
# Description
docs: corrected typo's in references
docs: added hyperlinks to references
docs: removed integration dependencies by mocking

Closes #3352 

**Type of change**

(Remember to title the PR according to the type of change)

- [X] Documentation update

**How Has This Been Tested**

(Please describe the tests that you ran to verify your changes.)

- [X] `sphinx-autobuild` (read [Developer Documentation](https://docs.argilla.io/en/latest/community/developer_docs.html#building-the-documentation) for more details)

**Checklist**

### DIFF
--- a/src/argilla/training/peft.py
+++ b/src/argilla/training/peft.py
@@ -32,9 +32,8 @@ class ArgillaPeftTrainer(ArgillaTransformersTrainer):
     if sys.version_info < (3, 9):
         raise Exception("Must be using Python 3.9 or higher or PEFT won't work.")
 
-    require_version("peft")
-
     def init_training_args(self):
+        require_version("peft")
         super().init_training_args()
 
         self.lora_kwargs = {
--- a/src/argilla/training/setfit.py
+++ b/src/argilla/training/setfit.py
@@ -26,12 +26,11 @@ class ArgillaSetFitTrainer(ArgillaTransformersTrainer):
     _logger = logging.getLogger("ArgillaSetFitTrainer")
     _logger.setLevel(logging.INFO)
 
-    require_version("torch")
-    require_version("datasets")
-    require_version("transformers")
-    require_version("setfit>=0.6")
-
     def __init__(self, *args, **kwargs):
+        require_version("torch")
+        require_version("datasets")
+        require_version("transformers")
+        require_version("setfit>=0.6")
         if kwargs.get("model") is None and "model" in kwargs:
             kwargs["model"] = "all-MiniLM-L6-v2"
         self.multi_target_strategy = None
--- a/src/argilla/training/spacy.py
+++ b/src/argilla/training/spacy.py
@@ -30,8 +30,6 @@ class _ArgillaSpaCyTrainerBase(ArgillaTrainerSkeleton):
     _logger = logging.getLogger("ArgillaSpaCyTrainer")
     _logger.setLevel(logging.INFO)
 
-    require_version("spacy")
-
     def __init__(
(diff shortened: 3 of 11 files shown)


src/argilla/training/peft.py
```diff
@@ -32,9 +32,8 @@ class ArgillaPeftTrainer(ArgillaTransformersTrainer):
     if sys.version_info < (3, 9):
         raise Exception("Must be using Python 3.9 or higher or PEFT won't work.")
 
-    require_version("peft")
-
     def init_training_args(self):
+        require_version("peft")
```
alvarobartt, line 36: Can we do something like the following instead?

```python
def __init__(self, *args, **kwargs) -> None:
    require_version("peft")
    super().__init__(*args, **kwargs)
```

src/argilla/training/setfit.py
```diff
@@ -26,12 +26,11 @@ class ArgillaSetFitTrainer(ArgillaTransformersTrainer):
...
-    require_version("torch")
-    require_version("datasets")
-    require_version("transformers")
-    require_version("setfit>=0.6")
-
     def __init__(self, *args, **kwargs):
+        require_version("torch")
+        require_version("datasets")
+        require_version("transformers")
+        require_version("setfit>=0.6")
```
alvarobartt, line 33: I can work on a PR to extend the `require_version` and `requires_version` decorator to accept also lists of dependencies rather than just one by one, as I think it would be more readable

src/argilla/training/spacy.py
```diff
@@ -267,21 +266,25 @@ def predict(
...
     def __init__(self, freeze_tok2vec: bool = False, **kwargs) -> None:
-        """Initialize the `ArgillaSpaCyTrainer` class.
+        """
+        Initialize the ArgillaSpaCyTrainer class.
 
         Args:
-            freeze_tok2vec: A `bool` indicating whether to freeze the `tok2vec` weights
-                during the training. Defaults to False.
-            **kwargs: The `ArgillaSpaCyTrainerBase` arguments.
+            freeze_tok2vec (bool, optional): A flag indicating whether to freeze the tok2vec weights
```
alvarobartt: Is the type here needed? I think this may be redundant as we're also defining it in the `__init__` args type-hints? I think we should decide either to keep it for every docstring or remove it from every docstring, but I assume that during the Python reference generation the type-hints are re-used if defined already, so there should be no need

alvarobartt: e.g.

```suggestion
            freeze_tok2vec: A flag indicating whether to freeze the tok2vec weights
```

alvarobartt (approved):
LGTM! Maybe we should think about the considerations of defining the dependencies as `autodoc_mock_imports` as it's going to be harder to maintain, but I guess it's better to make the docs building time shorter

davidberenstein1957:
> LGTM! Maybe we should think about the considerations of defining the dependencies as autodoc_mock_imports as it's going to be harder to maintain, but I guess it's better to make the docs building time shorter

Alternatively, we would need to add them to integrations extra's and install that, however, this takes twice as long to build the docs sadly.

PR #3529, opened 2023-08-09
chore: add new `FeedbackDataset` datasets in `quickstart.Dockerfile`

### LINKED ISSUE #3492: Update datasets for quickstart image
Add 2 new datasets in the quickstart docker images:

- https://huggingface.co/datasets/argilla/oasst_response_quality
- https://huggingface.co/datasets/argilla/oasst_response_comparison

Note: this may affect to the demo setup

alvarobartt:
# Description

This PR adds some new `FeedbackDataset` datasets in the `quickstart.Dockerfile` to be downloaded via `docker/scripts/load_data.py` when using the `"full"` value for loading datasets into Argilla as part of the quickstart.

Closes #3492

### DIFF
--- a/docker/scripts/load_data.py
+++ b/docker/scripts/load_data.py
@@ -147,18 +147,19 @@ class LoadDatasets:
         )
 
     @staticmethod
-    def load_curated_dolly_en():
-        print("Loading databricks-dolly-15k-curated-en dataset")
+    def load_feedback_dataset_from_huggingface(repo_id: str, split: str = "train", samples: int = 100):
+        print(f"Loading {repo_id} dataset")
 
         # Load dataset from the hub
-        dataset = rg.FeedbackDataset.from_huggingface("argilla/databricks-dolly-15k-curated-en", split="train[:100]")
+        dataset = rg.FeedbackDataset.from_huggingface(
+            repo_id=repo_id, split=f"{split}{f'[:{samples}]' if samples else ''}"
+        )
 
         # Remove the `responses` from every `FeedbackRecord` to upload just the `fields`
         for record in dataset.records:
             record.responses = []
 
-        # Read in dataset, assuming it's a dataset for token classification
-        dataset.push_to_argilla(name="databricks-dolly-15k-curated-en")
+        dataset.push_to_argilla(name=repo_id.split("/")[-1])
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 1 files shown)


alvarobartt:
@gabrielmbmb at #3492, @frascuchon mentioned that changing the `FeedbackDataset` datasets that were downloaded as part of the `quickstart.Dockerfile` may affect the demo setup, could you please check?

I've also included those when using the `"full"` arg, so `"simple"` will still just download `argilla/databricks-dolly-15k-curated-en`

gabrielmbmb (approved):

gabrielmbmb:
hey @alvarobartt, the demo instance won't be affected by this change! it has a different setup :)

nataliaElv:
Is it too difficult to point this PR to main instead of the release? Just so that we can have it available ASAP and not being dependent on the release. Thanks!

alvarobartt:
Hi @nataliaElv, yes :( that's an issue since the original branch was created on top of `releases/1.14.0` instead of `main`, so we cannot just change the base, I can either create a new PR or just merge main, rebase the commits and force push just the changes related to `load_data.py`. Anyway, what's your idea on using it? You can use it from this branch already, and, alternatively, we could just push it to the Docker Hub manually if that helps?

nataliaElv:
> Hi @nataliaElv, yes :( that's an issue since the original branch was created on top of `releases/1.14.0` instead of `main`, so we cannot just change the base, I can either create a new PR or just merge main, rebase the commits and force push just the changes related to `load_data.py`. Anyway, what's your idea on using it? You can use it from this branch already, and, alternatively, we could just push it to the Docker Hub manually if that helps?

In this case I would suggest to close this PR without merging and creating a new PR that points to `main`. Please check that `argilla/oasst_response_quality` gets loaded correctly (it's giving me issues when trying to load it in a notebook).

alvarobartt:
> In this case I would suggest to close this PR without merging and creating a new PR that points to `main`. Please check that `argilla/oasst_response_quality` gets loaded correctly (it's giving me issues when trying to load it in a notebook).

Hi @nataliaElv the issue that you mention is already solved at #3539, and regarding the PR, we cannot point to `main` as it won't work since it's relying on the fix created on top of `releases/1.14.0`, but we can guide you on how to build and run it locally if that works!

Issue #3395, opened 2023-07-13
[FEATURE] Manually change the width of record card & feedback form


nataliaElv:
**Is your feature request related to a problem? Please describe.**
Sometimes records are more crowded than the feedback form and in other datasets it might be the other way. 

**Describe the solution you'd like**
I'd like to see on hover a dragging bar between the record card and the feedback form so that I can adjust manually the width of these two elements.

davidberenstein1957:
@damianpumar @nataliaElv I noticed this got lost so I added it to the sprint again.

Amelie-V:
I close this issue which will be tackle with other layout improvements we are progressively integrating

PR #3536, opened 2023-08-09
✨ Add resizable annotation mode

damianpumar:
Here is the first approach for resize annotation mode sides, the issue is https://github.com/argilla-io/argilla/issues/3395

https://github.com/argilla-io/argilla/assets/7398909/eb10b61b-5f25-4837-af1b-4589dee934bb

### DIFF
--- a/frontend/components/commons/resizable/Resizable.vue
+++ b/frontend/components/commons/resizable/Resizable.vue
@@ -2,7 +2,7 @@
   <div class="resizable">
     <div class="resizable__left"><slot name="left" /></div>
 
-    <div class="resizable__bar" id="resizableBar">
+    <div class="resizable__bar" ref="resizableBar">
       <span class="resizable__bar-button">:::</span>
     </div>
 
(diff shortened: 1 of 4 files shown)


dvsrepo:
this is amazing @damianpumar 🚀

damianpumar:
> this is amazing @damianpumar 🚀

Is just the first approach, tomorrow I will present it to the team.

frontend/components/commons/resizable/Resizable.vue
```diff
@@ -0,0 +1,118 @@
...
+    return {
+      prevClientX: 0,
+      prevLeftWith: 0,
+      resizer: null,
+      leftSide: null,
+      rightSide: null,
+    };
+  },
+  mounted() {
+    this.resizer = document.getElementById("resizableBar");
```
leiyre: Could we use $ref here instead of id?

damianpumar: Yes, true! 
Thanks to catch it!

davidberenstein1957:
Looking great! Two minor comments. Would it be possible to:
- keep the state of the positon of the scrollbar during a page refresh?
- also change the position when hovering on the middle line apart from the button?

nataliaElv:
Closes #3395

PR #3541, opened 2023-08-10
✏️  Auto save record

### LINKED ISSUE #3648: 🔃 [UI] Support auto save record

damianpumar:
- [X] When the user applies any modification after 2 seconds, we save that record as a draft.
- [X] When the user press `Ctrl + s` or `Cmd + s` we save automatically the record as a draft.
- [X] When the user discards a record and change something in that record discarded, we change the status to pending and automatically mark as a draft.
- [X] For submitted records, never save as a draft.
- [X] If the user answers any question and at the moment that the draft is saving click on submit, the record must be saved as a submitted.
- [x] If the record has a draft status must show the spark icon when the answer is equal to suggested
- [x] For pending records, the user can move to prev or next without blocking (no showing the toaster)
- [x] Ensure the metrics works correctly, we need to show in a pending section all records pending and also the drafts one.
- [x] Show the toaster when the user goes outside this page (not for spa pages, in that case is not necessary).
- [x] E2E for all behavior
- [x] Prevent show the toaster when the user modifies something in `discard` status
- [x] Partial rank the ranking question
- [x] Change toaster message for submitted records


damianpumar:
@keithCuniah, In my last push I just modify something related to this task 
- For pending records, the user can move to prev or next without blocking (no showing the toaster)
Verify if it is necessary to add or modify something. I just remove one line and I think is working, but I'm not sure.

keithCuniah:
@damianpumar 

Just for adding more context. 
The toast was showing when the `form have been touched` and applying one of theses 5 actions : 
- going on `next/prev record` with pagination buttons
- going on `home page` throught clicking in breadcrumbs 
- going to `settings page` throught clicking on corresponding icon
- switching `record status` throught filter dropdown
- applying a `search` query.
- clicking on `refresh` from the sidebar

By adding a clause guard on the QuestionsForm to emit the form status (touched or not) to parents or by bus events affect all theses 6 actions in same time.

frontend/v1/infrastructure/services/useBeforeUnload.ts
```diff
@@ -0,0 +1,13 @@
+class BeforeUnload {
+  confirm() {
+    window.onbeforeunload = () => "NO!";
```
damianpumar: NO? please use the label defined

keithCuniah:
updates regarding toast:

These toast have been **removed** : 
- going on home page throught clicking in breadcrumbs : `Your changes will be lost if you leave the current page`
- going to settings page throught clicking on corresponding icon : `Your changes will be lost if you leave the current page`
- on refresh from sidebar : `Your changes will be lost if you refresh the page`

The next toast messages have been update with `You didn't submit your changes` and the button become `Ignore and continue` :
- going on next/prev record with pagination buttons : `Your changes will be lost if you move to another page`
- switching record status throught filter dropdown : `Your changes will be lost if you move to another view`
- applying a search query : `Your changes will be lost if you apply the search filter`

PR #3552, opened 2023-08-11
fix: cannot update ranking & rating question with `settings`

gabrielmbmb:
# Description

The input payload for updating a question was not consistent across the different types of questions. For the `text`, `label_selection` and `multi_label_selection`, it is necessary to provide the kind of question to be updated including `{"settings": {"type": "question_type"}` in the input payload, while for the `rating` and `ranking` question it was not needed to provide the `settings` and when provided it caused a validation error. This was because this last two questions types didn't have a `pydantic.BaseModel` included in the `QuestionUpdateSettings` discriminator causing the following payloads to not be valid when updating them:

```json
{
   "name": "New Name",
    "settings": {
        "type": "ranking"
    }
}
```

and the only valid payload was:

```json
{
   "name": "New Name"
}
```

To keep the consistency in the input payload for updating a question, this PR include two new `pydantic.BaseModel`, one for rating and one for ranking question, that are included in the `QuestionUpdateSettings` discriminator so the payload mentioned above is valid when updating a `ranking` or `rating` question.

**Type of change**

**How Has This Been Tested**

Unit tests has been updated and manual tests on local environment.

**Checklist**


CHANGELOG.md
```diff
@@ -14,11 +14,13 @@ These are the section headers that we use:
 * "Security" in case of vulnerabilities.
 -->
 
-## [Unreleased]
+## [Unreleased]()
```
alvarobartt, line 17:
```suggestion
## [Unreleased]
```

src/argilla/server/schemas/v1/questions.py
```diff
@@ -92,7 +92,13 @@ class Config:
 
 class TextQuestionSettingsUpdate(UpdateSchema):
     type: Literal[QuestionType.text]
-    use_markdown: bool
+    use_markdown: Optional[bool]
```
alvarobartt, line 95: Shouldn't this just be set to False?

```suggestion
    use_markdown: Optional[bool] = False
```

alvarobartt (approved):
LGTM!

src/argilla/server/schemas/v1/questions.py
alvarobartt: Ok it cannot be False because it's to update the existing, thanks for the internal clarification!

PR #3555, opened 2023-08-11
feat: allow `require_version` and `requires_version` to also accept list of requirements

### LINKED ISSUE #3554: [FEATURE] allow `require_version` and `requires_version` decorator to also accept lists of requirements
There are functions in the codebase that explicitly require a version to proceed, like for example when instantiating some [trainers](https://github.com/argilla-io/argilla/blob/5aa7096141abc94b19dadf0ed811a4a3fce1f1b3/src/argilla/training/setfit.py#L30-L33).
Updating `require_version` and `requires_version` could help improving readability as suggested in [this link](https://github.com/argilla-io/argilla/pull/3527#discussion_r1288090507).

Instead of calling a number of times the function (or the decorator):

```python
require_version("torch")
require_version("datasets")
require_version("transformers")
require_version("setfit>=0.6")
```

A list could be used for such cases:

```python
require_version(["torch", "datasets", "transformers", "setfit>=0.6"])
```

**Additional context**

plaguss:
# Description

This feature extends the functionality of `require_version` and `requires_version` to also accept list of requirements.

Closes #3554.

**Type of change**

**How Has This Been Tested**

Two new unit tests were added to check that both the function and the decorator work as expected with the new addition.

**Checklist**


plaguss:
Also if wanted I can update this PR with the changes to the files where this function is used multiple times in a row. The following program outputs show the files where these functions are called (not all of them should be modified).

```sh
❯ grep -Rc 'src/argilla' -e 'require_version(' --exclude-dir '__pycache__' | grep -e ':0' -v
src/argilla/monitoring/asgi.py:1
src/argilla/training/span_marker.py:3
src/argilla/training/openai.py:1
src/argilla/training/transformers.py:5
src/argilla/training/setfit.py:4
src/argilla/training/spacy.py:5
src/argilla/training/peft.py:1
src/argilla/training/autotrain_advanced.py:2
src/argilla/client/feedback/training/frameworks/setfit.py:4
src/argilla/client/feedback/training/frameworks/spacy.py:3
src/argilla/client/feedback/training/schemas.py:1
src/argilla/client/feedback/dataset/base.py:1
src/argilla/client/datasets.py:2
src/argilla/utils/dependency.py:4
```

```sh
❯ grep -Rc 'src/argilla' -e 'requires_version(' --exclude-dir '__pycache__' | grep -e ':0' -v
src/argilla/labeling/text_classification/label_models.py:6
src/argilla/labeling/text_classification/label_errors.py:1
src/argilla/client/feedback/training/schemas.py:3
src/argilla/client/feedback/integrations/huggingface/dataset.py:5
src/argilla/client/feedback/dataset/base.py:1
src/argilla/client/datasets.py:11
src/argilla/utils/dependency.py:2
src/argilla/server/services/tasks/text_classification/metrics.py:1
```

src/argilla/utils/dependency.py
```diff
@@ -63,22 +63,29 @@ def _compare_versions(
...
 
-def require_version(requirement: str, func_name: Optional[str] = None) -> None:
+def require_version(requirement: Union[str, List[str]], func_name: Optional[str] = None) -> None:
     """
     Perform a runtime check of the dependency versions, using the exact same syntax used by pip.
     The installed module version comes from the *site-packages* dir via *importlib_metadata*.
 
     Args:
-        requirement (`str`): pip style definition, e.g.,  "tokenizers==0.9.4", "tqdm>=4.27", "numpy"
+        requirement (`str` or `List[str]`): pip style definition, e.g.,  "tokenizers==0.9.4", "tqdm>=4.27", "numpy"
```
alvarobartt: I think we can rename this arg to `requirements` and remove the type-hints from the docstring, as those are automatically generated based on the function type-hints 👍🏻 

```suggestion
        requirements: pip style definition of dependencies, e.g. "tokenizers==0.9.4", "tqdm>=4.27", "numpy". Can either be a single dependency or a list of dependencies.
```

src/argilla/utils/dependency.py
```diff
@@ -63,22 +63,29 @@ def _compare_versions(
         )
 
 
-def require_version(requirement: str, func_name: Optional[str] = None) -> None:
+def require_version(requirement: Union[str, List[str]], func_name: Optional[str] = None) -> None:
```
alvarobartt:
```suggestion
def require_version(requirements: Union[str, List[str]], func_name: Optional[str] = None) -> None:
```

alvarobartt: Note that `func_name` should also be handled, so this may be tricky!

Plus ideally we can rename `func_name` to `fn_name` 👍🏻

alvarobartt: If the `fn_name` is not widely used along the codebase, I guess we can even separate it to a different decorator, as that's not the most common scenario, and under the condition of `require_version(requirements=["a", "b", "c", "d"], fn_name=["fn_a", None, None, "fn_d"])` we should define a matching list which is not ideal. So I prefer you to detach this functionality and the arg itself from this decorator and create a separate one named `require_dependency_function` or something similar

alvarobartt: Maybe this can be renamed from `require_version` to `require_dependency` or `require_dependencies`

src/argilla/utils/dependency.py
```diff
@@ -63,22 +63,29 @@ def _compare_versions(
...
     ```python
     require_version("pandas>1.1.2")
     require_version("datasets>1.17.0", "from_datasets")
+    require_version(["pandas>1.1.2", "datasets>1.17.0"])
     ```
     """
 
+    if isinstance(requirement, list):
+        if len(requirement) == 0:
+            return
```
alvarobartt: I think we can raise a `ValueError` exception here

src/argilla/utils/dependency.py
```diff
@@ -130,7 +137,7 @@ def require_version(requirement: str, func_name: Optional[str] = None) -> None:
 _R = TypeVar("_R")
 
 
-def requires_version(requirement: str) -> Callable[[Callable[_P, _R]], Callable[_P, _R]]:
+def requires_version(requirement: Union[str, List[str]]) -> Callable[[Callable[_P, _R]], Callable[_P, _R]]:
```
alvarobartt: I think I prefer `requires_depedencies` over `requires_version`

```suggestion
def requires_dependencies(requirement: Union[str, List[str]]) -> Callable[[Callable[_P, _R]], Callable[_P, _R]]:
```

alvarobartt (commented):
Great work so far @plaguss! Left some minor comments I'd like to discuss, we can chat early next week! Enjoy the weekend

src/argilla/utils/dependency.py
plaguss: Done! I missed the docstrings update. Regarding the autogenerated type-hints, I didn't know that, thank you!

src/argilla/utils/dependency.py
plaguss: makes sense the new name :+1:

src/argilla/utils/dependency.py
plaguss: I will let `require_dependencies` like the renamed decorator

src/argilla/utils/dependency.py
plaguss: The bare return is intended at this point, it was just to stop calling the function when the list was empty, and given the function shouldn't return anything when the dependencies are met, I thought it made sense like this. It's implemented like in the following snippet from the same function, but maybe I should the the list treatment?

```python
# special case
if package == "python":
    got_version = ".".join([str(x) for x in sys.version_info[:3]])
    for op, want_version in wanted.items():
        _compare_versions(op, got_version, want_version, requirement, package, func_name=func_name)
    return
```

src/argilla/utils/dependency.py
plaguss: I will rename `func_name` to `fn_name`! Regarding the handling of the argument, I think we can keep discussing it on the previous conversation.

plaguss: As you guessed, the `fn_name` is not widely used along the  codebase.

- I reviewed the uses of `require_version` and none is making use of `fn_name`.
- With the *decorator* version (`requires_version`) is a bit different as it just takes the `requirements` argument, and grabs the `fn_name` internally from the `func.__name__` attribute:

```python
    def decorator(func: Callable[_P, _R]) -> Callable[_P, _R]:
        @functools.wraps(func)
        def wrapper(*args: _P.args, **kwargs: _P.kwargs) -> _R:
            require_dependencies(requirements, func.__name__)
            return func(*args, **kwargs)

        return wrapper

    return decorator

```

So I think the current implementation assumes that when using the decorator version, it just applies to a single function, like in the following example (it corresponds to the method `score` in `Snorkel` [class](https://github.com/argilla-io/argilla/blob/develop/src/argilla/labeling/text_classification/label_models.py#L609C1-L615C1). 

```python
...
    @requires_version("scikit-learn")
    def score(
        self,
        tie_break_policy: Union[TieBreakPolicy, str] = "abstain",
        output_str: bool = False,
    ) -> Union[Dict[str, float], str]:
        """Returns some scores/metrics of the label model with respect to the annotated records.
...
```

I think that for those cases (the majority), the implementation as is works just fine taking a list of requirements, and a single `fn_name` instead of a list. 
The decorator is also used on `Snorkel` class instead of a method or function, but I think it's just the same logic that for a function.

I think that it could work using a list of `requirements` and a single `fn_name` for the current uses along the codebase, and the current implementation would still be available (single `requirement` with a `fn_name` for the help message), but I understand it may be confusing. 
Anyway, I could detach this to a new `require_dependency_function` but I could use some help with for example the expected use :)

plaguss:
Aside from the pending discussions, the new additions imply some changes to the tests and to the codebase, I will tackle those once the new changes are settled if it's ok.

src/argilla/utils/dependency.py
alvarobartt: Yes I get your point, but at then end we don't want to allow `requires_versions` or `requires_dependencies` to `requires_version([])` as that wouldn't make much sense, that's why I prefer to raise an exception if the decorator is not being used properly

alvarobartt:
Hi @plaguss we would like to merge this before the 1.16.0 release, so I've invited @tomaarsen to review as AFAIK he created the original decorators, so we can all discuss the changes.

* Renamed from `require_version` and `requires_version` to `require_dependency` and `requires_dependencies`, respectively
* Rename `func_name` arg to `fn_name`
* Accept a list of strings as requirements, instead of having to call `requires_version` multiple times

All the above plus we need to decide whether we want to keep the `fn_name` arg (formerly `func_name`) or not, as it's not used in our codebase, and probably won't be used anytime soon, so probably we can just remove it to simplify things.

src/argilla/utils/dependency.py
tomaarsen: I think the `ValueError` makes sense

src/argilla/utils/dependency.py
```diff
@@ -58,36 +58,43 @@ def _compare_versions(
...
+    require_dependencies("datasets>1.17.0", "from_datasets")
+    require_dependencies(["pandas>1.1.2", "datasets>1.17.0"])
     ```
     """
 
+    if isinstance(requirements, list):
+        if len(requirements) == 0:
+            return
+        require_dependencies(requirements[0], func_name=fn_name)
+        return require_dependencies(requirements[1:], func_name=fn_name)
```
tomaarsen: I have no real preference here, but we could also just loop over the requirements and call them with each element rather than this head-tails approach.
E.g.
```python
for requirement in requirements:
    require_dependencies(requirement, fn_name=fn_name)
```

tomaarsen:
```suggestion
        require_dependencies(requirements[0], fn_name=fn_name)
        return require_dependencies(requirements[1:], fn_name=fn_name)
```

src/argilla/utils/dependency.py
```diff
@@ -58,36 +58,43 @@ def _compare_versions(
...
-def require_version(requirement: str, func_name: Optional[str] = None) -> None:
+def require_dependencies(requirements: Union[str, List[str]], fn_name: Optional[str] = None) -> None:
     """
     Perform a runtime check of the dependency versions, using the exact same syntax used by pip.
     The installed module version comes from the *site-packages* dir via *importlib_metadata*.
 
     Args:
-        requirement (`str`): pip style definition, e.g.,  "tokenizers==0.9.4", "tqdm>=4.27", "numpy"
+        requirements: pip style definition of dependencies, e.g. "tokenizers==0.9.4", "tqdm>=4.27", "numpy". Can either be a single dependency or a list of dependencies.
         func_name (`str`, *optional*): what suggestion to print in case of requirements not being met
```
tomaarsen:
```suggestion
        fn_name: what suggestion to print in case of requirements not being met
```

tomaarsen:
Regarding all of the changes:

* Renamed from `require_version` and `requires_version` to `require_dependency` and `requires_dependencies`, respectively

This seems like a reasonable move.

* Rename func_name arg to fn_name

I also like this.

* Accept a list of strings as requirements, instead of having to call `requires_version` multiple times

Quite useful.

---

As for keeping `fn_name`, I think we should, for two reasons:
* Its existence isn't causing a lot of code complexity. It's limited to essentially just one method, where it's only used to format some strings.
* We do use it - we use `requires_version`, i.e. the decorator variant of `require_version`. With the decorator variant, the function/class name is passed to `require_version`.

I think all that remains at this point is replacing all `requires_version` with `requires_dependencies` and replacing `require_version` with `require_dependency`, and merging subsequent `require_dependency` into a list.

- Tom Aarsen

src/argilla/utils/dependency.py
plaguss: I personally think this approach reads better, I will update it

plaguss: Assuming your previous implementation is preferred, this change is not necessary anymore

src/argilla/utils/dependency.py
plaguss: Done!

tomaarsen (approved):
Looks good to me!

alvarobartt (approved):

alvarobartt:
Great job @plaguss! Once you resolve the conflicts I think we can go ahead with the merge, could you also update the PR title to be compliant with the rest of the PRs? i.e. `feat: allow ...` Thanks again 🎉

plaguss:
> Great job @plaguss! Once you resolve the conflicts I think we can go ahead with the merge, could you also update the PR title to be compliant with the rest of the PRs? i.e. `feat: allow ...` Thanks again 🎉

We are ready then!

PR #3582, opened 2023-08-18
feat: add `login` function for storing Argilla credentials

### LINKED ISSUE #3581: [FEATURE] Add `login` function to store credentials and update `rg.init` to use stored credentials

gabrielmbmb:
# Description

This PR includes a new function called `login` that will try to login to the Argilla server using the provided credentials. If it success, the provided credentials will be saved to `~/.cache/argilla/credentials.json` file. The `argilla.client.client.Argilla.__init__` method has been updated to try to use the stored credentials if the `api_url` and `api_key` arguments are not provided.

Closes #3581

**Type of change**

**How Has This Been Tested**

I've added unit tests and tried in a local development environment that `rg.init` works using the three options: providing credentials via function arguments, via environment variables and loading them from the stored file.

**Checklist**


CHANGELOG.md
```diff
@@ -16,6 +16,10 @@ These are the section headers that we use:
 
 ## [Unreleased]()
 
+### Added
+
+- Added `argilla.client.login.login` function to login in an Argilla server and store the credentials locally ([#3582](https://github.com/argilla-io/argilla/pull/3582)).
```
```suggestion
- Added `login` function in `argilla.client.login` to login into an Argilla server and store the credentials locally ([#3582](https://github.com/argilla-io/argilla/pull/3582)).
```

src/argilla/client/client.py
```diff
@@ -104,11 +104,28 @@ def __init__(
...
+            try:
+                credentials = ArgillaCredentials.load()
+                api_url = credentials.api_url
+                api_key = credentials.api_key
+                workspace = credentials.workspace
+                extra_headers = credentials.extra_headers
+            except FileNotFoundError:
+                pass
+
+        api_url = api_url or "http://localhost:6900"
```
alvarobartt: Should we defined the default host as `DEFAULT_API_URL`? I assume we won't change the port at any point, but since this is a default value, aren't we more safe if we just add it under `argilla._constants`?

src/argilla/client/login.py
```diff
@@ -0,0 +1,77 @@
...
+        workspace: The default workspace where the datasets will be created.
+        extra_headers: A dictionary containing extra headers that will be sent to the Argilla server.
+
+    Raises:
+        ValueError: If the login fails.
+    """
+    # Try to login to the server
+    try:
+        init(api_url=api_url, api_key=api_key, workspace=workspace, extra_headers=extra_headers)
+    except UnauthorizedApiError as e:
```
alvarobartt, line 75: This can also fail with `HttpResponseError` in case e.g. the server is not reachable?

alvarobartt: I would suggest something like

```suggestion
    except HttpResponseError as e:
        raise ValueError(f"Could not reach '{api_url}', make sure that the Argilla Server is running and working as expected") from e
    except UnauthorizedApiError as e:
```

src/argilla/client/login.py
```diff
@@ -0,0 +1,77 @@
...
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+from pathlib import Path
+from typing import Dict, Optional
+
+from pydantic import AnyHttpUrl, BaseModel
+
+from argilla.client.api import init
+from argilla.client.sdk.commons.errors import UnauthorizedApiError
```
alvarobartt: Related to a comment mentioned below:

```suggestion
from argilla.client.sdk.commons.errors import HttpResponseError, UnauthorizedApiError
```

src/argilla/client/client.py
```diff
@@ -104,11 +104,28 @@ def __init__(
...
         """
-        api_url = api_url or os.getenv("ARGILLA_API_URL", "http://localhost:6900")
+        from argilla.client.login import ArgillaCredentials
+
+        api_url = api_url or os.getenv("ARGILLA_API_URL")
+        api_key = api_key or os.getenv("ARGILLA_API_KEY")
+        workspace = workspace or os.getenv("ARGILLA_WORKSPACE")
+        extra_headers = extra_headers or {}
+
+        if api_url is None and api_key is None and workspace is None:
```
alvarobartt: Since the `workspace` is optional and has nothing to do with the connection to Argilla itself, shouldn't we just remove the condition from here? e.g. we store the credentials for the `owner` and `workspace=argilla`, but we want to connect to it using `workspace=gabriel`, then we won't be able to reuse those values, right? But we can indeed just assign a different workspace via the same `api_url` and `api_key`, does that make sense?

alvarobartt (approved):
LGTM! I just left some minor comments and suggestions before merging 👍🏻

src/argilla/client/client.py
gabrielmbmb: sure

src/argilla/server/schemas/batch.py
alvarobartt, line 1: I guess this push is unintended, right?

src/argilla/client/client.py
```diff
@@ -104,11 +110,28 @@ def __init__(
                 the headers of argilla client requests, like additional security restrictions. Default: `None`.
 
         """
-        api_url = api_url or os.getenv("ARGILLA_API_URL", "http://localhost:6900")
+        from argilla.client.login import ArgillaCredentials
+
+        api_url = api_url or os.getenv("ARGILLA_API_URL")
+        api_key = api_key or os.getenv("ARGILLA_API_KEY")
+        workspace = workspace or os.getenv("ARGILLA_WORKSPACE")
```
alvarobartt, line 117: Also I think this may be a breaking change if users previously were assuming that no `api_url` and no `api_key` implied the default (local) configuration was being used, while right now the configuration stored in `~/.cache/argilla/credentials.json` will be used instead

alvarobartt: Because the previous line could be translated to

```python
api_url = api_url or os.getenv("ARGILLA_API_URL") or DEFAULT_API_URL
api_key = api_key or os.getenv("ARGILLA_API_KEY") or DEFAULT_API_KEY
```

While the current is translated to

```python
api_url = api_url or os.getenv("ARGILLA_API_URL") or credentials.api_url or DEFAULT_API_URL
api_key = api_key or os.getenv("ARGILLA_API_KEY") or credentials.api_key or DEFAULT_API_KEY
```

So providing `api_url=None` and `api_key=None` implies that the configuration stored in `credentials.json` will be used if None, instead of the one intended for local usage. So either we update the docs to actually reflect this change, or just find a way to use the `credentials.json` over the default configuration e.g. `ARGILLA_USE_CREDENTIALS` env variable

gabrielmbmb: As far as the user doesn't use the `login` function then calling `rg.init()` will use the default credentials

PR #3604, opened 2023-08-21
🐛 Fix/3598 filtering does not work

### LINKED ISSUE #3598: [BUG] Workspace filtering doesn't work as intended
**Describe the bug**

When having more than one workspace starting with the same string e.g. `argilla`, the workspace filtering is not working as intended, since assuming we have the workspaces `argilla`, `argillaworkspace`, `argilla-workspace`, and `argilla_workspace`, the only filter that works as intended is the `argilla` one, while the rest of them will include `argilla` as part of the workspaces being filtered.

This means that the filtering is not looking for exact matches between the string to use as a filter and the workspace names, since when filtering by `argilla-workspace` only the datasets assigned to `argilla-workspace` should be shown, not also the ones in `argilla`.

Note that `argilla` is the one used for showcasing purposes, but it also happens with any workspace naming sharing the same starting string plus something else, which can either be another string, an allowed special symbol plus another string, etc.

Also note that the URL looks fine and the workspace filter is properly included e.g. `https://<URL>/datasets?workspaces=argilla-workspace`

**To Reproduce**

1. Create N workspaces and M datasets assigned to those workspaces
2. Go to Argilla UI where you created those workspaces and datasets
3. Click on the `Workspace` filter, and select the checkboxes individually to see that the results always include `argilla`, and not just the filtered one

**Expected behaviour**

The filter to work with an exact match to avoid including workspaces that are not selected.

**Screenshots**

ui-bug-workspace

**Video**

https://github.com/argilla-io/argilla/assets/36760800/7a440154-6f98-45e3-8fb7-4a211c679449

 - OS [e.g. iOS]: macOS
 - Browser [e.g. chrome, safari]: Arc (based on Chromium)
 - Argilla Version [e.g. 1.0.0]: 1.14.1
 - Docker Image (optional) [e.g. argilla:v1.0.0]: argilla:v1.14.1

damianpumar:


damianpumar:
Good morning @alvarobartt, could you please test your case here?
Obviously, I tested that fix with E2E tests, but if you want to do a double check, feel free!

alvarobartt:
> Good morning @alvarobartt, could you please test your case here? Obviously, I tested that fix with E2E tests, but if you want to do a double check, feel free!

Thanks @damianpumar! I've just tested it and I can confirm that it works 👍🏻 I tested it against https://argilla-quickstart-pr-3604-ki24f765kq-no.a.run.app/

https://github.com/argilla-io/argilla/assets/36760800/e725384b-cc91-449c-ba15-72f251b35304

alvarobartt (approved):

keithCuniah (approved):

Issue #3394, opened 2023-07-13
Signal required questions instead of optional


nataliaElv:
```[tasklist]
### Acceptance criteria
- [x] Signal required questions with a red asterisk `*`
- [x] Remove optional tag `(optional)` from optional questions
- [ ] https://github.com/argilla-io/argilla/issues/3639
```

PR #3608, opened 2023-08-22
:sparkles: FEAT: signal required questions instead optional

keithCuniah:
Show required question with a `red asterisk` instead of showing optional widget

See #3394 

**Type of change**

- New feature

**How Has This Been Tested**

Only feedback task => we should see a `red asterisk` at the end of the title from a required question instead of showing `(optional)` for optional question.

Note : 
All e2e images have to be regenerated since this affect all `feedback task` datasets

### DIFF
--- a/frontend/components/commons/forms/questions-form/fields/shared-components/question-header/QuestionHeader.component.vue
+++ b/frontend/components/commons/forms/questions-form/fields/shared-components/question-header/QuestionHeader.component.vue
@@ -3,7 +3,7 @@
     <span
       class="suggestion-info"
       v-text="title"
-      v-required-field="isRequired && { color: 'red' }"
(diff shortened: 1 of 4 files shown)


frontend/components/commons/forms/questions-form/fields/shared-components/question-header/QuestionHeader.component.vue
```diff
@@ -3,7 +3,7 @@
     <span
       class="suggestion-info"
       v-text="title"
-      v-optional-field="!isRequired"
+      v-required-field="isRequired && { color: 'red' }"
```
damianpumar: color: red?

damianpumar (approved):

PR #3610, opened 2023-08-22
feat: add `RemoteFeedbackDataset.filter_by` and `FilteredRemoteFeedbackDataset`

alvarobartt:
# Description

This PR adds a method named `filter_by` to `RemoteFeedbackDataset` so as to be able to filter based on the `response_status` of any given record in the dataset; as well as refactoring how the `RemoteFeedbackDataset` is structured to detach the base functionality. Additionally, `FilteredRemoteFeedbackDataset` has been created with certain constraints on what the users can do with it, while still being linked to Argilla, but using the specified filter.

**Type of change**

**How Has This Been Tested**

- [x] Add unit and integration tests for the newly included functionality, as well as refactoring the existing ones due to the module refactor applied to `remote.py` (now living under `remote/dataset.py`)

**Checklist**


alvarobartt:
Pending #3613 merge, so that we can directly include the `response_status` arg for the filtering over `RemoteFeedbackDataset`s to be included in the HTTP requests to the Argilla Server when fetching records

src/argilla/client/feedback/dataset/remote/dataset.py
```diff
@@ -0,0 +1,219 @@
...
+                datasets_api_v1.delete_records(
+                    client=self._client,
+                    id=self._dataset_id,
+                    record_ids=[record.id for record in records[start:end]],
+                )
+            except Exception as e:
+                raise RuntimeError("Failed to remove records from Argilla") from e
+
+
+class RemoteFeedbackDataset(RemoteFeedbackDatasetBase[RemoteFeedbackRecords]):
```
frascuchon, line 126: It should be possible to keep the `RemoteFeedbackDataset` as it was and extend the records class and move the filtered logic to a `FiltererdRecordsFeedbackDataset` class? Is any anything other than the filters and restricted method that is specific to this `FilteredFeedbackDataset`?

src/argilla/client/feedback/dataset/remote/base.py
```diff
@@ -248,6 +198,7 @@ def __init__(
         fields: List["AllowedFieldTypes"],
         questions: List["AllowedQuestionTypes"],
         guidelines: Optional[str] = None,
+        filters: Dict[str, Any] = {},
```
frascuchon: Since this `filters` argument will be used only for FilteredFeedbackDatase, can we move from the base and move to the proper class?  something like this:

```python
class RemoteFeedbackDatasetBase:
   pass

class RemoteFeedbackDataset(RemoteFeedbackDatasetBase):
  def __init__(...):
      self._records = RemoteRecords()

class FilteredRemoteFeedbackDataset(RemoteFeedbackDatasetBase):
  def __init__(..., filters):
      super().__init__(...)
      self._records = FiltteredRemoteRecords(filters)

```

alvarobartt: It's been done like this since the `filters` are retrieved internally when fetching records from Argilla and that's part of the base functionality, but we can figure out if there's a cleaner way, but IMO we should make those classes protected `_...` as those are not intended to be used by users

src/argilla/client/feedback/dataset/remote/dataset.py
alvarobartt: Could you elaborate a little bit more? I don't quite understand the question 😩

src/argilla/client/feedback/dataset/remote/base.py
```diff
@@ -30,16 +28,20 @@
...
 
 
 warnings.simplefilter("always", DeprecationWarning)
 
+T = TypeVar("T")
+
 
-class RemoteFeedbackRecords:
-    def __init__(self, dataset: "RemoteFeedbackDataset") -> None:
+class RemoteFeedbackRecordsBase:
```
frascuchon: Do we need this records Base class? Could be possible to define the `FilteredRemoteFeedbackRecords` as specialization of the  `RemoteFeedbackRecords` class?

```python
class FilteredRemoteFeedbackRecords(RemoteFeedbackRecords): 
   def __init__(self, dataset, filters):
      super().__init__(dataset)
      self._filters = filters

   def __len__(self) -> int:
      # specific implementation using filters

   # overriding other methods to use filters
```

src/argilla/client/feedback/dataset/remote/base.py
```diff
@@ -102,7 +105,9 @@ def __len__(self) -> int:
         return response.parsed.records.count
 
     @allowed_for_roles(roles=[UserRole.owner, UserRole.admin])
-    def __getitem__(self, key: Union[slice, int]) -> Union[RemoteFeedbackRecord, List[RemoteFeedbackRecord]]:
+    def _get_records(
+        self, key: Union[slice, int], filters: Dict[str, Any] = {}
```
frascuchon: If this is a base class, I'm not sure should have knowledge of `filters` argument, which is required only for one specific class implementation

src/argilla/client/feedback/dataset/remote/base.py
```diff
@@ -102,7 +105,9 @@ def __len__(self) -> int:
...
+    ) -> Union[RemoteFeedbackRecord, List[RemoteFeedbackRecord]]:
```
frascuchon:
```suggestion
    ) -> Union[RemoteFeedbackRecordBase, List[RemoteFeedbackRecordBase]]:
```

src/argilla/client/feedback/dataset/remote/base.py
```diff
@@ -155,12 +160,13 @@ def __getitem__(self, key: Union[slice, int]) -> Union[RemoteFeedbackRecord, Lis
...
                 offset=offset,
                 limit=limit,
+                **filters,
             ).parsed
             records.extend([self.__parse_record(record) for record in fetched_records.items])
         return records[0] if isinstance(key, int) else records
 
     @allowed_for_roles(roles=[UserRole.owner, UserRole.admin])
-    def __iter__(self) -> Iterator[RemoteFeedbackRecord]:
+    def _iter_records(self, filters: Dict[str, Any] = {}) -> Iterator[RemoteFeedbackRecord]:
```
frascuchon: same here

frascuchon:
```suggestion
    def _iter_records(self, filters: Dict[str, Any] = {}) -> Iterator[RemoteFeedbackRecordBase]:
```

src/argilla/client/feedback/dataset/remote/base.py
alvarobartt: We could do that, but we would need to overwrite every method in the base, which implies that there's not a clear base, just two classes doing different things, that's why I parametrised the base class, but I can turn it into a mixin if that makes more sense

src/argilla/client/feedback/dataset/remote/base.py
alvarobartt: With base here I meant not base functionality but with the default methods, I can rename that to be a mixin separately for the requests to the API

src/argilla/client/feedback/dataset/remote/base.py
alvarobartt: There's no `RemoteFeedbackRecordBase` and the `RemoteFeedbackRecord` is the same on both the `RemoteFeedbackDataset` and the `FilteredRemoteFeedbackDataset` as we want to keep the record-level operations (update and delete)

src/argilla/client/feedback/dataset/remote/base.py
alvarobartt: As previously mentioned, I can move that to a mixin instead to cover all the functionality supported by the API endpoint

alvarobartt: Same here, there's no `RemoteFeedbackRecordBase` class, and all these classes return `RemoteFeedbackRecord` as that's the same on both `RemoteFeedbackDataset` and `FilteredRemoteFeedbackDataset`

src/argilla/client/feedback/dataset/remote/base.py
frascuchon: ups sorry. I saw an extra `s`

src/argilla/client/feedback/dataset/remote/base.py
frascuchon: For me, the suffix `Base` refers to abstract classes, which contain a set of common methods and let the specific ones without implementation. A mixin is more like a way to extend behavior to different classes, right? So, yes if the base class is not defining all the methods, should be better a mixin

src/argilla/client/feedback/dataset/remote/filtered.py
```diff
@@ -0,0 +1,77 @@
...
+    from argilla.client.feedback.schemas.records import FeedbackRecord
+    from argilla.client.feedback.schemas.types import AllowedFieldTypes, AllowedQuestionTypes
+    from argilla.client.workspaces import Workspace
+
+
+class FilteredRemoteFeedbackRecords(RemoteFeedbackRecordsBase, ArgillaRecordsMixin):
+    def __init__(self, dataset: "RemoteFeedbackDataset", filters: Dict[str, Any]) -> None:
+        super().__init__(dataset=dataset)
+
+        self._filters = filters
```
frascuchon: Is there any way to move out from the `ArgillaRecordsMixin` all the `self._filters` related operations and keep them as part of the `FilteredRemoteFeedbackRecords` definition? It would be more readable to keep them here. Otherwise, we're coupling both `RemoteFeedbackRecords` and `FilteredRemoteFeedbackRecords` through the logic defined in the  `ArgillaRecordsMixin` class.

src/argilla/client/feedback/dataset/remote/filtered.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def add_records(
+        self,
+        records: Union["FeedbackRecord", Dict[str, Any], List[Union["FeedbackRecord", Dict[str, Any]]]],
+        show_progress: bool = True,
+    ) -> None:
+        raise NotImplementedError("`add_records` does not work for filtered datasets.")
+
+    def delete_records(self, records: Union["RemoteFeedbackRecord", List["RemoteFeedbackRecord"]]) -> None:
+        raise NotImplementedError("`delete_records` does not work for filtered datasets.")
```
frascuchon: since these methods are defined as shortcuts for `ds.records.add`, `ds.records.delete`  and to keep aligned with the `RemoteFeedbackDataset` implementation,  we could move this to the `FilteredRemoteFeedbackRecords`. 

Otherwise, executing directly `ds.records.add`  or `ds.records.delete` will raise a different error than the shortcut

src/argilla/client/feedback/dataset/remote/mixins.py
```diff
@@ -0,0 +1,120 @@
...
+    @allowed_for_roles(roles=[UserRole.owner, UserRole.admin])
+    def __len__(self: Union["RemoteFeedbackRecords", "FilteredRemoteFeedbackRecords"]) -> int:
+        """Returns the number of records in the current `FeedbackDataset` in Argilla."""
+        try:
+            response = datasets_api_v1.get_metrics(client=self._client, id=self._dataset_id)
+        except Exception as e:
+            raise Exception(
+                f"Failed while getting the metrics from the current `FeedbackDataset` in Argilla with exception: {e}"
+            ) from e
+        return response.parsed.records.count
```
frascuchon: This method does not work for filtered datasets, since the filtered must be applied. So, no much sense to include as part of the mixin

src/argilla/client/feedback/dataset/remote/mixins.py
```diff
@@ -0,0 +1,120 @@
...
+
+        records = []
+        for offset, limit in zip(offsets, limits):
+            fetched_records = datasets_api_v1.get_records(
+                client=self._client,
+                id=self._dataset_id,
+                offset=offset,
+                limit=limit,
+                **({} if not hasattr(self, "_filters") else self._filters),
+            ).parsed
```
frascuchon: If you extract this into an abstract method, and let FilteredRecords and RemoteRecords implement their own version, the rest of the code can be shared and without requiring the filters to check in the shared code.

src/argilla/client/feedback/dataset/remote/mixins.py
```diff
@@ -0,0 +1,120 @@
...
+        """Iterates over the `FeedbackRecord`s of the current `FeedbackDataset` in Argilla."""
+        current_batch = 0
+        while True:
+            batch = datasets_api_v1.get_records(
+                client=self._client,
+                id=self._dataset_id,
+                offset=FETCHING_BATCH_SIZE * current_batch,
+                limit=FETCHING_BATCH_SIZE,
+                **({} if not hasattr(self, "_filters") else self._filters),
+            ).parsed
```
frascuchon: You can use the same extracted method here. No need for logic duplication (filters) and the component becomes more extensible.

tests/integration/client/feedback/dataset/test_filtered.py
```diff
@@ -0,0 +1,89 @@
...
+        await RecordFactory.create_batch(dataset=dataset, size=10)
+        user = await UserFactory.create(role=role, workspaces=[dataset.workspace])
+
+        api.init(api_key=user.api_key)
+        remote_dataset = FeedbackDataset.from_argilla(id=dataset.id)
+        filtered_dataset = remote_dataset.filter_by(response_status=status)
+        assert isinstance(filtered_dataset, FilteredRemoteFeedbackDataset)
+
+        with pytest.raises(NotImplementedError):
+            filtered_dataset.delete_records(remote_dataset.records[0])
```
frascuchon: we should check also the shortcutted method `filtered_dataset.records.delete`

tests/integration/client/feedback/dataset/test_filtered.py
```diff
@@ -0,0 +1,89 @@
...
+
+        with pytest.raises(NotImplementedError):
+            filtered_dataset.add_records(FeedbackRecord(fields={text_field.name: "test"}))
```
frascuchon: `filtered_dataset.records.add`

src/argilla/client/feedback/dataset/remote/dataset.py
frascuchon: Forge this comment.

src/argilla/client/feedback/dataset/remote/base.py
```diff
@@ -280,10 +231,10 @@ def __init__(
         self._name = name
         self._workspace = workspace
 
-        self._records: RemoteFeedbackRecords = RemoteFeedbackRecords(dataset=self)
+        self._records = self.records_cls(dataset=self, **kwargs)
```
frascuchon, line 156: We should let each implementation initialize its own records class.

src/argilla/client/feedback/dataset/remote/mixins.py
alvarobartt: I'm adding the complete payload to the metrics endpoint to use the other values, not the `count`, as those can be easily combined with the filters e.g. `submitted` shows the count for the submitted responses

tests/integration/client/feedback/dataset/test_filtered.py
alvarobartt: Yes, I'll add some other `NotImplemented` exceptions for those

src/argilla/client/feedback/dataset/remote/base.py
alvarobartt: But then in the `RemoteFeedbackDatasetBase` we cannot include any method that relies on `_records` as we won't have those, and we would need to refine those twice

src/argilla/client/feedback/dataset/remote/filtered.py
alvarobartt: But why should we do that? I mean the `response_status` is an arg of the `get_records` method in the SDK; additionally, doing that would imply duplicating functionality and the base class ends up being non-sense

alvarobartt: IMO that was the point on using the mixin, to be able to reuse those, otherwise we would need to redefine the magic methods in the class inheriting from those, and also to define two alternate version of the same code where the only change is an extra arg

src/argilla/client/feedback/dataset/remote/mixins.py
```diff
@@ -0,0 +1,109 @@
...
+
+        records = []
+        for offset, limit in zip(offsets, limits):
+            fetched_records = datasets_api_v1.get_records(
+                client=self._client,
+                id=self._dataset_id,
+                offset=offset,
+                limit=limit,
+                **({} if not hasattr(self, "_filters") else self._filters),
+            ).parsed
```
alvarobartt:
```suggestion
            fetched_records = self._fetch_records(offset=..., limit=...)
```

tests/integration/client/feedback/dataset/test_filtered.py
```diff
@@ -0,0 +1,89 @@
...
+        await TextQuestionFactory.create(dataset=dataset, required=True)
+        await RecordFactory.create_batch(dataset=dataset, size=10)
+        user = await UserFactory.create(role=role, workspaces=[dataset.workspace])
+
+        api.init(api_key=user.api_key)
+        remote_dataset = FeedbackDataset.from_argilla(id=dataset.id)
+        filtered_dataset = remote_dataset.filter_by(response_status=status)
+        assert isinstance(filtered_dataset, FilteredRemoteFeedbackDataset)
+
+        with pytest.raises(NotImplementedError):
```
gabrielmbmb: can we add the `match` argument here?

tests/integration/client/feedback/dataset/test_filtered.py
```diff
@@ -0,0 +1,89 @@
...
+            filtered_dataset.delete_records(remote_dataset.records[0])
+
+        with pytest.raises(NotImplementedError):
```
gabrielmbmb: ditto

tests/integration/client/feedback/dataset/test_filtered.py
```diff
@@ -0,0 +1,89 @@
...
+            filtered_dataset.add_records(FeedbackRecord(fields={text_field.name: "test"}))
+
+        with pytest.raises(NotImplementedError):
```
gabrielmbmb: ditto

src/argilla/client/feedback/dataset/remote/base.py
```diff
@@ -75,7 +77,10 @@ def __repr__(self) -> str:
         )
         return f"[{','.join([str(record) for record in self][:2])}, ...]"
 
-    def __parse_record(self, record: "FeedbackItemModel") -> RemoteFeedbackRecord:
+    def __len__(self) -> None:
```
frascuchon:
```suggestion
    def __len__(self) -> int:
```

src/argilla/client/feedback/dataset/remote/dataset.py
```diff
@@ -0,0 +1,191 @@
...
+            raise Exception(
+                f"Failed while getting the metrics from the current `FeedbackDataset` in Argilla with exception: {e}"
+            ) from e
+        return response.parsed.records.count
+
+    def _fetch_records(self, offset: int, limit: int) -> "FeedbackRecordsModel":
+        """Fetches a batch of records from Argilla."""
+        return datasets_api_v1.get_records(
+            client=self._client,
+            id=self._dataset_id,
```
frascuchon: why not just self._dataset.id?

src/argilla/client/feedback/dataset/remote/base.py
```diff
@@ -90,154 +95,24 @@ def __parse_record(self, record: "FeedbackItemModel") -> RemoteFeedbackRecord:
...
 
-    @allowed_for_roles(roles=[UserRole.owner, UserRole.admin])
     def add(
         self,
-        records: Union[FeedbackRecord, Dict[str, Any], List[Union[FeedbackRecord, Dict[str, Any]]]],
+        records: Union["FeedbackRecord", Dict[str, Any], List[Union["FeedbackRecord", Dict[str, Any]]]],
         show_progress: bool = True,
     ) -> None:
-        """Pushes a list of `FeedbackRecord`s to Argilla.
+        raise NotImplementedError(f"`records.add` not implemented for `{self.__class__.__name__}`.")
```
frascuchon: could we define the `add`, `update`, and `delete` as abstract methods and let each implementation notify with a more friendly message? For example, for the filtered datasets, I can imagine a message informing users that filtered datasets do not support adding or deleting data.
```suggestion
    @abstract
    def add(...)
        pass
```

src/argilla/client/feedback/dataset/remote/filtered.py
```diff
@@ -0,0 +1,76 @@
...
+    def __init__(self, dataset: "RemoteFeedbackDataset", filters: Dict[str, Any]) -> None:
+        super().__init__(dataset=dataset)
+
+        self._filters = filters
+
+    def _fetch_records(self, offset: int, limit: int) -> "FeedbackRecordsModel":
+        """Fetches a batch of records from Argilla."""
+        return datasets_api_v1.get_records(
+            client=self._client,
+            id=self._dataset_id,
```
frascuchon: why not just `self._dataset.id`?

src/argilla/client/feedback/dataset/remote/filtered.py
```diff
@@ -0,0 +1,76 @@
...
+
+    import httpx
+
+    from argilla.client.feedback.dataset.remote.dataset import RemoteFeedbackDataset
+    from argilla.client.feedback.schemas.types import AllowedFieldTypes, AllowedQuestionTypes
+    from argilla.client.sdk.v1.datasets.models import FeedbackRecordsModel
+    from argilla.client.workspaces import Workspace
+
+
+class FilteredRemoteFeedbackRecords(RemoteFeedbackRecordsBase):
```
frascuchon, line 32: I think we can implement the `add` and `delete` notifying to users in the same fashion as for the `dataset.delete`: "Adding records do not work for filtered datasets"...

frascuchon (approved):
It's okay for me. Just one thing. Should we add some sections in docs apart from the Python client refs? Maybe we can refactor an existing tutorial applying filters and start using the `filter_by` method.

cc  @davidberenstein1957

davidberenstein1957:
@frascuchon I agree. I think every added functionality to the Python API should be represented in the docs and that this needs to be done before wrapping up the PR.

I see there also does not seem any reference to the inner workings of the `RemoteFeedbackDataset`besides declaring variables as `remote_dataset`. 

@alvarobartt Could you add:
- A reference to the existence/usage of the `RemoteFeedbackDataset` everywhere in the docs `/guides/llms/practical_guides/*.md` where push/pull(and later clone) is used? They can also have a reference to the URL mentioned in the point below. Similarly, we could add this to the lower/higher tabs in the docs.
image
- A more concrete definition of each of the pull/push/clone methods to `/guides/llms/practical_guides/export_dataset.md`. For example, the `pull` in this page still references a normal `FeedbackDataset`. Is that correct?
- A more explicit definition of the `filter_by` and a lack of a sort method here `/guides/llms/practical_guides/export_dataset.md`. 
- `filter_by` can also be referenced in `/guides/llms/practical_guides/collect_responses.md` and `/guides/llms/practical_guides/create_dataset.html`
- If possible, update the tutorials `https://docs.argilla.io/en/latest/guides/llms/examples/*.ipynb`

I will add a work item about representing the change above and adding it to the `ArgillaTrainer`

davidberenstein1957:
Thanks! LFTM. Not sure if we want to change Feedback Dataset => `FeedbackDataset` throughout all places in the docs?

PR #3620, opened 2023-08-23
refactor: move `Enum`s to `enums`

gabrielmbmb:
# Description

This PR moves all the `Enum` classes of the Argilla server to the `argilla.server.enums` module in order to avoid circular imports, and to have one single file in which find all the enums (they are used across all the server codebase).

**Type of change**

**How Has This Been Tested**

**Checklist**


alvarobartt (approved):
LGTM, makes a lot of sense! My only concern is whether we should also include the `V1` at the end of each `Enum` to differentiate v0 from v1 enums, or whether we should just split those in different modules, WDYT?

PR #3623, opened 2023-08-23
fix: response with `draft` status was returning `422`

### LINKED ISSUE #3612: [BUG] Update response endpoint returning 422 when status sent is `draft`
**Describe the bug**
@damianpumar detected a bug in which the `PUT /api/v1/responses/{response_id}` endpoint returns a `422` HTTP status code when trying to update a response and the sent status is `draft`.

This bug is being caused because we don't have a `DraftResponseStatusCode` class:

https://github.com/argilla-io/argilla/blob/aeb5b8e9d1dbb52429bc3d1bec508c44659e7e9b/src/argilla/server/schemas/v1/responses.py#L50-L60

**Expected behavior**
The endpoint return a `200` HTTP status code when called and the sent status is `draft`.

gabrielmbmb:
# Description

This PR fix an issue in which the server returned a `422` when trying to update the response of a record with `draft` status.

In addition, this PR updates `check_response` method used to validate the responses received for a question to also received the `status` of the response. This allows to execute partial validations of the responses depending on the received status. For example, for the `RankingQuestion` we only want to validate that all the possible options have been ranked if the status is `submitted`, otherwise, we just want to validate that the options that have been ranked are valid:

Closes #3612

**Type of change**

- [x] New feature

**How Has This Been Tested**

In a local development environment:

- [x] Record response can be updated with `draft` status

**Checklist**

### DIFF
--- a/src/argilla/server/models/questions.py
+++ b/src/argilla/server/models/questions.py
@@ -16,7 +16,7 @@ from typing import Any, Generic, List, Literal, Optional, Protocol, TypeVar, Uni
 
 from pydantic import BaseModel, Field
 
-from argilla.server.enums import QuestionType
+from argilla.server.enums import QuestionType, ResponseStatus
 
 try:
     from typing import Annotated
@@ -29,7 +29,7 @@ class ResponseValue(Protocol):
 
 
 class BaseQuestionSettings(BaseModel):
-    def check_response(self, response: ResponseValue):
+    def check_response(self, response: ResponseValue, status: Optional[ResponseStatus] = None):
         pass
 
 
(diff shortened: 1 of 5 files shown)


src/argilla/server/models/questions.py
```diff
@@ -29,15 +29,15 @@ class ResponseValue(Protocol):
 
 
 class BaseQuestionSettings(BaseModel):
-    def check_response(self, response: ResponseValue):
+    def check_response(self, response: ResponseValue, status: Optional[ResponseStatus] = None):
```
gabrielmbmb, line 32: @frascuchon I've added the `status` argument so partial validations depending on the response status can be done. What do you think?

PR #3641, opened 2023-08-28
fix: add `list_workspaces` under `rg.*` and fix some type-hints

alvarobartt:
# Description

This PR adds the function `list_workspaces` so that it can be used as `import argilla as rg; rg.list_workspaces` since it was the only function under `Argilla` not included as part of the `__init__`. Besides that, since we now have `GET /api/v1/me/workspaces` there's no need to filter the returned workspaces by the users on it, as we can use `/api/v1/me` instead.

Additionally, some type-hints have been included and some others have been fixed, specially the ones not putting `Optional` surrounding the actual type when the value defaults to None e.g. `workspace: str = None` when it should be `workspace: Optional[str] = None`; and also some cases where return type-hints were missing or were wrong.

**Type of change**

**How Has This Been Tested**

- [x] Add integration tests for `list_workspaces`

**Checklist**


frascuchon (approved):
👍

PR #3674, opened 2023-08-30
:construction: POC change component from option api to composition api

keithCuniah:
- Change QuestionsForm and QuestionHeader components from `optionApi` to `compositionApi`
- move business logic inside external composables in the domain application
Note:
- to get the value of a reactive `variable` inside the setup hook, we need to pass through `variable.value` (see `questionAreCompletedCorrectly.value`)
- In vue3, bus event have been removed. that's means it's not possible to use `$root.$on`, `$root.$emit`, `$root.$on`.
The vue community recomend to not use bus event or use an external librairy as `mitt` (see [official documentation](https://github.com/developit/mitt#install))
- between Vue 2 and 3, some hooks have been replaced : 
          `created` => `setup`
          `mounted` => `onMounted`
          `destroyed` => `onUnMounted`

Next step : 
=> replace native bus event from nuxt by `mitt` or manage logic with `store`

**Type of change**

**How Has This Been Tested**

- Feedback task


keithCuniah:
this PR will be replaced by this one #3778

PR #3677, opened 2023-08-30
🛣️ Feat/improve queue annotation

### LINKED ISSUE #3645: 🐛 [BUG] Improve annotation queue
When the user goes to the annotation page with specific criteria received from the backend a queue with records of the same status. However, when the user starts annotating, they might submit or discard records. During this process, the backend queue becomes different from what we have on the frontend side.

### Following the example below:
T0: (Initial state when the user arrives at the annotation page) (suppose that we receive a batch of three records from the backend)
Frontend queue: [pending1, pending2, pending3]
Backend queue: [pending1, pending2, pending3, pending4, pending5, ..., pendingN]

T1: (The user submits the first record in the queue)
Frontend queue: [submitted, pending, pending] ← When the user moves to page number four, we will send a request to the backend to get the content of page number four. Let's look at the state of the backend queue at that moment.
Backend queue: [submitted] - [pending2, pending3, pending4, pending5]  ← Here, pending4 never appears on the frontend because it is in the third position in the queue. We receive a new batch of records like that: [pending5, pending6, pending7]

### Proposed solution:
In this case, we send a request to the backend for the page based on the new queue. Assuming the previous example, we should request page number 3 instead of 4, since on the frontend we have only two records pending.

```
// Each record on the frontend corresponds to a page position. Sending the page number of the last record plus one could fix this issue when the user moves forward.
return lastRecord.page + 1

// If the user moves backward, we return the page number of the first record minus one. However, in this case, we only need to find one element, not a batch.
return firstRecord.page - 1
```

damianpumar:
- [x] Update changelog


keithCuniah (approved):

dvsrepo:
Looking forward to test this! It's highly important for annotation UX

keithCuniah (approved):

CHANGELOG.md
```diff
@@ -26,6 +26,10 @@ These are the section headers that we use:
 - Added `users list` command to list users ([#3688](https://github.com/argilla-io/argilla/pull/3688)).
 - Added `datasets push-to-hub` command to push a `FeedbackDataset` from Argilla into the HuggingFace Hub ([#3685](https://github.com/argilla-io/argilla/pull/3685)).
 
+### Fixed
+
+- Fixed `Annotation queue` ([#3677](https://github.com/argilla-io/argilla/pull/3677))
```
```suggestion
- Fixed `Pending queue` pagination problems when during data annotation ([#3677](https://github.com/argilla-io/argilla/pull/3677))
```

frascuchon (approved):

PR #3694, opened 2023-09-01
build: skip CI trigger when PR is draft

### LINKED ISSUE #3466: [CI] Skip CI trigger when PR is created as a draft
## Description

To avoid triggering the CI on draft PRs, we should be including something along the lines of `if: github.event.pull_request.draft == false`, this way we would avoid triggering the CI when not needed/desired. As in most of the cases creating a draft PR implies that the PR is still WIP, and the CI can be tackled later once the PR is published.

alvarobartt:
# Description

This PR skips the CI trigger from `package.yml` when the PR is still a draft, but keeps the trigger for opened, reopened, synchronize, and ready_for_review. The same filtering is applied for the `dependency_review.yml` file.

Closes #3466 

**Type of change**

**How Has This Been Tested**

- [X] Created this PR as a draft to see that the CI is skipped


gabrielmbmb (approved):
LGTM!

PR #3702, opened 2023-09-04
fix: default behaviour for `visible_labels` in `LabelQuestion` and `MultiLabelQuestion`

### LINKED ISSUE #3693: Align the default `visible_options`/`visible_labels` value with the options/labels list length
The default value for `visible_labels`  is set to `20` even when the number of configured labels is lower. This can be confusing and could generate some inconsistent scenarios.

The Python client should adapt the logic to improve this as follows:
- Set `visible_labels=20` when `labels.length > 20`
- Set `visible_labels=None` when `labels.length <=20`

alvarobartt:
# Description

This PR fixes the default behaviour for the `visible_labels` field of the `_LabelQuestion` schema parsing, so that instead of setting the default to 20, now it's set to the correct value which is 20 just if not provided (`visible_labels=None`) and `len(labels) > 20`.

So on, the behaviour is explained based on the following scenarios:

* `len(labels) > 20`, and `visible_labels = undefined` -> `warnings.warn` and `visible_options=20`
* `len(labels) < 21`, and `visible_labels = undefined` -> `visible_options=None`
* `len(labels) < 21`, and `visible_labels > 20` -> `warnings.warn` and `visible_options=None`
* `len(labels) > 20`, and `visible_labels < 21` -> `visible_options = visible_labels`
* `len(labels) < 3`, and `visible_labels > 3` -> `warnings.warn` and `visible_options = None`
* `len(labels) >= visible_labels` -> `visible_options = visible_labels`

Closes #3693

**Type of change**

**How Has This Been Tested**

- [X] Add unit tests to check that the scenarios mentioned above are successfully matched for different combinations of `labels` and `visible_labels`
- [X] Add unit tests to check that the expected `warnings.warn` messages are being shown

**Checklist**

### DIFF
--- a/src/argilla/client/feedback/schemas/questions.py
+++ b/src/argilla/client/feedback/schemas/questions.py
@@ -167,7 +167,7 @@ class _LabelQuestion(QuestionSchema, LabelMappingMixin):
                 visible_options = total_options
             else:
                 warnings.warn(
-                    f"`labels={values.get('labels')} has less than 3 labels, so `visible_labels`"
+                    f"`labels={values.get('labels')}` has less than 3 labels, so `visible_labels`"
                     " will be set to `None`, which means that all the labels will be visible.",
                     UserWarning,
                     stacklevel=1,
--- a/src/argilla/client/feedback/schemas/questions.py
+++ b/src/argilla/client/feedback/schemas/questions.py
@@ -170,16 +170,16 @@ class _LabelQuestion(QuestionSchema, LabelMappingMixin):
                 visible_labels = None
         else:
             visible_labels = values.get("visible_labels")
-            total_options = len(values.get("labels", []))
-            if visible_labels and visible_labels > total_options:
-                if total_options >= 3:
+            total_labels = len(values.get("labels", []))
+            if visible_labels and visible_labels > total_labels:
+                if total_labels >= 3:
                     warnings.warn(
                         f"`visible_labels={visible_labels}` is greater than the total number"
-                        f" of labels ({total_options}), so it will be set to `{total_options}`.",
+                        f" of labels ({total_labels}), so it will be set to `{total_labels}`.",
                         UserWarning,
                         stacklevel=1,
                     )
-                    visible_labels = total_options
+                    visible_labels = total_labels
                 else:
                     warnings.warn(
                         f"`labels={values.get('labels')}` has less than 3 labels, so `visible_labels`"
--- a/src/argilla/client/feedback/schemas/questions.py
+++ b/src/argilla/client/feedback/schemas/questions.py
@@ -131,14 +131,15 @@ class _LabelQuestion(QuestionSchema, LabelMappingMixin):
             the list must contain at least two unique labels. Additionally, `labels` can
             also be a dictionary of labels, where the keys are the labels, and the values
             are the labels that will be shown in the UI.
-        visible_labels: The number of visible labels in the UI. Defaults to 20, and must
-            be 3 or greater.
+        visible_labels: The number of visible labels in the UI. Defaults to undefined,
+            which means that it will be automatically set, otherwise it must be either None
+            which means all the labels will be shown, or 3 or greater.
     """
 
     labels: Union[conlist(str, unique_items=True, min_items=2), Dict[str, str]]
-    visible_labels: Optional[conint(ge=3)] = None
+    visible_labels: Union[Literal["undefined"], conint(ge=3), None] = "undefined"
 
-    @validator("labels", always=True)
+    @validator("labels", pre=True, always=True)
     def labels_dict_must_be_valid(cls, v: Union[List[str], Dict[str, str]]) -> Union[List[str], Dict[str, str]]:
         if isinstance(v, dict):
             assert len(v.keys()) > 1, "ensure this dict has at least 2 items"
@@ -148,40 +149,47 @@ class _LabelQuestion(QuestionSchema, LabelMappingMixin):
     @root_validator(skip_on_failure=True)
     def update_settings(cls, values: Dict[str, Any]) -> Dict[str, Any]:
         values["settings"]["type"] = values.get("type")
+
         if isinstance(values.get("labels"), dict):
             values["settings"]["options"] = [
                 {"value": key, "text": value} for key, value in values.get("labels").items()
             ]
-        if isinstance(values.get("labels"), list):
+        elif isinstance(values.get("labels"), list):
             values["settings"]["options"] = [{"value": label, "text": label} for label in values.get("labels")]
-        visible_options = values.get("visible_labels")
-        total_options = len(values["settings"]["options"])
-        if visible_options and visible_options > total_options:
-            if total_options >= 3:
+
+        if values.get("visible_labels") == "undefined":
+            if len(values.get("labels", [])) > 20:
                 warnings.warn(
-                    f"`visible_labels={visible_options}` is greater than the total number"
-                    f" of labels ({total_options}), so it will be set to `{total_options}`.",
+                    "Since `visible_labels` has not been provided and the total number"
+                    " of labels is greater than 20, `visible_labels` will be set to `20`.",
                     UserWarning,
                     stacklevel=1,
                 )
-                visible_options = total_options
+                visible_labels = 20
             else:
-                warnings.warn(
-                    f"`labels={values.get('labels')}` has less than 3 labels, so `visible_labels`"
-                    " will be set to `None`, which means that all the labels will be visible.",
-                    UserWarning,
-                    stacklevel=1,
-                )
-                visible_options = None
-        if not visible_options and total_options > 20:
-            warnings.warn(
-                "Since `visible_labels` has not been provided or is `None`, and the total number"
-                " of labels is greater than 20, `visible_labels` will be set to `20`.",
-                UserWarning,
-                stacklevel=1,
-            )
-            visible_options = 20
-        values["settings"]["visible_options"] = visible_options
+                visible_labels = None
+        else:
+            visible_labels = values.get("visible_labels")
+            total_options = len(values.get("labels", []))
+            if visible_labels and visible_labels > total_options:
+                if total_options >= 3:
+                    warnings.warn(
+                        f"`visible_labels={visible_labels}` is greater than the total number"
+                        f" of labels ({total_options}), so it will be set to `{total_options}`.",
+                        UserWarning,
+                        stacklevel=1,
+                    )
+                    visible_labels = total_options
+                else:
+                    warnings.warn(
+                        f"`labels={values.get('labels')}` has less than 3 labels, so `visible_labels`"
+                        " will be set to `None`, which means that all the labels will be visible.",
+                        UserWarning,
+                        stacklevel=1,
+                    )
+                    visible_labels = None
+        values["settings"]["visible_options"] = visible_labels
+
         return values
 
 
--- a/src/argilla/client/feedback/schemas/questions.py
+++ b/src/argilla/client/feedback/schemas/questions.py
@@ -135,7 +135,7 @@ class _LabelQuestion(QuestionSchema, LabelMappingMixin):
     """
 
     labels: Union[conlist(str, unique_items=True, min_items=2), Dict[str, str]]
-    visible_labels: Optional[conint(ge=3)] = 20
+    visible_labels: Optional[conint(ge=3)] = None
 
     @validator("labels", always=True)
     def labels_dict_must_be_valid(cls, v: Union[List[str], Dict[str, str]]) -> Union[List[str], Dict[str, str]]:
--- a/src/argilla/client/feedback/schemas/questions.py
+++ b/src/argilla/client/feedback/schemas/questions.py
@@ -160,13 +160,17 @@ class _LabelQuestion(QuestionSchema, LabelMappingMixin):
             if total_options >= 3:
                 warnings.warn(
                     f"`visible_labels={visible_options}` is greater than the total number"
-                    f" of labels ({total_options}), so it will be set to `{total_options}`."
+                    f" of labels ({total_options}), so it will be set to `{total_options}`.",
+                    UserWarning,
+                    stacklevel=1,
                 )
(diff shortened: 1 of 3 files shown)


src/argilla/client/feedback/schemas/questions.py
```diff
@@ -153,9 +154,26 @@ def update_settings(cls, values: Dict[str, Any]) -> Dict[str, Any]:
...
-        values["settings"]["visible_options"] = values.get(
-            "visible_labels"
-        )  # `None` is a possible value, which means all labels are visible
+        visible_options = values.get("visible_labels")
+        total_options = len(values["settings"]["options"])
+        if visible_options and visible_options > total_options:
+            if total_options >= 3:
+                warnings.warn(
+                    f"`visible_labels={visible_options}` is greater than the total number"
+                    f" of labels ({total_options}), so it will be set to `{total_options}`.",
```
frascuchon: This message can apply to whenever the `visible_label` is set to a value greater than `total_options`. Not sure to understand the `3`-number's logic

src/argilla/client/feedback/schemas/questions.py
```diff
@@ -153,9 +154,26 @@ def update_settings(cls, values: Dict[str, Any]) -> Dict[str, Any]:
...
+                visible_options = total_options
+            else:
+                warnings.warn(
+                    f"`labels={values.get('labels')}` has less than 3 labels, so `visible_labels`"
+                    " will be set to `None`, which means that all the labels will be visible.",
+                    UserWarning,
+                    stacklevel=1,
+                )
+                visible_options = None
+        values["settings"]["visible_options"] = visible_options
```
frascuchon: I'm missing the default value when `len(labels) > 20` and `visible_labels` is not provided. The behavior described in #3693 is:

```python 
if `visible_labels` not in `values`:
   visible_options = 20 if total_options > 20 else None
else:
   visible_options = values["visible_labels"]
   if  visible_options > total_options:
      visible_options = total_options:
      warnings.warn(...)
values["settings"]["visible_options"] = visible_options
```

tests/unit/client/feedback/schemas/test_questions.py
```diff
@@ -73,6 +74,30 @@ def test_label_question_errors(
...
+            "\`visible_labels=4\` is greater than the total number of labels \(3\), so it will be set to \`3\`.",
+        ),
+        (
+            {"name": "a", "description": "a", "required": True, "labels": ["a", "b"], "visible_labels": 3},
+            UserWarning,
+            "\`labels=\['a', 'b'\]\` has less than 3 labels, so \`visible_labels\` will be set to \`None\`, which means that all the labels will be visible.",
+        ),
+    ],
+)
+def test_label_question_warnings(
```
frascuchon: We should add a test using the default `20` value and options with fewer elements.

tests/unit/client/feedback/schemas/test_questions.py
```diff
@@ -81,23 +106,23 @@ def test_label_question_errors(
...
             {
                 "type": "label_selection",
                 "options": [{"value": "a", "text": "A"}, {"value": "b", "text": "B"}],
-                "visible_options": 20,
+                "visible_options": None,
             },
         ),
         (
-            {"name": "a", "description": "a", "required": True, "labels": ["a", "b"], "visible_labels": 5},
+            {"name": "a", "description": "a", "required": True, "labels": ["a", "b"], "visible_labels": 3},
```
frascuchon, line 124: this sample has already been tested in the previous test function, right? I would say to remove from here.

tests/unit/client/feedback/schemas/test_questions.py
```diff
@@ -73,6 +74,30 @@ def test_label_question_errors(
...
+        ),
+    ],
+)
+def test_label_question_warnings(
+    schema_kwargs: Dict[str, Any],
+    expected_warning: Warning,
+    expected_warning_message: str,
+) -> None:
+    with pytest.warns(expected_warning, match=expected_warning_message):
+        LabelQuestion(**schema_kwargs)
```
frascuchon: we should check also the expected value in these cases.

tests/unit/client/feedback/schemas/test_questions.py
```diff
@@ -108,6 +133,28 @@ def test_label_question_errors(
                 "visible_options": None,
             },
         ),
+        (
+            {"name": "a", "description": "a", "required": True, "labels": list(string.ascii_lowercase)},
+            {
+                "type": "label_selection",
+                "options": [{"value": a, "text": a} for a in string.ascii_lowercase],
+                "visible_options": None,
```
frascuchon: This expected value should be 20 instead of None since the number of different options is greater than 20

frascuchon: In terms of readability, I think it would be best to use a numeric range as sample data:
```python
..."labels": [str(i) for i in range(0, 21)],
{..., "options" : [{"value": str(i), "text": str(i)} for i range(0,21)],
```

tests/unit/client/feedback/schemas/test_questions.py
```diff
@@ -108,6 +133,28 @@ def test_label_question_errors(
...
+                "name": "a",
+                "description": "a",
+                "required": True,
+                "labels": list(string.ascii_lowercase),
+                "visible_labels": len(string.ascii_lowercase) * 2,
+            },
+            {
+                "type": "label_selection",
+                "options": [{"value": a, "text": a} for a in string.ascii_lowercase],
+                "visible_options": len(string.ascii_lowercase),
```
frascuchon: Also, I miss a sample for `len(labels) == 20` and `visible_labels` not filled

src/argilla/client/feedback/schemas/questions.py
alvarobartt: So I did it like this in case we allow modifications in the labels themselves, as we do for the moment with the `visible_options` field, so that the warning is directly related to the actual number of labels. Also because we set the minimum to 3 recently, and setting 3 may not be correct if we want to expand labels, at least that was what I thought, but maybe it's not necessary.

src/argilla/client/feedback/schemas/questions.py
alvarobartt: Ideally if `visible_labels` is not provided it means now that it's `None` which implies all the labels will be seen by default, otherwise there's no way for the users to specify all labels unless they explicitly set `visible_labels` to the length of the label list, which as I mentioned before may change in the future right? e.g. I have 10 labels initially, then `visible_options` is set to 10, but then I add a new one so I have 11 labels, but `visible_options` is still 10, which is not correct if the user initially specified `None`, right?

tests/unit/client/feedback/schemas/test_questions.py
alvarobartt: Yes, I was planning to combine both and directly test `_LabelQuestion`, which is the base model for both `LabelQuestion` and `MultiLabelQuestion`, as those just add the `type` value and rely on `_LabelQuestion` for the rest of the stuff

tests/unit/client/feedback/schemas/test_questions.py
alvarobartt: Sure, I'll change that!

src/argilla/client/feedback/schemas/questions.py
frascuchon: The thing is that we shouldn't change the default behavior. Before these changes, if `visible_labels` is `None`, the default value for `visible_options` was fixed to 20. What's means, "Okay, you will see 20 labels in any case".

But this can introduce some inconsistencies when `len(labels) < 20`. For those cases, setting `visible_options=None` or `visible_options=20` will have the same effect (If I have a question with 10 labels, `visible_options=20` let show all the configured labels as `visible_options=None`)

So, the idea should be to compute a default value for `visible_options` when `visible_labels` is not provided to result in the same behavior that we have currently but without confusing values for `visible_options` when `visible_options>len(labels)`:

* If `visible_labels` not provided (`not in values`) and `len(labels) > 20` -> `visible_options = 20`
* If `visible_labels` not provided (`not in values`) and `len(labels) <= 20` -> `visible_options = None`

frascuchon: > Ideally if `visible_labels` is not provided it means now that it's `None` which implies all the labels will be seen by default, otherwise there's no way for the users to specify all labels unless they explicitly set `visible_labels` to the length of the label list, which as I mentioned before may change in the future right? e.g. I have 10 labels initially, then `visible_options` is set to 10, but then I add a new one so I have 11 labels, but `visible_options` is still 10, which is not correct if the user initially specified `None`, right?

if `visible_labels` is not provided, should not be present in `values`, right (I think we should set the `pre=True` for root_validator). We could use this to separate from when `visible_labels` is forced to `None`

tests/unit/client/feedback/schemas/test_questions.py
alvarobartt: I've included those cases in the other unit test, so that this one just checks the `pytest.warns` message

src/argilla/client/feedback/schemas/questions.py
alvarobartt: I've addressed all the comments already, just wanted to mentioned that we need this logic here because the API is validating that `visible_labels>=3` to we cannot set `visible_labels` to a number lower than 3. Also because if I remember correctly we decided that if we have 40 labels but the user says `visible_labels=41`, we were automatically switching that to the maximum number of labels i.e. `visible_labels=40` which can't be done for the case where we have less than 3 labels, so we set `visible_labels=None` in those scenarios, right?

src/argilla/client/feedback/schemas/questions.py
```diff
@@ -153,9 +154,34 @@ def update_settings(cls, values: Dict[str, Any]) -> Dict[str, Any]:
...
+                visible_options = total_options
+            else:
+                warnings.warn(
+                    f"`labels={values.get('labels')}` has less than 3 labels, so `visible_labels`"
+                    " will be set to `None`, which means that all the labels will be visible.",
+                    UserWarning,
+                    stacklevel=1,
+                )
+                visible_options = None
+        if not visible_options and total_options > 20:
```
frascuchon: This will fail in one use case. I want to show all the labels and I have 25 different labels. I need to force `visible_labels=25` instead of forcing it to `None` to show all of them.

I would say to keep the same behavior for all cases. 

* If I want to show all the labels and `total_options  <= 20`, I can force by setting `visible_labels=None`
* If I want to show all the labels and `total_options > 20`, I cannot use `visible_options=None` I need to pass the exact number of labels.

I think we can differentiate if `visible_labels` is not notified from an explicit configuration to `None`. We could check if `visible_labels` is present in `values` or not. If is present and is `None` we should let the user decision to show all the labels.

src/argilla/client/feedback/schemas/questions.py
frascuchon: Yes. This logic is okay.

src/argilla/client/feedback/schemas/questions.py
```diff
@@ -153,9 +154,34 @@ def update_settings(cls, values: Dict[str, Any]) -> Dict[str, Any]:
...
+            warnings.warn(
+                "Since `visible_labels` has not been provided or is `None`, and the total number"
+                " of labels is greater than 20, `visible_labels` will be set to `20`.",
```
frascuchon: As I said before, as a user is weird to me to change the `visible_labels` value when I explicitly set it to `None` which means that I want to show all the labels indiferently of the number of labels I have.

src/argilla/client/feedback/schemas/questions.py
```diff
@@ -130,14 +131,15 @@ class _LabelQuestion(QuestionSchema, LabelMappingMixin):
...
+            which means that it will be automatically set, otherwise it must be either None
+            which means all the labels will be shown, or 3 or greater.
     """
 
     labels: Union[conlist(str, unique_items=True, min_items=2), Dict[str, str]]
-    visible_labels: Optional[conint(ge=3)] = 20
+    visible_labels: Union[Literal["undefined"], conint(ge=3), None] = "undefined"
 
-    @validator("labels", always=True)
+    @validator("labels", pre=True, always=True)
```
frascuchon, line 146: do we finally need this `pre=True`?

src/argilla/client/feedback/schemas/questions.py
```diff
@@ -147,15 +149,47 @@ def labels_dict_must_be_valid(cls, v: Union[List[str], Dict[str, str]]) -> Union
...
                 {"value": key, "text": value} for key, value in values.get("labels").items()
             ]
-        if isinstance(values.get("labels"), list):
+        elif isinstance(values.get("labels"), list):
             values["settings"]["options"] = [{"value": label, "text": label} for label in values.get("labels")]
-        values["settings"]["visible_options"] = values.get(
-            "visible_labels"
-        )  # `None` is a possible value, which means all labels are visible
+
+        if values.get("visible_labels") == "undefined":
```
frascuchon: I think we could create a constant for literal `"undefined"`

src/argilla/client/feedback/schemas/questions.py
alvarobartt: Yes, because we ideally want to validate the labels before actually validating everything via the `root_validator`, so that everything that comes into the `root_validator` is previously validated

src/argilla/client/feedback/schemas/questions.py
```diff
@@ -147,15 +149,47 @@ def labels_dict_must_be_valid(cls, v: Union[List[str], Dict[str, str]]) -> Union
...
+                    " of labels is greater than 20, `visible_labels` will be set to `20`.",
+                    UserWarning,
+                    stacklevel=1,
+                )
+                visible_labels = 20
+            else:
+                visible_labels = None
+        else:
+            visible_labels = values.get("visible_labels")
+            total_options = len(values.get("labels", []))
```
frascuchon: I think we can align the naming to `visible_labels` and call it `total_labels` instead of `total_options`. What do yo think?

frascuchon (approved):
Just some minor comments

PR #3703, opened 2023-09-04
feat: add `datasets delete` command

### LINKED ISSUE #3691: [FEATURE] Add delete dataset command `python -m argilla datasets delete`

gabrielmbmb:
# Description

This PR adds a new command `datasets delete` that allows to delete `FeedbackDataset`s. In addition, a custom callback has been added for the group of commands `datasets` in which a `--name` and a `--workspace` can be provided to load a `FeedbackDataset` that will be used for an operation. Apart from that, the `push-to-hugginface` command has been updated to display an spinner.

Closes #3691

**Type of change**

**How Has This Been Tested**

In a local development environment:

- [x] `python -m argilla datasets --name feedback-dataset --workspace gabriel delete`: deletes the `FeedbackDataset` successfully.

**Checklist**


CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ These are the section headers that we use:
 - Added `users delete` command to delete users ([#3671](https://github.com/argilla-io/argilla/pull/3671)).
 - Added `workspaces create` command to create an Argilla workspace ([#3676](https://github.com/argilla-io/argilla/pull/3676)).
 - Added `datasets push-to-hub` command to push a `FeedbackDataset` from Argilla into the HuggingFace Hub ([#3685](https://github.com/argilla-io/argilla/pull/3685)).
+- Added `datasets delete` command to delete a `FeedbackDataset` ([#3703](https://github.com/argilla-io/argilla/pull/3703)).
```
```suggestion
- Added `datasets delete` command to delete a `FeedbackDataset` from Argilla ([#3703](https://github.com/argilla-io/argilla/pull/3703)).
```

src/argilla/tasks/datasets/__main__.py
```diff
@@ -12,18 +12,59 @@
...
+) -> None:
+    init_callback()
+
+    from argilla.client.feedback.dataset.local import FeedbackDataset
+
+    if ctx.invoked_subcommand not in _COMMANDS_REQUIRING_DATASET:
+        return
+
+    if name is None:
+        raise typer.BadParameter("The command requires a workspace name provided using '--name' option")
```
alvarobartt, line 40: Doesn't the plain `typer.Option` handle this for us if we set the default to `...` instead of `None`? Or is this because we want to show a custom exception to guide the user?

src/argilla/tasks/datasets/delete.py
```diff
@@ -0,0 +1,34 @@
...
+if TYPE_CHECKING:
+    from argilla.client.feedback.dataset.remote.dataset import RemoteFeedbackDataset
+
+
+def delete_dataset(ctx: typer.Context) -> None:
+    dataset: "RemoteFeedbackDataset" = ctx.obj
+
+    try:
+        dataset.delete()
+    except RuntimeError as e:
```
alvarobartt, line 28: Here we can also handle `PermissionError` which is raised by `allow_for_roles` decorator if the user doesn't have `owner` role

src/argilla/tasks/datasets/push.py
```diff
@@ -69,3 +53,7 @@ def push_to_huggingface(
     except Exception as e:
         typer.echo("An unexpected error occurred when trying to push the `FeedbackDataset` to the HuggingFace Hub")
         raise typer.Exit(code=1) from e
+
+    console.print(
+        f":sparkles: `FeedbackDataset` successfully pushed to the :hugs: HuggingFace Hub at https://huggingface.co/{repo_id}"
```
alvarobartt: Should we keep the `:sparkles:` emoji here? Maybe we can just remove it, as we're not using emojis along the CLI so maybe this is a bit off

tests/unit/tasks/datasets/test_push.py
```diff
@@ -43,37 +43,35 @@ def test_push_to_huggingface(
...
+        mocker: "MockerFixture",
+        remote_feedback_dataset: "RemoteFeedbackDataset",
     ) -> None:
-        result = cli_runner.invoke(cli, f"datasets push-to-huggingface {args}")
+        mocker.patch(
+            "argilla.client.feedback.dataset.local.FeedbackDataset.from_argilla",
+            return_value=remote_feedback_dataset,
+        )
+
+        result = cli_runner.invoke(cli, "datasets --name my-dataset push-to-huggingface")
```
alvarobartt, line 65: With the current approach for the CLI as `datasets [NAME] [WORKSPACE] [CMD] [ARGS]` shouldn't we make `CMD` show the `--help` by default? Or do we want it to fail if no args are provided? I guess it would depend on whether we have commands without args or not, right?

alvarobartt (approved):
LGTM, just some minor comments/questions

src/argilla/tasks/datasets/__main__.py
gabrielmbmb: in this case `name` is not a mandatory argument for all the commands like `datasets list`. In the first `if` we check if the executed command needs the `FeedbackDataset` (returns if don't). If this `if` condition executed gets evaluated, then the executed command needs the `FeedbackDataset` and the `name` option has not been provided.

tests/unit/tasks/datasets/test_push.py
alvarobartt: I mean whether we should include `no_args_is_help=True` the the CMD e.g. `push-to-huggingface`

gabrielmbmb: ah, I see! I think it's better to not show by default the `--help`

PR #3731, opened 2023-09-07
:rotating_light: FEAT : highlight field after search

### LINKED ISSUE #3697: Basic highlight results for Feedback Datasets
# Description

The aim of this feature is to enable highlighting for the Feedback dataset annotations view. 

## Solution proposal

Since the search for Feedback datasets allows only searches by terms, and the search results do not include any info on highlighting results (and it's not planned since depends on the backed search engine and the current one could change in the future), the highlighting will be easily resolved on the frontend side (based on the input search text).

So text for record inputs will be processed to highlight terms matching the search.

Additionally, this should work when the record field shows a rendered markdown (`use_markdown=True`) (need to validate this with @damianpumar and @leiyre

keithCuniah:
Highlight record fields after applying a query search.

Note : 
- it works for fields with raw text (not for markdown).

Closes #3697 

**Type of change**

-  New feature

**How Has This Been Tested**

- only feedback task


frontend/components/feedback-task/field/TextField.component.vue
```diff
@@ -38,6 +38,19 @@ export default {
       default: false,
     },
   },
+  computed: {
+    searchValue() {
```
damianpumar: From my point of view, the best practice for that is receiving that value as a prop.

frontend/components/feedback-task/field/TextField.component.vue
```diff
@@ -18,21 +18,26 @@
         </BaseActionTooltip>
       </div>
       <div class="content-area --body1">
-        <div v-if="!useMarkdown" v-text="fieldText" />
-        <RenderMarkdownBaseComponent v-else :markdown="fieldText" />
+        <div v-if="!useMarkdown" v-text="text" />
```
frascuchon: If the highlight won't work for non markdown rendering, why we're passing the `text` instead the origin `fieldText`?

PR #3739, opened 2023-09-08
feat: add `sentence-similarity` support to the `ArgillaTrainer`

### LINKED ISSUE #3316: [FEATURE] add `sentence-similarity` support to the `ArgillaTrainer`
The ArgillaTrainer defined in the [docs](https://docs.argilla.io/en/latest/guides/llms/practical_guides/fine_tune.html), can be aligned with the common tasks described by HF, and some of them are missing.
I would like to see support for the https://huggingface.co/tasks/sentence-similarity task.

We want to have a workflow like shown underneath
```python
from argilla.feedback import ArgillaTrainer, FeedbackDataset, TrainingTask

dataset = FeedbackDataset.from_huggingface(
    repo_id="argilla/emotion"
)
task = TrainingTask.for_sentence_similarity(callable)
trainer = ArgillaTrainer(
    dataset=dataset,
    task=task,
    framework="sentence-transformers"
)
trainer.update_config(num_iterations=1)
trainer.predict("This is awesome!")
```

This means we should define:

- `src/argilla/client/feedback/training/schemas.py` should implement a `TrainingTaskForSentenceSimilarity` that can be initialized by a class-method `TrainingTask.for_sentence_similarity`. This `TrainingTask.for_sentence_similarity`-function should take a callable`formatting_func` which return a `str-str` `sentence-sentence`pair. 
- Passing `TrainingTaskForSentenceSimilarity` to the `ArgillaTrainer` should initiate a workflow for a training process for sentence similarity, which can either rely on a [bi-encoder](https://www.sbert.net/examples/training/sts/README.html) or [cross-encoder](https://www.sbert.net/docs/pretrained_cross-encoders.html).
-`src/argilla/client/feedback/training/frameworks/sentence_transformers.py` needs to be created with a `SentenceTransformersArgillaTrainer` that implements [the logic of fine-tuning](https://huggingface.co/blog/how-to-train-sentence-transformers).
- `tests/integration/client/feedback/training/test_sentence_transformers.py` should hold corresponding functionality tests.
- `docs/_source/_common/tabs/train_update_config.md` update the corresponding config.
- `docs/_source/guides/llms/practical_guides/fine_tune.md`update the docs with a section.

A comparable example can be found here: https://github.com/argilla-io/argilla/pull/3668

N.A.

**Additional context**
Each task should have a defined [task mapping schema](https://github.com/argilla-io/argilla/blob/develop/src/argilla/client/feedback/training/schemas.py) and a defined representation in their associated [framework](https://github.com/argilla-io/argilla/tree/develop/src/argilla/client/feedback/training/frameworks).

plaguss:
# Description

Includes integration with `sentence-transformers` to allow the `ArgillaTrainer` dealing with `sentence-similarity` tasks.

Closes #3316 

**Type of change**

**How Has This Been Tested**

The corresponding tests can be found at:

- [x] tests/integration/client/feedback/training/test_sentence_transformers.py

**Checklist**

### DIFF
--- a/src/argilla/client/feedback/training/frameworks/sentence_transformers.py
+++ b/src/argilla/client/feedback/training/frameworks/sentence_transformers.py
@@ -54,7 +54,7 @@ class ArgillaSentenceTransformersTrainer(ArgillaTrainerSkeleton):
             else:
                 self._model = "cross-encoder/ms-marco-MiniLM-L-6-v2"
 
-            self._logger.warning(f"No model was selected, using pre-defined `{'Cross-Encoder' if self._cross_encoder else 'Bi-Encoder'}` with `{self._model}`.")
+            self._logger.info(f"No model was selected, using pre-defined `{'Cross-Encoder' if self._cross_encoder else 'Bi-Encoder'}` with `{self._model}`.")
 
         else:
             self._model = model
@@ -104,7 +104,7 @@ class ArgillaSentenceTransformersTrainer(ArgillaTrainerSkeleton):
         self._logger.warning(f"`loss_cls` parameter set as default `{self._loss_cls}`.")
 
         if self._eval_dataset:
-            self._set_evaluator(sample.label, len(sample.texts))
+            self._set_evaluator(label=sample.label, number_of_sentences=len(sample.texts))
             self._logger.warning(f"`evaluator` parameter set to `{self.trainer_kwargs['evaluator']}` by default.")
 
     def _set_loss_cls(self, label: Optional[Union[int, float]], number_of_sentences: int) -> None:
@@ -167,17 +167,20 @@ class ArgillaSentenceTransformersTrainer(ArgillaTrainerSkeleton):
                     # a subset of the evaluators defined in sentence-transformers, inform
                     # the user the one selected is not allowed.
                     self.trainer_kwargs["evaluator"] = None
-                    self._logger.warning(f"Currently only developed evaluators that implement `cls.from_input_examples`.")
+                    self._logger.warning(f"Currently only developed evaluators that implement `cls.from_input_examples` can be set.")
         else:
             if label:
                 if number_of_sentences == 2:
                     if isinstance(label, int):
-                        if self._cross_encoder:
-                            from sentence_transformers.cross_encoder.evaluation import CEBinaryClassificationEvaluator
-                            self.trainer_kwargs["evaluator"] = CEBinaryClassificationEvaluator.from_input_examples(self._eval_dataset)
-                        else:
-                            from sentence_transformers.evaluation import BinaryClassificationEvaluator
-                            self.trainer_kwargs["evaluator"] = BinaryClassificationEvaluator.from_input_examples(self._eval_dataset)
+                        # This should work, but for some reason in the tests with default values
+                        # the classes cannot be instantiated, need some time to review.
+                        # if self._cross_encoder:
+                        #     from sentence_transformers.cross_encoder.evaluation import CEBinaryClassificationEvaluator
+                        #     self.trainer_kwargs["evaluator"] = CEBinaryClassificationEvaluator.from_input_examples(self._eval_dataset)
+                        # else:
+                        #     from sentence_transformers.evaluation import BinaryClassificationEvaluator
+                        #     self.trainer_kwargs["evaluator"] = BinaryClassificationEvaluator.from_input_examples(self._eval_dataset)
+                        self.trainer_kwargs["evaluator"] = None
 
                     else:
                         if self._cross_encoder:
--- a/src/argilla/client/feedback/training/frameworks/sentence_transformers.py
+++ b/src/argilla/client/feedback/training/frameworks/sentence_transformers.py
@@ -13,13 +13,12 @@
 #  limitations under the License.
 
 import logging
-from typing import List, Union, Optional, TYPE_CHECKING, Tuple
-
-from argilla.training.utils import get_default_args, filter_allowed_args
-from argilla.utils.dependency import require_version
+from typing import TYPE_CHECKING, List, Optional, Tuple, Union
 
 from argilla.client.feedback.training.base import ArgillaTrainerSkeleton
 from argilla.client.feedback.training.schemas import TrainingTaskForSentenceSimilarity
+from argilla.training.utils import filter_allowed_args, get_default_args
+from argilla.utils.dependency import require_version
 
 if TYPE_CHECKING:
     from argilla.client.feedback.dataset import FeedbackDataset
@@ -33,16 +32,18 @@ class ArgillaSentenceTransformersTrainer(ArgillaTrainerSkeleton):
         self,
         dataset: "FeedbackDataset",
         task: TrainingTaskForSentenceSimilarity,
-        prepared_data = None,
+        prepared_data=None,
         model: str = None,
         seed: int = None,
         train_size: Optional[float] = 1,
-        cross_encoder: bool = False
+        cross_encoder: bool = False,
     ) -> None:
         # TODO: Update with https://github.com/argilla-io/argilla/pull/3555
         require_version("sentence-transformers")
 
-        super().__init__(dataset=dataset, task=task, prepared_data=prepared_data, model=model, train_size=train_size, seed=seed)
+        super().__init__(
+            dataset=dataset, task=task, prepared_data=prepared_data, model=model, train_size=train_size, seed=seed
+        )
 
         # The prepared_data is lost in the TrainerSkeleton, is this intended?
         self._prepared_data = prepared_data
@@ -54,11 +55,13 @@ class ArgillaSentenceTransformersTrainer(ArgillaTrainerSkeleton):
             else:
                 self._model = "cross-encoder/ms-marco-MiniLM-L-6-v2"
 
-            self._logger.info(f"No model was selected, using pre-defined `{'Cross-Encoder' if self._cross_encoder else 'Bi-Encoder'}` with `{self._model}`.")
+            self._logger.info(
+                f"No model was selected, using pre-defined `{'Cross-Encoder' if self._cross_encoder else 'Bi-Encoder'}` with `{self._model}`."
+            )
 
         else:
             self._model = model
- 
+
         if isinstance(self._prepared_data, tuple):
             self._train_dataset = self._prepared_data[0]
             self._eval_dataset = self._prepared_data[1]
@@ -81,18 +84,22 @@ class ArgillaSentenceTransformersTrainer(ArgillaTrainerSkeleton):
         if self._cross_encoder:
             # Cross encoders don't support training with triplets
             if (n := len(sample.texts)) > 2:
-                raise ValueError(f"Cross-encoders don't support training with triplets, the dataset has `{n}` sentences per sample.")
+                raise ValueError(
+                    f"Cross-encoders don't support training with triplets, the dataset has `{n}` sentences per sample."
+                )
 
             from sentence_transformers import CrossEncoder
+
             self._trainer_cls = CrossEncoder
         else:
             from sentence_transformers import SentenceTransformer
+
             self._trainer_cls = SentenceTransformer
 
         self.model_kwargs = get_default_args(self._trainer_cls.__init__)
         self.trainer_kwargs = get_default_args(self._trainer_cls.fit)
         # These arguments don't exactly fit in either of the previous kwargs.
-        # We store here arguments that can be passed to the DataLoader or 
+        # We store here arguments that can be passed to the DataLoader or
         # related to the dataset types defined in sentence-transformers.
         self.data_kwargs = {}
         self.data_kwargs["batch_size"] = 32
@@ -120,30 +127,39 @@ class ArgillaSentenceTransformersTrainer(ArgillaTrainerSkeleton):
             if number_of_sentences == 2:
                 if isinstance(label, int):
                     from sentence_transformers.losses import ContrastiveLoss
+
                     self._loss_cls = ContrastiveLoss
                 else:
                     from sentence_transformers.losses import CosineSimilarityLoss
+
                     self._loss_cls = CosineSimilarityLoss
 
             else:
                 if isinstance(label, int):
                     from sentence_transformers.losses import BatchHardTripletLoss
+
                     self._loss_cls = BatchHardTripletLoss
                 else:
                     from sentence_transformers.losses import BatchAllTripletLoss
+
                     self._loss_cls = BatchAllTripletLoss
                     from sentence_transformers.datasets import SentenceLabelDataset
+
                     self.data_kwargs["dataset_type"] = SentenceLabelDataset
 
         else:
             if number_of_sentences == 3:
                 from sentence_transformers.losses import TripletLoss
+
                 self._loss_cls = TripletLoss
             else:
                 from sentence_transformers.losses import MultipleNegativesRankingLoss
+
                 self._loss_cls = MultipleNegativesRankingLoss
 
-    def _set_evaluator(self, label: Optional[Union[int, float]] = None, number_of_sentences: int  = None, evaluator_cls = None) -> None:
+    def _set_evaluator(
+        self, label: Optional[Union[int, float]] = None, number_of_sentences: int = None, evaluator_cls=None
+    ) -> None:
         """
         Helper method to choose an evaluator for the model based on the type
         of model, whether it has label or not and the number of sentences per example.
@@ -167,7 +183,9 @@ class ArgillaSentenceTransformersTrainer(ArgillaTrainerSkeleton):
                     # a subset of the evaluators defined in sentence-transformers, inform
                     # the user the one selected is not allowed.
                     self.trainer_kwargs["evaluator"] = None
-                    self._logger.warning(f"Currently only developed evaluators that implement `cls.from_input_examples` can be set.")
+                    self._logger.warning(
+                        f"Currently only developed evaluators that implement `cls.from_input_examples` can be set."
+                    )
         else:
             if label:
                 if number_of_sentences == 2:
@@ -185,17 +203,25 @@ class ArgillaSentenceTransformersTrainer(ArgillaTrainerSkeleton):
                     else:
                         if self._cross_encoder:
                             from sentence_transformers.cross_encoder.evaluation import CECorrelationEvaluator
-                            self.trainer_kwargs["evaluator"] = CECorrelationEvaluator.from_input_examples(self._eval_dataset)
+
+                            self.trainer_kwargs["evaluator"] = CECorrelationEvaluator.from_input_examples(
+                                self._eval_dataset
+                            )
                         else:
                             from sentence_transformers.evaluation import EmbeddingSimilarityEvaluator
-                            self.trainer_kwargs["evaluator"] = EmbeddingSimilarityEvaluator.from_input_examples(self._eval_dataset)
+
+                            self.trainer_kwargs["evaluator"] = EmbeddingSimilarityEvaluator.from_input_examples(
+                                self._eval_dataset
+                            )
 
                 else:
                     from sentence_transformers.evaluation import TripletEvaluator
+
                     self.trainer_kwargs["evaluator"] = TripletEvaluator.from_input_examples(self._eval_dataset)
             else:
                 if number_of_sentences == 3:
                     from sentence_transformers.evaluation import TripletEvaluator
+
                     self.trainer_kwargs["evaluator"] = TripletEvaluator.from_input_examples(self._eval_dataset)
                 else:
                     # Hard to choose one here
@@ -216,7 +242,7 @@ class ArgillaSentenceTransformersTrainer(ArgillaTrainerSkeleton):
             self.trainer_kwargs.pop("train_dataloader")
 
         self._trainer = self._trainer_cls(self._model, **self.model_kwargs)
-    
+
     def update_config(self, **kwargs) -> None:
         """
         Updates the configuration of the trainer, but the parameters depend on the trainer.subclass.
@@ -246,10 +272,13 @@ class ArgillaSentenceTransformersTrainer(ArgillaTrainerSkeleton):
         Trains the model.
         """
         from torch.utils.data import DataLoader
+
         self.init_model()
 
         if self.data_kwargs["dataset_type"]:
-            dataloader = DataLoader(dataset=self.data_kwargs["dataset_type"](self._train_dataset), batch_size=self.data_kwargs["batch_size"])
+            dataloader = DataLoader(
+                dataset=self.data_kwargs["dataset_type"](self._train_dataset), batch_size=self.data_kwargs["batch_size"]
+            )
         else:
             dataloader = DataLoader(dataset=self._train_dataset, batch_size=self.data_kwargs["batch_size"])
 
@@ -259,16 +288,15 @@ class ArgillaSentenceTransformersTrainer(ArgillaTrainerSkeleton):
             train_objectives = dataloader
         else:
             train_objectives = [(dataloader, train_loss)]
-        
-        self._trainer.fit(
-            train_objectives,
-            **self.trainer_kwargs
-        )
+
+        self._trainer.fit(train_objectives, **self.trainer_kwargs)
 
         if output_dir is not None:
             self.save(output_dir)
 
-    def predict(self, text: Union[List[List[str]], Tuple[str, List[str]]], as_argilla_records: bool = False, **kwargs) -> List[float]:
+    def predict(
+        self, text: Union[List[List[str]], Tuple[str, List[str]]], as_argilla_records: bool = False, **kwargs
+    ) -> List[float]:
         """
         Predicts the similarity of the sentences.
 
@@ -293,6 +321,7 @@ class ArgillaSentenceTransformersTrainer(ArgillaTrainerSkeleton):
 
         else:
             from sentence_transformers.util import cos_sim
+
             # Under this model we could potentially obtain the similarity of all vs all
             # the input sentences, but maybe is more intuitive to return only the similarities
             # between pairs as we do with the CrossEncoder predictions.
@@ -314,6 +343,6 @@ class ArgillaSentenceTransformersTrainer(ArgillaTrainerSkeleton):
             self._trainer.save(output_dir)
         else:
             # For the moment just save the in the simplest form, creating the model card or the
-            # dataset for example should be done taking the extra information from the argilla 
+            # dataset for example should be done taking the extra information from the argilla
             # dataset instead of the defaults
             self._trainer.save(output_dir, model_name=None, create_model_card=False, train_datasets=None)
--- a/src/argilla/client/feedback/training/frameworks/sentence_transformers.py
+++ b/src/argilla/client/feedback/training/frameworks/sentence_transformers.py
@@ -29,9 +29,6 @@ class ArgillaSentenceTransformersTrainer(ArgillaTrainerSkeleton):
     _logger = logging.getLogger("ArgillaSentenceTransformersTrainer")
     _logger.setLevel(logging.INFO)
 
-    # TODO: Update with https://github.com/argilla-io/argilla/pull/3555
-    require_version("sentence-transformers")
-
     def __init__(
         self,
         dataset: "FeedbackDataset",
@@ -42,6 +39,9 @@ class ArgillaSentenceTransformersTrainer(ArgillaTrainerSkeleton):
         train_size: Optional[float] = 1,
         cross_encoder: bool = False
     ) -> None:
+        # TODO: Update with https://github.com/argilla-io/argilla/pull/3555
+        require_version("sentence-transformers")
+
         super().__init__(dataset=dataset, task=task, prepared_data=prepared_data, model=model, train_size=train_size, seed=seed)
 
         # The prepared_data is lost in the TrainerSkeleton, is this intended?
--- a/src/argilla/client/feedback/training/frameworks/sentence_transformers.py
+++ b/src/argilla/client/feedback/training/frameworks/sentence_transformers.py
@@ -53,6 +53,8 @@ class ArgillaSentenceTransformersTrainer(ArgillaTrainerSkeleton):
                 self._model = "sentence-transformers/all-MiniLM-L6-v2"
             else:
                 self._model = "cross-encoder/ms-marco-MiniLM-L-6-v2"
+        else:
+            self._model = model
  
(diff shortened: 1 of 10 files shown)


src/argilla/client/feedback/training/frameworks/sentence_transformers.py
```diff
@@ -0,0 +1,210 @@
...
+        """
+        self.model_kwargs.update(filter_allowed_args(self._trainer_cls.__init__, **kwargs))
+        self.trainer_kwargs.update(filter_allowed_args(self._trainer_cls.fit, **kwargs))
+
+        if "batch_size" in kwargs:
+            self.dataloader_kwargs["batch_size"] = kwargs["batch_size"]
+
+        if "model" in self.model_kwargs:
+            self._model = self.model_kwargs["model"]
+            # self._model = self.model_kwargs["model_name_or_path"]
```
davidberenstein1957: Why is this commented out? Do we use `model_name_or_path` or `model`?

src/argilla/client/feedback/training/frameworks/sentence_transformers.py
```diff
@@ -0,0 +1,210 @@
...
+if TYPE_CHECKING:
+    from argilla.client.feedback.dataset import FeedbackDataset
+
+
+class ArgillaSentenceTransformersTrainer(ArgillaTrainerSkeleton):
+    _logger = logging.getLogger("ArgillaSentenceTransformersTrainer")
+    _logger.setLevel(logging.INFO)
+
+    # TODO: Update with https://github.com/argilla-io/argilla/pull/3555
+    require_version("sentence-transformers")
```
davidberenstein1957: I believe I moved this into the __init__ to ensure we can properly build the docs without needing to install all integration dependencies.

src/argilla/client/feedback/training/frameworks/sentence_transformers.py
```diff
@@ -0,0 +1,210 @@
...
+        super().__init__(
+            dataset=dataset, task=task, prepared_data=prepared_data, model=model, train_size=train_size, seed=seed
+        )
+
+        # The prepared_data is lost in the TrainerSkeleton, is this intended?
+        self._prepared_data = prepared_data
+        self._cross_encoder = cross_encoder
+
+        if model is None:
+            if not self._cross_encoder:
```
davidberenstein1957, line 53: Perhaps we could add a warning here about the usage of a pre-defined model

src/argilla/client/feedback/training/frameworks/sentence_transformers.py
```diff
@@ -0,0 +1,210 @@
...
+        Initializes the training arguments.
+        """
+        self.model_kwargs = {}
+
+        # Use a sample of the dataset to do some checking
+        sample = self._train_dataset[0]
+
+        if self._cross_encoder:
+            # Cross encoders don't support training with triplets
+            if (n := len(sample.texts)) > 2:
```
davidberenstein1957, line 86: Nice, this is the second `:=` in our codebase, just after @tomaarsen :)

src/argilla/client/feedback/training/frameworks/sentence_transformers.py
plaguss: `model` is the one used, I forgot to remove the `model_name_or_path`. Internally `SentenceTransformer` uses `model_name_or_path` as keyword argument, while `CrossEncoder` uses `model_name`.
I will take a closer look, the classes aren't completely equal and maybe I lost more arguments

src/argilla/client/feedback/training/schemas.py
```diff
@@ -1064,13 +1082,136 @@ def _dict_to_format(ds: datasets.Dataset) -> List[Dict[str, List[Dict[str, str]]
...
+
+class TrainingTaskForSentenceSimilarity(BaseModel, TrainingData):
+    """Training data for sentence similarity.
+
+    Args:
+        formatting_func: A formatting function converting a dictionary of records into
+            a dictionary of a pair of sentences, a pair of sentences and a label,
+            a sentence and a label or a triplet of sentences.
+
+    Examples:
```
davidberenstein1957, line 1443: I think we would require a different example per loss class and also a reference for the cross-encoder.

src/argilla/client/feedback/training/schemas.py
```diff
@@ -1064,13 +1082,136 @@ def _dict_to_format(ds: datasets.Dataset) -> List[Dict[str, List[Dict[str, str]]
...
+    Union[
+        Dict[str, Union[float, int]],  # case 1 with with two string elements and one int/float, case 3 with one or three strings and one int/float.
+        Dict[str, str],                # case 2 with two elements, case 4 with three elements
+    ]
+
+    For a reference of the different cases take a look at:
+    https://huggingface.co/blog/how-to-train-sentence-transformers#how-to-prepare-your-dataset-for-training-a-sentence-transformers-model
+    """
+
+    format: Union[Dict[str, Union[float, int]], Dict[str, str]]
```
davidberenstein1957: We also want to allow for returning a list of samples because each record might contain multiple comparisons.

src/argilla/client/feedback/training/schemas.py
```diff
@@ -1064,13 +1082,136 @@ def _dict_to_format(ds: datasets.Dataset) -> List[Dict[str, List[Dict[str, str]]
...
+        >>> def formatting_func(sample: Dict[str, Any]):
+        ...     return {"sentence": sample["text"], "label": int(sample["label"][0]["value"])}
+        >>> task = TrainingTaskForSentenceSimilarity(formatting_func=formatting_func)
+        >>> dataset.prepare_for_training(framework="...", task=task)
+    """
+
+    _formatting_func_return_types = TrainingTaskForSentenceSimilarityFormat
+    formatting_func: Callable[[Dict[str, Any]], Union[None, Tuple[str, str, str], Iterator[Tuple[str, str, str]]]]
+
+    def _format_data(self, dataset: "FeedbackDataset") -> List[Dict[str, Any]]:
```
davidberenstein1957: This should also be changed a bit to allow for returning a list of samples too from the formatting_func

src/argilla/client/feedback/training/schemas.py
```diff
@@ -1064,13 +1082,136 @@ def _dict_to_format(ds: datasets.Dataset) -> List[Dict[str, List[Dict[str, str]]
...
+        return [Framework(name) for name in names]
+
+    def __repr__(self) -> str:
+        return f"{self.__class__.__name__}\n\t formatting_func={self.formatting_func}"
+
+    @requires_version("scikit-learn")
+    def _train_test_split(self, data: List[dict], train_size: float, seed: int) -> Tuple[List[dict], List[dict]]:
+        from sklearn.model_selection import train_test_split
+
+        return train_test_split(
```
davidberenstein1957: Should we consider stratifying this split too? I can imagine the model might over-under estimate when not doing this?

src/argilla/client/feedback/training/frameworks/sentence_transformers.py
```diff
@@ -0,0 +1,210 @@
...
+                    self._loss_cls = MultipleNegativesRankingLoss
+
+    def init_model(self) -> None:
+        """
+        Initializes a model.
+        """
+        if "model_name_or_path" in self.model_kwargs:
+            self.model_kwargs.pop("model_name_or_path")
+        if "model_name" in self.model_kwargs:
+            self.model_kwargs.pop("model_name")
```
davidberenstein1957, line 237: why are we also using `model_name` and `model_name_or_path` here?

src/argilla/client/feedback/training/frameworks/sentence_transformers.py
```diff
@@ -0,0 +1,210 @@
...
+        """
+        if self._cross_encoder:
+            self._trainer.save(output_dir)
+        else:
+            # For the moment just save the in the simplest form, creating the model card or the
+            # dataset for example should be done taking the extra information from the argilla
+            # dataset instead of the defaults
+            self._trainer.save(output_dir, model_name=None, create_model_card=False, train_datasets=None)
+
+    def __repr__(self) -> str:
```
davidberenstein1957: why did you add this __repr__? One of the subclasses defines this already and outlines the arg and kwargs for the training config.

davidberenstein1957 (commented):
Thanks for the PR @plaguss it looks good already but I left some small initial remarks.

Also, I had some questions: 
Do you see this integration potentially also working with an out-of-box solution using Fields and Questions [as defined for TextClassification](https://docs.argilla.io/en/latest/guides/llms/practical_guides/fine_tune.html#training)? If so, why, else why not? 

What is, according to you, the most interesting Hugging Face [nlp task](https://huggingface.co/tasks) to add to the trainer and why?

plaguss:
> Thanks for the PR @plaguss it looks good already but I left some small initial remarks.
> 
> Also, I had some questions: Do you see this integration potentially also working with an out-of-box solution using Fields and Questions [as defined for TextClassification](https://docs.argilla.io/en/latest/guides/llms/practical_guides/fine_tune.html#training)? If so, why, else why not?

I think it it could work, but I didn't really think about it and just assumed the `formatting_func` would do the job. Sentence transformers allows training with different types of [datasets](https://huggingface.co/blog/how-to-train-sentence-transformers#how-to-prepare-your-dataset-for-training-a-sentence-transformers-model), which allow for [different training strategies](https://github.com/UKPLab/sentence-transformers/tree/master/examples/training) depending on the dataset.
~~I think it just needs some additional checking with respect to the `TextClassification` case, but let me take a look with some to some example datasets and think of an example.~~

It's working with [this commit](https://github.com/argilla-io/argilla/pull/3739/commits/3f61db4100dbda7ff39d31c87cf2a7fe46b3aac3), but I would like to see if it can be simplified a bit, I think I could simplify some loops that may be slow for big datasets.

> What is, according to you, the most interesting Hugging Face [nlp task](https://huggingface.co/tasks) to add to the trainer and why?

Regarding this one, I don't know that much about [table QA](https://huggingface.co/tasks/table-question-answering), (if they are mature or powerful enough for example...) but from were I come from where the world is built around csv files and excel sheets, I can assure you that a lot of people would freak out and be looking forward to using it :smiley:.

src/argilla/client/feedback/training/frameworks/sentence_transformers.py
plaguss: I thought I would need to adjust the subclass default a bit for the case like you did in the `openai.py` or `trl.py` cases,
and I let it there while I took a look, but I really don't now if it needs more info. I will remove it to let the subclass one.

src/argilla/client/feedback/training/frameworks/sentence_transformers.py
plaguss: Noted, its updated

src/argilla/client/feedback/training/frameworks/sentence_transformers.py
plaguss: Does something like this work?:

```python
self._logger.warning(f"No model was selected, using pre-defined `{'Cross-Encoder' if self._cross_encoder else 'Bi-Encoder'}` with `{self._model}`.")
```

src/argilla/client/feedback/training/frameworks/sentence_transformers.py
plaguss: To be fair, I was inspired after seeing him use it :smiley:

src/argilla/client/feedback/training/schemas.py
plaguss: Sure, I will add more examples for the different cases

src/argilla/client/feedback/training/schemas.py
plaguss: Sure, I will add it for the datasets that include integer labels.

src/argilla/client/feedback/training/frameworks/sentence_transformers.py
plaguss: When we grab the default args it is done for `SentenceTransformer` and `CrossEncoder` in a generic way:
```python
self.model_kwargs = get_default_args(self._trainer_cls.__init__)
self.trainer_kwargs = get_default_args(self._trainer_cls.fit)
```
If we don't remove those argument names we have problems during the __init__ call:

```console
>       self._trainer = self._trainer_cls(self._model, **self.model_kwargs)
E       TypeError: CrossEncoder.__init__() got multiple values for argument 'model_name'
...
>       self._trainer = self._trainer_cls(self._model, **self.model_kwargs)
E       TypeError: SentenceTransformer.__init__() got multiple values for argument 'model_name_or_path'
```

In the case of `CrossEncoder` the first argument on __init__ is positional (`model_name`), while `SentenceTransformer`
uses a keyword argument (`model_name_or_path`). As we are passing this argument by position in both cases I think
it's easier to remove from the general arguments.

The same happens with the `fit` method:

```python
if "train_objectives" in self.trainer_kwargs:
    self.trainer_kwargs.pop("train_objectives")
if "train_dataloader" in self.trainer_kwargs:
    self.trainer_kwargs.pop("train_dataloader")
```

```console
E       TypeError: SentenceTransformer.fit() got multiple values for argument 'train_objectives'
...
E       TypeError: CrossEncoder.fit() got multiple values for argument 'train_dataloader'
```

Maybe there is a better way of doing this that I am not seeing?

src/argilla/client/feedback/training/frameworks/sentence_transformers.py
```diff
@@ -175,11 +175,43 @@ def train(self, output_dir: Optional[str] = None) -> None:
...
+            # Under this model we could potentially obtain the similarity of all vs all
+            # the input sentences, but maybe is more intuitive to return only the similarities
+            # between pairs as we do with the CrossEncoder predictions.
+            sources, targets = zip(*sentences)
+            embeds_source = self._trainer.encode(sources, **kwargs)
+            embeds_target = self._trainer.encode(targets, **kwargs)
+            prediction = list(cos_sim(embeds_source, embeds_target).numpy()[0])
+
+        if as_argilla_records:
+            raise NotImplementedError("To be done")
```
plaguss, line 334: Do you think that as_argilla_records makes sense in this type of models?

src/argilla/client/feedback/training/schemas.py
plaguss: [This commit](https://github.com/argilla-io/argilla/pull/3739/commits/be99ae4d967ac7c881e76f3de2ed2229d8f3b92b) should address both this point and the [previous one](https://github.com/argilla-io/argilla/pull/3739#discussion_r1320079351). The functions from the tests have been modified to generate data following  `test_trl.py`.

src/argilla/client/feedback/training/schemas.py
```diff
@@ -91,7 +92,23 @@ def _format_data(self, dataset: "FeedbackDataset") -> List[Dict[str, Any]]:
                 # with default and formatting_func either one can be None
                 if pydantic_field[-1] is not None:
                     pydantic_field_name, pydantic_field_value = pydantic_field
-                    if isinstance(pydantic_field_value, (TextField,)):
+                    if isinstance(pydantic_field_value, list):
```
tomaarsen: This might be a bit restrictive, i.e. doesn't support tuples or sets.

src/argilla/client/feedback/training/schemas.py
```diff
@@ -381,6 +398,55 @@ def for_chat_completion(
...
+            [Dict[str, Any]], Union[None, Tuple[str, str, str], Iterator[Tuple[str, str, str]]]
+        ] = None,
+        label_strategy=None,
+    ) -> "TrainingTaskForSentenceSimilarity":
+        """
+        Return a task that can be used in `FeedbackDataset.prepare_for_training(framework="...", task)`
+        to extract data from the Feedback Dataset in a format suitable for sentence similarity.
+
+        Args:
+            formatting_func: A formatting function converting a dictionary of records into a dict.
```
tomaarsen: A dict of what?

src/argilla/client/feedback/training/schemas.py
```diff
@@ -381,6 +398,55 @@ def for_chat_completion(
...
+
+        Returns:
+            TrainingTaskForSentenceSimilarity: A task mapping instance to be used in `FeedbackDataset.prepare_for_training()`
+        """
+
+        # The label is optional, only check for texts
+        if texts and formatting_func is not None:
+            # if (texts and label) and formatting_func is not None:
```
tomaarsen: Dead code, can be removed presumably

src/argilla/client/feedback/training/schemas.py
```diff
@@ -381,6 +398,55 @@ def for_chat_completion(
...
+            raise ValueError("You must provide either `texts` and `label`, or a `formatting_func`, not both.")
+
+        if formatting_func is not None:
+            if texts or label:
```
tomaarsen: This conditional can only be reached if texts is falsy, so there's no need to include it in the conditional.

src/argilla/client/feedback/training/schemas.py
```diff
@@ -381,6 +398,55 @@ def for_chat_completion(
...
+                raise ValueError("`formatting_func` is already defined, so you cannot define `texts` and `label`.")
```
tomaarsen: Maybe this exception can just be removed and the one above it can be expanded? E.g. `if (texts or label) and formatting_func is not None` with the top exception. Then we only need the "else" branch below here.

src/argilla/client/feedback/training/schemas.py
tomaarsen: Can we get docstrings for texts, labels and label_strategy?

src/argilla/client/feedback/training/schemas.py
```diff
@@ -381,6 +398,55 @@ def for_chat_completion(
     ) -> "TrainingTaskForChatCompletion":
         return TrainingTaskForChatCompletion(formatting_func=formatting_func)
 
+    @classmethod
+    def for_sentence_similarity(
+        cls,
+        texts: Optional[List[TextField]] = None,
+        label: Optional[LabelQuestionUnification] = None,
+        formatting_func: Callable[
+            [Dict[str, Any]], Union[None, Tuple[str, str, str], Iterator[Tuple[str, str, str]]]
```
tomaarsen: Is this doctype correct? It doesn't match the "into a dict" comment below

src/argilla/client/feedback/training/schemas.py
```diff
@@ -381,6 +398,55 @@ def for_chat_completion(
...
+        ] = None,
+        label_strategy=None,
```
tomaarsen: What type is this meant to be?

src/argilla/client/feedback/training/schemas.py
```diff
@@ -381,6 +398,55 @@ def for_chat_completion(
...
+            if isinstance(label, LabelQuestionUnification):
+                if label_strategy is not None:
+                    raise ValueError("label_strategy is already defined via Unification class.")
+            else:
+                unification_kwargs = {"question": label}
+                if label_strategy is not None:
+                    unification_kwargs["strategy"] = label_strategy
+                else:
+                    _LOGGER.info(f"No label strategy defined. Using default strategy for {type(label)}.")
+                if isinstance(label, LabelQuestion):
```
tomaarsen, line 574: Is this possible? `label` is also a `LabelQuestionUnification` instance. I'm not very familiar with these classes so I'm not sure what the inheritance is like.

src/argilla/client/feedback/training/schemas.py
```diff
@@ -1064,13 +1130,208 @@ def _dict_to_format(ds: datasets.Dataset) -> List[Dict[str, List[Dict[str, str]]
...
+                    continue
+
+                self._test_output_formatting_func(output)
+
+                outputs.append(output)
+
+            if "label" in outputs[0]:
+                _all_labels = set()
+                for sample in outputs:
+                    if isinstance(sample, list):
```
tomaarsen: Could fail if labels are tuples or sets I think.

src/argilla/client/feedback/training/schemas.py
```diff
@@ -1064,13 +1130,208 @@ def _dict_to_format(ds: datasets.Dataset) -> List[Dict[str, List[Dict[str, str]]
...
+                    question=LabelQuestion(name="custom_func", labels=list(_all_labels))
+                )
+
+            return outputs
+
+        else:
+            formatted_data = super()._format_data(dataset)
+            # NOTE: Maybe this post processing of the formatted data can be simplified
+            # or directly done in super()._format_data(dataset).
+            new_keys = {field.name: f"sentence-{i+1}" for i, field in enumerate(self.texts)}
```
tomaarsen:
```suggestion
            new_keys = {field.name: f"sentence-{i}" for i, field in enumerate(self.texts, start=1)}
```
Extremely nitpicky, but I usually prefer this solution 😄

src/argilla/client/feedback/training/schemas.py
```diff
@@ -1064,13 +1130,208 @@ def _dict_to_format(ds: datasets.Dataset) -> List[Dict[str, List[Dict[str, str]]
...
+            for example in formatted_data:
+                record = {}
+                for k, v in new_keys.items():
+                    value = example[k]
+                    if v == "label":
+                        # At this point the label must be either an int or a float, determine which one is it.
+                        if value.lstrip("-").isdigit():
+                            value = int(value)
+                        else:
+                            value = float(value)
```
tomaarsen, line 1538: Is it important to determine which it is? Can't we always just cast it to a float? Or does that not work for the "sentence and label" case?

src/argilla/client/feedback/training/schemas.py
```diff
@@ -1064,13 +1130,208 @@ def _dict_to_format(ds: datasets.Dataset) -> List[Dict[str, List[Dict[str, str]]
...
+        return train_test_split(data, train_size=train_size, shuffle=True, random_state=seed, stratify=stratify)
+
+    @requires_version("sentence-transformers")
+    def _prepare_for_training_with_sentence_transformers(
+        self, data: List[dict], train_size: float, seed: int
+    ) -> Union["InputExample", Tuple["InputExample", "InputExample"]]:
+        from sentence_transformers import InputExample
+
+        # Use the first sample to decide what type of dataset to generate:
+        sample_keys = data[0].keys()
```
tomaarsen: Potential edge case: empty list

src/argilla/client/feedback/training/schemas.py
```diff
@@ -1064,13 +1130,208 @@ def _dict_to_format(ds: datasets.Dataset) -> List[Dict[str, List[Dict[str, str]]
...
+        if "label" in sample_keys:
+            if (len(sample_keys) == 3) and all(s in sample_keys for s in ["sentence-1", "sentence-2"]):
+                dataset_fields = lambda sample: {
+                    "texts": [sample["sentence-1"], sample["sentence-2"]],
+                    "label": sample["label"],
+                }
```
tomaarsen: My recommendation is to use set equality here, then we only need one if-elif-else tree. E.g.
```suggestion
        if set(sample_keys) == {"label", "sentence-1", "sentence-2"}:
            dataset_fields = lambda sample: {
                "texts": [sample["sentence-1"], sample["sentence-2"]],
                "label": sample["label"],
            }
```
And then the same for all cases here

src/argilla/client/feedback/training/schemas.py
```diff
@@ -1064,13 +1130,208 @@ def _dict_to_format(ds: datasets.Dataset) -> List[Dict[str, List[Dict[str, str]]
...
+                for record in sample:
+                    train_samples.append(InputExample(**dataset_fields(record)))
+            else:
+                train_samples.append(InputExample(**dataset_fields(sample)))
+
+        if train_size != 1:
+            stratify = None
+            if label := train_samples[0].label:
+                if isinstance(label, int):
+                    stratify = [example.label for example in train_samples]
```
tomaarsen: Nice, I love this operator. I think we can only do these since 2 months ago or so.
That said, I think we can simplify this:
```suggestion
            if (label := train_samples[0].label) and isinstance(label, int):
                stratify = [example.label for example in train_samples]
```
I'm not 100% sure on this though.

tomaarsen: Also, I see now that you need the labels to be an int here, so the "Can you just use float?" comment from above is a bit moot

tests/integration/client/feedback/training/test_sentence_transformers.py
```diff
@@ -0,0 +1,310 @@
...
+
+from typing import TYPE_CHECKING, Callable, List, Union
+
+import pytest
+from argilla.client.feedback.dataset import FeedbackDataset
+from argilla.client.feedback.schemas.records import FeedbackRecord
+from argilla.client.feedback.training.base import ArgillaTrainer
+from argilla.client.feedback.training.schemas import (
+    TrainingTask,
+)
```
tomaarsen, line 23:
```suggestion
from argilla.client.feedback.training.schemas import TrainingTask
```

tomaarsen (commented):
Hey! This is looking good. I haven't ran it myself locally, but it seems complete. Have you been able to test it out with any cool sentence similarity dataset on the Hub?
I found a few small nitpicks, mostly regarding formatting.

src/argilla/client/feedback/training/schemas.py
plaguss: updated

src/argilla/client/feedback/training/schemas.py
plaguss: For sure, I forgot it

src/argilla/client/feedback/training/schemas.py
plaguss: Updating all the docstrings and half-explanations in them

src/argilla/client/feedback/training/schemas.py
plaguss: Done, this and the previous point

src/argilla/client/feedback/training/schemas.py
plaguss: For sure, some things were not properly reviewed :smiley:

src/argilla/client/feedback/training/schemas.py
plaguss: It was copy-pasted, I forgot to update it, its updated now

src/argilla/client/feedback/training/schemas.py
plaguss: My bad, at this point this should be a `LabelQuestion`, the type hints weren't correct

src/argilla/client/feedback/training/schemas.py
plaguss: Updated with those cases

src/argilla/client/feedback/training/schemas.py
plaguss: In addition to the stratification you saw, its useful in case we want to determine internally the type of loss funaction to use during training, as well as an `evaluator` (if train/test sizes are used):
```python
def _set_loss_cls(self, label: Optional[Union[int, float]], number_of_sentences: int) -> None:
...
        if label:
            if number_of_sentences == 2:
                if isinstance(label, int):
                    from sentence_transformers.losses import ContrastiveLoss
                    self._loss_cls = ContrastiveLoss
                else:
                    from sentence_transformers.losses import CosineSimilarityLoss
                    self._loss_cls = CosineSimilarityLoss
...
```

src/argilla/client/feedback/training/schemas.py
plaguss: It wasn't, it is now :+1:

src/argilla/client/feedback/training/schemas.py
plaguss: I'll make a check first. Shouldn't this be checked in `FeedbackDatasetBase.prepare_for_training` so that all the frameworks can be sure they start with at least one example?
```python
if not len(data) > 0:
    raise ValueError("The dataset must contain at least one sample to be able to train.")
```

src/argilla/client/feedback/training/schemas.py
plaguss: Done, far more readable like that :smile:

plaguss:
> Hey! This is looking good. I haven't ran it myself locally, but it seems complete. Have you been able to test it out with any cool sentence similarity dataset on the Hub? I found a few small nitpicks, mostly regarding formatting.

You can see an example in the docs both for `Bi-Encoder` and `Cross-Encoder`, in a subset of the `snli` dataset :smiley: :

```python
from argilla.feedback import ArgillaTrainer, FeedbackDataset, TrainingTask

dataset = FeedbackDataset.from_huggingface("plaguss/snli-small")

task = TrainingTask.for_sentence_similarity(
    texts=[dataset.field_by_name("premise"), dataset.field_by_name("hypothesis")],
    label=dataset.question_by_name("label")
)
trainer = ArgillaTrainer(
    dataset=dataset,
    task=task,
    framework="sentence-transformers",
    framework_kwargs={"cross_encoder": False}
)
trainer.train(output_dir="my_sentence_transformer_model")
trainer.predict(
    [
        "Machine learning is so easy.",
        ["Deep learning is so straightforward.", "This is so difficult, like rocket science.", "I can't believe how much I struggled with this."]
    ]
)
# [0.77857256, 0.4587626, 0.29062212]
```

After training for an epoch and just a small number of samples in the dataset, the model yields quite similar results to those shown in this [hugging face blog post](https://huggingface.co/tasks/sentence-similarity) sentence examples.

src/argilla/client/feedback/training/schemas.py
plaguss: I checked the 'resolve conversation' here but didn't commit the suggestion. It does work so your simplified version wins

plaguss:
Sorry @tomaarsen I thought that I had answered all your points when I was closing them, but I didn't submit the reviews...

davidberenstein1957:
Hi @plaguss,

Thanks a lot for the PR. It is looking good. I transferred ownership to a new branch to work on the fine-tuning of the docs as a final wrapup.

Regards,
David

src/argilla/client/feedback/training/frameworks/sentence_transformers.py
davidberenstein1957: That is the best wai to do it. I believe I misunderstood the difference of these variable per task type.

src/argilla/client/feedback/training/frameworks/sentence_transformers.py
davidberenstein1957: No, this doesn't make sense. Before this was logical but within the FeedbackDataset it is less relevant to have this support. Good catch not to implement it.

PR #3747, opened 2023-09-11
feat: shared datasets base class

davidberenstein1957:
# Description

I added ad shared common `FeedbackDatasetSharedBase` ABC to ensure all methods are represented in all dataset types and no methods can be called outside those.

Closes #NA

**Type of change**

**How Has This Been Tested**
NA
**Checklist**

NA


davidberenstein1957:
@alvarobartt I aligned with Gabri a bit and decided to continue the effort and work towards this. Let me know what you think.

src/argilla/client/feedback/dataset/local/dataset.py
```diff
@@ -15,15 +15,15 @@
...
+from argilla.client.feedback.dataset.local.base import FeedbackDatasetBase
+from argilla.client.feedback.dataset.local.mixins import ArgillaMixin
 from argilla.client.feedback.schemas.types import AllowedFieldTypes, AllowedQuestionTypes
 
 if TYPE_CHECKING:
     from argilla.client.feedback.schemas import FeedbackRecord
 
 
-class FeedbackDataset(FeedbackDatasetBase, ArgillaMixin):
+class FeedbackDataset(ArgillaMixin, FeedbackDatasetBase):
```
alvarobartt: Can we revert the order here? I liked most the class/es first, then the mixins

```suggestion
class FeedbackDataset(FeedbackDatasetBase, ArgillaMixin):
```

src/argilla/client/feedback/dataset/local/dataset.py
```diff
@@ -155,5 +155,10 @@ def add_records(
         else:
             self._records = records
 
+    def delete_records(self, *args, **kwargs):
+        raise NotImplementedError(
+            "Deleting records for local datasets can be done by deleting record from the list FeedbackDataset._records."
```
alvarobartt: If we are planning to point users to `_records` then there's no point on making it protected, I think we can just make it public

src/argilla/client/feedback/dataset/remote/filtered.py
```diff
@@ -92,4 +94,24 @@ def __init__(
...
+            id=self.id,
+            name=self.name,
+            workspace=self.workspace,
+            created_at=self.created_at,
+            updated_at=self.updated_at,
+            fields=self.fields,
+            questions=self.questions,
+            guidelines=self.guidelines,
+            filters=[],
+        )
```
alvarobartt: Hmmm this seems a bit unclear to me, as ideally users should `filtered_dataset = remote_dataset.filter_by(...)` so there should be no need on adding a method to go back to `RemoteFeedbackDataset` IMO

src/argilla/client/feedback/dataset/remote/filtered.py
```diff
@@ -92,4 +94,24 @@ def __init__(
         )
 
     def delete(self) -> None:
-        raise NotImplementedError("`delete` does not work for filtered datasets.")
+        raise NotImplementedError(
+            "`delete` does not work for filtered datasets. First call `filter_reset()` and then `delete()`."
+        )
```
alvarobartt: I think we should be showing `warnings.warn` instead of raising `NotImplementedError` exceptions, as that may break the code and the method will never be implemented, so maybe we're just safer with a `warnings.warn`

alvarobartt: Plus I'm not sure about `filter_reset()` as mentioned below

src/argilla/client/feedback/dataset/remote/base.py
```diff
@@ -323,4 +331,13 @@ def push_to_argilla(
             UserWarning,
             stacklevel=2,
         )
-        pass
+        raise NotImplementedError(
+            "`push_to_argilla` doesn't work for neither `RemoteFeedbackDataset` nor `FilteredRemoteFeedbackDataset`.",
+            "First call `pull_from_argilla` to get a local instance of the dataset, and then call `push_to_argilla(*args, **kwargs)`",
+        )
```
alvarobartt: I'd say we don't need to `raise NotImplementedError`, I think the exception is far more than enough and it doesn't break any existing snippet

alvarobartt:
```suggestion
```

src/argilla/client/feedback/dataset/remote/base.py
```diff
@@ -323,4 +331,13 @@ def push_to_argilla(
...
+
+    def prepare_for_training(self):
+        """Prepares the dataset for training."""
+        raise NotImplementedError(
```
alvarobartt: Same here, I'd just show a `warnings.warn` instead

src/argilla/client/feedback/dataset/remote/base.py
```diff
@@ -312,7 +320,7 @@ def pull(self) -> "FeedbackDataset":
     def push_to_argilla(
         self,
         name: Optional[str] = None,
-        workspace: Optional[Union[str, Workspace]] = None,
+        workspace: Optional[Union[str, "Workspace"]] = None,
         show_progress: Optional[bool] = None,
```
alvarobartt: Indeed we can make this `**kwargs`, as it won't use them, so there's no need to pre-define them

src/argilla/client/feedback/dataset/remote/base.py
```diff
@@ -221,6 +225,10 @@ def __len__(self) -> int:
         """Returns the number of records in the dataset."""
         return self._records.__len__()
 
+    def iter(self) -> Iterator[RemoteFeedbackRecord]:
+        """Returns an iterator over the records in the dataset."""
+        return self.__iter__()
```
alvarobartt: How is this useful? I think we're not using `iter` along the codebase, and if so, we should remove it in favour of the magic method `__iter__`

alvarobartt:
```suggestion
```

src/argilla/client/feedback/dataset/local/base.py
```diff
@@ -0,0 +1,408 @@
...
+            >>> rg.init(...)
+            >>> dataset = rg.FeedbackDataset.from_argilla(name="my-dataset")
+            >>> huggingface_dataset = dataset.format_as("datasets")
+        """
+        if format == "datasets":
+            return self._huggingface_format(self)
+        raise ValueError(f"Unsupported format '{format}'.")
+
+    # TODO(davidberenstein1957): detatch unification into a mixin
+    def unify_responses(
```
alvarobartt: Maybe it's a nice time to detach this also from the current class and move it to a `UnificationMixin` or something similar?

src/argilla/client/feedback/dataset/local/base.py
```diff
@@ -0,0 +1,408 @@
...
+                strategy = RankingQuestionStrategy(strategy)
+            else:
+                raise ValueError(f"Question {question} is not supported yet")
+
+        strategy.unify_responses(self.records, question)
+
+    # TODO(alvarobartt,davidberenstein1957): we should consider having something like
+    # `export(..., training=True)` to export the dataset records in any format, replacing
+    # both `format_as` and `prepare_for_training`
+    def prepare_for_training(
```
alvarobartt: Same here, we can take the refactoring as a chance to move this to a separate mixin instead e.g. `ArgillaTrainDatasetMixin`

alvarobartt (commented):
Thanks! 👍🏻 While I agree with most of the core refactoring, I think that I would do some things differently, left my thoughts on this PR, so LMWYT

src/argilla/client/feedback/dataset/remote/filtered.py
davidberenstein1957: Yes, I also wasn't sure about this but normally when data wrangling, these kinds of operators are nice, however the duplicated `filter_by` I added, should do the job nicely, I agree with removing this too. Was just playing around.

src/argilla/client/feedback/dataset/remote/base.py
davidberenstein1957: That is fine. I noticed an `iter` somewhere else so I copied the behavior. But I agree removing it is better.

src/argilla/client/feedback/dataset/local/base.py
davidberenstein1957: I agree with this.

src/argilla/client/feedback/dataset/local/base.py
davidberenstein1957: I agree.

davidberenstein1957:
@alvarobartt I also had not gotten around to the test, but will push the feedback later today to ensure we can add it to this release.

src/argilla/client/feedback/dataset/remote/base.py
alvarobartt: Yes, I think it was originally for local datasets, but I thought it was already removed, if that was not the case, then feel free to remove it yourself 👍🏻

src/argilla/client/feedback/dataset/remote/filtered.py
davidberenstein1957: I removed `filter_reset`

davidberenstein1957:
@alvarobartt this should be ready for a final review

alvarobartt (commented):
LGTM, we can safely merge once the unit tests are refactored too to follow the same structure `client/feedback/dataset/local/...` which implies renaming directories plus checking that everything's covered (no need to cover the abc classes)

Issue #3028, opened 2023-05-30
👬 Similarity search in Feedback datasets


nataliaElv:
```[tasklist]
### Tasks
- [ ] https://github.com/argilla-io/argilla/pull/3750
- [ ] https://github.com/argilla-io/argilla/pull/3744
- [ ] https://github.com/argilla-io/argilla/pull/3799
- [x] [TEST] Make the `elasticsearch` engine tests work in the CI pipeline
- [ ] https://github.com/argilla-io/argilla/issues/3761
- [x] [FEATURE] Add `GET /api/v1/datasets/{dataset_id}/vectors-settings` endpoint
- [x] [FEATURE] Add `DELETE /api/v1/vectors-settings/{vector_settings_id}` endpoint
- [x] [FEATURE] Update `POST /api/v1/datasets/{dataset_id}/records` endpoint to include vectors in body (https://github.com/argilla-io/argilla/pull/4022)
- [ ] https://github.com/argilla-io/argilla/pull/3765
- [x] New endpoint `GET /api/v1/records/:record_id`  to fetch the reference record
- [ ] https://github.com/argilla-io/argilla/issues/4071
- [ ] https://github.com/argilla-io/argilla/issues/4066
- [x] [HIGH] Define `dataset.add_vector_settings` method in the Python SDK https://github.com/argilla-io/argilla/pull/4055
- [x] [HIGH] Allow to create records with vectors in Python SDK https://github.com/argilla-io/argilla/pull/4043
- [x] [HIGH][FEATURE] Update `POST /api/v1/me/datasets/{dataset_id}/records/search` endpoint to allow vector search (https://github.com/argilla-io/argilla/pull/4019)
- [x] Unify `index_record` and `set_record_vectors` for the search engine
- [x] [HIGH] Allow partial update for record vectors (not needed to pass all the vectors structure for a record) https://github.com/argilla-io/argilla/pull/4104
- [ ] https://github.com/argilla-io/argilla/issues/4085
- [ ] https://github.com/argilla-io/argilla/issues/4102
- [ ] https://github.com/argilla-io/argilla/issues/4020
- [ ] https://github.com/argilla-io/argilla/issues/4114
- [ ] https://github.com/argilla-io/argilla/issues/4051
- [ ] https://github.com/argilla-io/argilla/issues/4069
- [ ] https://github.com/argilla-io/argilla/issues/4105
- [ ] https://github.com/argilla-io/argilla/issues/4106
[...]

PR #3765, opened 2023-09-14
feat(search): Similarity search for Opensearch and Elasticsearch

frascuchon:
# Description

This PR includes a bunch of changes regarding the `search_engine` module: 

- The base class includes a new method to tackle vector generation and defines the `similarity_search` method.
- Two similar implementations of the `SearchEngine` class have been included `OpenSearchEngine` and `ElasticSearchEngine`. Since they share a lot of functionality, a common abstract class has been implemented for common code blocks, letting each implementation handle its own details.
- A new way for engine registration and generation has been developed. This new way is based on a class decorator. The aim of this approach is to simplify future engine implementations, keeping the code isolated for each new implementation.
The following example shows how a new engine could be implemented and registered as an available engine for the Argilla server:
```python
from argilla.server.search_engine import SearchEngine

@SearchEngine.register(engine_name="my-new-engine")
class NewSearchEngineImplementation(SearchEngine):
   # implemente all required method
   ...
```
Then, you can use your search engine by environment variable:
```bash
ARGILLA_SEARCH_ENGINE=my-new-engine python -m argilla server
```
The `ARGILLA_SEARCH_ENGINE` environment variable has been added as a common server setting. This is a first approach and may change in the future. This is the reason why I did not include it in the docs section.

Refs #3028

**Type of change**

**How Has This Been Tested**

- [ ] Test A
- [ ] Test B

**Checklist**

### DIFF
--- a/.github/workflows/run-python-tests.yml
+++ b/.github/workflows/run-python-tests.yml
@@ -74,19 +74,19 @@ jobs:
           key: ${{ runner.os }}-pip-${{ env.CACHE_NUMBER }}-${{ hashFiles('pyproject.toml') }}
       - name: Set huggingface hub credentials
         if: github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/heads/releases')
-        env:
-          HF_HUB_ACCESS_TOKEN: ${{ secrets.HF_HUB_ACCESS_TOKEN }}
-        run: echo "Enable HF access token"
+        run: |
+          echo "HF_HUB_ACCESS_TOKEN=${{ secrets.HF_HUB_ACCESS_TOKEN }}" >> "$GITHUB_ENV"
+          echo "Enable HF access token"
       - name: Set Argilla search engine env variable
         if: startsWith(inputs.searchEngineDockerImage, 'docker.elastic.co')
-        env:
-          ARGILLA_SEARCH_ENGINE: elasticsearch
-        run: echo "Configure elasticsearch engine"
+        run: |
+          echo "ARGILLA_SEARCH_ENGINE=elasticsearch" >> "$GITHUB_ENV"
+          echo "Configure elasticsearch engine"
       - name: Set Argilla search engine env variable
         if: startsWith(inputs.searchEngineDockerImage, 'opensearchproject')
-        env:
-          ARGILLA_SEARCH_ENGINE: opensearch
-        run: echo "Configure opensearch engine"
+        run: |
+          echo "ARGILLA_SEARCH_ENGINE=opensearch" >> "$GITHUB_ENV"
+          echo "Configure elasticsearch engine"
       - name: Run tests 📈
         run: |
           pip install -e ".[server,listeners]"
--- a/.github/workflows/run-python-tests.yml
+++ b/.github/workflows/run-python-tests.yml
@@ -77,6 +77,16 @@ jobs:
         env:
           HF_HUB_ACCESS_TOKEN: ${{ secrets.HF_HUB_ACCESS_TOKEN }}
         run: echo "Enable HF access token"
+      - name: Set huggingface hub credentials
+        if: startsWith(searchEngineDockerImage, 'docker.elastic.co')
+        env:
+          ARGILLA_SEARCH_ENGINE: elasticsearch
+        run: echo "Configure elasticsearch engine"
+      - name: Set huggingface hub credentials
+        if: startsWith(searchEngineDockerImage, 'opensearchproject')
+        env:
+          ARGILLA_SEARCH_ENGINE: opensearch
+        run: echo "Configure opensearch engine"
       - name: Run tests 📈
         run: |
           pip install -e ".[server,listeners]"
--- a/.github/workflows/run-python-tests.yml
+++ b/.github/workflows/run-python-tests.yml
@@ -78,12 +78,12 @@ jobs:
           HF_HUB_ACCESS_TOKEN: ${{ secrets.HF_HUB_ACCESS_TOKEN }}
         run: echo "Enable HF access token"
       - name: Set huggingface hub credentials
-        if: startsWith(searchEngineDockerImage, 'docker.elastic.co')
+        if: startsWith(inputs.searchEngineDockerImage, 'docker.elastic.co')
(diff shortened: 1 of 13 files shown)


src/argilla/server/search_engine/elasticsearch.py
```diff
@@ -0,0 +1,135 @@
...
+from argilla.server.search_engine import (
+    SearchEngine,
+    SearchResponses,
+    UserResponseStatusFilter,
+)
+from argilla.server.search_engine.commons import BaseElasticAndOpenSearchEngine, field_name_for_vector_settings
+from argilla.server.settings import settings
+
+
+def _compute_num_candidates_from_k(k: int):
```
gabrielmbmb:
```suggestion
def _compute_num_candidates_from_k(k: int) -> int:
```

src/argilla/server/search_engine/elasticsearch.py
```diff
@@ -0,0 +1,135 @@
...
+class ElasticSearchEngine(BaseElasticAndOpenSearchEngine):
+    config: Dict[str, Any]
+
+    es_number_of_shards: int
+    es_number_of_replicas: int
+
+    def __post_init__(self):
+        self.client = AsyncElasticsearch(**self.config)
+
+    def _configure_index_settings(self):
```
gabrielmbmb:
```suggestion
    def _configure_index_settings(self) -> Dict[str, Any]:
```

src/argilla/server/search_engine/elasticsearch.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def _mapping_for_vector_settings(self, vector_settings: VectorSettings) -> dict:
+        return {
+            f"vectors.{vector_settings.id}": {
+                "type": "dense_vector",
+                "dims": vector_settings.dimensions,
+                "index": True,
+                # can similarity property also be part of config @frascuchon ?
+                # relates vector search similarity metric
+                "similarity": "l2_norm",  ## default value regarding the knn best practices es documentation
```
gabrielmbmb, line 80: I think that we should let the user configure this when creating settings for a collection of vectors. In some cases, depending on the vectors used, one metric could be better than other.

gabrielmbmb (commented):
I think that we should allow the user to configure the similarity metric. Also, I think we could assume that no matter the vector database that Argilla is using, HNSW algorithm will be used.

HNSW is the default ANN algorithm in all the vector databases, and I think it's important also to let the user configure the parameters of the algorithm `ef`, `m` and `ef_construction`, as these would affect the results of the search.

src/argilla/server/search_engine/elasticsearch.py
frascuchon: Yes. We can iterate over it in future PRs. The `similarity` param could affect other default ones.

src/argilla/server/search_engine/base.py
```diff
@@ -96,3 +96,24 @@ async def search(
...
+
+    @abstractmethod
+    async def similarity_search(
+        self,
+        dataset: Dataset,
+        vector_settings: VectorSettings,
+        value: Optional[List[float]] = None,
+        record: Optional[Record] = None,
+        user_response_status_filter: Optional[UserResponseStatusFilter] = None,
+        max_results: conint(ge=2, le=500) = 100,
```
gabrielmbmb: No validation will be applied here by `pydantic`, so let's just use `int`

```suggestion
        max_results: int = 100,
```

.github/workflows/run-python-tests.yml
```diff
@@ -77,6 +77,16 @@ jobs:
         env:
           HF_HUB_ACCESS_TOKEN: ${{ secrets.HF_HUB_ACCESS_TOKEN }}
         run: echo "Enable HF access token"
+      - name: Set huggingface hub credentials
```
gabrielmbmb:
```suggestion
      - name: Set Argilla search engine env variable
```

.github/workflows/run-python-tests.yml
```diff
@@ -77,6 +77,16 @@ jobs:
...
+        if: startsWith(searchEngineDockerImage, 'docker.elastic.co')
+        env:
+          ARGILLA_SEARCH_ENGINE: elasticsearch
+        run: echo "Configure elasticsearch engine"
+      - name: Set huggingface hub credentials
```
gabrielmbmb:
```suggestion
      - name: Set Argilla search engine env variable
```

src/argilla/server/search_engine/elasticsearch.py
```diff
@@ -0,0 +1,132 @@
...
+            verify_certs=settings.elasticsearch_ssl_verify,
+            ca_certs=settings.elasticsearch_ca_path,
+            retry_on_timeout=True,
+            max_retries=5,
+        )
+        return ElasticSearchEngine(
+            config,
+            es_number_of_shards=settings.es_records_index_shards,
+            es_number_of_replicas=settings.es_records_index_replicas,
+        )
```
gabrielmbmb:
```suggestion
    @classmethod
    async def new_instance(cls) -> "ElasticSearchEngine":
        config = dict(
            hosts=settings.elasticsearch,
            verify_certs=settings.elasticsearch_ssl_verify,
            ca_certs=settings.elasticsearch_ca_path,
            retry_on_timeout=True,
            max_retries=5,
        )
        return cls(
            config,
            es_number_of_shards=settings.es_records_index_shards,
            es_number_of_replicas=settings.es_records_index_replicas,
        )
```

src/argilla/server/search_engine/elasticsearch.py
```diff
@@ -0,0 +1,132 @@
...
+
+def _compute_num_candidates_from_k(k: int) -> int:
+    if k < 50:
+        return 500
+    elif 50 <= k < 200:
+        return 100
+    return 2000
+
+
+@SearchEngine.register(engine_name="opensearch")
```
gabrielmbmb:
```suggestion
@SearchEngine.register(engine_name="elasticsearch")
```

gabrielmbmb (approved):

PR #3795, opened 2023-09-20
FIX: Correctly save a ArgillaTrainer + TRL model with peft_config 

tomaarsen:
# Description
If you pass `peft_config` to `trainer.update_config`, then it will currently still save the full (but broken) model with some LoRA (or other PEFT) parts interspersed that can't be loaded again. This is because the `self._transformers_model` was saved, but that model was the original model (e.g. Llama 2 7b) and not the PEFT model, which is created internally in e.g. the SFTTrainer.

In short, this now saves the actual model and tokenizer used by the underlying trainer, so that if that trainer make any other changes to the model, those changes will be saved too.

**Type of change**

**How Has This Been Tested**

Created a new test and verified that there's now `adapter_model.bin` and `adapter_config.json` files after training.

**Checklist**

---

- Tom Aarsen


tomaarsen:
This also works in my RLHF tutorial.

davidberenstein1957 (approved):

Issue #3748, opened 2023-09-12
⏳ Filtering and sorting using custom metadata info


nataliaElv:
```[tasklist]
### Tasks
- [ ] https://github.com/argilla-io/argilla/issues/3798
- [x] Get metadata fields from backend
- [ ] https://github.com/argilla-io/argilla/issues/3796
- [ ] https://github.com/argilla-io/argilla/pull/3841
- [ ] https://github.com/argilla-io/argilla/issues/3844
- [ ] https://github.com/argilla-io/argilla/pull/3949
- [x] Use metadata filter title instead of name [FRONTEND]
- [ ] https://github.com/argilla-io/argilla/pull/3811
- [ ] https://github.com/argilla-io/argilla/pull/3813
- [ ] https://github.com/argilla-io/argilla/pull/3816
- [x] Index metadata values on record creation (#3816)
- [x] Enable metadata filtering from `SearchEngine` (#3816)
- [x] Adapt search records endpoint including metadata query params (#3816)
- [ ] https://github.com/argilla-io/argilla/pull/3819
- [x] Enable metadata filtering from get records endpoints `GET /api/v1/datasets/:dataset_id/records` and `GET /api/v1/me/datasets/:dataset_id/records`  (#3819)
- [ ] https://github.com/argilla-io/argilla/pull/3825
- [x] Sort by `last_updated` and `inserted_at` values (https://github.com/argilla-io/argilla/pull/3825)
- [x] Add basic `settings` support for metadata properties, including `max` and `min` for integer and float types and a value list for terms
- [x] Dataset with metadata in quickstart script (https://github.com/argilla-io/argilla/pull/3830/commits/e421d77dd1c9258c3e5ee3112d903c725bbcb99f)
- [ ] https://github.com/argilla-io/argilla/issues/3835
- [x] Sort by metadata property values
- [ ] https://github.com/argilla-io/argilla/issues/3898
- [ ] https://github.com/argilla-io/argilla/issues/3859
- [x] Add `allow_extra_metadata` support (API + SDK)
- [x] Add new endpoint to delete metadata properties (https://github.com/argilla-io/argilla/pull/3911)
- [x] Add `visible_for_annotator` support (https://github.com/argilla-io/argilla/pull/3924)
- [x] Update policies to support metadata property visibility for annotators
- [x] Add client validations for SDK
- [x] Allow to update metadata for records (https://github.com/argilla-io/argilla/pull/3946)
[...]

PR #3811, opened 2023-09-22
db: create `MetadataProperty` ORM class  and `metadata_properties` table 

frascuchon:
# Description

This PR includes the new `MetadataProperty` class, the base class for the metadata filtering feature.

Refs [#3748](https://github.com/argilla-io/argilla/issues/3748)


src/argilla/server/models/database.py
```diff
@@ -167,6 +181,27 @@ def __repr__(self):
         )
```
gabrielmbmb, line 182:
```suggestion


MetadataPropertyTypeEnum = SAEnum(MetadataPropertyType, name="metadata_property_type")


```

src/argilla/server/models/database.py
```diff
@@ -167,6 +181,27 @@ def __repr__(self):
...
 
 
+class MetadataProperty(DatabaseModel):
+    __tablename__ = "metadata_properties"
+
+    name: Mapped[str] = mapped_column(Text, index=True)
+    type: Mapped[MetadataPropertyType] = mapped_column(Text)
```
gabrielmbmb, line 188:
```suggestion
    type: Mapped[MetadataPropertyType] = mapped_column(MetadataPropertyTypeEnum)
```

gabrielmbmb (commented):
It looks good to me! We just need to create the SQLAlchemy enum so a data type is created in PostgreSQL

src/argilla/server/models/database.py
frascuchon: I prefer to keep it as Text and check the enum at the application level. We had some problems updating enums in the past and this could be an entry point for future problems (with other db than SQLite)

PR #3851, opened 2023-09-28
fix: Allow pull datasets without records

frascuchon:
# Description

Fixed for `ds.pull()` with datasets without records.

Also, the `__len__` for the filtered dataset returns a 0 instead of `None`. Otherwise, the `len(ds)` fails with a confusing error to the user.

### DIFF
--- a/src/argilla/client/feedback/dataset/remote/base.py
+++ b/src/argilla/client/feedback/dataset/remote/base.py
@@ -277,7 +277,9 @@ class RemoteFeedbackDatasetBase(Generic[T], FeedbackDatasetBase):
             guidelines=self.guidelines,
             metadata_properties=self.metadata_properties,
         )
-        instance.add_records(
-            records=[record.to_local() for record in self._records],
-        )
+        records = [record.to_local() for record in self._records]
+
+        if records:
+            instance.add_records(records=records)
+
         return instance
--- a/src/argilla/client/feedback/dataset/remote/base.py
+++ b/src/argilla/client/feedback/dataset/remote/base.py
@@ -279,7 +279,7 @@ class RemoteFeedbackDatasetBase(Generic[T], FeedbackDatasetBase):
         )
         records = [record.to_local() for record in self._records]
 
-        if records:
+        if len(records) > 0:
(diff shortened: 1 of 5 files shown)


src/argilla/client/feedback/dataset/remote/filtered.py
```diff
@@ -52,14 +52,15 @@ def __init__(
...
-    def __len__(self) -> None:
+    def __len__(self) -> int:
         warnings.warn(
             "The `records` of a filtered dataset in Argilla are being lazily loaded"
             " and len computation may add undesirable extra computation. You can fetch"
             "records using\n`ds.pull()`\nor iterate over results to know the length of the result:\n"
             "`records = [r for r in ds.records]\n",
             stacklevel=1,
         )
+        return 0
```
alvarobartt, line 63: I think this can be misleading, I'm more comfortable with not retuning anything i.e. `None`

alvarobartt, line 63: I think this can be misleading, I'm more comfortable with not retuning anything i.e. `None`

src/argilla/client/feedback/dataset/remote/base.py
```diff
@@ -277,7 +277,9 @@ def pull(self) -> "FeedbackDataset":
             guidelines=self.guidelines,
             metadata_properties=self.metadata_properties,
         )
-        instance.add_records(
-            records=[record.to_local() for record in self._records],
-        )
+        records = [record.to_local() for record in self._records]
+
+        if records:
```
alvarobartt: Here we can just be more explicit:

```suggestion
        if len(records) > 0:
```

src/argilla/client/feedback/dataset/remote/filtered.py
frascuchon: Then, an error will be raised since `__len__` implementation does not support `None` values, which will be confusing to users.

alvarobartt: Then maybe we can just roll back to keep the `NotImplementedError` but with the updated message?

PR #3852, opened 2023-09-28
feat: update `load_data.py` with datasets for comparison

alvarobartt:
# Description

This PR adds a new dataset into the `load_data.py` script to be created by default on both the current alpha version for the `FeedbackDataset` updates, and also in the legacy/former format i.e. `DatasetForTextClassification`.

This is done for being able to compare the same dataset and the different features and visuals for both of them.

**Type of change**

### DIFF
--- a/docker/scripts/load_data.py
+++ b/docker/scripts/load_data.py
@@ -163,38 +163,6 @@ class LoadDatasets:
         dataset.push_to_argilla(name=repo_id.split("/")[-1])
 
     @staticmethod
-    def load_mock_dataset_for_filter_metadata():
-        print("Loading Mock dataset for metadata properties test")
-        import random
-
-        import argilla as rg
-
-        dataset = rg.FeedbackDataset.from_huggingface("argilla/oasst_response_quality", split="train")
-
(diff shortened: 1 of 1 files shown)


docker/scripts/load_data.py
```diff
@@ -274,8 +289,9 @@ def build_record(row):
                     ld = LoadDatasets(API_KEY)
                     # TODO(@frascuchon): Remove these datasets creation
                     ld.load_mock_dataset_for_filter_metadata()
```
frascuchon: I think we can remove also this one since we have a real one with metadata properties
```suggestion
```

docker/scripts/load_data.py
```diff
@@ -273,9 +288,9 @@ def build_record(row):
                 if response.status_code == 200:
                     ld = LoadDatasets(API_KEY)
                     # TODO(@frascuchon): Remove these datasets creation
```
frascuchon: Also this one, sorry
```suggestion
```

frascuchon (approved):

PR #3860, opened 2023-09-29
feat: add validation layer for `metadata_properties` and `metadata_filters`

### LINKED ISSUE #3859: [FEATURE] Add metadata validation in `add_records` and `filter_by`
## Description

A validation layer should be included for both the `metadata_properties` and the `metadata_filters` (linked with the pre-defined `metadata_properties`) on both `add_records` and `filter_by` methods, respectively.

This validation will help on detecting the problem on the user-side to avoid calling the API with wrong payloads, and also to help the user debug it better.

alvarobartt:
# Description

This PR adds a validation layer for both the `metadata_properties` when `metadata` is provided via `FeedbackRecord` when adding that record to either a local or existing `FeedbackDataset`, as well as validating the `metadata_filters` provided to the `filter_by` method for `RemoteFeedbackDataset`.

These validation layers would help reduce the API workload as the validation will be done client-side, which implies that the client will handle the validation first.

In this case, as we have the following properties: `TermsMetadataProperty`, `IntegerMetadataProperty` and `FloatMetadataProperty`; we'll need to ensure that the values provided for those are correct, both on `add_records` and also on `filter_by`.

The approach is similar to the one already implemented for the `fields` of a `FeedbackRecord`, but extending it to the `metadata_properties` to also include a validation layer that ensures not just the type of the value provided, but whether it's compliant with the pre-defined values for `TermMetadataProperty` or the pre-defined boundaries for both `IntegerMetadataProperty` and `FloatMetadataProperty`. The core difference is that besides the `pydantic` validation, there's also a method in each property that validates the values provided in the filters related to those metadata properties.

Closes #3859

**Type of change**

**How Has This Been Tested**

- [x] Add unit tests for the `generate_pydantic_schema_for_metadata` method
- [x] Add unit tests for both `_validate_filter` and `_pydantic_field_with_validator` in each property i.e. `TermsMetadataProperty`, `IntegerMetadataProperty`, and `FloatMetadataProperty`
- [x] Add unit tests for `add_records` with both correct and wrong `metadata`
- [x] Add integration tests for `add_records` and `filter_by` to ensure that the validation is correct both before and after applying those ops

**Checklist**

### DIFF
--- a/src/argilla/client/feedback/dataset/base.py
+++ b/src/argilla/client/feedback/dataset/base.py
@@ -284,7 +284,7 @@ class FeedbackDatasetBase(ABC, HuggingFaceDatasetMixin):
         if not hasattr(self, "_metadata_schema"):
             self._metadata_schema = None
 
-        if self._metadata_schema is None:
+        if self._metadata_schema is None and self.metadata_properties is not None:
             self._metadata_schema = generate_pydantic_schema_for_metadata(self.metadata_properties)
 
         for record in records:
--- a/src/argilla/client/feedback/dataset/base.py
+++ b/src/argilla/client/feedback/dataset/base.py
@@ -213,7 +213,27 @@ class FeedbackDatasetBase(ABC, HuggingFaceDatasetMixin):
     def metadata_properties(
         self,
     ) -> Union[List["AllowedMetadataPropertyTypes"], List["AllowedRemoteMetadataPropertyTypes"]]:
+        """Returns the metadata properties that will be indexed and could be used to filter the dataset."""
         return self._metadata_properties
+
+    def metadata_property_by_name(
+        self, name: str
+    ) -> Union["AllowedMetadataPropertyTypes", "AllowedRemoteMetadataPropertyTypes"]:
+        """Returns the metadata property by name if it exists. Othewise a `ValueError` is raised.
+
+        Args:
+            name: the name of the metadata property to return.
+
+        Raises:
+            ValueError: if the metadata property with the given name does not exist.
+        """
+        for metadata_property in self._metadata_properties:
+            if metadata_property.name == name:
+                return metadata_property
+        raise ValueError(
+            f"Metadata property with name='{name}' not found, available metadata property names are:"
+            f" {', '.join(m.name for m in self._metadata_properties)}"
+        )
 
     def _parse_records(
         self, records: Union[FeedbackRecord, Dict[str, Any], List[Union[FeedbackRecord, Dict[str, Any]]]]
--- a/src/argilla/client/feedback/dataset/base.py
+++ b/src/argilla/client/feedback/dataset/base.py
@@ -35,7 +35,7 @@ from argilla.client.feedback.training.schemas import (
     TrainingTaskForTextClassification,
     TrainingTaskTypes,
 )
-from argilla.client.feedback.utils import generate_pydantic_schema_for_fields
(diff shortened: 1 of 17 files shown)


src/argilla/client/feedback/schemas/metadata.py
```diff
@@ -98,6 +103,19 @@ def check_values(cls, terms_values: List[str], values: Dict[str, Any]) -> List[s
     def server_settings(self) -> Dict[str, Any]:
         return {"type": self.type, "values": self.values}
 
+    def _all_values_exist(self, introduced_value: str) -> None:
+        if introduced_value not in self.values:
+            raise ValueError(
+                f"Provided '{self.name}={introduced_value}' is not valid, only values in {self.values} are allowed."
+            )
```
frascuchon, line 126: Is this a pydantic validator?

alvarobartt: Well it's just a class-method in the `pydantic.BaseModel` which is later used as a validator in the property `_pydantic_field_with_validator`

alvarobartt:
FYI @frascuchon I had to revert the `return 0` in `FilteredRemoteRecords.__len__` magic method as returning 0 is misleading IMO, and we cannot return e.g. -1, anyway I assume this is temporarily as before the release we can just add an API call to get the total number of records for filtered datasets.

src/argilla/client/feedback/dataset/remote/dataset.py
```diff
@@ -177,12 +178,36 @@ def filter_by(
...
-        if response_status and not isinstance(response_status, list):
-            response_status = [response_status]
-        if metadata_filters and not isinstance(metadata_filters, list):
-            metadata_filters = [metadata_filters]
 
-        #  accessing records later
+        if response_status:
+            if not isinstance(response_status, list):
+                response_status = [response_status]
+            if not all(status in [arg.value for arg in FeedbackResponseStatusFilter] for status in response_status):
```
gabrielmbmb, line 185: I would create a constant called `_RESPONSE_STATUS_FILTER_VALUES` or something similar with value `[arg.value for arg in FeedbackResponseStatusFilter]`

src/argilla/client/feedback/dataset/remote/dataset.py
```diff
@@ -177,12 +178,36 @@ def filter_by(
...
+                raise ValueError(
+                    f"Invalid `response_status={response_status}` provided, must be one"
+                    f" of: {[arg.value for arg in FeedbackResponseStatusFilter]}"
+                )
+
+        if metadata_filters:
+            if not isinstance(metadata_filters, list):
+                metadata_filters = [metadata_filters]
+            if not all(
+                metadata_filter.name in [metadata_property.name for metadata_property in self.metadata_properties]
```
gabrielmbmb: I would create a variable `metadata_properties_names` just after the `if metadata_filters`

src/argilla/client/feedback/dataset/base.py
```diff
@@ -213,8 +213,28 @@ def question_by_name(self, name: str) -> Union[AllowedQuestionTypes, "AllowedRem
     def metadata_properties(
         self,
     ) -> Union[List["AllowedMetadataPropertyTypes"], List["AllowedRemoteMetadataPropertyTypes"]]:
+        """Returns the metadata properties that will be indexed and could be used to filter the dataset."""
         return self._metadata_properties
 
+    def metadata_property_by_name(
```
gabrielmbmb, line 219: maybe we could cache the return value of this function

PR #3885, opened 2023-10-06
fix: `sort_by` by metadata property validation

gabrielmbmb:
# Description

This PR fixes the expected value that has to be provided to the `sort_by` query param of the listing records endpoints when using a metadata property. Before, it was just expecting the name of the metadata property `sort_by=metadata-property-name:asc`, which is not correct. Now, it expects `sort_by=metadata.metadata-propert-name:asc`.

**Type of change**

**How Has This Been Tested**

**Checklist**


tests/unit/server/test_utils.py
```diff
@@ -37,6 +37,7 @@
         ({"max_keys": 1}, ["key1:value1,value2,value3"], {"key": "key1", "value": ["value1", "value2", "value3"]}),
         ({"max_keys": 1, "max_values_per_key": 1}, ["key1:value1"], {"key": "key1", "value": "value1"}),
         ({"max_values_per_key": 1}, ["key1:value1", "key2:value2"], {"key1": "value1", "key2": "value2"}),
+        ({"max_values_per_key": 1}, ["key1", "key2"], {"key1": None, "key2": None}),
```
frascuchon: should we also set the `group_keys_without_values=False` to achieve the expected result?

PR #3903, opened 2023-10-09
feat: add computing `total` in listing endpoints when no search engine

gabrielmbmb:
# Description

This PR adds counting the total records matching the query in the listing records endpoints when the search engine is not used. Therefore, the listing records endpoints will return an integer value in the `total` field of the payload no matter the search engine is used (metadata properties where used for filtering) or not.

**Type of change**

**How Has This Been Tested**

Unit tests has been updated.

**Checklist**

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -55,27 +55,27 @@ jobs:
     strategy:
       matrix:
         include:
-# TODO(@frascuchon): Apply this for separate Release pipelines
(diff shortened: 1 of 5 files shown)


alvarobartt:
AFAIK `total` was removed because it was expensive to compute, is this still the case? If not, is it worth to compute it as part of the same endpoint as the one listing the records or could we have a separate one? Asking because the client sends a lot of requests to the API when fetching records

gabrielmbmb:
@alvarobartt I've tested with a dataset with 100000 records, several fields, questions and metadata properties and adding the total query didn't add any noticeable latency

PR #3920, opened 2023-10-10
feat: add `PATCH /api/v1/records/{record_id}` endpoint

gabrielmbmb:
# Description

This PR adds a new endpoint `PATCH /api/v1/records/{record_id}` that allows to partially update a Feedback Dataset record. A new method called `SearchEngine.update_record_metadata` has been added so the record metadata can also be updated on the `SearchEngine`.

**Type of change**

**How Has This Been Tested**

Unit tests covering the additions have been added.

**Checklist**


src/argilla/server/contexts/datasets.py
```diff
@@ -256,7 +257,10 @@ async def get_record_by_id(
 ) -> Union[Record, None]:
     query = select(Record).filter_by(id=record_id)
     if with_dataset:
-        query = query.options(selectinload(Record.dataset).selectinload(Dataset.questions))
+        query = query.options(
+            selectinload(Record.dataset).selectinload(Dataset.questions),
+            selectinload(Record.dataset).selectinload(Dataset.metadata_properties),
```
jfcalvo, line 270: Do we always need `questions` and `metadata_properties` when getting a record via `get_record_by_id`?

src/argilla/server/policies.py
```diff
@@ -369,6 +369,16 @@ async def is_allowed(actor: User) -> bool:
 
 
 class RecordPolicyV1:
+    @classmethod
+    def update(cls, record: Record) -> PolicyAction:
+        async def is_allowed(actor: User) -> bool:
+            return actor.is_owner or (
+                actor.is_admin
+                or await _exists_workspace_user_by_user_and_workspace_id(actor, record.dataset.workspace_id)
```
jfcalvo: This should be `and` instead of `or` right?

```suggestion
                and await _exists_workspace_user_by_user_and_workspace_id(actor, record.dataset.workspace_id)
```

gabrielmbmb: yes!

src/argilla/server/contexts/datasets.py
gabrielmbmb: probably not, I would change this

PR #3924, opened 2023-10-11
Feature: add allowed roles to metadata properties

jfcalvo:
This PR add `allowed_roles` for metadata properties.

This has the potential to affect additional endpoints associated with metadata properties which are still not merged on the feature branch.

If you think this change will affect anything please say something to me and we can discuss about the possible/necessary refactors.

I have updated the draft with the following changes:
* The endpoint `/api/v1/datasets/{dataset.id}/metadata-properties` has been changed to `/api/v1/me/datasets/{dataset.id}/metadata-properties`. (this change should be reflected on the SDK if we are finally accepting the draft @alvarobartt).
* We are internally using `allowed_roles` as a JSON column to store a list of allowed roles to access one specific metadata property.
  * `allowed_roles` is right now an internal column and is not exposed in the API.
* A new `visible_for_annotators` boolean value is accepted to create metadata properties from the API.
  * When `visible_for_annotators` is set as `True` we store `[UserRole.admin, UserRole.annotator]` on `allowed_roles`.
  * When `visible_for_annotators` is set as `False` we store `[UserRole.admin]` on `allowed_roles`.
  * If not specified, `visible_for_annotators` is `True` so admins and annotators has access to the metadata property. (@frascuchon tell me if this is not the expected behavior).
  * `visible_for_annotators` is exposed in the API as a calculated value using `allowed_roles` (as a property on the database model).


src/argilla/server/apis/v1/handlers/datasets.py
```diff
@@ -292,15 +308,19 @@ async def list_dataset_questions(
     return Questions(items=dataset.questions)
 
 
-@router.get("/datasets/{dataset_id}/metadata-properties", response_model=MetadataProperties)
-async def list_dataset_metadata_properties(
+@router.get("/me/datasets/{dataset_id}/metadata-properties", response_model=MetadataProperties)
```
frascuchon, line 311: This change will affect the frontend part. We need to sync with @damianpumar once the PR is ready. We can eventually apply the change in this PR.

@damianpumar The idea regarding metadata property visibility is just to return those metadata properties accessible by the requested user. So your frontend logic won't change. (You will receive only the visible properties)

src/argilla/server/policies.py
```diff
@@ -379,6 +390,7 @@ async def is_allowed(actor: User) -> bool:
                 actor, metadata_property.dataset.workspace_id
             )
 
+            # TODO: Check if we need to change this to support allowed_roles
```
frascuchon: Yes, the metrics should apply the same policy logic as for the `get` action.

src/argilla/server/apis/v1/handlers/datasets.py
```diff
@@ -527,24 +547,26 @@ async def create_dataset_metadata_property(
     db: AsyncSession = Depends(get_async_db),
     search_engine: SearchEngine = Depends(get_search_engine),
     dataset_id: UUID,
-    metadata_prop_create: MetadataPropertyCreate,
+    metadata_property_create: MetadataPropertyCreate,
```
frascuchon, line 550: We should have an update endpoint to allow users change the property visibility (`visible_for_annotators` attribute)  Not sure if should tackled here or in a separate PR @jfcalvo

src/argilla/server/policies.py
jfcalvo: Then I guess we can remove this policy and use simply `get` policy?

jfcalvo: Done, now using `MetadataPropertyPolicyV1.get` instead.

src/argilla/server/apis/v1/handlers/datasets.py
jfcalvo: I will take care of that new update endpoint in a different PR.

src/argilla/server/apis/v1/handlers/datasets.py
alvarobartt: Also the Python SDK needs to be modified to send the requests to the new endpoint adding the `/me`

src/argilla/server/contexts/datasets.py
```diff
@@ -56,6 +56,9 @@
 
 LIST_RECORDS_LIMIT = 20
 
+VISIBLE_FOR_ANNOTATORS_ALLOWED_ROLES = [UserRole.admin, UserRole.annotator]
+NOT_VISIBLE_FOR_ANNOTATORS_ALLOWED_ROLES = [UserRole.admin]
```
alvarobartt, line 61: Just a nit, but maybe we could rename those to something like `METADATA_VISIBLE_ALLOWED_ROLES` and `METADATA_NOT_VISIBLE_ALLOWED_ROLES`, just mentioning this because I recall we decided not to name it `visible_for_annotators` (even this may be the case in the Python client exposed to the user), so maybe the rename makes it clearer in the backend, WDYT?

jfcalvo: We are actually exposing in the API a "virtual/calculated" boolean attribute that is `visible_for_annotators`. I can add `METADATA_PROPERTY` but then is really long like `METADATA_PROPERTY_VISIBLE_FOR_ANNOTATORS_ALLOWED_ROLES`.

src/argilla/server/apis/v1/handlers/datasets.py
frascuchon: Done!

frascuchon:
@jfcalvo I've already updated the endpoint calls in Python SDK and frontend.  The PR is ready to merge!

frascuchon (approved):

src/argilla/server/models/database.py
```diff
@@ -189,6 +189,7 @@ class MetadataProperty(DatabaseModel):
     type: Mapped[MetadataPropertyType] = mapped_column(Text)
     description: Mapped[str] = mapped_column(Text, nullable=True)
     settings: Mapped[dict] = mapped_column(MutableDict.as_mutable(JSON), default={})
+    allowed_roles: Mapped[List[UserRole]] = mapped_column(MutableList.as_mutable(JSON), default=[])
```
gabrielmbmb:
```suggestion
    allowed_roles: Mapped[List[UserRole]] = mapped_column(MutableList.as_mutable(JSON), default=[], server_default="[]")
```

src/argilla/server/alembic/versions/7cbcccf8b57a_create_metadata_properties_table.py
```diff
@@ -33,12 +33,14 @@ def upgrade() -> None:
...
         sa.Column("type", sa.Text(), nullable=False),
-        sa.Column("description", sa.Text(), nullable=True),
+        sa.Column("description", sa.Text()),
         sa.Column("settings", sa.JSON(), nullable=False),
+        sa.Column("allowed_roles", sa.JSON, server_default="[]"),
         sa.Column("dataset_id", sa.Uuid(), nullable=False),
-        sa.Column("id", sa.Uuid(), nullable=False),
         sa.Column("inserted_at", sa.DateTime(), nullable=False),
         sa.Column("updated_at", sa.DateTime(), nullable=False),
         sa.ForeignKeyConstraint(["dataset_id"], ["datasets.id"], ondelete="CASCADE"),
```
gabrielmbmb, line 46: we have to make sure that running `alembic -c src/argilla/alembic.ini revision --autogenerate -m "test"` generates an empty revision meaning that the metadata in the class models is aligned with the metadata in this migration file

src/argilla/server/alembic/versions/7cbcccf8b57a_create_metadata_properties_table.py
```diff
@@ -33,12 +33,14 @@ def upgrade() -> None:
     # ### commands auto generated by Alembic - please adjust! ###
     op.create_table(
         "metadata_properties",
+        sa.Column("id", sa.Uuid(), nullable=False),
         sa.Column("name", sa.Text(), nullable=False),
+        # TODO: We should move type column to settings as an attribute (as we are already doing for Fields and Questions).
         sa.Column("type", sa.Text(), nullable=False),
-        sa.Column("description", sa.Text(), nullable=True),
+        sa.Column("description", sa.Text()),
```
gabrielmbmb, line 40: why have we removed the `nullable=True`?

gabrielmbmb (changes requested):
Hi @jfcalvo, looks good to me! Just a comment about the migration files

src/argilla/server/alembic/versions/7cbcccf8b57a_create_metadata_properties_table.py
jfcalvo: I will fix the migration on the new PR with the changes for updating metadata properties. Thanks for the comment.

src/argilla/server/alembic/versions/7cbcccf8b57a_create_metadata_properties_table.py
jfcalvo: This should be fixed in the next PR with the new endpoint to update metadata properties.

PR #3715, opened 2023-09-06
feat: ensure alignment between `FeedbackDataset` and `RemoteFeedbackDataset`

alvarobartt:
# Description

This PR has been created as a follow up on the recent community feedback and internal discussions on trying to align a bit both the `FeedbackDataset` and the `RemoteFeedbackDataset`, so that there's a clear distinction of the local one from the remote, respectively.

Some of these changes imply:
* Add `sync` arg to `from_argilla` to either return a `FeedbackDataset` or a `RemoteFeedbackDataset` based on its value, can either be False or True, respectively
* Add a warning message every time `from_argilla` is called with `sync=True` to state that all the additions, deletions and updates will be automatically pushed to Argilla
* Add the `filter_by` method for `FeedbackDataset` in case users want to use the same methods on both dataset types, but raising a `NotImplementedError` for the moment
* Add the `push_to_argilla` method back to `RemoteFeedbackDataset` so that a warning message is shown to let users know that there's no need to call `push_to_argilla` after every update anymore
* Rename `pull` to `pull_from_argilla` while still keeping `pull` for `RemoteFeedbackDataset` and `FilteredRemoteFeedbackDataset`

**Type of change**

- [X] New feature (non-breaking change which adds functionality)
- [X] Refactor (change restructuring the codebase without changing functionality)
- [X] Improvement (change adding some improvement to an existing functionality)

**How Has This Been Tested**

- [x] Add unit/integration tests to capture the included warning messages

**Checklist**

- [ ] I added relevant documentation
- [X] follows the style guidelines of this project
- [X] I did a self-review of my code
- [ ] I made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I filled out [the contributor form](https://tally.so/r/n9XrxK) (see text above)
- [x] I have added relevant notes to the CHANGELOG.md file (See https://keepachangelog.com/)

PR #3925, opened 2023-10-11
feat: Support for sort by from Python SDK

frascuchon:
# Description

This PR introduces the `sort_by` method for the Python SDK.

Also, apply some code refactoring the the `FeedbackDataset` classes:
1. The remote dataset and remote records have been unified in `RemoteFeedbackDataset` and `RemoteFeedbackRecords` 
2. The `FeedbackDatasetBase` class defines **all** methods available for feedback datasets. The "local" version of the dataset just warns and returns the same dataset.
3. The integration tests have been reviewed, and the search engine is not mocked anymore.

Refs #3748


src/argilla/client/feedback/dataset/remote/dataset.py
```diff
@@ -43,29 +42,82 @@
...
-class RemoteFeedbackRecords(RemoteFeedbackRecordsBase):
-    def __init__(self, dataset: "RemoteFeedbackDataset") -> None:
-        super().__init__(dataset=dataset)
+class RemoteFeedbackRecords(ArgillaRecordsMixin):
+    def __init__(
+        self,
+        dataset: "RemoteFeedbackDataset",
+        response_status: Optional[Union["FeedbackResponseStatusFilter", List["FeedbackResponseStatusFilter"]]] = None,
+        metadata_filters: Optional[Union["MetadataFilters", List["MetadataFilters"]]] = None,
+        sort_by: Optional[List[SortBy]] = None,
```
davidberenstein1957, line 52: @frascuchon can we once again add a sort_by  datetime here?

davidberenstein1957: inserted_at I mean

tests/unit/client/feedback/dataset/test_local.py
```diff
@@ -94,3 +94,39 @@ def test_add_records_validation_error(record: FeedbackRecord, exception_cls: Exc
...
+    )
+    with pytest.warns(
+        UserWarning,
+        match="`sort_by` method only works for `FeedbackDataset` pushed to Argilla."
+        " Use `sorted` with dataset.records instead.",
+    ):
+        assert dataset.sort_by("field") == dataset
+
+
+def test_filter_by_for_local_is_not_implemented():
```
davidberenstein1957, line 135: can we create good first issue work items to support this?

tests/unit/client/feedback/dataset/test_local.py
```diff
@@ -94,3 +94,39 @@ def test_add_records_validation_error(record: FeedbackRecord, exception_cls: Exc
     with pytest.raises(exception_cls, match=exception_msg):
         dataset.add_records(record)
     assert len(dataset.records) == 0
+
+
+def test_sort_by_for_local_is_not_implemented():
```
davidberenstein1957, line 117: can we create good first issue work items to support this?

tests/unit/client/feedback/dataset/test_base.py
```diff
@@ -12,27 +12,43 @@
...
+    def sort_by(
+        self, field: Union[str, RecordSortField], order: Union[str, SortOrder] = SortOrder.asc
+    ) -> "TestFeedbackDataset":
+        return self
+
     @property
-    def records(self) -> None:
-        pass
+    def records(self) -> Iterable[FeedbackRecord]:
+        return []
```
davidberenstein1957, line 56: I don't see any tests with working sorts?

src/argilla/client/feedback/dataset/remote/dataset.py
```diff
@@ -141,27 +250,187 @@ def __init__(
...
+    def created_at(self) -> datetime:
+        """Returns the datetime when the dataset was created in Argilla."""
+        return self._created_at
+
+    @property
+    def updated_at(self) -> datetime:
+        """Returns the datetime when the dataset was last updated in Argilla."""
+        return self._updated_at
+
+    def __repr__(self) -> str:
```
davidberenstein1957, line 337: maybe this can be formatted a bit more nicely with multi_line. Also, I would say thing like the filters and sorts should be added, when present.

src/argilla/client/feedback/dataset/remote/dataset.py
```diff
@@ -141,27 +250,187 @@ def __init__(
...
+
+        instance = FeedbackDataset(
+            fields=self.fields,
+            questions=self.questions,
+            guidelines=self.guidelines,
+            metadata_properties=self.metadata_properties,
         )
+        records = [record.to_local() for record in self._records]
+
+        if len(records) > 0:
```
davidberenstein1957, line 422: else a warning might be good? Not sure, but when pulling a filtered dataset or something it might be nice to know if/why it is empty? However, perhaps it would be better to provide a warning when a filter_by returns an empty dataset?

davidberenstein1957 (commented):
Hi Paco, I also added a shared ABC here: https://github.com/argilla-io/argilla/pull/3715/files. Not sure what you think about that but for me it makes sense to enforce alignment between the different datasets a bit more strongly and work out these basic functionalities for both the local one via good first issues and thing the interns might be able to pick up to ensure API-alignment for the remote and local versions.

src/argilla/client/feedback/dataset/remote/dataset.py
frascuchon: Yes, The SortBy.fields is or str or `RecordSortBy.inserted_at` and `RecordSortBy.inserted_at` which can be configure by string directly:
```python
ds.sort_by([SortBy(field="inserted_at", order="desc")]
```

tests/unit/client/feedback/dataset/test_base.py
frascuchon: The related tests for sorting are here.

https://github.com/argilla-io/argilla/pull/3925/files#diff-9dc83346e5235364ae1e990daffc664e649bff9d04eb522bcb59ef382eec8607R199

src/argilla/client/feedback/dataset/remote/dataset.py
frascuchon: Make sense to include a warning if no data found when pulling

tests/unit/client/feedback/dataset/test_local.py
frascuchon: I would wait until we review the workflows working with datasets. Maybe later the local version will make less sense (or not)

tests/unit/client/feedback/dataset/test_local.py
frascuchon: ditto

src/argilla/client/feedback/dataset/remote/dataset.py
frascuchon: Agree, and I think this is a good first issue for someone to contribute for

src/argilla/client/feedback/schemas/records.py
```diff
@@ -19,6 +19,7 @@
 from pydantic import BaseModel, Extra, Field, PrivateAttr, StrictInt, StrictStr, conint, validator
 
 from argilla.client.feedback.schemas.enums import ResponseStatus
+from argilla.server.enums import RecordSortField, SortOrder
```
alvarobartt: We shouldn't be importing stuff from the `argilla.server` so I guess you meant

```suggestion
from argilla.client.feedback.schemas.enums import RecordSortField, SortOrder
```

src/argilla/client/feedback/schemas/records.py
```diff
@@ -251,3 +252,32 @@ def to_server_payload(self, question_name_to_id: Optional[Dict[str, UUID]] = Non
...
             payload["external_id"] = self.external_id
         return payload
+
+
+class SortBy(BaseModel):
+    field: Union[str, RecordSortField]
+    order: Union[str, SortOrder] = SortOrder.asc
+
+    @validator("field")
+    def check_field_name(cls, field: Union[str, RecordSortField]):
```
alvarobartt:
```suggestion
    def check_field_name(cls, field: Union[str, RecordSortField]) -> RecordSortField:
```

src/argilla/client/feedback/schemas/records.py
```diff
@@ -251,3 +252,32 @@ def to_server_payload(self, question_name_to_id: Optional[Dict[str, UUID]] = Non
...
+        if isinstance(field, str) and field.startswith("metadata."):
+            return field
+        elif isinstance(field, str):
+            return RecordSortField(field)
+        return field
```
alvarobartt:
```suggestion
        return RecordSortField(field) if isinstance(field, str) else field
```

src/argilla/client/feedback/schemas/records.py
```diff
@@ -251,3 +252,32 @@ def to_server_payload(self, question_name_to_id: Optional[Dict[str, UUID]] = Non
...
+    @validator("field", pre=True)
+    def check_record_field(cls, field):
+        try:
+            return RecordSortField(field)
+        except:
+            return field
+
+    @validator("order")
+    def check_order(cls, order):
+        return SortOrder(order)
```
alvarobartt: Why do we need these? Isn't it better to just add one validator per field i.e. one for `field` and another one for `order` with `pre=True` and just parsing the `str` to `RecordSortField` and `SortOrder`?

src/argilla/client/sdk/v1/datasets/api.py
```diff
@@ -191,12 +193,15 @@ def get_records(
 
     params = {"include": ["responses", "suggestions"], "offset": offset, "limit": limit}
 
-    if response_status is not None:
+    if response_status:
```
alvarobartt, line 201: I think `is not None` is more consistent here, is there any reason for this change?

tests/integration/client/feedback/dataset/remote/test_dataset.py
```diff
@@ -17,7 +17,14 @@
...
 import pytest
-from argilla import FeedbackRecord
+from argilla import (
+    FeedbackRecord,
+    FloatMetadataProperty,
+    IntegerMetadataProperty,
+    TermsMetadataProperty,
+    TextField,
+    TextQuestion,
+)
```
alvarobartt: Do we really want to import these `from argilla` instead of from their respective module? e.g.

```suggestion
from argilla.client.feedback.schemas.records import FeedbackRecord
from argilla.client.feedback.schemas.metadata import FloatMetadataProperty, IntegerMetadataProperty, TermsMetadataProperty
from argilla.client.feedback.schemas.fields import TextField
from argilla.client.feedback.schemas.questions import TextQuestion
```

tests/integration/client/feedback/dataset/remote/test_dataset.py
```diff
@@ -83,38 +101,48 @@ async def test_from_argilla_with_responses(self, owner: "User", statuses: List[s
...
 
         api.init(api_key=user.api_key)
-        remote_dataset = FeedbackDataset.from_argilla(id=dataset.id)
+        remote = test_dataset.push_to_argilla(name="test_dataset", workspace=ws)
+        remote_dataset = FeedbackDataset.from_argilla(id=remote.id)
         remote_dataset.delete()
 
-        datasets = api.active_api().http_client.get("/api/v1/me/datasets")["items"]
-        assert not any(ds["name"] == remote_dataset.name for ds in datasets)
+        with pytest.raises(Exception):
```
alvarobartt: Here we should handle the exception raised with a `match` arg

tests/integration/client/feedback/dataset/remote/test_filter_and_sorting.py
```diff
@@ -17,37 +17,52 @@
 from uuid import UUID
 
 import pytest
+from argilla import SortBy, TextField, TextQuestion
```
alvarobartt, line 20: Same here, do we want to import `from argilla` or from the specific module?

src/argilla/client/feedback/schemas/records.py
frascuchon: RecordSortField is only for `inserted_at` and `updated_at` values. For metadata you should pass a `metadata.your_prop` string

alvarobartt: Then the code should look like the following, right?

```suggestion
        return RecordSortField(field) if isinstance(field, str) and not field.startswith("metadata.") else field
```

src/argilla/client/sdk/v1/datasets/api.py
frascuchon: just to avoid passing a []

src/argilla/client/feedback/schemas/records.py
alvarobartt: What I mean is that if an `if-else-if` statement returns the same object inside one of the if-conditions than outside, it can be simplified as an `if-else` instead, in case you prefer that over the Python one-liner that can also be done as

```suggestion
        if isinstance(field, str) and not field.startswith("metadata."):
            return RecordSortField(field)
        return field
```

tests/integration/client/feedback/dataset/remote/test_dataset.py
frascuchon: Do you mean in tests? I would like to use them as a user should use them, so yes, they should be imported from argilla. Otherwise, we cannot check that expected schemas have been properly configured in the main module

alvarobartt: Ok cool, then we should probably create a separate PR at some point to refactor all the integration tests to use `import argilla as rg` instead, as that's what we're pointing users in the docs

tests/integration/client/feedback/dataset/remote/test_filter_and_sorting.py
frascuchon: same answer as before. We should define the integration tests as users should use the python sdk

src/argilla/client/feedback/dataset/base.py
```diff
@@ -161,7 +164,7 @@ def __init__(
         # self._extra_metadata_allowed = extra_metadata_allowed
 
     @property
-    @abstractproperty
+    @abstractmethod
     def records(self) -> Any:
         """Returns the records of the dataset."""
         pass
```
alvarobartt, line 171: Just as a side-comment, we should probably briefly discuss which standard do we want to apply for the `@property` + `@abstractmethod` and for the `@abstractmethod` as in some cases we're just adding the docstring, `pass`, `...` or `raise NotImplementedError`, so I think it would be nice to align here.

My personal preference is `...` or `pass`

src/argilla/client/feedback/dataset/local.py
```diff
@@ -34,7 +36,7 @@ def __init__(
         self,
         *,
         fields: List["AllowedFieldTypes"],
-        questions: List[AllowedQuestionTypes],
+        questions: List["AllowedQuestionTypes"],
```
alvarobartt, line 38: No need to use double-quotes as it's not under the `TYPE_CHECKING`

```suggestion
        questions: List[AllowedQuestionTypes],
```

src/argilla/client/feedback/dataset/local.py
```diff
@@ -188,3 +190,24 @@ def add_records(
...
+        warnings.warn(
+            "`sort_by` method only works for `FeedbackDataset` pushed to Argilla. "
+            "Use `sorted` with dataset.records instead."
+        )
+        return self
+
+    def filter_by(
+        self,
+        *,
+        response_status: Optional[Union[FeedbackResponseStatusFilter, List[FeedbackResponseStatusFilter]]] = None,
```
alvarobartt:
```suggestion
        response_status: Optional[Union[ResponseStatusFilter, List[ResponseStatusFilter]]] = None,
```

src/argilla/client/feedback/dataset/local.py
```diff
@@ -11,14 +11,16 @@
...
 from typing import TYPE_CHECKING, Any, Dict, Iterator, List, Optional, Union
 
 from argilla.client.feedback.constants import FETCHING_BATCH_SIZE
 from argilla.client.feedback.dataset.base import FeedbackDatasetBase
 from argilla.client.feedback.dataset.mixins import ArgillaMixin, UnificationMixin
-from argilla.client.feedback.schemas.fields import TextField
+from argilla.client.feedback.schemas.enums import RecordSortField, SortOrder
+from argilla.client.feedback.schemas.metadata import MetadataFilters
 from argilla.client.feedback.schemas.types import AllowedQuestionTypes
+from argilla.client.sdk.v1.datasets.models import FeedbackResponseStatusFilter
```
alvarobartt:
```suggestion
from argilla.client.feedback.schemas.enums import ResponseStatusFilter
```

src/argilla/client/feedback/dataset/remote/dataset.py
```diff
@@ -141,27 +250,187 @@ def __init__(
...
+                of records. But those need to be previously pushed to Argilla, otherwise
+                they won't be deleted.
+
+        Raises:
+            PermissionError: if the user does not have either `owner` or `admin` role.
+            RuntimeError: If the deletion of the records from Argilla fails.
+        """
+        self._records.delete(records=[records] if not isinstance(records, list) else records)
+
+    def pull(self) -> "FeedbackDataset":
```
alvarobartt, line 405: We could also probably rename this to `pull_from_argilla` to make it more clear as in the former PR and keep the `pull` with a `DeprecationWarning`

src/argilla/client/feedback/dataset/remote/dataset.py
```diff
@@ -239,3 +468,22 @@ def delete(self) -> None:
             datasets_api_v1.delete_dataset(client=self._client, id=self.id)
         except Exception as e:
             raise RuntimeError(f"Failed while deleting the `FeedbackDataset` from Argilla with exception: {e}") from e
+
+    @classmethod
+    def __copy_from(cls, dataset: "RemoteFeedbackDataset") -> "RemoteFeedbackDataset":
```
alvarobartt: Why is this private? Also the naming `copy_from` is not clear to me, maybe we can simplify the naming here?

src/argilla/client/feedback/dataset/remote/dataset.py
```diff
@@ -141,27 +250,187 @@ def __init__(
...
+
+        Returns:
+            The record(s) at the given index(es).
+        """
+        return self._records.__getitem__(key)
+
+    def sort_by(self, sort: List[SortBy]) -> "RemoteFeedbackDataset":
+        """Sorts the current `RemoteFeedbackDataset` based on the given sort fields and orders."""
+        sorted_dataset = self.__copy_from(self)
+        sorted_dataset._records = RemoteFeedbackRecords.copy_from(sorted_dataset, sort_by=sort)
```
alvarobartt: I think the previous approach using the `records_cls` as a global class attribute may be more consistent here as one should just `cls.records_cls = ...` and then instantiate `cls(...)`

frascuchon: We only need one single `RemoteFeedbackRecords` implementation for this kind of dataset, so the conditional record class is not needed anymore. Also, my feeling reading the code was that the logic was diluted in several places and it was hard to track what was going on.

src/argilla/client/feedback/dataset/remote/dataset.py
frascuchon: Agree, but this would be tackled in a separate PR

src/argilla/client/feedback/dataset/remote/dataset.py
frascuchon: Since this logic is not part of the API definition, I would avoid exposing it as public. This method just creates a dataset copy using as source the passed as argument. We can call it `__create_copy`

PR #3934, opened 2023-10-11
feat: add `PATCH /api/v1/dataset/{dataset_id}/records` endpoint

gabrielmbmb:
# Description

This PR adds the following:

- Updates `PATCH /api/v1/records/{record_id}` added in #3920 allowing also to update the suggestions of a record. The suggestions in the input payload will **replace** the old suggestions.
- Add new `PATCH /api/v1/datasets/{dataset_id}/records` endpoint allowing to batch/bulk update the records of a dataset. The endpoint allow to update the same attributes from the record as in the `PATCH /api/v1/records/{record_id}` endpoint.
- Slightly modify the `SearchDocument` getter dict to do not try to populate the `SearchDocument.responses` attribute if the relationship has not been loaded (this allows us to not to have to load the `Record.responses` when updating the record document in the `SearchEngine` using `add_records`)
- Removes `SearchEngine.update_record_metadata` method as the same logic is covered by `SearchEngine.add_records` method, which can be also used to update the fields of an existing document.
- Rename `SearchEngine.add_records` method to `SearchEngine.index_records` as it can be used to both add and update records.

**Type of change**

**How Has This Been Tested**

I made an small benchmark to test the latency of the new endpoint. I've created a dataset with 100000 records and all the possible questions and metadata properties. Then I built batches of 1000 records, updating all the responses and metadata fields, and sent them to the API. The average response time of the bulk `PATCH` endpoint was ~= 0.8 seconds. 

<details>
  <summary>Code used for benchmark</summary>

  ```python
import uuid
import random
import argilla as rg

LABELS = ["a", "b", "c"]
RANKS = ["top-1", "top-2", "top-3"]

dataset = rg.FeedbackDataset(
    fields=[rg.TextField(name="text")],
    questions=[
        rg.TextQuestion(name="text"),
        rg.RatingQuestion(name="rating", values=[1, 2, 3, 4, 5]),
        rg.LabelQuestion(name="label", labels=LABELS),
        rg.MultiLabelQuestion(name="multi-label", labels=LABELS),
        rg.RankingQuestion(name="ranking", values=RANKS),
    ],
    metadata_properties=[
        rg.TermsMetadataProperty(name="label", values=LABELS),
        rg.IntegerMetadataProperty(name="integer", min=0, max=10),
        rg.FloatMetadataProperty(name="float", min=0, max=10),
    ],
)

dataset.add_records([rg.FeedbackRecord(fields={"text": "Hello"}, metadata={"extra": "yes"}) for _ in range(100000)])

remote = dataset.push_to_argilla(name=f"benchmark-{uuid.uuid4()}", workspace="gabriel")


def random_rank_order():
    ranks = RANKS.copy()
    ranks.sort(key=lambda x: random.random())
    return [{"value": rank, "rank": i + 1} for i, rank in enumerate(ranks)]


def build_update_payload(record):
    return {
        "id": str(record.id),
        "external_id": str(uuid.uuid4()),
        "metadata": {
            "label": random.choice(["a", "b", "c"]),
            "integer": random.randint(0, 10),
            "float": random.uniform(0, 10),
        },
        "suggestions": [
            {"question_id": str(remote.questions[0].id), "value": "hello world" * random.randint(1, 15)},
            {"question_id": str(remote.questions[1].id), "value": random.randint(1, 5)},
            {"question_id": str(remote.questions[2].id), "value": random.choice(["a", "b", "c"])},
            {"question_id": str(remote.questions[3].id), "value": [random.choice(["a", "b", "c"])]},
            {"question_id": str(remote.questions[4].id), "value": random_rank_order()},
        ],
    }

http_client = rg.active_client().http_client

elapseds = []

batch = []
for record in remote.records:
    batch.append(build_update_payload(record))

    if len(batch) == 1000:
        response = http_client.httpx.patch(f"/api/v1/datasets/{remote.id}/records", json={"items": batch})
        elapseds.append(response.elapsed.total_seconds())
        batch = []

average_elapsed_time = sum(elapseds) / len(elapseds)

print("Average elapsed time", average_elapsed_time)


  ```
</details>

**Checklist**


src/argilla/server/contexts/datasets.py
```diff
@@ -536,24 +535,126 @@ async def create_records(
...
+            params.pop("suggestions")
+            for suggestion_i, suggestion in enumerate(record_update.suggestions):
+                try:
+                    questions = await _validate_suggestion(db, suggestion, questions)
+                    suggestions.append(Suggestion(record_id=record_update.id, **suggestion.dict()))
+                except ValueError as err:
+                    raise ValueError(
+                        f"Provided suggestion for record at position {record_i} and suggestion at position "
+                        f"{suggestion_i} is not valid: {err}"
+                    ) from err
```
gabrielmbmb, line 593: I'm going to add a check for checking that there is just one suggestion for each question_id per record, otherwise we can get an `IntegrityError` later

src/argilla/server/search_engine/commons.py
```diff
@@ -73,6 +73,11 @@ def _build_metadata_field_payload(dataset: Dataset, metadata: Union[Dict[str, An
 class SearchDocumentGetter(GetterDict):
     def get(self, key: Any, default: Any = None) -> Any:
         if key == "responses":
+            # `responses` of the record haven't been loaded, set the default value so when using
+            # `SearchDocument(...).dict(exclude_unset=True)` the field is not included.
+            if "responses" not in self._obj.__dict__:
```
frascuchon: Does make sense to move this logic to a computed property in the ORM model? 
```python
if self._obj.is_responses_loaded
# or
if self._obj.is_loaded("responses")
```

src/argilla/server/contexts/datasets.py
```diff
@@ -536,24 +535,126 @@ async def create_records(
...
+    records_ids = [record_update.id for record_update in records_update.items]
+
+    if len(records_ids) != len(set(records_ids)):
+        raise ValueError("Found duplicate records IDs")
+
+    existing_records_ids = await _exists_records_with_ids(db, dataset_id=dataset.id, records_ids=records_ids)
+    non_existing_records_ids = set(records_ids) - set(existing_records_ids)
+
+    if len(non_existing_records_ids) > 0:
+        raise ValueError("Found records that do not exist")
```
frascuchon: Does make sense to return the `non_existing_records_ids` list?

src/argilla/server/contexts/datasets.py
```diff
@@ -536,24 +535,126 @@ async def create_records(
...
+
+        records_update_objects.append(params)
+
+    async with db.begin_nested():
+        params = [Suggestion.record_id.in_(records_delete_suggestions)]
+        await Suggestion.delete_many(db, params=params, autocommit=False)
+        db.add_all(suggestions)
+        await Record.update_many(db, records_update_objects, autocommit=False)
+        records = await get_records_by_ids(db, dataset_id=dataset.id, records_ids=records_search_engine_update)
+        await search_engine.add_records(dataset, records)
```
frascuchon: Maybe now that we're using this method to add or update records, we could rename it to `index_records`, wdyt?

frascuchon (approved):

PR #3945, opened 2023-10-13
fix: create `inserted_at` and `update_at` attributes with proper precission

frascuchon:
# Description

This PR forces `inserted_at` and `updated_at` attribute creation using the `datetime.utcnow` factory to avoid some unexpected scenarios with the `func.now()` SQLAlchemy function.

**Type of change**

**How Has This Been Tested**

Locally tests with PostgreSQL and SQLite

**Checklist**

### DIFF
--- a/src/argilla/server/models/mixins.py
+++ b/src/argilla/server/models/mixins.py
@@ -160,7 +160,7 @@ class CRUDMixin:
 
 
 def _default_inserted_at(context: DefaultExecutionContext) -> datetime:
-    return context.get_current_parameters()["inserted_at"]
+    return context.get_current_parameters(isolate_multiinsert_groups=False)["inserted_at"]
 
 
 class TimestampMixin:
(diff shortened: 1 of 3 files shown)


src/argilla/server/models/mixins.py
```diff
@@ -158,6 +159,10 @@ async def save(self, db: "AsyncSession", autocommit: bool = True) -> Self:
         return self
 
 
+def _default_inserted_at(context: DefaultExecutionContext) -> datetime:
+    return context.get_current_parameters()["inserted_at"]
```
gabrielmbmb: To make it work in all the cases:

```suggestion
    return context.get_current_parameters(isolate_multiinsert_groups=False)["inserted_at"]
```

gabrielmbmb (changes requested):
Hi @frascuchon, thanks for the fix! I tried this a long time ago but it was not working for the upsert. I realise that we need to set the `isolate_multiinsert_groups=False` in order to make it work with the upsert.

src/argilla/server/models/mixins.py
frascuchon: Thanks, I saw this this morning, but I don't really understand why this solves the tests. Related to async connections?

PR #3946, opened 2023-10-14
feat: Support for update records from SDK

### LINKED ISSUE #3897: [FEATURE] Update `metadata` for a `FeedbackRecord` in Argilla

frascuchon:
# Description

This PR introduces support for updating records from Python SDK. The records update workflow has 2 ways to be implemented. In this PR both are supported:
Single update:
```python
records = ds.records[:100]
for record in records:
   record.metadata.update({"new": "metadata"})
   record.update() 
```
or batch update:
```python
records = ds.records[:100]
for record in records:
   record.metadata.update({"new": "metadata"})
   record.suggestions = [...]

ds.update_records(records)
```
This is still compatible with previous record update functionality (where only suggestions could be updated)

```python
for record in records:
   record.update(suggestions=[...])
```

In order to support this, some changes have been introduced affecting records immutability, which is now removed. The reason behind this is to provide a similar way to update local and remote records (change your data at record level and then call the `update_records` method) since the `record.update` method is only available for synced/remote entities.

The metadata validation is still missing on the record update and will be tackled as a separate PR since extra changes could be potentially introduced.

Refs #3748 

**Type of change**

**How Has This Been Tested**

All the flows described below have been tested locally.

**Checklist**


src/argilla/client/feedback/schemas/remote/records.py
```diff
@@ -117,37 +120,29 @@ class RemoteFeedbackRecord(FeedbackRecord, RemoteSchema):
...
-    suggestions: Union[Tuple[RemoteSuggestionSchema], List[RemoteSuggestionSchema]] = Field(
-        default_factory=tuple, allow_mutation=False
-    )
+    suggestions: Union[Tuple[AllowedSuggestionSchema], List[AllowedSuggestionSchema]] = Field(default_factory=tuple)
 
     class Config:
+        allow_mutation = True
         validate_assignment = True
 
-    def __update_suggestions(
```
frascuchon, line 130: Here, the old `__update_suggestions` method has been split into 2 steps: 1. validate and normalize/filter suggestions and 2. prepare and call the update endpoints.

src/argilla/client/feedback/dataset/remote/dataset.py
```diff
@@ -311,6 +319,14 @@ def records(self) -> RemoteFeedbackRecords:
...
         return self._records
 
+    def update_records(self, records: Union[RemoteFeedbackRecord, List[RemoteFeedbackRecord]]) -> None:
+        if not isinstance(records, list):
+            records = [records]
+
+        # TODO: Use the batch version of endpoint once is implemented
+        for record in records:
+            record.update()
+
```
alvarobartt, line 329: Which is the scenario where someone modifies a `RemoteFeedbackRecord` and then pushes it if not via `RemoteFeedbackRecord.update`? Are we allowing the assignment there? e.g. `record.metadata = {"a": 1}`, if so, won't this be conflictive?

src/argilla/client/feedback/dataset/remote/dataset.py
```diff
@@ -105,6 +105,14 @@ def _client(self) -> "httpx.Client":
...
         return self.dataset._client
 
+    @property
+    def _question_id_to_name(self) -> Dict["UUID", str]:
+        return self.dataset._question_id_to_name_id
+
+    @property
+    def _question_name_to_id(self) -> Dict[str, "UUID"]:
+        return self.dataset._question_name_to_id
+
```
alvarobartt, line 115: Do we actually need to wrap those properties under the same name? IMO we can just re-use those from `self.dataset`

src/argilla/client/feedback/schemas/records.py
```diff
@@ -88,6 +88,7 @@ def to_server_payload(self) -> Dict[str, Any]:
         """Method that will be used to create the payload that will be sent to Argilla
         to create a `ResponseSchema` for a `FeedbackRecord`."""
         return {
+            # UUID is not json serializable!!!
```
alvarobartt, line 91:
```suggestion
```

Yes, it's not, for the moment we're checking the `user_id` in the Python SDK in `add_records`, but we can review this later to just parse it as a `str` in the `to_server_payload` method

src/argilla/client/feedback/schemas/records.py
```diff
@@ -194,9 +195,7 @@ class FeedbackRecord(BaseModel):
     fields: Dict[str, Union[str, None]]
     metadata: Dict[str, Any] = Field(default_factory=dict)
     responses: List[ResponseSchema] = Field(default_factory=list)
-    suggestions: Union[Tuple[SuggestionSchema], List[SuggestionSchema]] = Field(
-        default_factory=tuple, allow_mutation=False
-    )
+    suggestions: Union[Tuple[SuggestionSchema], List[SuggestionSchema]] = Field(default_factory=tuple)
```
alvarobartt, line 198: I think I already mentioned this, but the `tuple` may be confusing, I'm more comfortable with the `list`, in any case, `responses` is still a list, so we should align that at some point

src/argilla/client/feedback/schemas/remote/records.py
```diff
@@ -117,37 +120,29 @@ class RemoteFeedbackRecord(FeedbackRecord, RemoteSchema):
...
+    def __normalize_suggestions_to_update(
```
alvarobartt, line 133: I think we should carefully review the `suggestions` update/addition workflow, because I think we're adding too much complexity here that can probably be simplified

src/argilla/client/feedback/schemas/remote/shared.py
```diff
@@ -21,6 +21,7 @@
 
 
 class RemoteSchema(BaseModel, ABC):
+    # TODO(@alvarobartt): Review optional id configuration for remote schemas
```
alvarobartt, line 24: It's optional for the moment, as we're not using those neither for the `fields` nor for the `questions`, but sure, we should make it mandatory once it's fully covered, as well as the `client`

src/argilla/client/feedback/schemas/remote/shared.py
```diff
@@ -30,6 +31,7 @@ def _client(self) -> Optional[httpx.Client]:
         return self.client
 
     class Config:
+        # TODO(@alvarobart) Not sure if we need this at this level
```
alvarobartt, line 34: `Config` is inherited AFAIK, but we do need to `arbitrary_types_allowed = True` here because otherwise we cannot add the `httpx.Client` type-hint

src/argilla/client/sdk/v1/records/api.py
```diff
@@ -12,16 +12,41 @@
...
+from argilla.client.sdk.v1.datasets.models import FeedbackSuggestionModel
 from argilla.client.sdk.v1.records.models import FeedbackItemModel
 
 
+def update_record(
+    # TODO: Use the proper sdk API Model instead of the dict
+    client: httpx.Client,
+    id: Union[str, UUID],
+    data: Dict[str, Any],
+) -> Response[Union[FeedbackItemModel, ErrorMessage, HTTPValidationError]]:
```
alvarobartt, line 31: Missing docstrings here!

src/argilla/client/workspaces.py
```diff
@@ -262,7 +262,7 @@ def __active_client() -> "httpx.Client":
             raise RuntimeError(f"The `rg.active_client()` is not available or not respoding.") from e
 
     @classmethod
-    def __new_instance(
+    def _new_instance(
```
alvarobartt, line 265: Why is this change? Is there any strong reason for it?

tests/integration/client/feedback/dataset/remote/test_dataset.py
```diff
@@ -119,6 +120,62 @@ async def test_add_records(self, owner: "User", test_dataset: FeedbackDataset, r
 
         assert len(remote_dataset.records) == 1
 
+    async def test_update_records(self, owner: "User", test_dataset: FeedbackDataset):
+        import argilla as rg
```
alvarobartt, line 124: Why don't we import this globally on top of the file?

tests/integration/client/feedback/dataset/remote/test_dataset.py
```diff
@@ -119,6 +120,62 @@ async def test_add_records(self, owner: "User", test_dataset: FeedbackDataset, r
...
+        remote.update_records(first_record)
+
+        assert first_record == remote[0]
+
+        first_record = remote[0]
+        assert first_record.external_id == "new-external-id"
+        assert first_record.metadata["terms-metadata"] == "a"
+
+    async def test_update_records_with_suggestions(self, owner: "User", test_dataset: FeedbackDataset):
+        import argilla as rg
```
alvarobartt, line 151: Ditto

tests/unit/client/feedback/dataset/remote/test_dataset.py
```diff
@@ -0,0 +1,251 @@
...
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+from datetime import datetime
+from uuid import uuid4
+
+import httpx
+import pytest
+from argilla import Workspace
```
alvarobartt, line 20: As this is a unit test maybe for consistency we can just rename this import to
```suggestion
from argilla.client.workspaces import Workspace
```

alvarobartt: Note that the `pre-commit` will fail if you apply the suggestion above, because that import should be placed lower

tests/unit/client/feedback/dataset/test_base.py
```diff
@@ -37,6 +36,9 @@
 
 
 class TestFeedbackDataset(FeedbackDatasetBase):
+    def update_records(self, records: Union[FeedbackRecord, List[FeedbackRecord]]) -> None:
```
alvarobartt: As this is unused we can just 
```suggestion
    def update_records(self, **kwargs: Dict[str, Any]) -> None:
```

alvarobartt (approved):

src/argilla/client/feedback/schemas/records.py
```diff
@@ -88,6 +88,7 @@ def to_server_payload(self) -> Dict[str, Any]:
         """Method that will be used to create the payload that will be sent to Argilla
         to create a `ResponseSchema` for a `FeedbackRecord`."""
         return {
+            # UUID is not json serializable!!!
             "user_id": self.user_id,
```
gabrielmbmb, line 92: this should be `str(self.user_id)`, right?

gabrielmbmb (approved):

src/argilla/client/feedback/dataset/remote/dataset.py
frascuchon: Once we have the batch version of records update, the updates should be done using the batch version, since it has a better performance than the per-record update. The `record.update` is a way to support the current behaviour but i think it should be deprecated and removed

src/argilla/client/feedback/dataset/remote/dataset.py
frascuchon: There are several places using this variable. The idea should be to remove them and start using the dataset ones, but for this, we need to refactor some remote schemas first. I would like to keep this PR with minimal changes

src/argilla/client/feedback/schemas/records.py
frascuchon: I agree, but we need to do this taking into account that tuples have been used in current releases. My change here just removes the `allow_mutation=True`. Other extra things should be tackled in separate PRs. Otherwise, a lot of changes could be included here without a need.

src/argilla/client/feedback/schemas/remote/records.py
frascuchon: yes, but separate PR. I didn't change the internal logic, just separated the method in 2 different ones.

src/argilla/client/workspaces.py
frascuchon: we need to create workspaces in unit tests, a there is no easy way since the `__init__` cannot be used.

src/argilla/client/feedback/schemas/records.py
frascuchon: It should be, but a lot of tests must be changed to support this. I put the comment to don't forget and tackle in a separate PR

tests/unit/client/feedback/dataset/remote/test_dataset.py
frascuchon: The pre-commit should apply a commit fixing these kinds of things. As I mentioned, I prefer to use Argilla models as a user should do, so it's better to  import as `from argilla import Workspace`

tests/integration/client/feedback/dataset/remote/test_dataset.py
frascuchon: I will change this in the validation PR

PR #3952, opened 2023-10-16
feat: add endpoint to update metadata properties

jfcalvo:
# Description

This PR adds the following changes:

* Add a new endpoint `PATCH /api/v1/metadata-properties/{metadata_property_id}` allowing the update of metadata properties. Right now only `title` and `visible_for_annotators` attributes are allowed in the update.
* `title` has been added to metadata properties.
* `description` has been removed from metadata properties.

**Type of change**

- [x] New feature (non-breaking change which adds functionality)

**How Has This Been Tested**

- [x] Adding unit tests for the new endpoint.
- [x] Running unit tests using PostgreSQL.

**Checklist**

- [ ] I added relevant documentation
- [x] follows the style guidelines of this project
- [x] I did a self-review of my code
- [ ] I made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I filled out [the contributor form](https://tally.so/r/n9XrxK) (see text above)
- [x] I have added relevant notes to the CHANGELOG.md file (See https://keepachangelog.com/)

PR #3953, opened 2023-10-16
feat: add `title` and drop `description` from Python SDK

alvarobartt:
# Description

This PR adds the `title` field to the `TermsMetadataProperty`, `IntegerMetadataProperty`, and `FloatMetadataProperty` schemas, and drops the `description`. Also the docstrings have been updated accordingly.

This PR needs to be merged after #3952.

**Type of change**

**How Has This Been Tested**

- [X] Re-run `pytest -s tests/unit`

**Checklist**

### DIFF
--- a/src/argilla/client/feedback/schemas/metadata.py
+++ b/src/argilla/client/feedback/schemas/metadata.py
@@ -66,6 +66,10 @@ class MetadataPropertySchema(BaseModel, ABC):
         extra = Extra.forbid
         exclude = {"type"}
 
+    @validator("title", always=True)
+    def title_must_have_value(cls, v: Optional[str], values: Dict[str, Any]) -> str:
+        return values.get("name") if not v else v
+
     @property
     @abstractmethod
     def server_settings(self) -> Dict[str, Any]:
--- a/src/argilla/client/feedback/schemas/metadata.py
+++ b/src/argilla/client/feedback/schemas/metadata.py
@@ -43,7 +43,8 @@ class MetadataPropertySchema(BaseModel, ABC):
 
     Args:
         name: The name of the metadata property.
-        title: A title of the metadata property (what's shown in the UI). Defaults to `None`.
+        title: A title of the metadata property (what's shown in the UI). Defaults to `None`,
+            which means that if not provided then the `name` will be used as the `title`.
         visible_for_annotators: Whether the metadata property should be visible for
             users with the `annotator` role. Defaults to `True`.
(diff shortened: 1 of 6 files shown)


src/argilla/client/feedback/schemas/metadata.py
```diff
@@ -43,7 +43,7 @@ class MetadataPropertySchema(BaseModel, ABC):
 
     Args:
         name: The name of the metadata property.
-        description: A description of the metadata property. Defaults to `None`.
+        title: A title of the metadata property (what's shown in the UI). Defaults to `None`.
```
jfcalvo: `description` was optional but `title` is not (in the same fashion with `fields` and `questions`). So probably you need to change the documentation, models and tests.

alvarobartt: Hmm ok, then we should default it internally to `title = name` if `title is None`?

alvarobartt: That's what we are doing for the `title` in fields and questions

jfcalvo: If that's what we are doing for fields and questions then I guess we can do the same. Can you confirm @frascuchon ?

frascuchon: Yes, the same behavior that we have already for `fields`. Anyway, this should be reflected in the docstring

src/argilla/client/sdk/v1/datasets/models.py
```diff
@@ -112,7 +112,7 @@ class FeedbackQuestionModel(BaseModel):
 class FeedbackMetadataPropertyModel(BaseModel):
     id: UUID
     name: str
-    description: Optional[str] = None
+    title: Optional[str] = None
```
jfcalvo: If we are using this model to represent information from the API then I guess then `title` is not optional (in the same way that is not for `FeedbackFieldModel` and `FeedbackQuestionModel`

```suggestion
    title: str
```

src/argilla/client/sdk/v1/datasets/models.py
```diff
@@ -112,7 +112,7 @@ class FeedbackQuestionModel(BaseModel):
...
     visible_for_annotators: Optional[bool] = True
```
jfcalvo: Something similar for this one, it should always have a value:

```suggestion
    visible_for_annotators: bool
```

PR #3956, opened 2023-10-16
feat: remove type from metadata properties

jfcalvo:
# Description

This PR removes the `type` column from the `metadata_properties` and left only the `type` available inside `settings`.

**How Has This Been Tested**

- [ ] Fixing some unit tests.
- [ ] Running unit tests using PostgreSQL.

**Checklist**

### DIFF
--- a/src/argilla/server/models/database.py
+++ b/src/argilla/server/models/database.py
@@ -187,7 +187,6 @@ class MetadataProperty(DatabaseModel):
 
     name: Mapped[str] = mapped_column(String, index=True)
     title: Mapped[str] = mapped_column(Text)
-    type: Mapped[MetadataPropertyType] = mapped_column(Text)
     settings: Mapped[dict] = mapped_column(MutableDict.as_mutable(JSON), default={})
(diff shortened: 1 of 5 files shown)


gabrielmbmb (approved):

src/argilla/server/models/database.py
```diff
@@ -196,6 +195,10 @@ class MetadataProperty(DatabaseModel):
 
     __table_args__ = (UniqueConstraint("name", "dataset_id", name="metadata_property_name_dataset_id_uq"),)
 
+    @property
+    def type(self) -> MetadataPropertyType:
+        return self.settings["type"]
```
frascuchon: The raw value in settings should be parsed, right?
```suggestion
        return MetadataPropertyType(self.settings["type"])
```

frascuchon: The type conversion is not needed. But in that case, I would say to return an `str` instead of a `MetadataPropertyType`

frascuchon (approved):

PR #3957, opened 2023-10-16
Feature: validate records metadata on record updates

frascuchon:
# Description

This PR adds the record metadata validation when records are updated using `update_records`. 

This validation is not included in `record.update` since the dataset must be visible at a record level and many changes must be applied.  It will be applied in a separate PR

Also, the server metadata property validation for numeric properties has been fixed when passing unparseable numbers ("wrong-float" or "wrong-integer", for example)

Refs #3748 

**Type of change**

**How Has This Been Tested**

- [ ] Test A
- [ ] Test B

**Checklist**


src/argilla/client/feedback/dataset/base.py
```diff
@@ -349,36 +356,35 @@ def _validate_records(self, records: List[FeedbackRecord]) -> None:
...
-
-        # TODO: review up to what extent we want to apply the validation here
-        existing_metadata_properties = self.metadata_properties
-        if self._metadata_schema is None and existing_metadata_properties:
-            self._metadata_schema = generate_pydantic_schema_for_metadata(
-                existing_metadata_properties, allow_extra_metadata=self._allow_extra_metadata
-            )
+        # WE build once the schemas to avoid building it for each record
+        fields_schema = self._build_fields_schema()
+        metadata_schema = self._build_metadata_schema()
```
alvarobartt, line 361: This should just be built if there are any `metadata_properties`, otherwise there's no need to

frascuchon: We need to check the  `allow_extra_metadata` which is configured as part of the `MetaConfig` inside the metadata schema. If we generate the schema only when there are metadata properties, we need to add extra steps checking the allow_metadata_extra in the validation which makes the code very dirty

src/argilla/server/models/metadata_properties.py
```diff
@@ -66,10 +66,20 @@ def check_metadata(self, value: NT) -> None:
...
+            raise ValueError(f"'{value}' is not an integer.")
+        return super().check_metadata(value)
+
 
 class FloatMetadataPropertySettings(NumericMetadataPropertySettings[float]):
     type: Literal[MetadataPropertyType.float]
 
+    def check_metadata(self, value: float) -> None:
+        if not isinstance(value, float):
+            raise ValueError(f"'{value}' is not an float.")
```
gabrielmbmb:
```suggestion
            raise ValueError(f"'{value}' is not a float.")
```

gabrielmbmb (approved):

PR #3967, opened 2023-10-17
fix: add missing question update validations

jfcalvo:
# Description

* Question `title` and `description` are using the same validations for update and create.
* Question `settings` is now a non nullable field when updating.
* Metadata Property `title` and `visible_for_annotators` are now non nullable fields when updating.

**Type of change**

**How Has This Been Tested**

- [x] Running unit tests locally.
- [x] Running unit tests locally with PostgreSQL.

**Checklist**


src/argilla/server/schemas/v1/datasets.py
```diff
@@ -313,22 +314,36 @@ class Questions(BaseModel):
...
-    name: constr(
+QuestionName = Annotated[
+    constr(
         regex=QUESTION_CREATE_NAME_REGEX,
         min_length=QUESTION_CREATE_NAME_MIN_LENGTH,
         max_length=QUESTION_CREATE_NAME_MAX_LENGTH,
-    )
-    title: constr(
+    ),
+    PydanticField(..., description="The name of the question"),
```
gabrielmbmb, line 327: maybe we can move the `regex`, `min_length` and `max_length` to the `PydanticField` and use `str` instead of `constr`

gabrielmbmb: at some point we're going to have to do it anyway: https://docs.pydantic.dev/latest/migration/#constrained-types

gabrielmbmb (approved):

src/argilla/server/schemas/v1/datasets.py
jfcalvo: Yes, we should do it for all of the annotated fields once we move to v2.

tests/unit/server/api/v1/test_metadata_properties.py
```diff
@@ -286,6 +286,27 @@ async def test_update_metadata_property_disabling_visible_for_annotators(
     assert metadata_property.allowed_roles == [UserRole.admin]
 
 
+@pytest.mark.asyncio
```
frascuchon, line 289: I'm missing a test updating only the `visible_for_annotators` attribute.

jfcalvo: They were created in a previous PR when adding `visible_for_annotators` feature to metadata properties. You can see them here: https://github.com/argilla-io/argilla/blob/ec2af2ac7b892d65e1f949d2f4036d6555e64350/tests/unit/server/api/v1/test_metadata_properties.py#L242

frascuchon (approved):

PR #3983, opened 2023-10-18
fix(chore): add `argilla_model_card.md` to `pyproject.toml`

plaguss:
# Description

This PR addresses some pending fixes as a follow up of #3969 (as pointed out by @alvarobartt), which were preventing the `ModelCard` to be generated once the package is uploaded to PyPI as the file was kept locally as not as part of the `package-files`.

Besides that, this PR also fixes some minor style improvements to keep consistency with the rest of the codebase.

**Type of change**


davidberenstein1957 (commented):
LFTM 

@alvarobartt, would you check and to the honors?

PR #4005, opened 2023-10-20
refactor: enhance the metadata validation error messages

frascuchon:
# Description

This PR enhances how type checking errors are shown to the users for metadata validations. It will show the expected value but also the provided one, which can be very helpful to users fixing the problem

**Type of change**


dvsrepo:
Looks good! Could you open another issue to make sure we apply this (showing the provided value) for other components (names in dataset, fields, questions, etc.) I've encountered this with metadata property names for example in a loop and was difficult debug which field was causing this.

PR #4006, opened 2023-10-20
fix: `FeedbackDataset` fields with `required=False` failing on `push_to_huggingface`

alvarobartt:
# Description

This PR addresses an issue reported by @dvsrepo, that was preventing from uploading a `FeedbackDataset` to the Hugging Face Hub via `push_to_huggingface` when there were any optional fields not provided.

**Type of change**

**How Has This Been Tested**

- [x] Add unit tests for `FeedbackDataset` with optional i.e. `required=False` fields

**Checklist**

### DIFF
--- a/src/argilla/client/feedback/integrations/huggingface/dataset.py
+++ b/src/argilla/client/feedback/integrations/huggingface/dataset.py
@@ -114,7 +114,7 @@ class HuggingFaceDatasetMixin:
 
         for record in dataset.records:
             for field in dataset.fields:
-                hf_dataset[field.name].append(record.fields[field.name] if field.name in record.fields else None)
+                hf_dataset[field.name].append(record.fields.get(field.name, None))
             for question in dataset.questions:
                 if not record.responses:
                     hf_dataset[question.name].append([])
--- a/src/argilla/client/feedback/integrations/huggingface/dataset.py
+++ b/src/argilla/client/feedback/integrations/huggingface/dataset.py
@@ -114,7 +114,7 @@ class HuggingFaceDatasetMixin:
 
         for record in dataset.records:
             for field in dataset.fields:
-                hf_dataset[field.name].append(record.fields[field.name])
+                hf_dataset[field.name].append(record.fields[field.name] if field.name in record.fields else None)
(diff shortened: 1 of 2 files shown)


src/argilla/client/feedback/integrations/huggingface/dataset.py
```diff
@@ -114,7 +114,7 @@ def _huggingface_format(dataset: Union["FeedbackDataset", "RemoteFeedbackDataset
 
         for record in dataset.records:
             for field in dataset.fields:
-                hf_dataset[field.name].append(record.fields[field.name])
+                hf_dataset[field.name].append(record.fields[field.name] if field.name in record.fields else None)
```
frascuchon: Could we write like this?
```suggestion
                hf_dataset[field.name].append(record.fields.get(field.name, None))
```

tests/unit/client/feedback/integrations/huggingface/test_dataset.py
```diff
@@ -0,0 +1,47 @@
...
+from argilla.client.feedback.schemas.fields import TextField
+from argilla.client.feedback.schemas.questions import TextQuestion
+from argilla.client.feedback.schemas.records import FeedbackRecord
+
+
+class TestSuiteHuggingFaceDatasetMixin:
+    @pytest.mark.parametrize(
+        "record, hf_record",
+        [
+            (FeedbackRecord(fields={"required-field": "value"}), {"required-field": "value", "optional-field": None}),
```
frascuchon, line 29:
```suggestion
            (FeedbackRecord(fields={"required-field": "value"}), {"required-field": "value", "optional-field": None}),
            (FeedbackRecord(fields={"required-field": "value", "optional-field": None}), {"required-field": "value", "optional-field": None}),
```

alvarobartt: This is not needed because what we want to test here is that either the value is propagated to the `datasets.Dataset` or that `None` is set if not providing, so setting `None` is the same as setting any other value

frascuchon (approved):

PR #4019, opened 2023-10-23
feature: search records using vectors

frascuchon:
# Description

This PR updates the endpoint `POST /api/me/datasets/:dataset_id/records/search` to accept vector searches:
```bash
POST /api/me/datasets/:dataset_id/records/search
{
  "query": {
    "vector": {
      "name": "my-vector-name",
      "record_id": ...,
      "order": "most_similar"
    }
  }
}
````

Also, the minimal OpenSearch version has been upgraded to 2.5, since versions <=2.4 do not support filtering + similarity search properly.

Finally, the default Argilla search engine is configured to Elasticsearch (⚠️  this change may affect users using OpenSearch)

## TODO
- [x] Add missing tests
- [x] Filter using metadata + response status 
- [x] Apply most/least similarity order
- [x] Validations (record exists by id)
- [x] Adding search text as filter

PR https://github.com/argilla-io/argilla/pull/4043 must be merged before

Refs: https://github.com/argilla-io/argilla/issues/3028

**Type of change**

**How Has This Been Tested**

- [ ] Test A
- [ ] Test B

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,8 @@ These are the section headers that we use:
 - Added support to include vectors for `POST /api/v1/me/datasets/:dataset_id/records/search` endpoint response using `include` query param. ([#4063](https://github.com/argilla-io/argilla/pull/4063))
 - Added `add_vector_settings` method to local and remote FeedbackDataset. ([#4055](https://github.com/argilla-io/argilla/pull/4055))
 - Added `vector_settings_by_name` method to local and remote Feedback datasets. ([#4055](https://github.com/argilla-io/argilla/pull/4055))
-
+- Added `ARGILLA_SEARCH_ENGINE` environment variable to configure the search engine to use. ([#4019](https://github.com/argilla-io/argilla/pull/4019))
+ 
 ### Changed
 
 - Updated `POST /api/v1/datasets/:dataset_id/records` endpoint to allow to create records with `vectors` ([#4022](https://github.com/argilla-io/argilla/pull/4022))
@@ -37,6 +38,7 @@ These are the section headers that we use:
 - Update `FeedbackDataset.__init__` to allow passing a vector_settings list. ([#4055](https://github.com/argilla-io/argilla/pull/4055))
 - Update `FeedbackDataset.push_to_argilla` to also push vector settings. ([#4055](https://github.com/argilla-io/argilla/pull/4055))
 - Update `FeedbackDatasetRecord` to allow to create records with vectors. ([#4043](https://github.com/argilla-io/argilla/pull/4043))
+- Update `POST /api/v1/me/datasets/:dataset_id/records/search` endpoint to allow to search records with vectors. ([#4019](https://github.com/argilla-io/argilla/pull/4019))
 
 ## [1.18.0]()
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,7 @@ These are the section headers that we use:
 - Added `add_vector_settings` method to local and remote FeedbackDataset. ([#4055](https://github.com/argilla-io/argilla/pull/4055))
 - Added `vector_settings_by_name` method to local and remote Feedback datasets. ([#4055](https://github.com/argilla-io/argilla/pull/4055))
 - Added `ARGILLA_SEARCH_ENGINE` environment variable to configure the search engine to use. ([#4019](https://github.com/argilla-io/argilla/pull/4019))
- 
+
 ### Changed
 
 - Updated `POST /api/v1/datasets/:dataset_id/records` endpoint to allow to create records with `vectors` ([#4022](https://github.com/argilla-io/argilla/pull/4022))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ These are the section headers that we use:
 - Update `FeedbackDataset.push_to_argilla` to also push vector settings. ([#4055](https://github.com/argilla-io/argilla/pull/4055))
 - Update `FeedbackDatasetRecord` to allow to create records with vectors. ([#4043](https://github.com/argilla-io/argilla/pull/4043))
 - Update `POST /api/v1/me/datasets/:dataset_id/records/search` endpoint to allow to search records with vectors. ([#4019](https://github.com/argilla-io/argilla/pull/4019))
+- [breaking] Users working with OpenSearch engines must use version >=2.5 and set `ARGILLA_SEARCH_ENGINE=opensearch`. ([#4019](https://github.com/argilla-io/argilla/pull/4019))
 
 ## [1.18.0]()
 
--- a/src/argilla/server/apis/v1/handlers/datasets.py
+++ b/src/argilla/server/apis/v1/handlers/datasets.py
@@ -221,23 +221,10 @@ async def _get_search_responses(
     record = None
 
     if vector_query:
+        vector_settings = await _get_vector_settings_by_name_or_raise(db, dataset, vector_query.name)
         if vector_query.record_id is not None:
-            record = await datasets.get_record_by_id(db, vector_query.record_id)
-            if record is None:
-                raise HTTPException(
-                    status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
-                    detail=f"Record with id `{vector_query.record_id}` not found in dataset `{dataset.id}`.",
-                )
+            record = await _get_dataset_record_by_id_or_raise(db, dataset, vector_query.record_id)
             await record.awaitable_attrs.vectors
-
-        vector_settings = await datasets.get_vector_settings_by_name_and_dataset_id(
-            db, name=vector_query.name, dataset_id=dataset.id
-        )
-        if vector_settings is None:
-            raise HTTPException(
-                status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
-                detail=f"Vector `{vector_query.name}` not found in dataset `{dataset.id}`.",
-            )
 
     if (
         text_query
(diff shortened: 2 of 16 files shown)

### CODE REVIEW

File: src/argilla/server/schemas/v1/datasets.py [hunk 1]
```diff
@@ -615,12 +615,42 @@ def metadata_parsed(self) -> List[MetadataParsedQueryParam]:
         return [MetadataParsedQueryParam(q) for q in self.metadata]
 
 
-class TextQuery(BaseModel):
-    text: StringQuery
+class VectorQuery(BaseModel):
+    name: str
+    record_id: Optional[UUID] = None
+    value: Optional[List[float]] = None
+    max_results: conint(gt=0, le=500) = PydanticField(default=50)
```
jfcalvo: Correct me if I'm wrong but I believe that `VectorQuery` should be free of `max_results`, the query should be limited in the number of results by the `limit` parameter that we are receiving in the endpoint right?

frascuchon: Totally. We can use the `limit` parameter to provide the number of results to return and remove the `max_results` attribute

damianpumar:
Hi @frascuchon 
I don't see in the example you have shared the possibility to change the order of "Most" | "Least".

Thanks!

jfcalvo:
> Hi @frascuchon I don't see in the example you have shared the possibility to change the order of "Most" | "Least".
> 
> Thanks!

You're right @damianpumar we need some way to specify the ordering for the score.

frascuchon:
Yes @damianpumar, we will include an `order` attribute for controlling the similarity order. 

@jfcalvo is not really a score order, since fetching the N least similar records will compute the score in a different way as the most N similar.

src/argilla/server/schemas/v1/datasets.py
```diff
@@ -621,12 +621,44 @@ def metadata_parsed(self) -> List[MetadataParsedQueryParam]:
...
+
+    order: Union[Literal["most_similar"], Literal["least_similar"]] = "most_similar"
+
+    @root_validator
+    def check_required(cls, values: dict) -> dict:
+        """Check that either 'record_id' or 'value' is provided"""
+        record_id = values.get("record_id")
+        value = values.get("value")
+
+        if record_id is None and value is None:
```
jfcalvo: I was thinking that we must check that one of the two attributes are present but also raise an error when the two of them are present (equivalent to a logical xor condition).

jfcalvo:
```suggestion
        if bool(record_id) == bool(value):
```

src/argilla/server/search_engine/elasticsearch.py
```diff
@@ -98,7 +98,7 @@ async def _request_similarity_search(
             # TODO: Add metadata !!!!
             knn_query["filter"] = self._build_response_status_filter(user_response_status_filter)
 
-        return await self.client.search(index=index, knn=knn_query, _source=False, track_total_hits=True)
+        return await self.client.search(index=index, knn=knn_query, _source=False, track_total_hits=True, size=k)
```
jfcalvo, line 104: It is possible to rename `k` to `size`? If `k` has some clear meaning (like in k-clustering or something) in the context of this function then forget about this suggestion.

```suggestion
        return await self.client.search(index=index, knn=knn_query, _source=False, track_total_hits=True, size=size)
```

damianpumar:
I found two little bugs.
- The order "most_similar" "least_similar" does not change my response.
- When I set the limit to 200, always receive 100.

@gabrielmbmb @jfcalvo @frascuchon

frascuchon:
> I found two little bugs.
> 
> * The order "most_similar" "least_similar" does not change my response.
> * When I set the limit to 200, always receive 100.
> 
> @gabrielmbmb @jfcalvo @frascuchon

Thanks @damianpumar. The first one is normal since is not yet implemented. we're not processing this parameter yet, we will do it ASAP. For the second one, I think the dataset you're pushing has only 99 records (the loop creating data in the script I provided you defines a `range(1, 100)`) You can try incrementing that range.

src/argilla/server/search_engine/elasticsearch.py
frascuchon: Related capabilities for OpenSearch and Elasticsearch refer explicitly to the `k` parameter. I would keep it aligned since `size` may be confusing in this context.

src/argilla/server/apis/v1/handlers/datasets.py
```diff
@@ -210,24 +211,70 @@ async def _get_search_responses(
...
     user: Optional[User] = None,
     response_statuses: Optional[List[ResponseStatusFilter]] = None,
     sort_by_query_param: Optional[Dict[str, str]] = None,
 ) -> "SearchResponses":
+    vector_settings = None
+    record = None
+
+    if vector_query:
+        if vector_query.record_id is not None:
+            record = await datasets.get_record_by_id(db, vector_query.record_id)
```
jfcalvo: You need to get the record by `id` and `dataset_id` if not you could potentially search by any `record_id` (even those outside of the dataset).

An alternative to this is to add/use a policy.

A good idea to check this is to add a new test trying to find by a record belonging to a different dataset and see if it's failing. After that fix the problem and make the test passing.

src/argilla/server/apis/v1/handlers/datasets.py
```diff
@@ -210,24 +211,70 @@ async def _get_search_responses(
...
+            await record.awaitable_attrs.vectors
+
+        vector_settings = await datasets.get_vector_settings_by_name_and_dataset_id(
+            db, name=vector_query.name, dataset_id=dataset.id
+        )
+        if vector_settings is None:
+            raise HTTPException(
+                status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+                detail=f"Vector `{vector_query.name}` not found in dataset `{dataset.id}`.",
+            )
```
jfcalvo: I would move this to the beginning of the if condition (after `if vector_query`). This is something that has a big potential to fail so I think it's better to put it first.

src/argilla/server/schemas/v1/datasets.py
```diff
@@ -673,12 +673,44 @@ def metadata_parsed(self) -> List[MetadataParsedQueryParam]:
...
 
-class TextQuery(BaseModel):
-    text: StringQuery
+class VectorQuery(BaseModel):
+    name: str
+
+    record_id: Optional[UUID] = None
+    value: Optional[List[float]] = None
+
+    order: Union[Literal["most_similar"], Literal["least_similar"]] = "most_similar"
```
jfcalvo:
```suggestion
    name: str
    record_id: Optional[UUID] = None
    value: Optional[List[float]] = None
    order: Union[Literal["most_similar"], Literal["least_similar"]] = "most_similar"
```

src/argilla/server/search_engine/base.py
```diff
@@ -266,8 +266,11 @@ async def similarity_search(
         vector_settings: VectorSettings,
         value: Optional[List[float]] = None,
         record: Optional[Record] = None,
+        query: Optional[Union[TextQuery, str]] = None,
         user_response_status_filter: Optional[UserResponseStatusFilter] = None,
+        metadata_filters: Optional[List[MetadataFilter]] = None,
         max_results: int = 100,
+        order: Union[Literal["most_similar"], Literal["least_similar"]] = "most_similar",
```
jfcalvo: Maybe we should use an Enum to be used here and in the schemas?

src/argilla/server/search_engine/commons.py
```diff
@@ -270,42 +270,53 @@ async def similarity_search(
...
             for vector in record.vectors:
                 if vector.vector_settings_id == vector_settings.id:
                     vector_value = vector.value
+                    record_id = record.id
+                    break
 
         if not vector_value:
             raise ValueError("Cannot find a vector value to apply with provided info")
 
+        if order == "least_similar":
```
jfcalvo: We clearly need and Enum to avoid this type of checks.

tests/unit/server/api/v1/test_datasets.py
```diff
@@ -5878,6 +5880,181 @@ async def test_search_dataset_records_with_response_status_filter(
         )
         assert response.status_code == 200
 
+    async def test_search_dataset_records_with_record_vector(
+        self, async_client: "AsyncClient", mock_search_engine: SearchEngine, owner: User, owner_auth_header: dict
+    ):
+        workspace = await WorkspaceFactory.create()
+        dataset, _, records, *_ = await self.create_dataset_with_user_responses(owner, workspace)
```
jfcalvo, line 5893: We should start thinking about not using these kind of helpers anymore. In the future we should create isolated tests that are easy to change and to understand.

tests/unit/server/api/v1/test_datasets.py
```diff
@@ -5878,6 +5880,181 @@ async def test_search_dataset_records_with_response_status_filter(
...
+            headers=owner_auth_header,
+            json=query_json,
+            params={"offset": 0, "limit": 10},
+        )
+
+        assert response.status_code == 422
+        response_json = response.json()
+        assert response_json == {"detail": f"Vector `wrong_vector` not found in dataset `{dataset.id}`."}
+
+    async def test_search_dataset_records_with_wrong_vector_record_id(
```
jfcalvo:
```suggestion
    async def test_search_dataset_records_with_nonexistent_vector_record_id(
```

tests/unit/server/api/v1/test_datasets.py
```diff
@@ -5878,6 +5880,181 @@ async def test_search_dataset_records_with_response_status_filter(
...
+            f"/api/v1/me/datasets/{dataset.id}/records/search",
+            headers=owner_auth_header,
+            json=query_json,
+            params={"offset": 0, "limit": 10},
+        )
+
+        assert response.status_code == 422
+        response_json = response.json()
+        assert response_json == {"detail": f"Record with id `{wrong_record_id}` not found in dataset `{dataset.id}`."}
+
```
jfcalvo, line 6063: We should create here a `test_search_dataset_records_with_record_id_from_different_dataset` or something similar.

CHANGELOG.md
```diff
@@ -37,6 +38,7 @@ These are the section headers that we use:
 - Update `FeedbackDataset.__init__` to allow passing a vector_settings list. ([#4055](https://github.com/argilla-io/argilla/pull/4055))
 - Update `FeedbackDataset.push_to_argilla` to also push vector settings. ([#4055](https://github.com/argilla-io/argilla/pull/4055))
 - Update `FeedbackDatasetRecord` to allow to create records with vectors. ([#4043](https://github.com/argilla-io/argilla/pull/4043))
+- Update `POST /api/v1/me/datasets/:dataset_id/records/search` endpoint to allow to search records with vectors. ([#4019](https://github.com/argilla-io/argilla/pull/4019))
```
jfcalvo, line 43: We should add the breaking change here somehow about using ElasticSearch by default.

jfcalvo (changes requested):

src/argilla/server/apis/v1/handlers/datasets.py
frascuchon: Thanks, I will do it

src/argilla/server/apis/v1/handlers/datasets.py
frascuchon: Yes, I will take a look this code an try to refactor/better organize it

src/argilla/server/search_engine/commons.py
```diff
@@ -270,42 +270,53 @@ async def similarity_search(
...
+            for i in range(0, len(vector_value)):
+                vector_value[i] *= -1
```
gabrielmbmb: maybe we can put this in a function called `inverse_vector`

tests/unit/server/api/v1/test_datasets.py
frascuchon: Not in this PR :-)

gabrielmbmb (approved):

jfcalvo (approved):

src/argilla/server/search_engine/commons.py
frascuchon: Done

PR #3799, opened 2023-09-20
feat: add vectors endpoints

gabrielmbmb:
# Description

This PR adds and updates the existing endpoints that will allow to manage the `Vector`s for a `Feedback Dataset`.

**Type of change**

- [x] New feature (non-breaking change which adds functionality)
- [ ] Refactor (change restructuring the codebase without changing functionality)
- [ ] Improvement (change adding some improvement to an existing functionality)

**How Has This Been Tested**

(Please describe the tests that you ran to verify your changes. And ideally, reference `tests`)

- [ ] Test A
- [ ] Test B

**Checklist**

- [ ] I added relevant documentation
- [ ] I followed the style guidelines of this project
- [ ] I did a self-review of my code
- [ ] I made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I filled out [the contributor form](https://tally.so/r/n9XrxK) (see text above)
- [ ] I have added relevant notes to the `CHANGELOG.md` file (See https://keepachangelog.com/)

PR #4022, opened 2023-10-23
feat: Create records with vectors

frascuchon:
# Description

In this PR the `POST /api/v1/datasets/:dataset_id/records` endpoint is modified to support vector creation. The `vectors` attribute is an optional dictionary where the key is the vector (settings) name and the value is the vector value.

The changes included here are based on PR #3799, but only apply the record creation modifications. Future PRs will include listing records with vector info and updating records with vector info, but separately, since a lot of non-trivial changes must by applied.

Refs #3799 

**Type of change**

**How Has This Been Tested**

Creating record with vectors locally 

**Checklist**

### DIFF
--- a/src/argilla/server/alembic/versions/bda6fe24314e_create_vectors_table.py
+++ b/src/argilla/server/alembic/versions/bda6fe24314e_create_vectors_table.py
@@ -14,16 +14,16 @@
 
 """create vectors table
 
-Revision ID: 66e63faf4a33
+Revision ID: bda6fe24314e
 Revises: 7850ab5b42d9
-Create Date: 2023-09-12 16:22:04.279445
+Create Date: 2023-09-21 15:41:59.642012
 
 """
 import sqlalchemy as sa
 from alembic import op
 
 # revision identifiers, used by Alembic.
-revision = "66e63faf4a33"
+revision = "bda6fe24314e"
 down_revision = "7850ab5b42d9"
 branch_labels = None
 depends_on = None
@@ -34,14 +34,17 @@ def upgrade() -> None:
     op.create_table(
         "vectors",
         sa.Column("value", sa.JSON(), nullable=False),
-        sa.Column("vector_settings_id", sa.Uuid(), nullable=False),
+        sa.Column("dataset_id", sa.Uuid(), nullable=False),
         sa.Column("record_id", sa.Uuid(), nullable=False),
+        sa.Column("vector_settings_id", sa.Uuid(), nullable=False),
         sa.Column("id", sa.Uuid(), nullable=False),
         sa.Column("inserted_at", sa.DateTime(), nullable=False),
         sa.Column("updated_at", sa.DateTime(), nullable=False),
+        sa.ForeignKeyConstraint(["dataset_id"], ["datasets.id"], ondelete="CASCADE"),
         sa.ForeignKeyConstraint(["record_id"], ["records.id"], ondelete="CASCADE"),
         sa.ForeignKeyConstraint(["vector_settings_id"], ["vectors_settings.id"], ondelete="CASCADE"),
         sa.PrimaryKeyConstraint("id"),
+        sa.UniqueConstraint("record_id", "vector_settings_id", name="vector_record_id_vector_settings_id_uq"),
     )
     op.create_index(op.f("ix_vectors_record_id"), "vectors", ["record_id"], unique=False)
     op.create_index(op.f("ix_vectors_vector_settings_id"), "vectors", ["vector_settings_id"], unique=False)
--- a/src/argilla/server/alembic/versions/bda6fe24314e_create_vectors_table.py
+++ b/src/argilla/server/alembic/versions/bda6fe24314e_create_vectors_table.py
@@ -34,13 +34,11 @@ def upgrade() -> None:
     op.create_table(
         "vectors",
         sa.Column("value", sa.JSON(), nullable=False),
-        sa.Column("dataset_id", sa.Uuid(), nullable=False),
         sa.Column("record_id", sa.Uuid(), nullable=False),
         sa.Column("vector_settings_id", sa.Uuid(), nullable=False),
         sa.Column("id", sa.Uuid(), nullable=False),
         sa.Column("inserted_at", sa.DateTime(), nullable=False),
         sa.Column("updated_at", sa.DateTime(), nullable=False),
-        sa.ForeignKeyConstraint(["dataset_id"], ["datasets.id"], ondelete="CASCADE"),
         sa.ForeignKeyConstraint(["record_id"], ["records.id"], ondelete="CASCADE"),
         sa.ForeignKeyConstraint(["vector_settings_id"], ["vectors_settings.id"], ondelete="CASCADE"),
         sa.PrimaryKeyConstraint("id"),
--- a/src/argilla/server/contexts/datasets.py
+++ b/src/argilla/server/contexts/datasets.py
@@ -544,7 +544,7 @@ async def validate_vector(
     vector_settings = vectors_settings.get(vector_name, None)
     if not vector_settings:
         vector_settings = await get_vector_settings_by_name_and_dataset_id(db, vector_name, dataset_id)
-        if not vector_settings or vector_settings.dataset_id != dataset_id:
+        if not vector_settings:
             raise ValueError(f"vector with name={str(vector_name)} does not exist for dataset_id={str(dataset_id)}")
 
         vector_settings = VectorSettingsSchema.from_orm(vector_settings)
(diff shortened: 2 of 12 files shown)


src/argilla/server/alembic/versions/bda6fe24314e_create_vectors_table.py
```diff
@@ -34,14 +34,17 @@ def upgrade() -> None:
     op.create_table(
         "vectors",
         sa.Column("value", sa.JSON(), nullable=False),
-        sa.Column("vector_settings_id", sa.Uuid(), nullable=False),
+        sa.Column("dataset_id", sa.Uuid(), nullable=False),
```
jfcalvo: I already commented this to Gabriel. Not sure why we are duplicating the association to the dataset when vector is already associated to the dataset through the record, maybe I'm missing something.

Ted Codd would not have approved this change. 😝

src/argilla/server/alembic/versions/bda6fe24314e_create_vectors_table.py
```diff
@@ -34,14 +34,17 @@ def upgrade() -> None:
...
         sa.Column("record_id", sa.Uuid(), nullable=False),
+        sa.Column("vector_settings_id", sa.Uuid(), nullable=False),
```
jfcalvo, line 38: I saw this also before on PR made by Gabriel. I prefer to think in vectors like not associated with the settings in any way. But I understand that this adds flexibility to rename vectors and other stuff.

src/argilla/server/contexts/datasets.py
```diff
@@ -509,17 +518,51 @@ async def _validate_suggestion(
...
+
+        vector_settings = VectorSettingsSchema.from_orm(vector_settings)
+        vectors_settings[vector_name] = vector_settings
+
+    vector_settings.check_vector(vector_value)
+
+    return vectors_settings
+
+
 async def create_records(
```
jfcalvo, line 558: This function is really complex. We should think into create a `create_record` for a single record if possible and additional ones that process responses, suggestions, metadata and vectors (like `create_record_responses`or similar).

src/argilla/server/contexts/datasets.py
```diff
@@ -531,12 +574,15 @@ async def create_records(
...
-            for response in record_create.responses:
-                # TODO(gabrielmbmb): the result of this query should be cached
-                if not await accounts.get_user_by_id(db, response.user_id):
-                    raise ValueError(f"Provided user_id: {response.user_id!r} is not a valid user id")
-
-                validate_response_values(dataset, values=response.values, status=response.status)
+            users_ids_cache: Set[UUID] = set()
+            for i, response in enumerate(record_create.responses):
+                try:
+                    users_ids_cache = await validate_user(db, response.user_id, users_ids_cache)
```
jfcalvo: Why do we need to validate that the response user exists?

I can understand a check to know if the user belongs to the dataset workspace (we're not doing that right?) but only for existence? (The foreign key is already doing that).

src/argilla/server/alembic/versions/bda6fe24314e_create_vectors_table.py
frascuchon: We have 2 kinds of entities related to vectors, the vector configuration/settings (at the dataset level) and the vector value itself (at the record level). Let's see the `vector_settings` as the vector schema and `vectors` as the value for those for each record

src/argilla/server/contexts/datasets.py
frascuchon: Admins can provide multiple responses for a record (and each one from a different user). This validation tries to check that responses were properly configured without raising an undescribable `IntegrityError`

src/argilla/server/contexts/datasets.py
frascuchon: Now,  the logic inside the `create_records` method is separated per sub-entity (`create_records_responses`, `*_suggestions`, and `*_vectors`).

Still needed to review validators and additional cache logic (datasets are loaded with all that info, so we're probably doing unnecessary queries). Anyway I propose to review it  in a separate PR

src/argilla/server/contexts/datasets.py
```diff
@@ -509,83 +519,190 @@ async def _validate_suggestion(
...
+    vector_value: List[float],
+    vectors_settings: Optional[Dict[str, VectorSettingsSchema]] = None,
+) -> Dict[str, VectorSettingsSchema]:
+    if vectors_settings is None:
+        vectors_settings = {}
+
+    vector_settings = vectors_settings.get(vector_name, None)
+    if not vector_settings:
+        vector_settings = await get_vector_settings_by_name_and_dataset_id(db, vector_name, dataset_id)
+        if not vector_settings or vector_settings.dataset_id != dataset_id:
```
jfcalvo: I'm missing something or the second condition will never be `True` ? I mean you are finding the vector_settings by `dataset_id` and `name` so if found it will always belongs to that dataset.

```suggestion
        if not vector_settings:
```

src/argilla/server/contexts/datasets.py
```diff
@@ -509,83 +519,190 @@ async def _validate_suggestion(
...
+    if not metadata:
+        return
+
+    try:
+        await _validate_metadata(db, dataset=dataset, metadata=metadata, metadata_properties=cache)
+    except ValueError as e:
+        raise ValueError(f"Provided metadata for record at position {record_idx} is not valid: {e}") from e
+
+
+async def create_record_vectors(
```
jfcalvo: I believe we should use `build` (or similar) for these functions instead of `create`. `create` should be used when we insert a new row/rows into the database.

These functions should be "private" also.

```suggestion
async def _build_record_vectors(
```

src/argilla/server/contexts/datasets.py
```diff
@@ -509,83 +519,190 @@ async def _validate_suggestion(
...
+            vectors.append(Vector(value=vector_value, vector_settings_id=cache[vector_name].id))
+        except ValueError as e:
+            raise ValueError(
+                f"Provided vector with name={vector_name} of record at position {record_idx} is not valid: {e}"
+            ) from e
+
+    return vectors
+
+
+async def create_record_responses(
```
jfcalvo:
```suggestion
async def _build_record_responses(
```

src/argilla/server/contexts/datasets.py
```diff
@@ -509,83 +519,190 @@ async def _validate_suggestion(
...
+            )
+        except ValueError as e:
+            raise ValueError(
+                f"Provided response at position {idx} of record at position {record_idx} is not valid: {e}"
+            ) from e
+
+    return responses
+
+
+async def create_record_suggestions(
```
jfcalvo:
```suggestion
async def _build_record_suggestions(
```

src/argilla/server/contexts/datasets.py
```diff
@@ -509,83 +519,190 @@ async def _validate_suggestion(
...
+    db: "AsyncSession", record_idx: int, record_create: "RecordCreate", cache: Dict[UUID, Question] = {}
```
jfcalvo: Maybe `record_position` is better? I was thinking that it was a typo.

```suggestion
    db: "AsyncSession", record_position: int, record_create: "RecordCreate", cache: Dict[UUID, Question] = {}
```

src/argilla/server/contexts/datasets.py
```diff
@@ -509,83 +519,190 @@ async def _validate_suggestion(
...
+                Suggestion(
+                    type=suggestion.type,
+                    score=suggestion.score,
+                    value=suggestion.value,
+                    agent=suggestion.agent,
+                    question_id=suggestion.question_id,
+                )
+            )
+        except ValueError as e:
+            raise ValueError(f"Provided suggestion for record at position {record_idx} is not valid: {e}") from e
```
jfcalvo:
```suggestion
            raise ValueError(f"Provided suggestion for record at position {record_position} is not valid: {e}") from e
```

src/argilla/server/contexts/datasets.py
```diff
@@ -803,10 +920,14 @@ async def delete_response(db: "AsyncSession", search_engine: SearchEngine, respo
     return response
 
 
-def validate_response_values(dataset: Dataset, values: Dict[str, ResponseValue], status: ResponseStatus):
+def validate_response_values(
```
jfcalvo: If it's only used internally:

```suggestion
def _validate_response_values(
```

src/argilla/server/contexts/datasets.py
frascuchon: Yes, not needed the second condition. Changed

src/argilla/server/contexts/datasets.py
frascuchon: Done

src/argilla/server/contexts/datasets.py
frascuchon: done

jfcalvo (approved):

PR #4104, opened 2023-10-31
feat: enhance `PATCH` record endpoints for updating vectors

gabrielmbmb:
# Description

This PR updates the `PATCH /api/v1/datasets/{dataset_id}/records` and `PATCH /api/v1/records/{record_id}` endpoints to selectively update or create the records vectors using the provided vectors in the payload of the request. If there is a vector that has not been provided in the payload, it won't be removed.

**Type of change**

**How Has This Been Tested**

Unit tests have been updated

**Checklist**


src/argilla/server/contexts/datasets.py
```diff
@@ -723,13 +737,17 @@ async def _build_record_suggestions(
     return suggestions
 
 
+R = TypeVar("R")
+
+
 async def _build_record_vectors(
     db: "AsyncSession",
     dataset: Dataset,
     record_schema: Union["RecordCreate", "RecordUpdateWithId"],
```
frascuchon: We're using `RecordUpdate` schemas in from the `update_record` method
```suggestion
    record_schema: Union["RecordCreate", "RecordUpdate", "RecordUpdateWithId"],
```

tests/unit/server/api/v1/test_datasets.py
```diff
@@ -4820,11 +4819,10 @@ async def test_update_dataset_records(
         assert len(records[1].suggestions) == 1
         assert records[1].suggestions[0].value == "suggestion updated 1 1"
         for vector in records[1].vectors:
-            assert records_1_updated_vectors[vector.vector_settings.name] == vector.value
+            if vector.vector_settings.name in records_1_updated_vectors:
+                assert records_1_updated_vectors[vector.vector_settings.name] == vector.value
```
frascuchon: We should check also non-updated vectors, right?

tests/unit/server/api/v1/test_datasets.py
```diff
@@ -4802,11 +4802,10 @@ async def test_update_dataset_records(
         assert records[0].suggestions[1].value == "suggestion updated 0 2"
         assert records[0].suggestions[2].value == "suggestion updated 0 3"
         for vector in records[0].vectors:
-            assert records_0_updated_vectors[vector.vector_settings.name] == vector.value
+            if vector.vector_settings.name in records_0_updated_vectors:
```
frascuchon: It's a bit hard to understand completely this test. Maybe it makes sense to separate into different simpler tests, checking each of the updatable entities (update_suggestions, update_vectors, update_metadata,....) 

I feel as watching a David Lynch film, concretely, Eraserhead 😅

frascuchon (approved):

tests/unit/server/api/v1/test_datasets.py
gabrielmbmb: agree, I'll split the test

src/argilla/server/contexts/datasets.py
```diff
@@ -723,13 +737,17 @@ async def _build_record_suggestions(
     return suggestions
 
 
+R = TypeVar("R")
```
jfcalvo: Maybe it's an idiomatic Python thingy but I don't understand this. Can we try to improve the name from `R` to something more descriptive?

src/argilla/server/contexts/datasets.py
jfcalvo, line 1: I believe that having a `_create_record` and `_update_record` functions where you have the functionality for only one record will improve readability a lot. Not sure if it's a simple thing to do.

src/argilla/server/contexts/datasets.py
gabrielmbmb: yes, usually `R` is used for `TypeVar`s that are used in the return type hint of a function, but it doesn't hurt anyone to write something more meaningful like `VectorClass`. I'll do it!

src/argilla/server/contexts/datasets.py
gabrielmbmb: done!

src/argilla/server/contexts/datasets.py
```diff
@@ -591,55 +589,61 @@ async def _validate_vector(
...
+    record = Record(
+        fields=record_create.fields,
+        metadata_=record_create.metadata,
+        external_id=record_create.external_id,
+        dataset_id=dataset.id,
+        responses=record_responses,
+        suggestions=record_suggestions,
+        vectors=record_vectors,
+    )
+    return record, record_vectors
```
jfcalvo: It is really necessary to return `record_vectors`? You are already returning the record with them inside.

src/argilla/server/contexts/datasets.py
```diff
@@ -746,35 +737,72 @@ async def _build_record_vectors(
...
+) -> Tuple[Dict[str, Any], List[Suggestion], List[VectorSchema], bool, Dict[str, Any]]:
+    if caches is None:
+        caches = {
+            "metadata_properties": {},
+            "questions": {},
+            "vector_settings": {},
+        }
+
+    params = record_update.dict(exclude_unset=True)
+    needs_search_engine_update = False
```
jfcalvo, line 769: Do we really need this? It's a little bit error-prone (like forgetting set it to `True` and then not indexing). I would always reindex after an update instead.

src/argilla/server/contexts/datasets.py
gabrielmbmb: yes, totally unnecessary

src/argilla/server/contexts/datasets.py
gabrielmbmb: I was doing it like this to avoid querying all the records with their vectors from the database and then call the `index_records` with all of them as this can be a bit heavy operation, right?

jfcalvo: I see. Let's continue with that then, we can try to refactor this in a near future. Thanks for the explanation.

jfcalvo (approved):

PR #4109, opened 2023-10-31
chore: Upgrade minimal OpenSearch version to 2.6.0

frascuchon:
# Description

Doing some tests with OpenSearch 2.5 I realized some cases where similarity search fails when adding as part of the `knn` filter an existing field clause (for user responses). This is documented and fixed in the Lucene 9.5.0 version (See here: https://github.com/apache/lucene/pull/12072). So the minimal required OpenSearch version must use the Lucene 9.5.0, which is 2.6.0.

Using versions under  2.6.0 an error `"failed to create query: Rewrite first"` using similarity search can be potentially raised.

**Type of change**

**How Has This Been Tested**

Local tests 

**Checklist**


frascuchon:
This still happens for the latest OpenSearch versions. This PR will be dismissed

PR #4110, opened 2023-11-01
feat: added `__get_property_by_name`-method and return `None` for `FeedbackDataset.*_by_name()`

### LINKED ISSUE #4101: [FEATURE] Align `FeedbackDataset.*_by_name` methods for better integration with conditional creation flows
The current behavior for the several `FeedbackDataset.*_by_name` method is to raise a `KeyError` when the entity is not found by name. Instead of that, we could just return a `None` value. This scenario could work when admin users want to control when a new entity must be created for an existing dataset:

#### What we need to do right now
```python
try:
   dataset.question_by_name("label_question")
except KeyError:
   dataset.add_question(rg.LabelQuestion(name="label_question",...)
```

#### What I propose
```python
question = dataset.question_by_name("label_question")
if question is None:
   dataset.add_question(rg.LabelQuestion(name="label_question",...)
```
We're using questions as an example, but this can work with other entities.

N/A

**Additional context**
N/A

davidberenstein1957:
# Description

I agree with Paco that the implementation was a bit blocking. Instead we show a warning if not present and not block running code.

Closes #4101 

**Type of change**

**How Has This Been Tested**

N.A.

**Checklist**

### DIFF
--- a/src/argilla/client/feedback/dataset/base.py
+++ b/src/argilla/client/feedback/dataset/base.py
@@ -11,7 +11,7 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
-
+import warnings
 from abc import ABC, ABCMeta, abstractmethod
 from typing import TYPE_CHECKING, Any, Dict, Generic, Iterable, List, Literal, Optional, Type, TypeVar, Union
 
@@ -119,6 +119,18 @@ class FeedbackDatasetBase(ABC, Generic[R], metaclass=ABCMeta):
         """Returns whether if adding extra metadata to the records of the dataset is allowed"""
         return self._allow_extra_metadata
 
+    def __get_property_by_name(self, item_name: str, iterable_items, item_type: str):
+        for item in iterable_items:
+            if item.name == item_name:
+                return item
+            if hasattr(item, "title"):
+                if item.title == item_name:
+                    return item
+        warnings.warn(
+            f"{item_type} with name='{item_name}' not found, available {item_type} names are:"
+            f" {', '.join(item.name for item in iterable_items)}"
+        )
+
     @property
     def fields(self) -> Union[List[AllowedFieldTypes], List["AllowedRemoteFieldTypes"]]:
         """Returns the fields that define the schema of the records in the dataset."""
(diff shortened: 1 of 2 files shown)


src/argilla/client/feedback/dataset/base.py
```diff
@@ -119,49 +119,43 @@ def allow_extra_metadata(self) -> bool:
         """Returns whether if adding extra metadata to the records of the dataset is allowed"""
         return self._allow_extra_metadata
 
+    def __get_property_by_name(self, item_name: str, iterable_items, item_type: str):
+        for item in iterable_items:
+            if item.name == item_name:
+                return item
+            if hasattr(item, "title"):
+                if item.title == item_name:
```
davidberenstein1957: @frascuchon, I added this in the assumption that names can't overlap with titles. Is that correct?

davidberenstein1957:
I still need to add tests

frascuchon:
I see this change when I update the similarity search feature branch from develop.  We must be conscience that changes on this PR are breaking changes and should be informed properly (in CHANGELOG and documentation). 

The reason behind the breaking changes is that we're changing the current behavior of the existing method and some code using previous versions could fail (if users are currently capturing the `KeyError`). 

In my opinion, in order to accept this change we should 
- 1. Include a [breaking] tag in the CHANGELOG
- 2. Update documentation sections referring to how to get fields/questions/...by name (and show how users should be from these changes.
- 3. Prepare a section for the Release notes about his breaking change (The `field_by_name` method has changed, and now should handle by using `None` bla blab lab...)

Please @davidberenstein1957, take care of this to not create unexpected confusion with upgrading the version

cc @nataliaElv

davidberenstein1957:
Hi Paco,Thanks for the feedback. I didn't look at it this way. I will make sure to change this. Also, it might be an idea to slowly build the release highlight in the changelog during PRs and then just copy/paste when applicable?On 7 Nov 2023, at 00:37, Francisco Aranda ***@***.***> wrote:﻿
I see this change when I update the similarity search feature branch from develop.  We must be conscience that changes on this PR are breaking changes and should be informed properly (in CHANGELOG and documentation).
The reason behind the breaking changes is that we're changing the current behavior of the existing method and some code using previous versions could fail (if users are currently capturing the KeyError).
In my opinion, in order to accept this change we should



Include a [breaking] tag in the CHANGELOG




Update documentation sections referring to how to get fields/questions/...by name (and show how users should be from these changes.




Prepare a section for the Release notes about his breaking change (The field_by_name method has changed, and now should handle by using None bla blab lab...)



Please @davidberenstein1957, take care of this to not create unexpected confusion with upgrading the version
cc @nataliaElv

nataliaElv:
Has this been properly documented? Docs should be reviewed so that there is no code that becomes irrelevant.

PR #4161, opened 2023-11-07
feat: add a maximum limit of 5 vectors settings that can be created per dataset

jfcalvo:
# Description

This PR add changes to the endpoint `POST /api/v1/datasets/:dataset_id/vectors-settings` only allowing a maximum of `5` vector settings to be created by dataset. 

**Type of change**

**How Has This Been Tested**

- [x] Adding unit tests.

**Checklist**

### DIFF
--- a/src/argilla/server/apis/v1/handlers/datasets.py
+++ b/src/argilla/server/apis/v1/handlers/datasets.py
@@ -80,6 +80,8 @@ LIST_DATASET_RECORDS_LIMIT_LE = 1000
 LIST_DATASET_RECORDS_DEFAULT_SORT_BY = {RecordSortField.inserted_at.value: "asc"}
 DELETE_DATASET_RECORDS_LIMIT = 100
 
+CREATE_DATASET_VECTOR_SETTINGS_MAX_COUNT = 5
+
 router = APIRouter(tags=["datasets"])
 
 parse_record_include_param = parse_query_param(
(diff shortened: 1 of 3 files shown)


src/argilla/server/apis/v1/handlers/datasets.py
```diff
@@ -656,6 +658,13 @@ async def create_dataset_vector_settings(
 
     await authorize(current_user, DatasetPolicyV1.create_vector_settings(dataset))
 
+    count_vectors_settings_by_dataset_id = await datasets.count_vectors_settings_by_dataset_id(db, dataset_id)
+    if count_vectors_settings_by_dataset_id >= CREATE_DATASET_VECTOR_SETTINGS_MAX_COUNT:
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
```
gabrielmbmb: maybe a 409 is better here?

gabrielmbmb (approved):

src/argilla/server/apis/v1/handlers/datasets.py
frascuchon: Not sure if 409. In any case, the 403 status is normally related to resource access permissions, and maybe in this case the problem is more related to an internal status of the dataset. Does make sense to return a 422? It's what we're doing when adding a field on published datasets, and this case is quite similar

jfcalvo: To be honest it's open to interpretation, let's see the possibilities:
* `429` too many requests: some people use this status code when they want to communicate that the resource is limited and you already created too many of them. For me this status code it's more linked to rate-limiting and I don't think it's a good idea to use it.
* `409` conflict: I think it's a little bit misleading to say to the API user that there is a conflict with the resource, but it could be a good option too. According to w3c "This code is used in situations where the user might be able to resolve the conflict and resubmit the request. " so the user can delete a vector setting and try to execute the request again.
* `403` forbidden: I believe is the one that best communicate the intentions of failing request (in fact I believe these checks could be moved to the policies) but it's true that users could think that for some reason they don't have enough permissions to execute the request (and it's not the case).

jfcalvo: Let's have a super quick discussion today and take a decision if we need to do something similar to this in the future.

jfcalvo: We will use `422` by now and if it's necessary we can open a new discussion about how we use status codes.

PR #4196, opened 2023-11-10
feat: add helper function to keep track of the tutorial usage

### LINKED ISSUE #4189: [FEATURE] create a telemetry integration for keeping track of tutorials
it might be nice to add an [optional telemetry module](https://haystack.deepset.ai/tutorials/14_query_classifier) to see the number of usages for tutorials.

plaguss:
# Description

This PR includes a function to keep track of the tutorials running, similar to the format used in haystack:

```python
from argilla.utils.telemetry import tutorial_running

tutorial_running("tutorial_filename")
```

Closes #4189

**Type of change**

**How Has This Been Tested**

- [ ] Test A
- [ ] Test B

**Checklist**

### DIFF
--- a/src/argilla/utils/telemetry.py
+++ b/src/argilla/utils/telemetry.py
@@ -103,5 +103,14 @@ async def track_login(request: "Request", username: str):
     )
 
 
+def tutorial_running(tutorial_id: int) -> None:
+    """Can be called when a tutorial is executed so that the tutorial_id is used to identify the tutorial and send an event.
+
+    Args:
+        tutorial_id: ID number of the tutorial.
+    """
+    _CLIENT.track_data(action="TutorialRunning", data={"tutorial_id": tutorial_id})
+
+
 def get_telemetry_client() -> TelemetryClient:
     return _CLIENT
--- a/src/argilla/utils/telemetry.py
+++ b/src/argilla/utils/telemetry.py
@@ -103,11 +103,11 @@ async def track_login(request: "Request", username: str):
     )
 
 
-def tutorial_running(tutorial_id: int) -> None:
+def tutorial_running(tutorial_id: str) -> None:
     """Can be called when a tutorial is executed so that the tutorial_id is used to identify the tutorial and send an event.
 
     Args:
-        tutorial_id: ID number of the tutorial.
+        tutorial_id: ID number of the tutorial, use the filename of the file without the extension.
     """
     _CLIENT.track_data(action="TutorialRunning", data={"tutorial_id": tutorial_id})
 
--- a/src/argilla/utils/telemetry.py
+++ b/src/argilla/utils/telemetry.py
@@ -103,13 +103,43 @@ async def track_login(request: "Request", username: str):
     )
 
 
-def tutorial_running(tutorial_id: str) -> None:
-    """Can be called when a tutorial is executed so that the tutorial_id is used to identify the tutorial and send an event.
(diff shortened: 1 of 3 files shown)


plaguss:
@frascuchon, @davidberenstein1957, what do you think?
The idea is like the one used in [haystack's](https://haystack.deepset.ai/tutorials/14_query_classifier) as suggested by David. Would we need more information?

I see we already have [telemetry docs](https://docs.argilla.io/en/latest/reference/telemetry.html) so we could just reference that in the tutorials.

davidberenstein1957:
I think we might also just use `Path(__file__).name`?

plaguss:
> I think we might also just use `Path(__file__).name`?

To keep track of the tutorial instead of the *int* id?

davidberenstein1957:
@plaguss, could you align with one of the interns about this usage in the tutorial and potentially we want to do some Argilla version check?

davidberenstein1957:
@plaguss, perhaps using `__file__` out of the box woul be better to not allow user to pass any argument?

plaguss:
> @plaguss, perhaps using `__file__` out of the box woul be better to not allow user to pass any argument?

I saw that `__file__` wasn't working on the notebook, and it seems that the variable is not available:
- https://stackoverflow.com/questions/39125532/file-does-not-exist-in-jupyter-notebook,
- https://github.com/ipython/ipython/issues/10123
Others tried to work with the `cwd`, but that would not work either... Do you have a better solution to avoid this?

plaguss:
@davidberenstein1957 could you check the last addition? I tested the function in the different environments and seems to work... but needs an extra dependency: [`ipynbname`](https://pypi.org/project/ipynbname/). Let's me know what you think (if we want this option we should add the dependency).

src/argilla/utils/telemetry.py
```diff
@@ -103,13 +103,43 @@ async def track_login(request: "Request", username: str):
...
+        import urllib.parse
+
+        import ipynbname
+
+        return Path(urllib.parse.unquote_plus(ipynbname.name())).stem
+    except Exception as e:
+        import warnings
+
+        warnings.warn("Couldn't determine the filename.")
+        return
```
davidberenstein1957: I think we can remove this `Return`

davidberenstein1957 (commented):
Hi, thanks @plaguss . Works like a charm :)

src/argilla/utils/telemetry.py
plaguss: Done!

src/argilla/utils/telemetry.py
```diff
@@ -103,5 +103,39 @@ async def track_login(request: "Request", username: str):
...
+            globals_ = globals()
+            return Path(globals_["__vsc_ipynb_file__"]).stem
+    except KeyError as e:
+        # This should work for notebooks running locally or using google colab
+        import urllib.parse
+
+        import ipynbname
+
+        return Path(urllib.parse.unquote_plus(ipynbname.name())).stem
+
```
davidberenstein1957, line 132: I think we should have a final except to not break the user when it does not run.

plaguss: With the last change it won't break, instead it wont track the tutorial (without warning at all). Do we want this behavior?

davidberenstein1957: Yes, than it is alright.

PR #4201, opened 2023-11-10
fix: Add missing validation when records does not contain vector for search

frascuchon:
# Description

Return status code 422 if the selected record for find similarity has no vector value for the selected vector name.

### DIFF
--- a/src/argilla/server/apis/v1/handlers/datasets.py
+++ b/src/argilla/server/apis/v1/handlers/datasets.py
@@ -229,10 +229,10 @@ async def _get_search_responses(
             record = await _get_dataset_record_by_id_or_raise(db, dataset, vector_query.record_id)
             await record.awaitable_attrs.vectors
 
-            if record.vector_value_by_vector_settings(vector_settings) is None:
+            if not record.vector_value_by_vector_settings(vector_settings):
                 raise HTTPException(
                     status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
-                    detail=f"Record {record.id} does not have a vector for vector settings {vector_settings.name}",
+                    detail=f"Record `{record.id}` does not have a vector for vector settings `{vector_settings.name}`",
                 )
 
     if (
--- a/src/argilla/server/apis/v1/handlers/datasets.py
+++ b/src/argilla/server/apis/v1/handlers/datasets.py
@@ -229,7 +229,7 @@ async def _get_search_responses(
             record = await _get_dataset_record_by_id_or_raise(db, dataset, vector_query.record_id)
             await record.awaitable_attrs.vectors
 
-            if record.vector_by_vector_settings(vector_settings) is None:
+            if record.vector_value_by_vector_settings(vector_settings) is None:
                 raise HTTPException(
                     status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
                     detail=f"Record {record.id} does not have a vector for vector settings {vector_settings.name}",
--- a/src/argilla/server/apis/v1/handlers/datasets.py
+++ b/src/argilla/server/apis/v1/handlers/datasets.py
@@ -229,6 +229,12 @@ async def _get_search_responses(
             record = await _get_dataset_record_by_id_or_raise(db, dataset, vector_query.record_id)
             await record.awaitable_attrs.vectors
 
+            if record.vector_by_vector_settings(vector_settings) is None:
+                raise HTTPException(
+                    status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+                    detail=f"Record {record.id} does not have a vector for vector settings {vector_settings.name}",
+                )
+
     if (
         text_query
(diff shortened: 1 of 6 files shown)


src/argilla/server/models/database.py
```diff
@@ -207,6 +207,12 @@ def __repr__(self):
             f"inserted_at={str(self.inserted_at)!r}, updated_at={str(self.updated_at)!r})"
         )
 
+    def vector_by_vector_settings(self, vector_settings: "VectorSettings") -> Union[Vector, None]:
+        for vector in self.vectors:
+            if vector.vector_settings_id == vector_settings.id:
+                return vector.value
+        return None
```
jfcalvo: Maybe instead of this method we should create inside the context file a `get_record_vector_for_vector_settings(record_id, vector_settings_ids)` or similar (doing only a query instead of iterating through all vectors collection).

Apart from that if we don't want to do that previous change we can remove the explicit `return None` and change the name of the method because we are not returning a vector but it's value instead:

```suggestion
    def vector_value_by_vector_settings(self, vector_settings: "VectorSettings") -> Union[List[float], None]:
        for vector in self.vectors:
            if vector.vector_settings_id == vector_settings.id:
                return vector.value
```

jfcalvo (changes requested):

src/argilla/server/apis/v1/handlers/datasets.py
```diff
@@ -229,6 +229,12 @@ async def _get_search_responses(
             record = await _get_dataset_record_by_id_or_raise(db, dataset, vector_query.record_id)
             await record.awaitable_attrs.vectors
 
+            if record.vector_value_by_vector_settings(vector_settings) is None:
```
jfcalvo:
```suggestion
            if not record.vector_value_by_vector_settings(vector_settings):
```

src/argilla/server/apis/v1/handlers/datasets.py
```diff
@@ -229,6 +229,12 @@ async def _get_search_responses(
...
+                raise HTTPException(
+                    status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+                    detail=f"Record {record.id} does not have a vector for vector settings {vector_settings.name}",
```
jfcalvo:
```suggestion
                    detail=f"Record `{record.id}` does not have a vector for vector settings `{vector_settings.name}`",
```

src/argilla/server/models/database.py
frascuchon: Renamed. About moving to a context function, my feelings are that the current code base is not yet prepared for that. We're working with ORM classes in other parts than DB repositories. We should use agnostic domain classes shared by all the layers, feeding them from the proper places. 

For example, this specific logic is needed in the search engine module, which should be totally agnostic about the ORM classes

src/argilla/server/models/database.py
```diff
@@ -207,6 +207,12 @@ def __repr__(self):
             f"inserted_at={str(self.inserted_at)!r}, updated_at={str(self.updated_at)!r})"
         )
 
+    def vector_value_by_vector_settings(self, vector_settings: "VectorSettings") -> Union[List[float], None]:
+        for vector in self.vectors:
+            if vector.vector_settings_id == vector_settings.id:
+                return vector.value
+        return None
```
jfcalvo:
```suggestion
    def vector_value_by_vector_settings(self, vector_settings: "VectorSettings") -> Union[List[float], None]:
        for vector in self.vectors:
            if vector.vector_settings_id == vector_settings.id:
                return vector.value
```

src/argilla/server/schemas/v1/suggestions.py
```diff
@@ -23,6 +23,7 @@
 class BaseSuggestion(BaseModel):
     question_id: UUID
     type: Optional[SuggestionType]
+    # TODO: score must be a float between 0 and 1
```
jfcalvo:
```suggestion
    # TODO: we should add a validation to score only allowing values between 0 and 1
```

tests/unit/server/api/v1/datasets/test_search_dataset_records.py
```diff
@@ -203,3 +213,27 @@ async def test_with_non_existent_dataset(self, async_client: AsyncClient, owner_
...
+
+        response = await async_client.post(
+            f"/api/v1/me/datasets/{dataset.id}/records/search",
+            headers=owner_auth_header,
+            json={"query": {"vector": {"name": vector_settings.name, "record_id": str(record_without_vector.id)}}},
+        )
+
+        assert response.status_code == 422
+        assert response.json() == {
+            "detail": f"Record {record_without_vector.id} does not have a vector for vector settings {vector_settings.name}"
```
jfcalvo:
```suggestion
            "detail": f"Record `{record_without_vector.id}` does not have a vector for vector settings `{vector_settings.name}`"
```

PR #4203, opened 2023-11-10
fix: safer dump of `vectors` in the HuggingFace Hub

alvarobartt:
# Description

This PR solves an issue that was happening (and is still happening) related to the naming we're using for the columns in the HuggingFace Hub dataset, as we're exporting each of the fields and responses using their names in Argilla, and appending `-suggestion` or whatever feels necessary, but we can run into conflicts in any `field` is name as any other `question`, named `metadata` or named `external_id`, as well as the other way around for `question`.

So on, this PR fixes that for the upcoming `vectors`, as those are usually named after the field that those refer to, so instead of dumping those under the same name, we're dumping a Python dict in a column named `vectors` where each key-value pair is a vector name and its value. Similarly to what was done previously for the former Argilla datasets.

**Type of change**

**How Has This Been Tested**

- [x] Add integration tests for duplicated name on `vectors_settings` for a given vector and ensured that it's dumped and restored properly

**Checklist**

- [x] I filled out [the contributor form](https://tally.so/r/n9XrxK) (see text above)

### DIFF
--- a/src/argilla/client/feedback/integrations/huggingface/dataset.py
+++ b/src/argilla/client/feedback/integrations/huggingface/dataset.py
@@ -115,8 +115,10 @@ class HuggingFaceDatasetMixin:
 
         vectors_settings = dataset.vectors_settings
         if vectors_settings:
+            hf_features["vectors"] = {}
             for vector_settings in vectors_settings:
-                hf_features[vector_settings.name] = Sequence(Value(dtype="float32"), id="vectors")
+                hf_features["vectors"].update({vector_settings.name: Sequence(Value(dtype="float32"), id="vectors")})
+            hf_dataset["vectors"] = []
 
         for record in dataset.records:
             for field in dataset.fields:
@@ -160,10 +162,10 @@ class HuggingFaceDatasetMixin:
             hf_dataset["external_id"].append(record.external_id or None)
 
             if vectors_settings:
+                vectors = {}
                 for vector_settings in vectors_settings:
-                    if vector_settings.name not in hf_dataset:
-                        hf_dataset[vector_settings.name] = []
-                    hf_dataset[vector_settings.name].append(record.vectors.get(vector_settings.name, None))
+                    vectors.update({vector_settings.name: record.vectors.get(vector_settings.name, None)})
+                hf_dataset["vectors"].append(vectors)
 
         return Dataset.from_dict(hf_dataset, features=Features(hf_features))
 
@@ -439,9 +441,13 @@ class HuggingFaceDatasetMixin:
                 metadata = json.loads(hfds[index]["metadata"])
 
             vectors = {}
-            for vector_settings in dataset.vectors_settings:
-                if vector_settings.name in hfds[index] and hfds[index][vector_settings.name] is not None:
-                    vectors.update({vector_settings.name: hfds[index][vector_settings.name]})
+            if "vectors" in hfds[index] and hfds[index]["vectors"]:
+                for vector_settings in dataset.vectors_settings:
+                    if (
+                        vector_settings.name in hfds[index]["vectors"]
+                        and hfds[index]["vectors"][vector_settings.name] is not None
+                    ):
+                        vectors.update({vector_settings.name: hfds[index]["vectors"][vector_settings.name]})
 
             records.append(
                 FeedbackRecord(
--- a/src/argilla/client/feedback/integrations/huggingface/dataset.py
+++ b/src/argilla/client/feedback/integrations/huggingface/dataset.py
@@ -443,10 +443,7 @@ class HuggingFaceDatasetMixin:
             vectors = {}
             if "vectors" in hfds[index] and hfds[index]["vectors"]:
                 for vector_settings in dataset.vectors_settings:
-                    if (
-                        vector_settings.name in hfds[index]["vectors"]
-                        and hfds[index]["vectors"][vector_settings.name] is not None
-                    ):
+                    if hfds[index]["vectors"].get(vector_settings.name, None) is not None:
                         vectors.update({vector_settings.name: hfds[index]["vectors"][vector_settings.name]})
 
             records.append(
--- a/tests/integration/client/feedback/integrations/huggingface/test_dataset.py
+++ b/tests/integration/client/feedback/integrations/huggingface/test_dataset.py
@@ -24,7 +24,7 @@ import pytest
 class TestSuiteHuggingFaceDatasetMixin:
     @classmethod
     def setup_class(cls: "TestSuiteHuggingFaceDatasetMixin") -> None:
-        github_ref_name = re.sub(r"[^a-zA-Z0-9-]", "-", os.getenv("GITHUB_REF_NAME", "fix/hello-world"))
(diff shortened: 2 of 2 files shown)


src/argilla/client/feedback/integrations/huggingface/dataset.py
```diff
@@ -160,10 +162,10 @@ def _huggingface_format(dataset: Union["FeedbackDataset", "RemoteFeedbackDataset
             hf_dataset["external_id"].append(record.external_id or None)
 
             if vectors_settings:
+                vectors = {}
                 for vector_settings in vectors_settings:
-                    if vector_settings.name not in hf_dataset:
-                        hf_dataset[vector_settings.name] = []
-                    hf_dataset[vector_settings.name].append(record.vectors.get(vector_settings.name, None))
+                    vectors.update({vector_settings.name: record.vectors.get(vector_settings.name, None)})
```
frascuchon, line 167: Here we can have a problem when loading the dataset from huggingface with `None`. After loading this, when pushing to argilla, the vector validation will raise an error with a `None` value that is not a valid `list[float]` or something similar.

frascuchon:
```suggestion
                    vector_value = record.vectors.get(vector_settings.name, None)
                    if vector_value:
                       vectors.update({vector_settings.name: vector_value})
```

frascuchon, line 167: Do we need to provide the same dict structure for vectors in HF for ALL the records? I mean, having 2 vectors,  `v1` and `v2` and record[0] only with `v1`, and record[1] only with `v2`. Do we still need to create the whole dictionary even if there is partial info?

src/argilla/client/feedback/integrations/huggingface/dataset.py
```diff
@@ -439,9 +441,13 @@ def from_huggingface(
...
-            for vector_settings in dataset.vectors_settings:
-                if vector_settings.name in hfds[index] and hfds[index][vector_settings.name] is not None:
-                    vectors.update({vector_settings.name: hfds[index][vector_settings.name]})
+            if "vectors" in hfds[index] and hfds[index]["vectors"]:
+                for vector_settings in dataset.vectors_settings:
+                    if (
+                        vector_settings.name in hfds[index]["vectors"]
+                        and hfds[index]["vectors"][vector_settings.name] is not None
+                    ):
+                        vectors.update({vector_settings.name: hfds[index]["vectors"][vector_settings.name]})
```
frascuchon:
```suggestion

                    vector_value = hfds[index]["vectors"].get(vector_settings.name, None)
                    if vector_value:
                        vectors.update({vector_settings.name: vector_value})
```

alvarobartt:
```suggestion
                    if hfds[index]["vectors"].get(vector_settings.name, None) is not None:
                        vectors.update({vector_settings.name: hfds[index]["vectors"][vector_settings.name]})
```

src/argilla/client/feedback/integrations/huggingface/dataset.py
alvarobartt: Yes, we do need it before dumping into `datasets.Dataset` format, but we're skipping those when `None` while fetching them

tests/integration/client/feedback/integrations/huggingface/test_dataset.py
```diff
@@ -44,7 +44,9 @@ def setup_class(cls: "TestSuiteHuggingFaceDatasetMixin") -> None:
                 rg.FloatMetadataProperty(name="float-metadata-property", min=0.0, max=100.0),
             ],
             vectors_settings=[
-                rg.VectorSettings(name="float-vector", dimensions=2),
+                # Named `text-field` to ensure that even if name is duplicated with a field/question, the
+                # vector is properly serialized under the `vectors` column in the `datasets.Dataset`
+                rg.VectorSettings(name="text-field", dimensions=2),
```
frascuchon: I would prefer seeing a specific test checking the error instead of changing the commong base feedback dataset. Something like:

```python
def test_pushing_vectors_with_same_name_as_fields(self) -> None:
   first_field = self.dataset.fields[0]
   v_settings = self.dataset.add_vector_settings(rg.VectorSettings(name=first_field.name, dimensions=2))
   
   for record in dataset:
      record.vectors.update({ v_settings.name: [0, 1]}
   
   self.dataset.push_to_huggingface(repo_id=self.repo_id, private=True, token=os.getenv("HF_HUB_ACCESS_TOKEN"))

   pushed_dataset = = rg.FeedbackDataset.from_huggingface(repo_id=self.repo_id, token=os.getenv("HF_HUB_ACCESS_TOKEN"))
   assert("All records contains the expected vectors" )
   assert("All records contains the expected fields")
```

frascuchon (changes requested):
Some comments. Mostly related to a proper missing test

tests/integration/client/feedback/integrations/huggingface/test_dataset.py
alvarobartt: Fair, I'll do so!

src/argilla/client/feedback/integrations/huggingface/dataset.py
jfcalvo: If you are skipping then later if they are `None` you can avoid them to be added on first instance too right?

tests/integration/client/feedback/integrations/huggingface/test_dataset.py
```diff
@@ -87,3 +87,49 @@ def test_from_huggingface(self) -> None:
         assert dataset.vectors_settings == self.dataset.vectors_settings
         assert dataset.guidelines == self.dataset.guidelines
         assert dataset.records == self.dataset.records
+
+
+# Separate edge cases for integration tests
+@pytest.mark.skipif(os.getenv("HF_HUB_ACCESS_TOKEN") is None, reason="`HF_HUB_ACCESS_TOKEN` is not set")
+def test_push_to_huggingface_duplicated_vectors_name() -> None:
```
jfcalvo:
```suggestion
def test_push_to_huggingface_with_field_and_vector_using_same_name() -> None:
```

tests/integration/client/feedback/integrations/huggingface/test_dataset.py
```diff
@@ -87,3 +87,49 @@ def test_from_huggingface(self) -> None:
...
+        fields=[
+            rg.TextField(name="text-field", required=True),
+        ],
+        questions=[
+            rg.TextQuestion(name="text-question", required=True),
+        ],
+        vectors_settings=[
+            # Named `text-field` to ensure that even if name is duplicated with a field/question, the
+            # vector is properly serialized under the `vectors` column in the `datasets.Dataset`
+            rg.VectorSettings(name="float-vector", dimensions=2),
```
jfcalvo: Why is this `VectorSettings` not named `text-field`?

frascuchon (approved):
I launched tests locally and is working. The ci is skipping these tests for this kind of branches. See https://github.com/argilla-io/argilla/issues/4205

PR #4211, opened 2023-11-13
fix: correct bug on sentence-transformers trainer with a list of values in the records

### LINKED ISSUE #4088: [FEATURE] formatting_func for sentence similarity return list
From PR #3955, Regarding the relation between the dataset for rag and the 'formatting func'. What it makes more sense to me is to use the FeedbackTemplate with the query and all  the retrieved documents, indicating, for instance, `number_of_retrievals=3`, and then after annotating in the ArgillaUI, when using the formatting_func, be able to make the matches: query-retrieved_document_1-label1, query-retrieved_document_2,-label2 query-retrieved_document_3-label3.

```python
# Adding records to the dataset
records = [
    rg.FeedbackRecord(
        fields={"query": question, "retrieved_document_1": context[0], "retrieved_document_2": context[1], "retrieved_document_3": context[2]},
    )
    for question, answer, context in tqdm(zip(questions, contexts))
]

# Add records to the dataset
dataset_rag.add_records(records)

```
Then, we would annotate using the ArgillaUI and add the LabelQuestion. So, now to define the task we would have to use a `formatting_func` to create the appropriate data matching the query with each of the retrieved_documents and the corresponding label.

Using a list

```python
def formatting_func(sample):

    records = []

    for i in range(1, 4):
        record = {"sentence-1": sample["query"], "sentence-2": sample[f"retrieved_document_{i}"]}
        values = [resp["value"] for resp in sample[f"label{i}"]]
        label = int(values[0])
        record["label"] = label
        records.append(record)

    return records
   ```
Or yield

```python
def formatting_func(sample):
    for i in range(1, 4):
        record = {
            "sentence-1": sample["query"],
            "sentence-2": sample[f"retrieved_document_{i}"]
        }
        values = [resp["value"] for resp in sample[f"label{i}"]]
        label = int(values[0])
        record["label"] = label
        yield record
```
Even if we could return a dictionary of dictionaries.

**Additional context**

I also checked the code and, for instance, in test_model_card.py, they seem to use something similar to what I want in the case 'c', but it doesn't work:

```python
    def formatting_func(sample):
        labels = [
            annotation["value"]
            for annotation in sample["question-3"]
            if annotation["status"] == "submitted" and annotation["value"] is not None
        ]
        if labels:
            # Three cases for the tests: None, one tuple and yielding multiple tuples
            if labels[0] == "a":
                return None
            elif labels[0] == "b":
                return {"sentence-1": sample["text"], "sentence-2": sample["text"], "label": 1}
            elif labels[0] == "c":
                return [
                    {"sentence-1": sample["text"], "sentence-2": sample["text"], "label": 1},
                    {"sentence-1": sample["text"], "sentence-2": sample["text"], "label": 0},
                ]

    task = TrainingTask.for_sentence_similarity(formatting_func=formatting_func)

```

plaguss:
# Description

This PR solves an error in the `ArgillaTrainer` for `sentence-similarity` when the records include nested lists.

Closes #4088

**Type of change**

**How Has This Been Tested**

Tested locally

**Checklist**

### DIFF
--- a/src/argilla/client/feedback/training/schemas.py
+++ b/src/argilla/client/feedback/training/schemas.py
@@ -133,7 +133,6 @@ class TrainingData(ABC):
                         explode_columns.add(pydantic_field_name)
             formatted_data.append(data)
         df = pd.DataFrame(formatted_data)
-
         if explode_columns:
             df = df.explode(list(explode_columns))
         # In cases of MultiLabel datasets the label column contains a list,
@@ -1594,7 +1593,7 @@ class TrainingTaskForSentenceSimilarity(BaseModel, TrainingData):
 
     @requires_dependencies("sentence-transformers")
     def _prepare_for_training_with_sentence_transformers(
-        self, data: List[dict], train_size: float, seed: int
+        self, data: Union[List[dict], List[List[dict]]], train_size: float, seed: int
     ) -> Union["InputExample", Tuple["InputExample", "InputExample"]]:
         from sentence_transformers import InputExample
 
@@ -1615,7 +1614,7 @@ class TrainingTaskForSentenceSimilarity(BaseModel, TrainingData):
             def dataset_fields(sample):
                 return {"texts": [sample["sentence-1"], sample["sentence-2"]], "label": sample["label"]}
 
-        elif sample_keys == sample_keys == {"label", "sentence-1", "sentence-2", "sentence-3"}:
+        elif sample_keys == {"label", "sentence-1", "sentence-2", "sentence-3"}:
 
             def dataset_fields(sample):
                 return {
--- a/src/argilla/client/feedback/training/schemas.py
+++ b/src/argilla/client/feedback/training/schemas.py
@@ -1596,13 +1596,22 @@ class TrainingTaskForSentenceSimilarity(BaseModel, TrainingData):
     def _prepare_for_training_with_sentence_transformers(
         self, data: List[dict], train_size: float, seed: int
     ) -> Union["InputExample", Tuple["InputExample", "InputExample"]]:
+        from types import GeneratorType
+
         from sentence_transformers import InputExample
 
         if not len(data) > 0:
             raise ValueError("The dataset must contain at least one sample to be able to train.")
 
         # Use the first sample to decide what type of dataset to generate:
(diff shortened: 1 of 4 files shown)


davidberenstein1957:
@plaguss, could you add some small tests for this too?

src/argilla/client/feedback/training/schemas.py
```diff
@@ -1602,7 +1602,14 @@ def _prepare_for_training_with_sentence_transformers(
...
-        sample_keys = set(data[0].keys())
+        if isinstance(data[0], list):
+            # In case we are returning lists, extract the first element of that list to check the fields.
+            sample_keys = set(data[0][0].keys())
+        elif isinstance(data[0], dict):
+            sample_keys = set(data[0].keys())
+        else:
+            raise ValueError(f"The type is not supported: {type(data[0])}.")
+
         if sample_keys == {"label", "sentence-1", "sentence-2"}:
```
davidberenstein1957, line 1612: maybe we should review these check too in order to align with a `set`? also, there seems to be a faulty line in 1618 checking for ` sample_keys == sample_keys`

plaguss: > maybe we should review these check too in order to align with a `set`? also, there seems to be a faulty line in 1618 checking for ` sample_keys == sample_keys`

I didn't noticed the error in line 1618.. thanks

davidberenstein1957:
@plaguss could you also check if the other implementations allow for passing generators too?

plaguss:
> @plaguss could you also check if the other implementations allow for passing generators too?

We only have tests for generators in the `text-classification` frameworks. I'm not sure if trl/openai work, let me check

plaguss:
> > @plaguss could you also check if the other implementations allow for passing generators too?
> 
> We only have tests for generators in the `text-classification` frameworks. I'm not sure if trl/openai work, let me check

I was speaking with @sdiazlor and it seems that the generators are not working properly with the other frameworks (she mentioned the `chat-completion` task I think). Maybe we can close this one and tackle that in a different issue?

davidberenstein1957:
@plaguss, I agree. Feel free to close this and create a separate issue.

sdiazlor (approved):

PR #4270, opened 2023-11-17
feat: define filters and sort schemas for search

jfcalvo:
# Description

This is a proposal for adding `filters` and `sort` attributes on the endpoints for search records.

The changes allow us to define the following schema for searching:

```python
from argilla.server.schemas.v1.datasets import SearchRecordsQuery

SearchRecordsQuery.parse_obj({
  "query": {
    "text": {
      "q": "query"
    },
  },
  "filters": {
    "and": [
      {"type": "terms", "scope": {"entity": "suggestion", "question": "sentiment"}, "values": ["positive"]},
      {"type": "terms", "scope": {"entity": "suggestion", "question": "sentiment", "property": "agent"}, "values": ["chat-gpt3.5", "chat-gpt-4.0"]},
      {"type": "terms", "scope": {"entity": "response", "question": "topic"}, "values": ["politics", "news"]},
      {"type": "range", "scope": {"entity": "response", "question": "quality"}, "gte": 0.5, "lte": 0.9},
      {"type": "range", "scope": {"entity": "metadata", "metadata_property": "price"}, "gte": 100.0},
    ]
  },
  "sort": [
    {"scope": {"entity": "suggestion", "question": "sentiment", "property": "score"}, "order": "asc"},
    {"scope": {"entity": "response", "question": "quality"}, "order": "desc"},
  ]
})

# This is an old proposal of the schema leaving it here for reference
SearchRecordsQuery.parse_obj({
  "query": {
    "text": {
      "q": "query"
    },
  },
  "filters": {
    "and": [
      {"type": "terms", "field": "suggestion.sentiment.value", "values": ["positive"]},
      {"type": "terms", "field": "suggestion.sentiment.agent", "values": ["chat-gpt3.5", "chat-gpt-4.0"]},
      {"type": "terms", "field": "response.topic.value", "values": ["politics", "news"]},
      {"type": "range", "field": "response.quality.value", "gte": 0.5, "lte": 0.9},
    ]
  },
  "sort": [
    {"field": "suggestion.sentiment.score", "order": "asc"},
    {"field": "response.quality.value", "order": "desc"},
  ]
})
```

Notice some changes from the Notion documentation:
* I'm using `field` instead of `path` because we are already using `field` in the `query` parameter.
* I have added the possibility of specify `float` in the values in the case that we want to specify numbers.
* <del>I'm adding a new `term` (singular) filter so is more explicit to filter by one specific value (this is optional we can remove it if not needed).</del>

With this PR we should ask several questions:
- [X] Is this functionality enough for the changes that are required on the frontend for filtering responses and suggestions?
  - Reviewed with @damianpumar and feedback applied.
- [ ] Is this format flexible enough so in the future we can add more boolean operators and extend the search without deprecate anything?.
- [x] Do we need additional boolean operators by now or it's enough with `and`?
  - `and` will be enough for the frontend requirements by now.
- [x] What `field`format should we follow?
  - <del>Making the "value" of the field to filter implicit like `suggestion.sentiment` instead of `suggestion.sentiment.value`?</del>
  - Making the "value" of the field to filter explicit like `suggestion.sentiment.value` instead of `suggestion.sentiment`?
    - We choose this solution and we will always use `value` (instead of `values`) at the end. 

@damianpumar should help us to answer the first question of this list.

**Type of change**

**How Has This Been Tested**

- [x] Running tests locally.

**Checklist**

### DIFF
--- a/src/argilla/server/schemas/v1/datasets.py
+++ b/src/argilla/server/schemas/v1/datasets.py
@@ -97,6 +97,18 @@ RECORDS_CREATE_MAX_ITEMS = 1000
 RECORDS_UPDATE_MIN_ITEMS = 1
 RECORDS_UPDATE_MAX_ITEMS = 1000
 
+TERMS_FILTER_VALUES_MIN_ITEMS = 1
+TERMS_FILTER_VALUES_MAX_ITEMS = 250
+
+FILTER_FIELD_MIN_LENGTH = 1
+FILTER_FIELD_MAX_LENGTH = 250
+
+FILTERS_AND_MIN_ITEMS = 1
+FILTERS_AND_MAX_ITEMS = 50
+
+SEARCH_RECORDS_QUERY_SORT_MIN_ITEMS = 1
+SEARCH_RECORDS_QUERY_SORT_MAX_ITEMS = 10
+
 
 class Dataset(BaseModel):
     id: UUID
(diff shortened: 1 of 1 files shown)


damianpumar:
> ```python
> terms
> ```

damianpumar:
@jfcalvo Yes, for me it's ok.

Thanks Jose!

damianpumar:
I found some issues in your object scheme,

  {"type": "terms", "field": "suggestion.sentiment.value**s**", "values": ["positive"]},

This example does not exist 👇
 {"type": "range", "field": "suggestion.sentiment.score", "gte": 0.8},

This example 👇
{"type": "term",  "field": "suggestion.sentimen.score", "value": 0.6},
must be
{"type": "range",  "field": "suggestion.sentimen**t**.score", **"gte": 0.5, "lte": 0.9**},

This example must be: 
{"type": "terms", "field": "suggestion.sentiment.value**s**", "values": ["positive"]},

damianpumar:
- For suggestions:
 - Score:
     - ```{"type": "range",  "field": "suggestion.sentiment.score", "gte": 0.5, "lte": 0.9}```
 - Agent:
     - ```{type": "terms", "field": "suggestion.sentiment.agent", "values": ["chat-gpt3.5", "chat-gpt-4.0"]}``` 
 - Values:
     - OR: 
        - ```{type": "terms", "field": "suggestion.sentiment.values", "values": ["positive", "negative"]}``` TBD: **value** or **values**
     - AND: 
       -  ```{"type": "terms", "field": "suggestion.sentiment.values", "values": ["positive"]}```
       -  ```{"type": "terms", "field": "suggestion.sentiment.values", "values": ["negative"]}```
 - For responses:
      - ```{"type": "terms", "field": "response.topic.values", "values": ["politics", "news"]}```

The **sort** key is not necessary for this iteration.

src/argilla/server/schemas/v1/datasets.py
```diff
@@ -714,8 +726,44 @@ def check_required(cls, values: dict) -> dict:
         return values
 
 
+FilterField = Annotated[str, PydanticField(..., min_length=FILTER_FIELD_MIN_LENGTH, max_length=FILTER_FIELD_MAX_LENGTH)]
```
frascuchon, line 729: We can align this to the constr we are already using for other str types
```suggestion

```

frascuchon (approved):

src/argilla/server/schemas/v1/datasets.py
```diff
@@ -714,8 +726,44 @@ def check_required(cls, values: dict) -> dict:
...
+    values: List[str] = PydanticField(
+        ..., min_items=TERMS_FILTER_VALUES_MIN_ITEMS, max_items=TERMS_FILTER_VALUES_MAX_ITEMS
+    )
+
+
+class RangeFilter(BaseModel):
+    type: Literal["range"]
+    field: FilterField
+    gte: float
+    lte: float
```
frascuchon: I see this now. We should be able to let users providing only one value (gte or lte) for open ranges

jfcalvo: Just added the validation for that.

frascuchon (approved):
I think this way is much more readable, comprehensible, parseable, and easy to validate

PR #4297, opened 2023-11-21
feat: Add a general `text_wrapper` to create `repr` methods

### LINKED ISSUE #2259: [FEATURE] add `__repr__` for the SDK classes
N.A.

The method should expose the active workspace and potentially available workspaces, URL/host and available datasets.
Suggestions are welcome:)

N.A.

**Additional context**
N.A.

kursathalat:
# Description

This PR adds a `text_wrapper` method that creates object representations automatically for classes that inherits from pydantic `BaseModel`. It also includes changes in repr methods of some other classes that are not inheriting from BaseModel.

Closes #2259 

**Type of change**

**How Has This Been Tested**

- [ ] Test A
- [ ] Test B

**Checklist**

### DIFF
--- a/src/argilla/client/users.py
+++ b/src/argilla/client/users.py
@@ -12,6 +12,7 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+import textwrap
 import warnings
 from datetime import datetime
 from typing import TYPE_CHECKING, Iterator, List, Optional, Union
@@ -149,11 +150,19 @@ class User:
         return self.role == UserRole.annotator
 
     def __repr__(self) -> str:
+        indent = "   "
+        first_name = self.first_name or None
         return (
-            f"User(id={self.id}, username={self.username}, role={self.role},"
-            f" api_key={self.api_key}, first_name={self.first_name},"
-            f" last_name={self.last_name}, inserted_at={self.inserted_at},"
-            f" updated_at={self.updated_at})"
+            "User("
+            + textwrap.indent(f"\nid={self.id}", indent)
+            + textwrap.indent(f"\nusername={self.username}", indent)
+            + textwrap.indent(f"\nrole={self.role}", indent)
+            + textwrap.indent(f"\napi_key={self.api_key}", indent)
+            + textwrap.indent(f"\nfirst_name={first_name}", indent)
+            + textwrap.indent(f"\nlast_name={self.last_name}", indent)
+            + textwrap.indent(f"\ninserted_at={self.inserted_at}", indent)
+            + textwrap.indent(f"\nupdated_at={self.updated_at}", indent)
+            + "\n)"
         )
 
     @staticmethod
--- a/src/argilla/client/workspaces.py
+++ b/src/argilla/client/workspaces.py
@@ -12,6 +12,7 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+import textwrap
 import warnings
 from datetime import datetime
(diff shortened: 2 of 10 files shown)


src/argilla/textwrapping.py
```diff
@@ -0,0 +1,59 @@
...
+    Returns:
+        A string for the __repr__ method of the class that inherits from pydantic.BaseModel.
+    """
+
+    if attr_ignore == None:
+        attr_ignore = []
+    if attr_multiline == None:
+        attr_multiline = []
+
+    def create_wrapped_list(self, attr_ignore=None, attr_multiline=None):
```
davidberenstein1957, line 36: would it be possible to check if an attribute is a `str` and then limit the entire printed text? To the first n characters ... last n characters?

src/argilla/listeners/models.py
```diff
@@ -36,6 +37,15 @@ class Search:
     total: int
     query_params: Optional[Dict[str, Any]] = None
 
+    def __repr__(self) -> str:
```
davidberenstein1957, line 40: we added a textwrapper to avoid this right?

src/argilla/listeners/models.py
```diff
@@ -67,6 +77,16 @@ class RGListenerContext:
     metrics: Optional[Metrics] = None
     query_params: Optional[Dict[str, Any]] = None
 
+    def __repr__(self) -> str:
```
davidberenstein1957, line 80: we added a textwrapper to avoid this right?

src/argilla/client/workspaces.py
```diff
@@ -119,9 +120,14 @@ def users(self) -> List["UserModel"]:
         return workspaces_api.list_workspace_users(self._client, self.id).parsed
 
     def __repr__(self) -> str:
```
davidberenstein1957, line 122: we added a textwrapper to avoid this right?

src/argilla/client/users.py
```diff
@@ -149,11 +150,19 @@ def is_annotator(self) -> bool:
         return self.role == UserRole.annotator
 
     def __repr__(self) -> str:
```
davidberenstein1957, line 152: we added a textwrapper to avoid this right?

davidberenstein1957 (commented):
LFTM but some minor remarks.

@kursathalat, as discussed we might make a share `Basemodel` across all schemas which we can overwrite. I think we can add this shared `TextWrappedBaseModel` and the `textwrapping` to `src/argilla/utils/utils.py`. 

Additionally, can you add some examples to the different representations using the arguments in the functiosns?

src/argilla/textwrapping.py
kursathalat: Currently, it is limiting `str` longer than 50, we can change it. However some strings like name are `ConstrainedStrValue`, so they are not limited and they should be also be addressed.

kursathalat:
Currently it is 
```
repr(rg.SuggestionSchema(question_name="quest")) # type is not shown
repr(rg.VectorSettings(name="vectorsetting", title="An example of a string that is longer than 50 characters so this one is limited as the first and second part which are connected with three dots.", dimensions=5))
```
and output

```
SuggestionSchema(
   question_name=quest
   score=None
   value=None
   agent=None
)
VectorSettings(
   name=vectorsetting
   title=An example of a string ... ected with three dots.
   dimensions=5
)
```

The problem is that the process of ignoring attributes is done via the `default` parameter of `Field` at the moment, which is temporary and should be changed. Ideally, we want to access the `repr` parameter of `Field`, but have not been able to do it.

PR #4303, opened 2023-11-22
feat: new pagination UI

leiyre:
# Description

This PR includes new pagination UI

**Type of change**

### DIFF
--- a/frontend/components/feedback-task/footer/Pagination.component.vue
+++ b/frontend/components/feedback-task/footer/Pagination.component.vue
@@ -1,5 +1,8 @@
 <template>
   <div class="pagination">
+    <span class="pagination__info" v-if="!!items">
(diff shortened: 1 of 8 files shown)


frontend/components/feedback-task/footer/Pagination.component.vue
```diff
@@ -41,21 +42,34 @@ export default {
...
   computed: {
     isFirstPage() {
       return this.currentPage === 1;
     },
-    pageFromRoute() {
-      return parseFloat(this.$route.query?._page) || 1;
+    totalPages() {
+      return Math.ceil(this.items / this.itemsPerPage);
+    },
+    currentPaginationRange() {
```
damianpumar, line 61: I think we need to do a PageCriteria to support this, because, now the pagination could be modify by the user

Issue #4228, opened 2023-11-14
[FEATURE] Integrate new filter and sort in search engine


frascuchon:
```[tasklist]
### Tasks
- [ ] https://github.com/argilla-io/argilla/pull/4292
- [ ] https://github.com/argilla-io/argilla/pull/4305
- [ ] https://github.com/argilla-io/argilla/pull/4309
```

PR #4305, opened 2023-11-22
feat: Using new filters mechanism with `search` method

frascuchon:
# Description

This PR adds support for the new `filter` and `sort` argument in the `search` method. 

Also, the deprecated arguments `metadata_filters`, `user_response_status_filter`, and  `sort_by` are parsed and unified with the new filter definition schema, so all filter and sort configurations are generated in the same way.

Base filter classes have been also rewritten to avoid using the pydantic `BaseModel` class. This change is motivated by weird behavior detected when using `Union`-definition types in pydantic schemas, which are parsing init arguments and generating unexpected objects.

As an illustrative example:

```python
from pydantic import BaseModel
from typing import Union

class A(BaseModel):
   property: str

class B(BaseModel):
   property: str

UnionClass = Union[A, B]

class C(BaseModel):
   p: UnionClass

b = B(property=200)
c = C(p=b)

assert type(b) == type(c.p) # This will fail
```

Refs https://github.com/argilla-io/argilla/issues/4228

### DIFF
--- a/src/argilla/server/search_engine/base.py
+++ b/src/argilla/server/search_engine/base.py
@@ -72,44 +72,57 @@ __all__ = [
 ]
 
 
-class SuggestionFilterScope(BaseModel):
-    question: str
-    property: str
+class SuggestionFilterScope:
+    def __init__(self, question: str, property: str):
+        self.question = question
+        self.property = property
 
 
-class ResponseFilterScope(BaseModel):
-    question: Optional[str]
-    property: Optional[str]
+class ResponseFilterScope:
+    def __init__(self, question: Optional[str] = None, property: Optional[str] = None, user: Optional[User] = None):
+        self.question = question
+        self.property = property
+        self.user = user
 
 
-class MetadataFilterScope(BaseModel):
-    metadata_property: str
+class MetadataFilterScope:
+    def __init__(self, metadata_property: str):
+        self.metadata_property = metadata_property
 
 
-FilterScope = Union[SuggestionFilterScope, ResponseFilterScope, MetadataFilterScope]
+class RecordFilterScope:
+    def __init__(self, property: str):
+        self.property = property
 
 
-class TermsFilter(BaseModel):
-    scope: FilterScope
-    values: List[str]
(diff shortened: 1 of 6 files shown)


src/argilla/server/search_engine/base.py
```diff
@@ -72,44 +72,57 @@
...
 
 
-class TermsFilter(BaseModel):
-    scope: FilterScope
-    values: List[str]
+FilterScope = Union[SuggestionFilterScope, ResponseFilterScope, MetadataFilterScope, RecordFilterScope]
+
+
+class TermsFilter:
+    def __init__(self, scope: FilterScope, values: List[str]):
```
jfcalvo, line 102: I know `TermsFilter` handlers schema has `values` defined as `List[str]` but I'm thinking that maybe for `RatingQuestions` this could be not true. I mean `RatingQuestions` have `values` as `List[int]` so maybe we need to change it.

src/argilla/server/search_engine/commons.py
```diff
@@ -114,15 +124,58 @@ class Config:
...
+def es_index_name_for_dataset(dataset: Dataset):
     return f"rg.{dataset.id}"
 
 
-def field_name_for_vector_settings(vector_settings: VectorSettings) -> str:
+def es_terms_query(field_name: str, values: List[str]) -> dict:
+    return {"terms": {field_name: values}}
+
+
+def es_range_query(field_name: str, gte: Optional[int] = None, lte: Optional[int] = None) -> dict:
```
jfcalvo: In `RangeFilter` we are using `float` instead of `int` for `ge` and `le`. Maybe you need to use `float` here as well?

```suggestion
def es_range_query(field_name: str, gte: Optional[float] = None, lte: Optional[float] = None) -> dict:
```

jfcalvo (changes requested):

src/argilla/server/search_engine/base.py
jfcalvo: I discussed this point with @frascuchon and we will use `RangeFilter`for `RatingQuestions` so by now this change is not necessary.

jfcalvo (approved):

PR #4325, opened 2023-11-27
fix: handling errors for non-existing endpoints

### LINKED ISSUE #4073: [BUG-python/deployment] backwards compatibility `src/argilla/client/sdk/v1/datasets/api.py` `get_metadata_properties`
**Describe the bug**
Whenever retrieving a dataset from server 1.17 with client 1.19 there seems to be an incompatibility issue when calling `response.json`. I think this might be a more recurrent issue, where for some reason the response does return with `response.status_code==200` even though the endpoint is not defined in FastAPI. Hence this might also occur with the semantic search update. 

https://github.com/argilla-io/argilla/blob/38c279984b007576880118e2ab8b66451bac60b5/src/argilla/client/sdk/v1/datasets/api.py#L423
Try loading a dataset from the 1.17 server with a 1.19 client.

**Expected behavior**

**Environment:**
 - Argilla Version [e.g. 1.0.0]: 1.19
 - ElasticSearch Version [e.g. 7.10.2]: N.A.
 - Docker Image (optional) [e.g. argilla:v1.0.0]: 1.17

**Additional context**
N.A.

frascuchon:
# Description

This PR adds a callback controller to return a proper status code and error message when a client requests a non-existing endpoint.

This change will help identify incompatibility problems when using new clients over old server instances as suggested in #4073 

Closes #4073

**Type of change**

**How Has This Been Tested**

Calling unexisting endpoints locally 

**Checklist**


src/argilla/server/routes.py
```diff
@@ -76,3 +76,8 @@
 api_router.include_router(users_v1.router, prefix="/v1")
 api_router.include_router(vectors_settings_v1.router, prefix="/v1")
 api_router.include_router(workspaces_v1.router, prefix="/v1")
+
+
+@api_router.route("/{full_path:path}")
```
davidberenstein1957: does this handle both GET POST etc?

src/argilla/server/routes.py
```diff
@@ -76,3 +76,8 @@
...
+def endpoint_not_found_controller(request: Request):
+    raise HTTPException(status_code=404, detail=f"Endpoint {request.url.path!r} not found")
```
davidberenstein1957, line 83: maybe we can urge mention potential server and client version misalignment?

jfcalvo, line 83: Maybe it's interesting to don't add this to swagger API schema.

jfcalvo (approved):

src/argilla/server/routes.py
frascuchon: Yes. Done. Thanks!

src/argilla/server/routes.py
frascuchon: Done. Using `include_in_schema=False`

frascuchon: This error should be processed on the client side. The server cannot assume that these kinds of errors are related to a misalignment between the server and the client.

davidberenstein1957 (approved):

Issue #4227, opened 2023-11-14
[FEATURE] Support filter and sort as part of the request body for search endpoints


frascuchon:
```[tasklist]
### Tasks
- [x] Define `filters` and `sort` pydantic schemas #4270
- [x] Change `POST /api/v1/me/datasets/:dataset_id/records/search`
- [x] Change `POST /apo/v1/datasets/:dataset_id/records/search`
- [ ] https://github.com/argilla-io/argilla/issues/4228
- [ ] https://github.com/argilla-io/argilla/issues/4229
- [ ] https://github.com/argilla-io/argilla/pull/4327
```

PR #4327, opened 2023-11-27
feat: add new filter support for search records endpoint

frascuchon:
# Description

This PR integrates the `POST /api/v1/dataset/:dataset_id/records/search`  with search engine to support `filter` and `sort` body attributes.

Also, the base filter classes are decorated as `dataclass`  in order to provide an easy way to compare args in tests (`dataclass` provide automatic `eq` method) 

Refs https://github.com/argilla-io/argilla/issues/4227

**Type of change**

- [] New feature (non-breaking change which adds functionality)

**Checklist**


src/argilla/server/schemas/v1/datasets.py
```diff
@@ -787,7 +787,7 @@ class Order(BaseModel):
 
 
 class SearchRecordsQuery(BaseModel):
-    query: Optional[Query]
+    query: Optional[Query] = PydanticField(default_factory=Query)
```
jfcalvo: Why is this change needed? Is `None` not a valid default value?

src/argilla/server/search_engine/base.py
```diff
@@ -71,53 +71,54 @@
...
 class AndFilter:
-    def __init__(self, filters: List["Filter"]):
-        self.filters = filters
+    filters: List["Filter"]
 
 
 Filter = Union[AndFilter, TermsFilter, RangeFilter]
 
 
+@dataclasses.dataclass
```
jfcalvo, line 121: You added this line but the class still has a `__init__` method. Did you miss to change this class?

jfcalvo (changes requested):

src/argilla/server/schemas/v1/datasets.py
frascuchon: It's true that `None` is a valid value. I reverted the change and refactored a bit how this object is parsed inside the endpoints handlers.

src/argilla/server/search_engine/base.py
frascuchon: Yes. removed! Thanks

jfcalvo (approved):

PR #4332, opened 2023-11-27
fix: preserve `TextClassificationSettings.label_schema` order

gabrielmbmb:
# Description

This PR fixes a bug where the order of the labels for a Text Classification dataset provided in the class `TextClassificationSettings` was not preserved. This was happening because the `labels_schema` attribute had the `Set[str]` type to ensure there is no duplicate labels, but `set` doesn't preserver the order.

Instead of using `set` to ensure there is no duplicates, `labels_schema` now has the `List[str]` type and a basic for loop has been added to ensure there is no duplicates.

Closes #3828

**Type of change**

- [x] Bug fix (non-breaking change which fixes an issue)

**How Has This Been Tested**

```python
import argilla as rg

rg.set_workspace("argilla")

settings = rg.TextClassificationSettings(label_schema=[
    "1 (extremely positive/supportive)",
    "2 (positive/supportive)",
    "3 (neutral)",
    "4 (hateful/unsupportive)",
    "5 (extremely hateful/unsupportive)",
    "6 (can't say!)",
    "6 (can't say!)",
    "6 (can't say!)",
])

rg.log(rg.TextClassificationRecord(text="blablabla"), name="test-order-labels")

rg.configure_dataset_settings(name="test-order-labels", settings=settings)
```

After that, go to the UI and check the labels appears in the provided order.

**Checklist**

- [x] I followed the style guidelines of this project
- [x] I did a self-review of my code
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I filled out [the contributor form](https://tally.so/r/n9XrxK) (see text above)
- [x] I have added relevant notes to the `CHANGELOG.md` file (See https://keepachangelog.com/)

PR #4347, opened 2023-11-29
feat: update `rg.{load,delete,list_datasets}` functions to work w/ `FeedbackDataset`s

### LINKED ISSUE #3915: [FEATURE] align `rg.*` default methods for old and new datasets

This feature has been requested by @dvsrepo so that we're able to list all the datasets for a given workspace via an intuitive method such as `list_datasets`.

The current approach is `rg.FeedbackDataset.list(workspace="argilla")` but that's to list just the `FeedbackDataset` datasets and may be not as intuitive as `Workspace(name="argilla").list_datasets()`.

Ideally we should add a `list_datasets` method to list all the datasets from a given workspace.

**Additional context**

We should clearly see how to combine both the `FeedbackDataset` with the previous datasets i.e. `DatasetForTextClassification`, `DatasetForTokenClassification` and `DatasetForText2Text`. And also evaluate whether we should be returning just the `name` and `type` of those or a class instance if working with `FeedbackDataset` datasets, etc.

gabrielmbmb:
# Description

This PR updates the `argilla.client.{load, delete, list_datasets}` to also work w/ `FeedbackDataset`. The changes made are the following:

- Move `ArgillaSingleton`, `init` and `active_client` to a new module called `singleton`. This was done to avoid circular imports caused when trying to import the `FeedbackDataset` class from the `argilla.client.api` module.
- Update `load` function to first try to load an old dataset, and if it fails try to load a `FeedbackDataset`.
- Update `delete` function to first try to delete an old dataset, and if it fails try to delete a `FeedbackDataset`. **BREAKING CHANGE**: `delete` function was not raising any kind of exception even if the dataset trying to delete didn't exist, this is not the case anymore.
- Update `list_datasets` function to also include `RemoteFeedbackDataset` in the returned list. In addition, an argument `type` has been added to filter by the kind of dataset.
- Refactor `argilla datasets list` command to use the updated `list_datasets` function as the logic was duplicate.
- Fix an error introduced in https://github.com/argilla-io/argilla/pull/4332 for which `label_schema` was a `list` of `dict`s instead of a `list` of `str`s
- Fix integration tests that were failing because the dataset was not being indexed in ElasticSearch because of usage of factories instead of proper dataset creation workflow.
- Fix the rest of integration tests that were failing in `develop`
- Pin to `setfit>=0.7.0,<1.0.0`

Closes #3915

**Type of change**

**How Has This Been Tested**

I've added tests covering the changes and additions.

**Checklist**


src/argilla/client/__init__.py
```diff
@@ -13,4 +13,4 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-from .api import active_client
+from .singleton import active_client
```
davidberenstein1957: maybe we want to use the full import path?

src/argilla/client/api.py
```diff
@@ -327,14 +236,21 @@ def load(
...
         workspace = workspace or argilla.get_workspace()
-        if _check_if_feedback_dataset_exists(client=argilla, name=name, workspace=workspace):
-            raise ValueError(
-                f"The dataset '{name}' exists but it is a `FeedbackDataset`. Use `rg.FeedbackDataset.from_argilla`"
-                " instead to load it."
-            )
-        raise e
+        try:
+            dataset = FeedbackDataset.from_argilla(name=name, workspace=workspace)
+        except ValueError:
```
davidberenstein1957, line 243: is it intended not to show the exception of this `ValueError`?

src/argilla/client/api.py
```diff
@@ -327,14 +236,21 @@ def load(
...
+            raise e
+
+        warnings.warn(
+            "Loaded dataset is a `FeedbackDataset`. It's recommended to use `rg.FeedbackDataset.from_argilla` class "
```
davidberenstein1957: can't we also dynamically pass the correct parameters here already?

src/argilla/client/api.py
```diff
@@ -360,18 +276,45 @@ def copy(dataset: str, name_of_copy: str, workspace: Optional[str] = None) -> No
...
+    except errors.ArApiResponseError as e:
+        try:
+            dataset = FeedbackDataset.from_argilla(name=name, workspace=workspace)
+        except ValueError:
+            raise e
+
+        dataset.delete()
+
+        warnings.warn(
+            "Removed dataset was a `FeedbackDataset`. It's recommended to use `rg.FeedbackDataset.delete` instead.",
```
davidberenstein1957: shouldn't this include the `from_argilla` methods and potentially params.

src/argilla/client/api.py
```diff
@@ -360,18 +276,45 @@ def copy(dataset: str, name_of_copy: str, workspace: Optional[str] = None) -> No
...
+    argilla = ArgillaSingleton.get()
+    workspace = workspace or argilla.get_workspace()
+    try:
+        # `delete` method is always successful, even if the dataset does not exist and that's why we need the extra
+        # call to `get_dataset` to check if the dataset exists. If it doesn't exist, then we try to delete a `FeedbackDataset`.
+        argilla.get_dataset(name=name, workspace=workspace)
+        argilla.delete(name=name, workspace=workspace)
+    except errors.ArApiResponseError as e:
+        try:
+            dataset = FeedbackDataset.from_argilla(name=name, workspace=workspace)
```
davidberenstein1957, line 307: don't we want to show the `ValueError`?

src/argilla/client/client.py
```diff
@@ -297,6 +297,10 @@ def copy(self, dataset: str, name_of_copy: str, workspace: Optional[str] = None)
             json_body=CopyDatasetRequest(name=name_of_copy, target_workspace=workspace),
         )
 
+    def get_dataset(self, name: str, workspace: Optional[str] = None) -> Dataset:
```
davidberenstein1957: docstring

tests/integration/client/feedback/test_unification.py
```diff
@@ -190,6 +190,7 @@ def test_multi_label_question_strategy_without_overlap(strategy, unified_respons
     records_payload = {
         "fields": {"text": "This is the first record", "label": "positive"},
         "responses": [
+            {"values": {question_name: {"value": ["label1"]}}},
```
davidberenstein1957, line 193: is this intended?

tests/integration/utils.py
```diff
@@ -0,0 +1,24 @@
...
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+from typing import Any
+
+from argilla.client.api import delete
+
+
+def delete_ignoring_errors(*args: Any, **kwargs: Any) -> None:
```
davidberenstein1957, line 20: What is this doing exactlyy, shouldn't we specific the `Exception` a bit better and pass a docstring about the usage?

gabrielmbmb: this is doing what the old `delete` was doing, which was to not return any exception. We were extensively using the `delete` as the first step in our unit tests, to remove datasets that could have beee not removed from previous unit tests executions. As `delete` now raises an exception, we cannot use it directly, and I had to create this function which tries to delete the dataset and if it doesn't exist, then just ignores the exception.

tests/integration/client/feedback/test_unification.py
gabrielmbmb: yes, otherwise the test would fail sometimes because it was not deterministic.

src/argilla/client/api.py
gabrielmbmb: yes, it is intended.

These functions were created to be used with the old datasets, adding the logic to be used with the `FeedbackDataset`s is secondary.

If that `ValueError` occurs, means there is no old dataset nor FeedbackDataset with that name and we raise the original exception to not introduce more breaking changes or confuse the users even more.

src/argilla/client/api.py
gabrielmbmb: ditto

PR #4397, opened 2023-12-11
fix: Fix `prepare_dataset_for_annotation_task` to return correct format

### LINKED ISSUE #4396: [BUG-python/deployment] `prepare_dataset_for_annotation_task` returns wrong format for NLTK metrics
**Describe the bug**
In `AgreementMetrics`, the alpha is calculated to be 0 because the format returned is not in line with what NLTK expects. Currently, it returns (user_id, question_name, value), where `question_name` is always the same, i.e. "label". But it should return the question id or the text itself so that each question should be differentiated from each other.

```python
ds_tiny = rg.FeedbackDataset.from_huggingface("argilla/go_emotions_raw", split="train[:10000]")
metric = AgreementMetric(dataset=ds_tiny, question_name="label")
metric.compute("alpha")
```
```bash
AgreementMetricResult(metric_name='alpha', count=35457, result=0.000)
```

**Expected behavior**
The function should return the correct calculation of the Krippendorf's Alpha.

**Environment:**
 - Argilla Version [e.g. 1.0.0]:

**Additional context**

kursathalat:
# Description

As described in the issue, the method returns question name for all records. In this PR, the question texts are returned so that the Alpha is correctly corrected. The method can be further improved with a better alternative.

Closes #4396 

**Type of change**

**How Has This Been Tested**

- [ ] Test A
- [ ] Test B

**Checklist**

### DIFF
--- a/src/argilla/client/feedback/metrics/agreement_metrics.py
+++ b/src/argilla/client/feedback/metrics/agreement_metrics.py
@@ -94,7 +94,10 @@ def prepare_dataset_for_annotation_task(
 
     formatted_responses: FormattedResponses = []
 
+    len_dataset = iter(range(len(hf_dataset[question_name])))
+
     for responses_ in hf_dataset[question_name]:
+        question_text = hf_dataset[next(len_dataset)]["text"]
         for response in responses_:
             user_id = response["user_id"]
             if user_id is None:
@@ -114,7 +117,7 @@ def prepare_dataset_for_annotation_task(
             elif question_type == MultiLabelQuestion:
                 value = frozenset(value)
 
-            formatted_responses.append((user_id, question_name, value))
+            formatted_responses.append((user_id, question_text, value))
 
     return formatted_responses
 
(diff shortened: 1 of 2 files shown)


plaguss (changes requested):
Thank you for tackling this @kursathalat! I think the function could be rewritten to make it more clear. For example in the loop:

```python
for responses_ in hf_dataset[question_name]:
     ...
```

Could be rewritten a bit:

```python
for row in hf_dataset:
    responses = row[question_name]
    question_text = row["text"]
```

Also, is this checked in the tests?

kursathalat:
> Thank you for tackling this @kursathalat! I think the function could be rewritten to make it more clear. For example in the loop:
> 
> ```python
> for responses_ in hf_dataset[question_name]:
>      ...
> ```
> 
> Could be rewritten a bit:
> 
> ```python
> for row in hf_dataset:
>     responses = row[question_name]
>     question_text = row["text"]
> ```
> 
> Also, is this checked in the tests?

Great, will be changing it. Currently, it is not checking the expected behavior in the test but will add some lines for that as well.

plaguss (approved):
Nice, thanks!

davidberenstein1957 (approved):

PR #4404, opened 2023-12-12
feat: add search engine reindexing cli task

### LINKED ISSUE #4335: [FEATURE] Add new argilla cli option to reindex all entities into search engine

Sometimes our search engine mapping is updated/changed and we don't have an easy way to allow a general reindex of previous created data on our database.

For example after a new version of Argilla is released and the search engine mapping is changed for some entities, users don't have an easy way to reindex all the old data to have support for searching with the new mappings.

A possible solution could be to create an Argilla cli option that reindex all the Argilla entities into the search engine.

Something similar to the following:

```sh
$ argilla server reindex
reindexing datasets...
reindexing records...
reindexing responses...
# Will reindex datasets, records, responses, suggestions, and the rest of entities etc.
```

Optionally we should show some way of progress or at least some messages showing which entities are reindexed.

Once that we have this Argilla cli option we can use it on our docker images so every time that the server is up we reindex or if that's too much we can add notes to the release saying that a manual reindex is necessary (in the case that we don't want to automatically do it).

jfcalvo:
# Description

This PR adds a new cli task to reindex datasets and records so we can use it once ElasticSearch/OpenSearch mappings are updated.

To execute the cli task use the following command:

```sh
$ argilla server search_engine reindex
```

This task will iterate over all the datasets in the database, reindexing them and doing the same for all records for each dataset.

> [!NOTE]
> We are using server-side cursors (streams) for fetch the collection of datasets and records in the cli task. For more information take a look to [SQLAlchemy documentation](https://docs.sqlalchemy.org/en/20/core/connections.html#engine-stream-results).

Closes #4335 

**Type of change**

**How Has This Been Tested**

- [x] Tested manually.
- [x] Tested manually using PostgreSQL.

**Checklist**


src/argilla/cli/server/search_engine/reindex.py
```diff
@@ -0,0 +1,113 @@
...
+        return (await db.execute(select(func.count(Dataset.id)))).scalar()
+
+    @classmethod
+    async def count_dataset_records(cls, db: AsyncSession, dataset: Dataset) -> int:
+        return (await db.execute(select(func.count(Record.id)).filter_by(dataset_id=dataset.id))).scalar()
+
+
+async def _reindex_datasets(db: AsyncSession, search_engine: SearchEngine, progress: Progress) -> None:
+    task = progress.add_task(
+        f"reindexing datasets...",
```
frascuchon: Here you can specify that only the Feedback datasets will be reindexed. Just to clarify to users

src/argilla/cli/server/search_engine/reindex.py
```diff
@@ -0,0 +1,113 @@
...
+                selectinload(Dataset.questions),
+                selectinload(Dataset.metadata_properties),
+                selectinload(Dataset.vectors_settings),
+            )
+            .execution_options(yield_per=cls.YIELD_PER)
+        )
+
+        async for dataset in stream.scalars():
+            await search_engine.delete_index(dataset)
+            await search_engine.create_index(dataset)
```
frascuchon, line 71: For "soft" changes we can use the update mappings and the update settings endpoints:

See docs:
- https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-put-mapping.html
- https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-update-settings.html

If we take care of mapping changes and we do not modify the field type, we could always update mappings without deleting the index.  See [here](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-put-mapping.html#updating-field-mappings)

tests/unit/cli/server/search_engine/test_reindex.py
```diff
@@ -0,0 +1,29 @@
...
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+import pytest
+from typer import Typer
+from typer.testing import CliRunner
+
+
+@pytest.mark.asyncio
```
frascuchon:
```suggestion
```

src/argilla/cli/server/search_engine/reindex.py
```diff
@@ -0,0 +1,115 @@
...
+async def _reindex_datasets(db: AsyncSession, search_engine: SearchEngine, progress: Progress) -> None:
+    task = progress.add_task(
+        f"reindexing datasets...",
+        total=await Reindexer.count_datasets(db),
+    )
+
+    async for dataset in Reindexer.reindex_datasets(db, search_engine):
+        await _reindex_dataset_records(db, search_engine, progress, dataset)
+
+        progress.advance(task)
```
gabrielmbmb, line 133: is this `progress.advance` needed?

jfcalvo: Yes, I think so. We will have two progress bars one for datasets and another one for records (per dataset).

src/argilla/cli/server/search_engine/reindex.py
jfcalvo: Done.

tests/unit/cli/server/search_engine/test_reindex.py
jfcalvo: Done.

src/argilla/cli/server/search_engine/reindex.py
gabrielmbmb: ahh, got it!

frascuchon:
It would be nice to add some references in docs about this new task. I think the better place should be [here](https://docs.argilla.io/en/latest/getting_started/installation/configurations/database_migrations.html#reindex-a-dataset)

jfcalvo:
> It would be nice to add some references in docs about this new task. I think the better place should be [here](https://docs.argilla.io/en/latest/getting_started/installation/configurations/database_migrations.html#reindex-a-dataset)

Ok, I will add a new section in the docs about this task.

jfcalvo:
> > It would be nice to add some references in docs about this new task. I think the better place should be [here](https://docs.argilla.io/en/latest/getting_started/installation/configurations/database_migrations.html#reindex-a-dataset)
> 
> Ok, I will add a new section in the docs about this task.

Done.

frascuchon (approved):

Issue #4377, opened 2023-12-05
[BUG-python/deployment] Potential issue with sorting with question name and metadata property name being the same


dvsrepo:
This needs reproduction from someone else but I have created a dataset where there's a question name (`score-chosen`) being the same as a metadata property and the sort functionality by the `score-chosen` metadata property seems to stop working.

1. I have checked the payloads and the backend seems to send an unordered list of records.
With name clash:
https://dvilasuero-ultrafeedback-curator.hf.space/dataset/1cf998cf-aebc-434a-9dd4-b883c3a7fb14/annotation-mode?_page=1&_status=pending&_sort=suggestion.score-chosen.score.desc

(test with sorting by metadata score-chosen)

2. I have created a new dataset changing the name of the metadata property to score and it seems to work well.

No name clash: https://dvilasuero-ultrafeedback-curator.hf.space/dataset/f1451716-d321-4fbd-b679-0f985cb70858/annotation-mode?_page=1&_status=pending&_sort=metadata.score.asc

This might or might not be the root cause but worth checking what's going on

jfcalvo:
I have been reproducing it locally with the following dataset (notice that I have a `rating-question` question and a metadata property using the same name):

```python
import argilla as rg

dataset = rg.FeedbackDataset(
  fields=[rg.TextField(name="text")],
  questions=[
    rg.RatingQuestion(name="rating-question", values=[1, 2, 3, 4]),
    rg.TextQuestion(name="text-question"),
    rg.LabelQuestion(name="label-question", labels=["one", "two"]),
    rg.MultiLabelQuestion(name="multi-label-question", labels=["one", "two", "three"]),
    rg.RankingQuestion(name="ranking-question", values=["ranking-1", "ranking-2", "ranking-3"]),
  ],
  metadata_properties=[
    rg.IntegerMetadataProperty(name="rating-question"),
  ],
)

dataset.add_records(
  records=[
    rg.FeedbackRecord(
      fields={"text": "record-1"},
      suggestions=[
        {"question_name": "rating-question", "value": 1},
        {"question_name": "text-question", "value": "suggestion-1",}
      ],
      metadata={
        "rating-question": 99,
      },
    ),
    rg.FeedbackRecord(
      fields={"text": "record-2"},
      suggestions=[
        {"question_name": "label-question", "value": "one"},
        {"question_name": "multi-label-question", "value": ["one", "two"]},
        {
          "question_name": "ranking-question",
          "value": [
[...]

PR #4411, opened 2023-12-14
🚀 Fix sort with same name

damianpumar:
Fix this issue https://github.com/argilla-io/argilla/issues/4377


jfcalvo:
I have tested it inside the QuickStart environment with the same dataset that used on #4377 and looks like working to me. Thanks!

leiyre (approved):
🙌

PR #4426, opened 2023-12-18
refactor: review security module

frascuchon:
# Description

In this PR, the `argilla.server.security` module is reviewed, by changing the `local` provider to `db` since is using db for validation. Also, the base `AuthProvider` methods are decorated as  `abstract`, and a new class method is used to create provider instances, to normalize the provider creation process.

**Type of change**

**How Has This Been Tested**

Test with a local deployment

**Checklist**


src/argilla/server/security/auth_provider/db/provider.py
```diff
@@ -52,36 +54,31 @@ async def login_for_access_token(
...
-        Parameters
-        ----------
-        username:
-            The user name
-        expires_delta:
-            Token expiration
+        if api_key:
+            user = await accounts.get_user_by_api_key(db, api_key)
+        elif token:
+            user = await self.fetch_token_user(db, token)
```
jfcalvo, line 76: To be a little bit more clear about the flow

```suggestion
        if api_key:
            user = await accounts.get_user_by_api_key(db, api_key)
        elif token:
            user = await self.fetch_token_user(db, token)
        else:
            raise UnauthorizedError()
```

With this you can delete the previous lines with:

```python
user = None
```

And also delete:

```python
if user is None:
    raise UnauthorizedError()
```

But still do the `return user` at the end.

jfcalvo (approved):
Looks good to me. I have added a small suggestion in the case that you think it's a good idea to do the change.

src/argilla/server/security/auth_provider/db/provider.py
frascuchon: Not really, both methods `accounts.get_user_by_api_key` and `self.fetch_token_user` can return `None` if the user is not found by provided info. We should refactor first those methods to apply your suggestion. I will skip for the moment.

PR #4427, opened 2023-12-18
feat: python-rq integration using datasets reindex as proof of concept

jfcalvo:
# Description

This PR include changes as a proof of concept to check how to integrate `rq` background processor with Argilla.

The changes include also two new endpoints:
* `PUT /api/v1/datasets/:dataset_id/reindex`
  * This endpoint will return a HTTP `202 (Accepted)` status.
  * A background job will be enqueue to reindex the dataset.
  * The response body will include the `id` of the job and its `status` (`queued` in this case if everything was fine).
  * Users can use the `id` of the job to get information about what is the status of the job.  
* `GET /api/v1/jobs/:job_id`
  * This endpoint is used to obtain information about one specific job (returning the `id` and `status`).
  * Jobs are right now not stored on database and I'm using `rq` API to get information about its status.
  * `rq` is saving job information for `500` seconds on Redis, so after a job is finished or failed the user has 500 seconds to get information about it.

Posible improvements:
* <del>Define a proper Redis connection using a pool of connections and getting settings from environment variables.</del> Redis is using a pool of connections by default and I have added a new environment variable to set the connection (`ARGILLA_REDIS_URL`).
* <del>Define a better way to store our jobs, maybe using a new `jobs` table on Argilla database and allowing to save results of the jobs there.</del> We will start with this approach of using `rq` results stored in Redis and in the future for more complex flows we will think into adding some data if necessary to our database.
* <del>Define a `rq` queue only for search engine purposes.</del> We will use `default` queue for now.
* <del>Once `Reindexer` class code is merged from PR adding reindex cli task we can remove it from the code in this PR.</del> We already merge the PR adding the reindex cli task and now the jobs are importing it and using it.
* Add a `result` field to `Job` schema so we can include the result of the job inside it. (Useful to know if there are errors or additional information about the process)

Things to investigate/discuss:
* How to use Redis on our docker images, specially with QuickStart images on HF.
* Alternatives to use Redis using `fakeredis` python library instead.

**Type of change**

- [ ] Documentation update

**How Has This Been Tested**

- [ ] Test A
- [ ] Test B

**Checklist**


src/argilla/server/jobs/search_engine_jobs.py
```diff
@@ -0,0 +1,37 @@
...
+
+# NOTE:
+# This could be executed as a normal function:
+# > await reindex_dataset(UUID("cfcc028a-1669-4b6a-baf5-9b50e0538e44"))
+# Or with rq in background:
+# > reindex_dataset.delay(UUID("cfcc028a-1669-4b6a-baf5-9b50e0538e44"))
+@job(default_queue, retry=Retry(max=3, interval=60))
+async def reindex_dataset(dataset_id: UUID):
+    async with AsyncSessionLocal() as db:
+        async for search_engine in get_search_engine():
```
gabrielmbmb:
```suggestion
    async with AsyncSessionLocal() as db:
        async with SearchEngine.get_by_name(settings.search_engine) as engine:
```

It's better to use the `SearchEngine` context manager directly than `get_search_engine` function which is intended for use with `fastapi.Depends`

gabrielmbmb: we could also combine the two lines above into one:

```python
async with AsyncSessionLocal() as db, SearchEngine.get_by_name(settings.search_engine) as engine:
    ...
```

src/argilla/server/settings.py
```diff
@@ -73,6 +73,9 @@ class Settings(BaseSettings):
     elasticsearch_ca_path: Optional[str] = None
     cors_origins: List[str] = ["*"]
 
+    redis_host: str = "localhost"
+    redis_port: int = 6379
```
gabrielmbmb: Here we could use `pydantic.RedisDsn` type

docker/quickstart.requirements.txt
```diff
@@ -1 +1,2 @@
 datasets
+rq == 1.15.1
```
gabrielmbmb, line 2: We can remove this from here.

We have this requirements file just to install the `datasets` required by the `docker/scripts/load_data.py` of the Docker quickstart image.

gabrielmbmb (changes requested):

src/argilla/server/jobs/search_engine_jobs.py
jfcalvo: Done!

src/argilla/server/settings.py
jfcalvo: I have found the `RedisDsn` to be not really flexible if we want to specify attributes to the Redis URL so instead I'm using a `str` `redis_url` settings attribute where we can set the entire URL to connect to.

docker/quickstart.requirements.txt
jfcalvo: As far as I understand these are the requirements of the image so I don't see any problem adding an additional requirement. Remember that we need `rq` to run the worker that will consume the enqueued jobs (as a different process).

PR #4450, opened 2023-12-21
feat: relax `pydantic` constraint

### LINKED ISSUE #3369: [FEATURE] Upgrade to `pydantic>=2.0.0`
Pydantic V2 has been released recently. It brings with a lot of changes in its core, mainly that it has been rewritten to Rust which makes it much much faster (between 5x - 50x faster).

We're using `pydantic.BaseModel` pretty much everywhere... The `FeedbackRecord` is a `pydantic.BaseModel`. One `FeedbackDataset` can have more than 1000 records easily, so I think we would be able to see some performance improvement once we update to `pydantic>=2.0`

Thing to take into consideration:

1. Many methods names of the `BaseModel` class have been changed. There is a v1 to v2 migration guide https://docs.pydantic.dev/latest/migration/ and also a tool https://github.com/pydantic/bump-pydantic to help with it.
2. If we update to `pydantic>=2.0`, then we need to update the API to use `fastapi>=0.100.0` which is the release that comes with `pydantic>=2.0` support.

gabrielmbmb:
# Description

This PR relax the `pydantic` constraint so `pydantic>=2.0.0` works also. In order to make `argilla` work with both pydantic versions:

1. Create `argilla.utils.pydantic` module that tries to import everything from `pydantic.v1`. If it fails imports everything from `pydantic`.
2. Update all the imports `from pydantic import ...` -> `from argilla.utils.pydantic import ...`
3. Relax `spacy` constraint that hardly depended on `pydantic<2.0.0`

We would still need to update the code to use the V2, but with this we allow users to have `pydantic>=2.0.0` and `argilla` installed in the same virtual environment.

Closes #3369

**Type of change**

**How Has This Been Tested**

- [ ] Test A
- [ ] Test B

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ These are the section headers that we use:
 - Updated `argilla.load` functions to also work with `FeedbackDataset`s. ([#4347](https://github.com/argilla-io/argilla/pull/4347))
 - [breaking] Updated `argilla.delete` functions to also work with `FeedbackDataset`s. It now raises an error if the dataset does not exist. ([#4347](https://github.com/argilla-io/argilla/pull/4347))
 - Updated `argilla.list_datasets` functions to also work with `FeedbackDataset`s. ([#4347](https://github.com/argilla-io/argilla/pull/4347))
+- Updated `pydantic` dependency constraint to `>=2.0.0`. ([#4450](https://github.com/argilla-io/argilla/pull/4450))
 
 ### Fixed
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,7 @@ dependencies = [
     "packaging >= 20.0",
     # pandas -> For data loading
     "pandas >=1.0.0,<2.0.0",
-    "pydantic >=2.0.0",
+    "pydantic >=1.10.7",
     # monitoring
     "wrapt >= 1.13,< 1.15",
     # weaksupervision
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,8 +28,7 @@ dependencies = [
     "packaging >= 20.0",
     # pandas -> For data loading
     "pandas >=1.0.0,<2.0.0",
-    # Aligned pydantic version with server fastAPI
-    "pydantic >= 1.10.7,< 2.0",
+    "pydantic >=2.0.0",
     # monitoring
     "wrapt >= 1.13,< 1.15",
     # weaksupervision
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -97,7 +97,7 @@ integrations = [
     "pgmpy",
     "plotly >= 4.1.0",
     "snorkel >= 0.9.7",
-    "spacy == 3.5.3",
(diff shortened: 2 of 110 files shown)


pyproject.toml
```diff
@@ -28,8 +28,7 @@ dependencies = [
     "packaging >= 20.0",
     # pandas -> For data loading
     "pandas >=1.0.0,<2.0.0",
-    # Aligned pydantic version with server fastAPI
-    "pydantic >= 1.10.7,< 2.0",
+    "pydantic >=2.0.0",
```
frascuchon: We should also support users using pydantic < 2.0. So, these changes won't be compatible with pydantic 1.x users, right?

gabrielmbmb: yes, that's true. With my latest change `argilla` should be compatible with both versions.

src/argilla/cli/server/database/users/create.py
```diff
@@ -27,6 +26,7 @@
     UserCreate,
     WorkspaceCreate,
 )
+from argilla.utils.pydantic import constr
```
frascuchon: Since this `utils.pydantic` is used just to contro how pydantic is imported, I would propose a different naming. `argilla.imports.pydantic` or just `argilla._import_pydantic`

src/argilla/server/apis/v0/handlers/datasets.py
```diff
@@ -28,6 +27,7 @@
 from argilla.server.schemas.v0.datasets import CopyDatasetRequest, CreateDatasetRequest, Dataset, UpdateDatasetRequest
 from argilla.server.security import auth
 from argilla.server.services.datasets import DatasetsService
+from argilla.utils.pydantic import parse_obj_as
```
frascuchon: Even if the client and server are sharing this new behavior, I would duplicate the new pydantic.py module for both, client and server separately, so in the future we can separate both packages

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ These are the section headers that we use:
 - Updated `argilla.load` functions to also work with `FeedbackDataset`s. ([#4347](https://github.com/argilla-io/argilla/pull/4347))
 - [breaking] Updated `argilla.delete` functions to also work with `FeedbackDataset`s. It now raises an error if the dataset does not exist. ([#4347](https://github.com/argilla-io/argilla/pull/4347))
 - Updated `argilla.list_datasets` functions to also work with `FeedbackDataset`s. ([#4347](https://github.com/argilla-io/argilla/pull/4347))
+- Updated `pydantic` dependency constraint to `>=2.0.0`. ([#4450](https://github.com/argilla-io/argilla/pull/4450))
```
davidberenstein1957, line 33: @gabrielmbmb I believe this should be rephrased a bit. Besides that, it LGTM.

frascuchon:
The final work has been tackled at https://github.com/argilla-io/argilla/pull/4459. This PR can be closed once the new one is approved and merged

PR #4451, opened 2023-12-22
perf: improve upsert responses bulk

### LINKED ISSUE #4449: [FEATURE] Improve response upsert bulk endpoint performance
The new bulk endpoint for response creation takes a long time to process the request (around 6 seconds for 100 responses). This may affect the global server performance. The bottleneck must be identified and fixed as soon as possible.

cc @jfcalvo

frascuchon:
# Description

This PR applies some performance improvements related to the bulk upsert responses endpoint since some environments are experimenting with a high execution time when running this operation with 100 items in the request.

The performance improvements are:

1. Read all requested records at once instead of once at a time.
2. Stop checking the existence of the search index since is taking longer.

These improvements are applied based on the profiling results using the test ...

Also, since the current API controller is doing many things, a new design proposal is applied in the implementation by defining a simple use case implementation for this functionality. The use case wraps all the needed steps to cover the functionality and make the controller implementation quite simple, focusing only on exception handling and request/response serialization.

Closes #4449 

**Type of change**

**How Has This Been Tested**

Test locally

**Checklist**

### DIFF
--- a/src/argilla/server/apis/v1/handlers/responses.py
+++ b/src/argilla/server/apis/v1/handlers/responses.py
@@ -17,23 +17,23 @@ from uuid import UUID
 from fastapi import APIRouter, Depends, HTTPException, Security, status
 from sqlalchemy.ext.asyncio import AsyncSession
 
-import argilla.server.errors.future as errors
 from argilla.server.contexts import datasets
 from argilla.server.database import get_async_db
-from argilla.server.models import Record, Response, User
-from argilla.server.policies import RecordPolicyV1, ResponsePolicyV1, authorize
+from argilla.server.errors.future import NotFoundError
+from argilla.server.models import Response, User
+from argilla.server.policies import ResponsePolicyV1, authorize
 from argilla.server.schemas.v1.responses import (
     Response as ResponseSchema,
-)
-from argilla.server.schemas.v1.responses import (
-    ResponseBulk,
-    ResponseBulkError,
+    ResponseUpdate,
     ResponsesBulk,
     ResponsesBulkCreate,
-    ResponseUpdate,
 )
 from argilla.server.search_engine import SearchEngine, get_search_engine
 from argilla.server.security import auth
+from argilla.server.use_cases.responses.upsert_responses_in_bulk import (
+    UpsertResponsesInBulkUseCase,
+    UpsertResponsesInBulkUseCaseFactory,
+)
 
 router = APIRouter(tags=["responses"])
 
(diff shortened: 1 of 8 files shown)


damianpumar:
Great work @frascuchon discarding 100 records.
Screenshot 2023-12-22 at 23 39 41

src/argilla/server/apis/v1/handlers/responses.py
```diff
@@ -49,36 +51,19 @@ async def _get_response(db: AsyncSession, response_id: UUID) -> Response:
...
-        else:
-            responses_bulk_items.append(ResponseBulk(item=ResponseSchema.from_orm(response), error=None))
-
-    return ResponsesBulk(items=responses_bulk_items)
+    try:
+        responses_bulk_items = await use_case.execute(body.items, user=current_user)
+    except NotFoundError as err:
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=str(err))
+    else:
+        return ResponsesBulk(items=responses_bulk_items)
```
gabrielmbmb, line 66:
```suggestion
return ResponsesBulk(items=responses_bulk_items)
```
I'm an else hater, feel free to apply or discard

src/argilla/server/use_cases/responses/upsert_responses_in_bulk.py
```diff
@@ -0,0 +1,60 @@
...
+            else:
+                responses_bulk_items.append(ResponseBulk(item=Response.from_orm(response), error=None))
+
+        return responses_bulk_items
+
+
+class UpsertResponsesInBulkUseCaseFactory:
+    def __call__(
+        self, db: AsyncSession = Depends(get_async_db), search_engine: SearchEngine = Depends(get_search_engine)
+    ):
```
gabrielmbmb, line 59: I like this pattern for the cloud.

src/argilla/server/use_cases/responses/upsert_responses_in_bulk.py
```diff
@@ -0,0 +1,60 @@
...
+        self.db = db
+        self.search_engine = search_engine
+
+    async def execute(self, responses: List[ResponseUpsert], user: User) -> List[ResponseBulk]:
+        responses_bulk_items = []
+
+        all_records = await datasets.get_records_by_ids(self.db, [item.record_id for item in responses])
+        non_empty_records = [r for r in all_records if r is not None]
+
+        await datasets.preload_records_relationships_before_validate(self.db, non_empty_records)
```
gabrielmbmb, line 40: In which cases the endpoint will receive responses for records of different datasets? Maybe we're complicating this endpoint logic assuming we will receive responses for records of different datasets and the endpoint could be `/api/v1/me/datasets/{dataset_id}/responses/bulk`.

gabrielmbmb: The preload query would be much lighter as we would only need to load one dataset and its questions with a simple select query.

src/argilla/server/contexts/datasets.py
```diff
@@ -821,6 +826,16 @@ async def _preload_record_relationships_before_index(db: "AsyncSession", record:
...
 
 
+async def preload_records_relationships_before_validate(db: "AsyncSession", records: List[Record]) -> None:
+    await db.execute(
+        select(Record)
+        .filter(Record.id.in_([record.id for record in records]))
+        .options(
+            selectinload(Record.dataset).selectinload(Dataset.questions),
+        )
+    )
```
gabrielmbmb, line 836: I think this could be faster getting the unique datasets ids from the list of provided records, and the selecting the datasets instead of the records.

gabrielmbmb (changes requested):

src/argilla/server/use_cases/responses/upsert_responses_in_bulk.py
frascuchon: I see your point, but ideally, endpoints should be defined in a more REST definition approach. Defining a bulk endpoint under `/api/v1/.../responses` makes sense. We could limit response processing to one dataset on each request. Anyway, changes in this PR are improving the endpoint performance x2. So, I would say to review these improvements later.

src/argilla/server/contexts/datasets.py
frascuchon: We already reduce the time consumption of the endpoint x2. I would say not to apply more optimization on fetching data. It would be interesting to improve the time consumption on the update part. But this can be tackled in future work.

Screenshot 2024-01-08 at 11 45 34

src/argilla/server/apis/v1/handlers/responses.py
frascuchon: I think the same, but this case makes clearer the `else` case when nothing unexpected occurs. I will keep it as is.

gabrielmbmb (approved):

PR #4459, opened 2024-01-04
feat: Allow support to pydantic v2 dependency

### LINKED ISSUE #3369: [FEATURE] Upgrade to `pydantic>=2.0.0`

### LINKED ISSUE #4466: [FEATURE] Add support using pydantic v2 with Argilla
Argilla users cannot work with pydantic v2, since dependency requirements are fixed by Argilla. We can adapt imports to support both, v1 and v2 versions, without a full v2 migration (proposed in https://github.com/argilla-io/argilla/issues/3369).

frascuchon:
# Description

This PR continues the work started by @gabrielmbmb in #4450 (thanks a lot for your insights!) and allows pydantic v2 support when using the argilla client. Users running the Argilla server in the same environment should experience compatibility problems with pydantic v2, since the server side is not fully compatible with the new pydantic major version.

Some considerations about this PR regarding the compatibility changes:

- `root_validator` decorators must always include the `skip_on_failure` attribute.
- `constr` pydantic types with regex constraints are not supported and must be replaced by standard pydantic `Field`
-  The `const` `Field` attribute is not supported anymore. We've just removed it, but a `Literal` should be used instead

In this PR, the similar approach used by @gabrielmbmb in #4450 is used, the mirror `pydantic_v1` modules define the same import structure - instead of importing everything in the [base module](https://github.com/argilla-io/argilla/pull/4450/files#diff-ebbde3234094c2d752047255ac4f8f423767fcd58e749fdd2ba1f3ce390130ccR17). This change is motivated trying to remark how the original pydantic module should be imported.

Since server-related - `unit/server` and majority of `integration` -  tests are not compatible when pydantic v2 is installed, some changes have been also included to skip them when v2 is detected.

In future works, the integration tests would be changed to running using a real argilla server connection.

Refs https://github.com/argilla-io/argilla/issues/3369

**Type of change**

**How Has This Been Tested**

Running some client code examples using pydantic v2 

**Checklist**

### DIFF
--- a/.github/workflows/end2end-examples.yml
+++ b/.github/workflows/end2end-examples.yml
@@ -85,7 +85,7 @@ jobs:
         env:
           ARGILLA_ENABLE_TELEMETRY: 0
         run: |
-          pip install -e .
+          pip install -e ".[server]"
           python -m argilla server database migrate
           python -m argilla server database users create_default
           uvicorn argilla.server.app:app --port 6900 --host 0.0.0.0 &
--- a/.github/workflows/run-python-tests.yml
+++ b/.github/workflows/run-python-tests.yml
@@ -95,6 +95,12 @@ jobs:
         run: |
           pip install -e ".[server,listeners]"
           pytest --cov=argilla --cov-report=xml:${{ env.COVERAGE_REPORT }}.xml ${{ inputs.pytestArgs }} -vs
+      - name: Run tests with pydantic v2 📈
+        env:
+          ARGILLA_ENABLE_TELEMETRY: 0
+        run: |
+          pip install "pydantic~=2.0"
+          pytest ${{ inputs.pytestArgs }} -vs
       - name: Upload coverage report artifact
         uses: actions/upload-artifact@v3
         with:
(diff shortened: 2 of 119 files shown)


gabrielmbmb (approved):
LGTM! Thanks @frascuchon ❤️

PR #4477, opened 2024-01-10
feature interpret metadata entries with nonenan as non existent instead of returning an error

### LINKED ISSUE #4300: [FEATURE] interpret metadata entries with `None`/`nan` as non-existent instead of returning an error
At the moment I've seen some users struggle with this because it is not very clear we should remove the entire key.

I think we can do something like that. 
```python
filtered_dict = {key: value for key, value in my_dict.items() if value is not None and not (isinstance(value, float) and math.isnan(value))}
```

N.A.

**Additional context**
N.A.

frascuchon:
# Description

This PR resolves problems working with `None` and `nan` values in metadata property values. 

- The `nan` values are ignored when detected on the server side
- The `None` values reset the metadata property value. So, if users want to reset the `prop`metadata property, they can set it up with a `None` value.

## UPDATE

New changes have been introduced to improve API behaviour:

- `Nan` values will raise a 422 error, which makes sense since is not a valid value.
- `None` values are accepted as valid values. But the behaviour can be equivalent to avoid passing the metadata value for that property since the metadata whole data is replaced/overwritten on a record update.

Saying this, the client will parse metadata values to detect `nan` and raise a proper. The `None` ones are fully accepted.

cc @davidberenstein1957 @sdiazlor @jfcalvo 

Closes #4300 

**Type of change**

**How Has This Been Tested**

This feature has been tested locally using the Python SDK.

**Checklist**


CHANGELOG.md
```diff
@@ -19,23 +19,24 @@ These are the section headers that we use:
 ### Added
 
 - Restore filters from feedback dataset settings ([#4461])(https://github.com/argilla-io/argilla/pull/4461)
-- Warning on feedback dataset settings when leaving page with unsaved changes ([#4461])(https://github.com/argilla-io/argilla/pull/4461)
-- Added pydantic v2 support using the python SDK ([#4459](https://github.com/argilla-io/argilla/pull/4459))
+- Warning on feedback dataset settings when leaving page with unsaved changes ([#4461](https://github.com/argilla-io/argilla/pull/4461))
```
jfcalvo:
```suggestion
- Warning on feedback dataset settings when leaving page with unsaved changes. ([#4461](https://github.com/argilla-io/argilla/pull/4461))
```

src/argilla/client/feedback/schemas/metadata.py
```diff
@@ -161,21 +164,33 @@ def _all_values_exist(self, introduced_value: Optional[Union[str, List[str]]] =
...
+        if value is not None:
+            if isinstance(value, str):
+                return value
+            if isinstance(value, list):
+                return [self._check_allowed_value_type(v) for v in value]
+            raise TypeError(
+                f"Provided '{self.name}={value}' of type {type(value)} is not valid, "
+                "only values of type `str` or `list` are allowed."
+            )
+        return value
```
jfcalvo: A possible simplification/refactor:

```suggestion
    def _check_allowed_value_type(self, value: Any) -> Any:
        if value is None or isinstance(value, str):
            return value
        
        if isinstance(value, list):
            return [self._check_allowed_value_type(v) for v in value]
        
        raise TypeError(
            f"Provided '{self.name}={value}' of type {type(value)} is not valid, "
            "only values of type `str` or `list` are allowed."
        )
```

src/argilla/client/feedback/schemas/metadata.py
```diff
@@ -318,6 +347,23 @@ class FloatMetadataProperty(_NumericMetadataPropertySchema):
...
 
+    def _check_allowed_value_type(self, value: Any) -> Any:
+        if value is not None:
+            if isinstance(value, (int, float)):
+                return value
+            raise TypeError(
+                f"Provided '{self.name}={value}' of type {type(value)} is not valid, "
+                "only values of type `int` or `float` are allowed."
+            )
+        return value
```
jfcalvo: Another similar suggestion:

```suggestion
    def _check_allowed_value_type(self, value: Any) -> Any:
        if value is None or isinstance(value, (int, float)):
            return value
            
        raise TypeError(
            f"Provided '{self.name}={value}' of type {type(value)} is not valid, "
            "only values of type `int` or `float` are allowed."
        )
```

src/argilla/server/schemas/v1/datasets.py
```diff
@@ -524,6 +524,13 @@ def check_user_id_is_unique(cls, values: Optional[List[UserResponseCreate]]) ->
 
         return values
 
+    @validator("metadata", pre=True)
+    def skip_nan_values(cls, metadata: Optional[Dict[str, Any]]) -> Optional[Dict[str, Any]]:
+        if metadata is None:
+            return metadata
+
+        return {k: v for k, v in metadata.items() if v == v}  # By definition, NaN != NaN
```
jfcalvo: It is possible to use `math.isnan(value)` to do the check more explicit than `v == v` but I'm not sure if it's really useful because the function raises a `TypeError` exception if it's not a number and here you're checking for other types as well.

davidberenstein1957:
@frascuchon, what do you mean by reset? I am unsure of the behaviour of `None` and `np.nan`. I would maybe go for ignoring `np.nan` and using `None`as allowed `TermMetadata`. But I might also opt for aligning the behaviour where bot scenarios reset (assuming the key-value pair is overwritten and removed).

Not sure what @sdiazlor thinks of this and what the reasoning of @frascuchon was behind the differentiation.

jfcalvo (changes requested):

tests/integration/client/feedback/dataset/remote/test_dataset.py
```diff
@@ -841,3 +842,51 @@ async def test_warning_local_methods(self, role: UserRole) -> None:
...
+                metadata={"float-metadata": float("nan")},
+            )
+            for i in range(1, 20)
+        ]
+        remote_dataset.add_records(records)
+
+        assert len(remote_dataset.records) == len(records)
+        assert all("float-metadata" not in record.metadata for record in remote_dataset.records)
+
+    async def test_add_records_with_metadata_including_none_values(
```
davidberenstein1957, line 866: How would the `None` `np.nan` difference work within the UI. Also, what happens if we upload `np.nan` to a `TermsMetadataProperty` or `IntMetadataProperty`.

src/argilla/server/schemas/v1/records.py
```diff
@@ -51,3 +51,11 @@ class RecordUpdate(UpdateSchema):
...
     suggestions: Optional[List[SuggestionCreate]] = None
     vectors: Optional[Dict[str, List[float]]]
+
+    @validator("metadata_", pre=True)
+    @classmethod
+    def skip_nan_values(cls, metadata: Optional[Dict[str, Any]]) -> Optional[Dict[str, Any]]:
+        if metadata is None:
+            return metadata
+
+        return {k: v for k, v in metadata.items() if v == v}  # By definition, NaN != NaN
```
davidberenstein1957, line 65: Not that the comment from @jfcalvo applies here too https://github.com/argilla-io/argilla/pull/4477/files#r1448542743

tests/integration/client/feedback/dataset/remote/test_dataset.py
```diff
@@ -18,6 +18,7 @@
```
davidberenstein1957, line 18: Maybe I'm missing some tests for filtering and sorting.

src/argilla/server/schemas/v1/datasets.py
frascuchon: `math.isnan(value)` will fail with string types and `None`. So we need to add extra conditions.

tests/integration/client/feedback/dataset/remote/test_dataset.py
frascuchon: `np.nan` will be ignored. `None` values will be accepted as `None`. I would not assume a similar treatment between `nan` and `None` values. This is the reason of the difference

src/argilla/server/schemas/v1/records.py
frascuchon: ditto

frascuchon:
> @frascuchon, what do you mean by reset? I am unsure of the behaviour of `None` and `np.nan`. I would maybe go for ignoring `np.nan` and using `None`as allowed `TermMetadata`. But I might also opt for aligning the behaviour where bot scenarios reset (assuming the key-value pair is overwritten and removed).
> 
> Not sure what @sdiazlor thinks of this and what the reasoning of @frascuchon was behind the differentiation.

By reset I mean set the value to `None`. And I would use reset only with `None` values. In my opinion, nan values can be a result of an unexpected argument in a math operation, while a `None` value is the absence of a value. This is the reason behind the differentiation.

sdiazlor (approved):
It looks good.  And the reasoning behind treat `nan`and `None` differently it makes sense now to me. Maybe the option of using `None` to reset should be also documented in any file.

davidberenstein1957:
> > @frascuchon, what do you mean by reset? I am unsure of the behaviour of `None` and `np.nan`. I would maybe go for ignoring `np.nan` and using `None`as allowed `TermMetadata`. But I might also opt for aligning the behaviour where bot scenarios reset (assuming the key-value pair is overwritten and removed).
> > Not sure what @sdiazlor thinks of this and what the reasoning of @frascuchon was behind the differentiation.
> 
> By reset I mean set the value to `None`. And I would use reset only with `None` values. In my opinion, nan values can be a result of an unexpected argument in a math operation, while a `None` value is the absence of a value. This is the reason behind the differentiation.

@frascuchon In that case, we might also want to allow filtering on `None` values. Kind of aligns with the mention of the missing tests for checking this behaviour when combined with filters and sorts. Not sure what your thoughts are on this.

This can of course be covered in a separate PR/issue because it will also require UI updates.

frascuchon:
> > > @frascuchon, what do you mean by reset? I am unsure of the behaviour of `None` and `np.nan`. I would maybe go for ignoring `np.nan` and using `None`as allowed `TermMetadata`. But I might also opt for aligning the behaviour where bot scenarios reset (assuming the key-value pair is overwritten and removed).
> > > Not sure what @sdiazlor thinks of this and what the reasoning of @frascuchon was behind the differentiation.
> > 
> > 
> > By reset I mean set the value to `None`. And I would use reset only with `None` values. In my opinion, nan values can be a result of an unexpected argument in a math operation, while a `None` value is the absence of a value. This is the reason behind the differentiation.
> 
> @frascuchon In that case, we might also want to allow filtering on `None` values. Kind of aligns with the mention of the missing tests for checking this behaviour when combined with filters and sorts. Not sure what your thoughts are on this.
> 
> This can of course be covered in a separate PR/issue because it will also require UI updates.

Okay, I got it. The current search endpoint does not allow queries by `nil`/ `null` / `None` values on a metadata property. This should be properly defined in the query DSL to support this kind of "missing" value query. And this is out of the scope of this PR.

davidberenstein1957:
@frascuchon will you create a follow up issue for the search and query support?

jfcalvo (approved):

frascuchon:
> @frascuchon will you create a follow up issue for the search and query support?

Done https://github.com/argilla-io/argilla/issues/4486

PR #4505, opened 2024-01-18
🐞 fix bulk annotation similarity pagination

damianpumar:


damianpumar:
@frascuchon could you try it and valida if it's working ?
I would like to merge it.

Thanks

leiyre (approved):

PR #4509, opened 2024-01-19
Update login.py - ARGILLA_CACHE_DIR environment variable

7flash:

I think sometimes it's important to be able provide custom path for cache directory, particularly in my case since it's running in runpod container with limited container storage, I must provide a path to cache directory in mounted volume, otherwise when running `argilla login` I get error `OSError: [Errno 28] No space left on device: '/root/.cache/argilla'`


frascuchon (commented):
Thanks for this, @7flash.

I'm missing some tests checking this behaviour. I include @gabrielmbmb in the review since he has more context on this feature.

Thanks again!

gabrielmbmb (changes requested):
LGTM! As @frascuchon commented, it would be good to have a simple unit test.

Issue #4510, opened 2024-01-19
[BUG-python/deployment] ArgillaTrainer.__init__() got an unexpected keyword argument 'tokenizer'


7flash:
Following this tutorial:

https://docs.argilla.io/en/latest/tutorials_and_integrations/tutorials/feedback/training-llm-mistral-sft.html#ArgillaTrainer-&-hyperparameters

I am trying to initialize ArgillaTrainer passing tokenizer as an argument:

```
from argilla.feedback import ArgillaTrainer

trainer = ArgillaTrainer(
    dataset=dataset,
    model=model,
    tokenizer=tokenizer,
    task=task,
    framework="trl",
    train_size=0.99,
)
```

It gives me an error:

```
ArgillaTrainer.__init__() got an unexpected keyword argument 'tokenizer'
```

It seems like ArgillaTrainer does not expect tokenizer argument and tutorial might be outdated. Please can you suggest the correct way to proceed on this step following a tutorial?

7flash:
Actually tokenizer argument was added in v1.16.0 accordingly to changelog

https://github.com/argilla-io/argilla/blob/4e02c3cb931e2edeafc470bc03bd2e1af3c36efb/CHANGELOG.md#1160

And seems to be present in develop branch

https://github.com/argilla-io/argilla/blob/4e02c3cb931e2edeafc470bc03bd2e1af3c36efb/src/argilla/client/feedback/training/base.py#L46

Although I am on version 1.16.0

```
pip show argilla
Name: argilla
Version: 1.16.0
```

I am still getting that error

7flash:
Updating to 1.17.0 has resolved the issue

```
pip install --upgrade argilla==1.17.0
```

PR #4511, opened 2024-01-19
Update training-llm-mistral-sft.ipynb - upgrade argilla to 1.17.0 #4510

7flash:


davidberenstein1957:
@7flash , Thanks for the contribution. Why did you bump this version? Also, would you be able to update the changelog.md so we can reference you as outlined here :) https://docs.argilla.io/en/latest/community/contributing.html

PR #4541, opened 2024-01-29
feat: include new HF login page

leiyre:


frascuchon:
Fixed in https://github.com/argilla-io/argilla/pull/4547

PR #4562, opened 2024-02-01
feat: add more telemetry metrics

frascuchon:
# Description

Improve telemetry:
- Register new `UserCreated`  event
- Add user id to the `login` event
- Include extra system-related attributes such us running on docker flag, deployment type (`quickstart` vs `server`)
 
**Type of change**

**How Has This Been Tested**

This feature has been tested locally using different types of deployments. Tests related to system-related attributes are not implemented due extra complexity to reproduce proper isolated environments for testing.

- [ ] Test A
- [ ] Test B

**Checklist**

### DIFF
--- a/docs/_source/reference/telemetry.md
+++ b/docs/_source/reference/telemetry.md
@@ -41,7 +41,7 @@ The following usage and error information is reported:
 * The machine type, e.g. `AMD64`
 * The underlying platform spec with as much useful information as possible. (eg. `macOS-10.16-x86_64-i386-64bit`)
 * The type of deployment: `quickstart` or `server`
-* The dockerized deployment flag: `true` or `false`
+* The dockerized deployment flag: `True` or `False`
 
 This is performed by registering information from the following API methods:
 
--- a/docs/_source/reference/telemetry.md
+++ b/docs/_source/reference/telemetry.md
@@ -41,7 +41,7 @@ The following usage and error information is reported:
 * The machine type, e.g. `AMD64`
 * The underlying platform spec with as much useful information as possible. (eg. `macOS-10.16-x86_64-i386-64bit`)
 * The type of deployment: `quickstart` or `server`
-* The dockerized deployment flag: `yes` or `no`
+* The dockerized deployment flag: `true` or `false`
 
 This is performed by registering information from the following API methods:
 
--- a/docs/_source/reference/telemetry.md
+++ b/docs/_source/reference/telemetry.md
@@ -40,12 +40,14 @@ The following usage and error information is reported:
 * The system’s release version, e.g. `Darwin Kernel Version 21.5.0: Tue Apr 26 21:08:22 PDT 2022; root:xnu-8020`
 * The machine type, e.g. `AMD64`
 * The underlying platform spec with as much useful information as possible. (eg. `macOS-10.16-x86_64-i386-64bit`)
-
+* The type of deployment: `quickstart` or `server`
+* The dockerized deployment flag: `yes` or `no`
 
 This is performed by registering information from the following API methods:
 
 * `/api/me`
 * `/api/dataset/{name}/{task}:bulk`
+* `POST /api/users`
 * Raised server API errors
 
 Additionally, we report the usage of integration for our Python library:
--- a/docs/_source/reference/telemetry.md
+++ b/docs/_source/reference/telemetry.md
@@ -45,8 +45,8 @@ The following usage and error information is reported:
 
 This is performed by registering information from the following API methods:
 
-* `/api/me`
-* `/api/dataset/{name}/{task}:bulk`
+* `GET /api/me`
+* `POST /api/dataset/{name}/{task}:bulk`
 * `POST /api/users`
 * Raised server API errors
 
--- a/src/argilla/server/apis/v1/handlers/oauth2.py
+++ b/src/argilla/server/apis/v1/handlers/oauth2.py
@@ -16,6 +16,7 @@ from fastapi import APIRouter, Depends, HTTPException, Request
 from fastapi.responses import RedirectResponse
 from sqlalchemy.ext.asyncio import AsyncSession
 
+from argilla.server import telemetry
 from argilla.server.contexts import accounts
 from argilla.server.database import get_async_db
 from argilla.server.enums import UserRole
@@ -49,7 +50,11 @@ def get_authentication(request: Request, provider: str) -> RedirectResponse:
 
 
 @router.get("/providers/{provider}/access-token", response_model=Token)
-async def get_access_token(request: Request, provider: str, db: AsyncSession = Depends(get_async_db)) -> Token:
+async def get_access_token(
+    request: Request,
+    provider: str,
+    db: AsyncSession = Depends(get_async_db),
+) -> Token:
     _check_oauth_enabled_or_raise()
 
     try:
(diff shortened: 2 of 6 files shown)


docs/_source/reference/telemetry.md
```diff
@@ -40,12 +40,14 @@ The following usage and error information is reported:
...
 * The machine type, e.g. `AMD64`
 * The underlying platform spec with as much useful information as possible. (eg. `macOS-10.16-x86_64-i386-64bit`)
-
+* The type of deployment: `quickstart` or `server`
+* The dockerized deployment flag: `yes` or `no`
 
 This is performed by registering information from the following API methods:
 
 * `/api/me`
 * `/api/dataset/{name}/{task}:bulk`
```
jfcalvo: Maybe you want to add the HTTP verb in these two?

tests/unit/server/commons/test_telemetry.py
```diff
@@ -52,3 +56,20 @@ async def test_track_bulk(test_telemetry):
     test_telemetry.track_data.assert_called_once_with(
         action="LogRecordsRequested", data={"task": task, "records": records}
     )
+
+
+@pytest.mark.parametrize("is_oauth", [True, False])
+@pytest.mark.parametrize("username", ["argilla", "john"])
+def test_default_user_created(test_telemetry, username: str, is_oauth: bool):
```
frascuchon:
```suggestion
def test_user_created(test_telemetry, username: str, is_oauth: bool):
```

src/argilla/server/apis/v1/handlers/oauth2.py
```diff
@@ -61,13 +66,14 @@ async def get_access_token(request: Request, provider: str, db: AsyncSession = D
...
         if user is None:
-            await accounts.create_user_with_random_password(
+            user = await accounts.create_user_with_random_password(
                 db,
                 username=username,
                 first_name=user_info.name,
                 role=_USER_ROLE_ON_CREATION,
                 workspaces=[workspace.name for workspace in settings.oauth.allowed_workspaces],
             )
+            telemetry.track_user_created(user, is_oauth=True)
```
jfcalvo, line 76: If the user is not created for some reason we are sending telemetry data (empty I guess)?. If that's the case maybe It makes sense to move the telemetry call to be inside `accounts.create_user_with_random_password` function.

src/argilla/server/telemetry.py
```diff
@@ -64,9 +67,14 @@ def __post_init__(self, enable_telemetry: bool, disable_send: bool, api_key: str
             "platform": platform.platform(),
             "python_version": platform.python_version(),
             "sys_version": platform.version(),
+            "deployment": "quickstart" if is_quickstart_server() else "server",
+            "docker": "yes" if is_running_on_docker_container() else "no",
```
jfcalvo: It's not supported to send a boolean value? like:
```suggestion
            "docker": is_running_on_docker_container(),
```

src/argilla/server/telemetry.py
```diff
@@ -64,9 +67,14 @@ def __post_init__(self, enable_telemetry: bool, disable_send: bool, api_key: str
             "platform": platform.platform(),
             "python_version": platform.python_version(),
             "sys_version": platform.version(),
+            "deployment": "quickstart" if is_quickstart_server() else "server",
```
jfcalvo: If the function is not used for other things maybe you can changed it to return an string with the type of server something like:

```suggestion
            "deployment": server_type(),
```

src/argilla/server/utils/_telemetry.py
```diff
@@ -0,0 +1,92 @@
...
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+import os
+
+import psutil
+
+
+def is_quickstart_server():
```
jfcalvo:
```suggestion
def is_quickstart_server() -> bool:
```

src/argilla/server/utils/_telemetry.py
```diff
@@ -0,0 +1,92 @@
...
+    """Returns True if the current process is running on the Quickstart server, False otherwise."""
+    global _is_quickstart_server
+
+    if _is_quickstart_server is None:
+        _is_quickstart_server = is_running_on_docker_container() and _is_quickstart_env()
+    return _is_quickstart_server
+
+
+def is_running_on_docker_container():
```
jfcalvo:
```suggestion
def is_running_on_docker_container() -> bool:
```

nataliaElv (commented):
Docs seem ok!

jfcalvo (changes requested):

src/argilla/server/apis/v1/handlers/oauth2.py
frascuchon: If the user is because the function `accounts.create_user_with_random_password` fails, so the next line won't be executed.

The user-created event will raised only the the user was properly created

docs/_source/reference/telemetry.md
```diff
@@ -40,12 +40,14 @@ The following usage and error information is reported:
 * The system’s release version, e.g. `Darwin Kernel Version 21.5.0: Tue Apr 26 21:08:22 PDT 2022; root:xnu-8020`
 * The machine type, e.g. `AMD64`
 * The underlying platform spec with as much useful information as possible. (eg. `macOS-10.16-x86_64-i386-64bit`)
-
+* The type of deployment: `quickstart` or `server`
+* The dockerized deployment flag: `true` or `false`
```
jfcalvo: If using booleans I guess we will have `True` or `False` instead?

```suggestion
* The dockerized deployment flag: `True` or `False`
```

src/argilla/server/telemetry.py
```diff
@@ -91,16 +99,29 @@ async def track_bulk(task: TaskType, records: int):
...
 
 
-async def track_login(request: Request, username: str):
+async def track_login(request: Request, user: User):
     _CLIENT.track_data(
         action="UserInfoRequested",
         data={
-            "is_default_user": username == "argilla",
-            "user_hash": str(uuid.uuid5(namespace=_CLIENT.server_id, name=username)),
+            "is_default_user": user.username == "argilla",
```
jfcalvo: Maybe we can use here the constant from `argilla/server/constants.py`?

```suggestion
            "is_default_user": user.username == DEFAULT_USERNAME,
```

src/argilla/server/telemetry.py
```diff
@@ -91,16 +99,29 @@ async def track_bulk(task: TaskType, records: int):
...
     )
 
 
+def track_user_created(user: User, is_oauth: bool = False):
+    _CLIENT.track_data(
+        action="UserCreated",
+        data={
+            "user_id": str(user.id),
+            "role": user.role,
+            "is_default_user": user.username == "argilla",
```
jfcalvo: Same for this line:

```suggestion
            "is_default_user": user.username == DEFAULT_USERNAME,
```

src/argilla/server/utils/_telemetry.py
```diff
@@ -0,0 +1,96 @@
...
+        "ADMIN_USERNAME",
+        "ADMIN_PASSWORD",
+        "ADMIN_API_KEY",
+        "ANNOTATOR_USERNAME",
+        "ANNOTATOR_PASSWORD",
+        "LOAD_DATASETS",
+    ]:
+        if env_var not in os.environ:
+            return False
+    return True
```
jfcalvo, line 91: Maybe for another PR but it should be easier to define a environment variable in the dockerfile for the QuickStart image with something like `SERVER_TYPE=quickstart` and return `True` only when the environment var is defined and have the value `quickstart`. (or maybe I'm missing something here)

jfcalvo (changes requested):

src/argilla/server/utils/_telemetry.py
frascuchon: Yes, I thought the env var solution, but this approach is really easy to hack.

jfcalvo (approved):

PR #4605, opened 2024-02-27
bug: FloatMetadataProperty: value is not a valid float when it is an rounded integer

### LINKED ISSUE #4570: [BUG-python] `FloatMetadataProperty`: value is not a valid `float` when it is an rounded integer
**Describe the bug**
Using `distilabel` and `to_argilla` method raises this error regarding metadata and `text-descriptives`.

```bash
ValueError: `FeedbackRecord.metadata` {'length-input': 19, 'length-instructions': 106, 'generation-model': 
'gpt-3.5-turbo', 'input_n_tokens': 2, 'input_n_unique_tokens': 2, 'input_n_sentences': 1, 'input_perplexity': 1, 
'input_entropy': 0, 'input_flesch_reading_ease': -48.99, 'instructions_n_tokens': 18, 
'instructions_n_unique_tokens': 18, 'instructions_n_sentences': 1, 'instructions_perplexity': 1.79, 
'instructions_entropy': 0.58, 'instructions_flesch_reading_ease': 61.67} does not match the expected schema, with 
exception: 2 validation errors for MetadataSchema
input_perplexity
  value is not a valid float (type=type_error.float)
input_entropy
  value is not a valid float (type=type_error.float)
```

**Expected behavior**
I would expect this check to pass given that any integer is a rounded of float.

**Environment:**
 - Argilla Version [e.g. 1.0.0]: 1.23

**Additional context**

sdiazlor:
# Description

With the removal of this line,  x.0 values will not be rounded so that it passes the FloatMetadataProperty validation. The error comes from the initial definition of the metadata property as float, but then casting and adding as value an integer.

NOTE: This fixes the bug error directly, although this question is still pending @alvarobartt:
>However, from a usability perspective there is also the issue of not allowing for passing a rounded float like 1.0 to an IntegerMetadataProperty and perhaps also not an integer like1 to a FloatMetadataProperty.

Closes #4570 

**Type of change**

**How Has This Been Tested**

- [ ] Test A
- [ ] Test B

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,10 @@ These are the section headers that we use:
 
 - Reorder labels in `dataset settings page` for single/multi label questions ([#4598](https://github.com/argilla-io/argilla/pull/4598))
 
+### Fixed
+
+- Fixed FloatMetadataProperty: value is not a valid float ([#4570](https://github.com/argilla-io/argilla/pull/4605))
+
 ## [1.24.0](https://github.com/argilla-io/argilla/compare/v1.23.0...v1.24.0)
 
 > [!NOTE]
--- a/src/argilla/client/feedback/integrations/textdescriptives.py
+++ b/src/argilla/client/feedback/integrations/textdescriptives.py
@@ -302,10 +302,6 @@ class TextDescriptivesExtractor:
             )
             for record, metrics in zip(records, df.to_dict("records")):
                 filtered_metrics = {key: value for key, value in metrics.items() if not pd.isna(value)}
-                filtered_metrics = {
-                    key: int(value) if isinstance(value, float) and value.is_integer() else value
-                    for key, value in filtered_metrics.items()
-                }
                 record.metadata.update(filtered_metrics)
                 modified_records.append(record)
                 progress_bar.update(task, advance=1)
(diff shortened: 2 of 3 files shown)


alvarobartt:
Hi here @sdiazlor, thanks for the PR! w.r.t. @davidberenstein1957's question, I'm not sure, because if we have an `IntegerMetadataProperty` we should enforce the values to be `int`, not casting those to `int` i.e. `1.0 -> 1`; same thing for the `FloatMetadataProperty`, which shouldn't cast to `float` by default. So IMO we can keep the casting within the `text-descriptives` integration while keeping `argilla` as is, also what do @jfcalvo and @frascuchon think about that?

davidberenstein1957 (approved):
@sdiazlor could you also check if this work with the error Dani was experiencing?

davidberenstein1957:
> Hi here @sdiazlor, thanks for the PR! w.r.t. @davidberenstein1957's question, I'm not sure, because if we have an `IntegerMetadataProperty` we should enforce the values to be `int`, not casting those to `int` i.e. `1.0 -> 1`; same thing for the `FloatMetadataProperty`, which shouldn't cast to `float` by default. So IMO we can keep the casting within the `text-descriptives` integration while keeping `argilla` as is, also what do @jfcalvo and @frascuchon think about that?

The proposed changed avoids the casting to `int` right? For me, it was more of a philosophical question about whether or not it made sense to allow for this, outside of this basic PR. I understand why but from a user experience it might be weird that it is impossible to add an `1.0` to an `IntegerMetadataProperty` or a `1` to a `FloatMetadataProperty`.

sdiazlor:
> @sdiazlor could you also check if this work with the error Dani was experiencing?

@davidberenstein1957 Yes, I already did. And with a similar case (using to_argilla method), it didn't raise the validation issue. Below, you can see that the values are floats.
image

PR #4623, opened 2024-03-04
feat: create responses and suggestions with spans

frascuchon:
# Description

This PR defines a basic flow to support responses and suggestions with span values.

```python
import spacy

import argilla as rg
from argilla.client.feedback.schemas import SpanValueSchema
from argilla.client.feedback.schemas.responses import ResponseSchema, ResponseBuilder


api_url = ...
api_key = ...

rg.init(api_url, api_key)

nlp = spacy.load("en_core_web_sm")
labels = nlp.get_pipe("ner").labels

gutenberg_config = rg.FeedbackDataset(
    fields=[rg.TextField(name="text")],
    questions=[rg.SpanQuestion(name="spans", title="Name entities recognition", field="text", labels=labels)]
)

ds = gutenberg_config.push_to_argilla("gutenberg_spacy-ner", workspace="argilla")
hf_ds = load_dataset("argilla/gutenberg_spacy-ner", split="train")

user = rg.User.me()

def to_response_spans(doc) -> list[SpanValueSchema]:
    """Converts a spacy doc to a list of response spans."""
    return [
        SpanValueSchema(start=ent.start_char, end=ent.end_char,label=ent.label_)
        for ent in doc.ents
    ]

question = ds.question_by_name("spans")
records = [
    rg.FeedbackRecord(
        fields={"text": r["text"]},
        responses=[
            ResponseSchema(
                status="draft",
                user_id=user.id,
                values=[question.response(to_response_spans(nlp(r["text"])))]
            )
        ],
        suggestions=[
            question.suggestion(value=to_response_spans(nlp(r["text"])), agent="spacy", type="model")
        ]
    )
    for r in hf_ds
]
ds.add_records(records)
```

Closes https://github.com/argilla-io/argilla/issues/4618

**Type of change**

**How Has This Been Tested**

- [ ] Test A
- [ ] Test B

**Checklist**

### DIFF
--- a/src/argilla/client/feedback/integrations/huggingface/dataset.py
+++ b/src/argilla/client/feedback/integrations/huggingface/dataset.py
@@ -428,10 +428,11 @@ class HuggingFaceDatasetMixin:
                     f"{question.name}-suggestion" in hfds[index]
                     and hfds[index][f"{question.name}-suggestion"] is not None
                 ):
-                    suggestion = {
-                        "question_name": question.name,
-                        "value": hfds[index][f"{question.name}-suggestion"],
-                    }
+                    value = hfds[index][f"{question.name}-suggestion"]
+                    if question.type == QuestionTypes.ranking:
+                        value = [{"rank": r, "value": v} for r, v in zip(value["rank"], value["value"])]
+
+                    suggestion = {"question_name": question.name, "value": value}
                     if hfds[index][f"{question.name}-suggestion-metadata"] is not None:
                         suggestion.update(hfds[index][f"{question.name}-suggestion-metadata"])
                     suggestions.append(suggestion)
--- a/src/argilla/client/feedback/integrations/huggingface/dataset.py
+++ b/src/argilla/client/feedback/integrations/huggingface/dataset.py
@@ -11,11 +11,11 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
-
 import json
 import logging
 import tempfile
 import warnings
+from copy import copy
 from typing import TYPE_CHECKING, Any, Optional, Type, Union
 
 from packaging.version import parse as parse_version
@@ -50,7 +50,7 @@ class HuggingFaceDatasetMixin:
             questions, and metadata_properties formatted as `datasets.Features`.
 
         Examples:
-            >>> from argilla.client.feedback.integrations.dataset import HuggingFaceDatasetMixin
+            >>> from argilla.client.feedback.integrations.huggingface import HuggingFaceDatasetMixin
             >>> dataset = FeedbackDataset(...) or RemoteFeedbackDataset(...)
             >>> huggingface_dataset = HuggingFaceDatasetMixin._huggingface_format(dataset)
         """
@@ -71,17 +71,38 @@ class HuggingFaceDatasetMixin:
         for question in dataset.questions:
             if question.type in [QuestionTypes.text, QuestionTypes.label_selection]:
                 value = Value(dtype="string", id="question")
+                suggestion_value = copy(value)
             elif question.type == QuestionTypes.rating:
                 value = Value(dtype="int32", id="question")
+                suggestion_value = copy(value)
             elif question.type == QuestionTypes.ranking:
                 value = Sequence({"rank": Value(dtype="uint8"), "value": Value(dtype="string")}, id="question")
+                suggestion_value = copy(value)
             elif question.type in QuestionTypes.multi_label_selection:
                 value = Sequence(Value(dtype="string"), id="question")
+                suggestion_value = copy(value)
+            elif question.type in QuestionTypes.span:
+                value = Sequence(
+                    {
+                        "start": Value(dtype="int32"),
+                        "end": Value(dtype="int32"),
+                        "label": Value(dtype="string"),
+                    },
+                    id="question",
+                )
+                suggestion_value = Sequence(
+                    {
+                        "start": Value(dtype="int32"),
+                        "end": Value(dtype="int32"),
+                        "label": Value(dtype="string"),
+                        "score": Value(dtype="float32"),
+                    }
+                )
             else:
                 raise ValueError(
                     f"Question {question.name} is of type `{question.type}`,"
                     " for the moment only the following question types are supported:"
-                    f" `{'`, `'.join([arg.value for arg in QuestionTypes])}`."
+                    f" `{'`, `'.join(QuestionTypes.values())}`."
                 )
 
             hf_features[question.name] = [
@@ -94,8 +115,8 @@ class HuggingFaceDatasetMixin:
             if question.name not in hf_dataset:
                 hf_dataset[question.name] = []
 
-            value.id = "suggestion"
-            hf_features[f"{question.name}-suggestion"] = value
+            suggestion_value.id = "suggestion"
+            hf_features[f"{question.name}-suggestion"] = suggestion_value
             if f"{question.name}-suggestion" not in hf_dataset:
                 hf_dataset[f"{question.name}-suggestion"] = []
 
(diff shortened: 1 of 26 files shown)


src/argilla/client/feedback/integrations/huggingface/dataset.py
```diff
@@ -421,6 +451,11 @@ def from_huggingface(
                         if value is not None:
                             if question.type == QuestionTypes.ranking:
                                 value = [{"rank": r, "value": v} for r, v in zip(value["rank"], value["value"])]
+                            elif question.type == QuestionTypes.span:
+                                value = [
+                                    {"start": s, "end": e, "label": l}
+                                    for s, e, l in zip(value["start"], value["end"], value["label"])
+                                ]
                             responses[user_id or "user_without_id"]["values"].update({question.name: {"value": value}})
```
davidberenstein1957, line 459: perhaps we should add else statements here to be sure it raises errors/has behaviour when we add question types?

src/argilla/client/feedback/integrations/huggingface/dataset.py
```diff
@@ -431,6 +466,11 @@ def from_huggingface(
                     value = hfds[index][f"{question.name}-suggestion"]
                     if question.type == QuestionTypes.ranking:
                         value = [{"rank": r, "value": v} for r, v in zip(value["rank"], value["value"])]
+                    elif question.type == QuestionTypes.span:
+                        value = [
+                            {"start": s, "end": e, "label": l}
+                            for s, e, l in zip(value["start"], value["end"], value["label"])
+                        ]
```
davidberenstein1957, line 473: perhaps we should add else statements here to be sure it raises errors/has behaviour when we add question types?

src/argilla/client/feedback/integrations/huggingface/dataset.py
```diff
@@ -431,6 +466,11 @@ def from_huggingface(
                     value = hfds[index][f"{question.name}-suggestion"]
                     if question.type == QuestionTypes.ranking:
                         value = [{"rank": r, "value": v} for r, v in zip(value["rank"], value["value"])]
+                    elif question.type == QuestionTypes.span:
+                        value = [
+                            {"start": s, "end": e, "label": l}
+                            for s, e, l in zip(value["start"], value["end"], value["label"])
```
davidberenstein1957, line 472: in terms of human-readability, I also like to add the extracted text normally. Something like`value["text"][value["start"]:value["end"]]` but perhaps this is difficult top map back into the correct format  when calling `from_huggingface`?

src/argilla/client/feedback/integrations/huggingface/dataset.py
```diff
@@ -421,6 +451,11 @@ def from_huggingface(
                         if value is not None:
                             if question.type == QuestionTypes.ranking:
                                 value = [{"rank": r, "value": v} for r, v in zip(value["rank"], value["value"])]
+                            elif question.type == QuestionTypes.span:
+                                value = [
+                                    {"start": s, "end": e, "label": l}
+                                    for s, e, l in zip(value["start"], value["end"], value["label"])
```
davidberenstein1957, line 457: in terms of human-readability, I also like to add the extracted text normally. Something like`value["text"][value["start"]:value["end"]]` but perhaps this is difficult top map back into the correct format  when calling `from_huggingface`?

src/argilla/client/feedback/schemas/questions.py
```diff
@@ -334,21 +347,35 @@ class SpanQuestion(QuestionSchema):
...
+    def parse_labels_dict(cls, labels) -> List[SpanLabelOption]:
+        if isinstance(labels, dict):
+            return [SpanLabelOption(value=label, text=text) for label, text in labels.items()]
+        return labels
 
     @validator("labels", always=True)
     def normalize_labels(cls, v: List[Union[str, SpanLabelOption]]) -> List[SpanLabelOption]:
         return [SpanLabelOption(value=label, text=label) if isinstance(label, str) else label for label in v]
 
+    @validator("labels")
```
davidberenstein1957, line 368: wasn't there a max for the labels too which we defined on the server side? perhaps we can use it here as well?

src/argilla/client/feedback/schemas/response_values.py
```diff
@@ -0,0 +1,112 @@
...
+        score: The score of the span.
+    """
+
+    label: constr(min_length=1)
+    start: conint(ge=0)
+    end: conint(ge=0)
+    score: Optional[confloat(ge=0.0, le=1.0)] = None
+
+    @root_validator
+    def check_span(cls, values):
```
davidberenstein1957, line 62: I saw we already tested for larger than 0 but do we also validate if the span is within the overall text field, i.e., they are not larger than the used text?

src/argilla/client/feedback/schemas/responses.py
```diff
@@ -0,0 +1,96 @@
...
+
+    @validator("user_id", always=True)
+    def user_id_must_have_value(cls, v):
+        if not v:
+            warnings.warn(
+                "`user_id` not provided, so it will be set to `None`. Which is not an"
+                " issue, unless you're planning to log the response in Argilla, as"
+                " it will be automatically set to the active `user_id`.",
+            )
+        return v
```
davidberenstein1957, line 86: perphaps we should return `None` explicitly here? `not 1` and `not True` also pass the `if-statement` and perhaps we require an explicit `None` to set the `user_id` automatically?

src/argilla/client/feedback/schemas/suggestions.py
```diff
@@ -0,0 +1,68 @@
...
+    _normalize_value = validator("value", allow_reuse=True, always=True)(normalize_response_value)
+
+    class Config:
+        extra = Extra.forbid
+        validate_assignment = True
+
+    def to_server_payload(self, question_name_to_id: Dict[str, UUID]) -> Dict[str, Any]:
+        """Method that will be used to create the payload that will be sent to Argilla
+        to create a `SuggestionSchema` for a `FeedbackRecord`."""
+        payload = self.dict(exclude_unset=True, include={"type", "score", "value", "agent"})
```
davidberenstein1957, line 65: to be clear, all of "type", "score", "value", "agent" are stored but not all of them are used, correct?

src/argilla/client/feedback/integrations/huggingface/dataset.py
frascuchon: I'm using the current approach to support span values mappings. I wouldn't add cross-cutting solution in the current version of the SDK. This is something that we can add for new SDK implementation

src/argilla/client/feedback/integrations/huggingface/dataset.py
frascuchon: I'm using the current approach to support span values mappings. I wouldn't add cross-cutting solution in the current version of the SDK. This is something that we can add for new SDK implementation

src/argilla/client/feedback/integrations/huggingface/dataset.py
frascuchon: This change is not hard to add, if it makes sense.

src/argilla/client/feedback/integrations/huggingface/dataset.py
frascuchon: This change is not hard to add, if it makes sense.

src/argilla/client/feedback/schemas/responses.py
frascuchon: This code was already here. I just moved it. I wouldn't change anything except the span feature. This can be tackled in the new SDK.

src/argilla/client/feedback/schemas/suggestions.py
frascuchon: 'score' is used to filter records at least from UI. The rest are not used yet, even if they're available for searches.

src/argilla/client/feedback/schemas/response_values.py
frascuchon: We're covering basic data validation from SDK, adding more complex validations is not quite simple to implement for this SDK. Since there are these kinds of validations from the server, and the rest of the suggestions/responses didn't include validations, I would keep it simple for this version.

src/argilla/client/feedback/schemas/questions.py
frascuchon: Yes, this is not hard to implement.

frascuchon: Anyway, this is not the current behaviour that exists for single and multi-label settings. In both cases, there is a min validation but not a max one.

Also, if we plan to support a configurable value for this, adding a hard validation here may introduce workflow problems. So, I will let as is.

PR #4628, opened 2024-03-05
fix: format suggestions for ranking values

frascuchon:
# Description

This PR fixes problems with HF format when passing ranking value objects (instead of raw dictionaries).

**Type of change**

**How Has This Been Tested**

Tested locally

- [ ] Test A
- [ ] Test B

**Checklist**

### DIFF
--- a/src/argilla/client/feedback/integrations/huggingface/dataset.py
+++ b/src/argilla/client/feedback/integrations/huggingface/dataset.py
@@ -148,7 +148,10 @@ class HuggingFaceDatasetMixin:
                 if record.suggestions:
                     for suggestion in record.suggestions:
                         if question.name == suggestion.question_name:
-                            suggestion_value = suggestion.value
+                            if question.type == QuestionTypes.ranking:
+                                suggestion_value = [r.dict() for r in suggestion.value]
+                            else:
+                                suggestion_value = suggestion.value
                             suggestion_metadata = {
                                 "type": suggestion.type,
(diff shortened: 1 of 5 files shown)


src/argilla/client/feedback/integrations/huggingface/dataset.py
```diff
@@ -148,7 +148,10 @@ def _huggingface_format(dataset: Union["FeedbackDataset", "RemoteFeedbackDataset
                 if record.suggestions:
                     for suggestion in record.suggestions:
                         if question.name == suggestion.question_name:
-                            suggestion_value = suggestion.value
+                            if question.type == QuestionTypes.ranking:
+                                suggestion_value = [r.dict() for r in suggestion.value]
+                            else:
+                                suggestion_value = suggestion.value
```
alvarobartt:
```suggestion
                            suggestion_value = suggestion.dict()["value"]
```

alvarobartt: I think that would do the work, so that everything's a `dict` in advance and then the value is retrieved, this way you ensure the nested `pydantic.BaseModel`s are casted to `dict` too, and may be more extensible if we add new questions IMO

frascuchon: It's working with this change. Thanks!

alvarobartt (approved):

PR #4644, opened 2024-03-11
chore: apply some of suggested changes

frascuchon:
This PR adds the `text` value for span suggestions and responses when publishing HF datasets. Commented by @davidberenstein1957  [here](https://github.com/argilla-io/argilla/pull/4623#discussion_r1519318718)


davidberenstein1957:
@frascuchon do we test if this works when we load data form the Hub again?

frascuchon:
> @frascuchon do we test if this works when we load data form the Hub again?

Yes. [This test](https://github.com/argilla-io/argilla/blob/feat/apply-some-of-suggested-changes/tests/integration/client/feedback/dataset/local/test_dataset.py#L575) is taking care of that

PR #4660, opened 2024-03-14
🎉 Allow external links for markdown

### LINKED ISSUE #4659: [FEATURE] Allow URL hyperlinks in markdown to open in a new tab or window in `FeedbackRecord`s
Currently when using text (markdown) fields in `FeedbackRecord`, it can be useful to add external URLs that provide additional context to the annotator. Currently, URLs are clickable, but open by default in the same browser tab that Argilla is using, thus disrupting the annotation flow.

Take for instance

```python
import argilla as rg
api_url = "http://localhost:6900"
api_key = "<my-api-key>"

rg.init(api_url=api_url, api_key=api_key)

# rg.Workspace.create('argilla')

rg.set_workspace("argilla")

num_labels = 250
labels = ["label " + str(i) for i in range(0, num_labels)]

dataset = rg.FeedbackDataset(
    fields=[
        rg.TextField(name="text", required= True, use_markdown=True),
        rg.TextField(name="url", required= True, use_markdown=True)
    ],
    questions=[
        rg.MultiLabelQuestion(name="question-multi", title="Which labels are correct?", labels=labels, required=True, visible_labels = 3)],
)

records = []

record = rg.FeedbackRecord(
    fields={
        "text": "tierra finamente dividida, constituida por agregados de silicatos de aluminio hidratados",
        "url": "<a href='https://github.com/argilla-io/argilla' target='_blank'>I'd love to open this in another tab</a>"
    }
)
records.append(record)

dataset.add_records(records)

dataset.push_to_argilla(name="workaround-dataset", workspace="argilla")
```

Here, the URL is using the syntax in https://stackoverflow.com/a/4425223/4240413 .
This renders as 

image

Ideally, when clicking the URL in the argilla frontend, the target link should open in another tab or window.

**Additional context**
I'm using `argilla=1.24.0` and Firefox 123.0, using other solutions like https://stackoverflow.com/a/4705645/4240413 didn't seem to work either.

damianpumar:
- [x] Test this PR with this case -> https://github.com/argilla-io/argilla/issues/4659

### DIFF
--- a/frontend/components/base/base-render-markdown/RenderMarkdown.base.component.vue
+++ b/frontend/components/base/base-render-markdown/RenderMarkdown.base.component.vue
@@ -27,7 +27,7 @@ DOMPurify.addHook("beforeSanitizeAttributes", (node) => {
       node.setAttribute("viewBox", `0 0 ${width} ${height}`);
     }
   }
-  if (node.tagName === "A") {
+  if (node instanceof HTMLAnchorElement) {
     node.setAttribute("target", "_blank");
   }
 });
--- a/frontend/components/base/base-render-markdown/RenderMarkdown.base.component.vue
+++ b/frontend/components/base/base-render-markdown/RenderMarkdown.base.component.vue
@@ -13,7 +13,7 @@ const preprocess = (html) => {
 const postprocess = (html) => {
   return DOMPurify.sanitize(html, {
     ADD_TAGS: ["embed", "object"],
-    ADD_ATTR: ["data"],
+    ADD_ATTR: ["data", "target"],
     ADD_URI_SAFE_ATTR: ["data"],
   });
 };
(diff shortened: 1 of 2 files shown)


frontend/components/base/base-render-markdown/RenderMarkdown.base.component.vue
```diff
@@ -27,6 +27,9 @@ DOMPurify.addHook("beforeSanitizeAttributes", (node) => {
       node.setAttribute("viewBox", `0 0 ${width} ${height}`);
     }
   }
+  if (node.tagName === "A") {
```
jfcalvo: You've probably already checked it out but it's not possible to check if the `node` is of type anchor or something similar instead of comparing to a string?

jfcalvo (approved):

frontend/components/base/base-render-markdown/RenderMarkdown.base.component.vue
jfcalvo: I mean something like this:

```suggestion
  if (node instanceof HTMLAnchorElement) {
```

damianpumar: Great idea Jose, I'll try it!

leiyre (approved):

PR #4672, opened 2024-03-19
Fix Client's handling of Enum Values, to support Python >=3.11 enum handling.

tim-win:
# Description

The argilla client for v0 Search and Metrics appears to be broken in Python 3.11 and greater. This comes from a simple mis-resolution of enum values, due to some of the poorly documented python 3.11. In short, given this definition of the TaskType enum:

```
class TaskType(str, Enum):
    text_classification = "TextClassification"
    token_classification = "TokenClassification"
    text2text = "Text2Text"
```

in Python 3.10 and before, format strings and the format method will return the value of the target enum:

```
>>> f'{TaskType.text_classification}'
'TextClassification'
```

However, in python 3.11 and later, format strings will produce a more verbose description:

```
>>> f'{TaskType.text_classification}'
'TaskType.text_classification'
```

The client's Search and Metrics code both use the same pattern, producing an incorrect _API_URL and thereby running into 404's:

```
class Search(AbstractApi):
    _API_URL_PATTERN = "/api/datasets/{name}/{task}:search"

    def search_records(
        self,
        name: str,
        task: TaskType,
        size: Optional[int] = None,
        **query,
    ):
    ...
    url = Search._API_URL_PATTERN.format(name=name, task=task)
    ...
    response = self.http_client.post(
            path=url,
            json={"query": query} if query else None,
        )
    ...
```

This PR just updates the way task is handled by the url format function in both APIs, a method that produces the correct value in 3.8, 3.10, 3.11, and 3.12.0a7:

```
>>> f'{TaskType.text_classification.value}'                                                                                            
'TextClassification'
```

```
    url = Search._API_URL_PATTERN.format(name=name, task=task.value)
```

**Type of change**

**How Has This Been Tested**

This has been tested locally using pyenv to swap between python 3.8, 3.10, 3.11, and 3.12.0a7 to confirm 3.11 is the change point. Also, tested this in-situ by trying to run AutoPrompt (which only supports 3.10 and below) with this change, and it appears this change is the last thing stopping autoprompt from supporting 3.11!

**Checklist**


frascuchon (approved):
Thanks @tim-win for your contribution!!!

tim-win:
:woozy_face: my first open source PR in over a year, I think! Thank you for accepting my change.

PR #4685, opened 2024-03-26
✨ Support RTL languages

### LINKED ISSUE #4683: [FEATURE] Right-to-left text direction in the text editor
Hi 👋

Many languages such as Arabic, Persian, and Hebrew use Right-to-left script. The problem is that the text editor (as for now) does not support right-to-left text direction. This makes editing RTL languages really tricky and confusing especially if the text had contained some LTR (such as Latin concepts) words in the mix. 

**Additional context**
Here are some screenshots of the way it looks like right now:

arg001
arg002

damianpumar:
Screenshot 2024-03-26 at 09 21 57
Screenshot 2024-03-26 at 09 22 07
Screenshot 2024-03-26 at 09 22 20
Screenshot 2024-03-26 at 09 22 28
Screenshot 2024-03-26 at 10 12 48
Screenshot 2024-03-26 at 10 13 18
Screenshot 2024-03-26 at 10 15 12
Screenshot 2024-03-26 at 10 15 27


davidberenstein1957:
@damianpumar 
"
OH! In that case, it'd work just fine 👌. The main problem with the unmatched text direction is the (somewhat) unpredicted cursor. Like having the last character at the beginning of the sentence, for example. With that being out of the equation now, we get a more reliable cursor and span selection should work as expected.
Also the snap feature is really great 💥
"

nataliaElv (approved):

PR #4702, opened 2024-04-04
🚀 Improve performance

damianpumar:
After trying this PR https://github.com/argilla-io/argilla-server/pull/85, I have improved the performance for Single/Multi/Span questions.

**The maximum supported value of options to guarantee good performance is 1000.**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ These are the section headers that we use:
 
 ## [Unreleased]()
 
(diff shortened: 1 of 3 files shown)


leiyre (approved):
🚀

jfcalvo:
@damianpumar can you explain a little bit how is the performance improved? I don't understand the improvement seeing the code. Thanks.

PR #4749, opened 2024-04-25
ci: Allow manual execution for deploy argilla environment

frascuchon:
This PR configs the manual execution of the PR environments.

Also, clean running environments once per day.

### DIFF
--- a/.github/workflows/deploy-environment.yml
+++ b/.github/workflows/deploy-environment.yml
@@ -1,13 +1,15 @@
 name: Deploy Argilla environment
 
 on:
+  workflow_dispatch:
   workflow_call:
     inputs:
       image-name:
         description: The name of the Docker image to deploy.
         type: string
+        default: argilla/argilla-quickstart-for-dev
       image-version:
-        description: The version of the Docker image to deploy.
+        description: The version of the Docker image to deploy. In the form pr-<PR_NUMBER>.
         type: string
 
 jobs:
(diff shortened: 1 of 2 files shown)


gabrielmbmb (approved):

.github/workflows/deploy-environment.yml
```diff
@@ -1,13 +1,15 @@
 name: Deploy Argilla environment
 
 on:
+  workflow_dispatch:
```
jfcalvo, line 4: So with this you can click on run and manually create a pr environment for the PR on gcloud?

.github/workflows/deploy-environment.yml
```diff
@@ -1,13 +1,15 @@
...
   workflow_call:
     inputs:
       image-name:
         description: The name of the Docker image to deploy.
         type: string
+        default: argilla/argilla-quickstart-for-dev
```
jfcalvo, line 10: Why did you change this? It's an image that still doesn't exists?

frascuchon: I added the image name as default. This will simplify the manual deployment execution. This image is created in the [previous step](https://github.com/argilla-io/argilla/blob/ci/manual-pr-deploy-and-cleaning-services-one-by-day/.github/workflows/package.yml#L155). At this point, it will be available to use.

.github/workflows/deploy-environment.yml
frascuchon: Exactly.

jfcalvo (approved):

PR #4753, opened 2024-04-26
🐛 Fix reversed string

### LINKED ISSUE #4733: [BUG-UI/UX] Search queries in korean get flipped (FeedbackDataset)
**Describe the bug**
On a feedback dataset (using SpanQuestion if relevant, older dataset types don't have this problem) if I type in the search bar in Korean it flips the text like:
type `한국어`-> press_enter-> `어국한`. There should be no flip. Seems to be happening on Chrome only(?), it works fine on safari. (& also doesn't happen with english)

- This doesn't happen when copy pasting text, only when typing.
- & it shows the wrong number of results compared to safari, seems to be analyzing the text differently.

**Screenshots**

https://github.com/argilla-io/argilla/assets/15624271/6ea53635-bdcb-4a2a-a5df-dbdde027acff

 - OS [e.g. iOS]: Mac
 - Browser [e.g. chrome, safari]: chrome
 - Argilla: 1.26 ~ 1.27

damianpumar:


nataliaElv (approved):
Looks like it's working! Thanks @damianpumar ! 🙌

PR #4771, opened 2024-05-06
✨ Search text filtering by field

damianpumar:


frontend/components/feedback-task/header/DatasetFilters.component.vue
```diff
@@ -84,6 +81,9 @@ export default {
     isSortedBy() {
       return this.recordCriteria.isSortedBy;
     },
+    fields() {
```
damianpumar: We should pass to `SearchBarBase` the entire "fields" object because we need to use the "name" of fields to our backend.

leiyre (approved):

PR #4792, opened 2024-05-07
Allow overlap equal to number of users in assign_users

kdutia:
# Description

In `assign_records` it's currently not possible to assign _all records to all users_. I can't see a reason why this would be the case (it's something my team need at the moment), so I've updated the code to allow this and to warn if all records will be assigned to all users.

Note I'm still setting up a dev environment on my machine (with `pip install` the integrations dependencies take ages to resolve), so I haven't confirmed that my test works.

**Type of change**

**How Has This Been Tested**

Updated existing test with a new test case

**Checklist**

### DIFF
--- a/src/argilla/client/feedback/utils/assignment.py
+++ b/src/argilla/client/feedback/utils/assignment.py
@@ -142,8 +142,11 @@ def assign_records_to_individuals(
     Raises:
         ValueError: If `overlap` is higher than the number of users or negative.
     """
-    if overlap < 0 or overlap >= len(users):
-        raise ValueError("Overlap must be less than the number of users and must not be negative.")
+    if overlap < 0 or overlap > len(users):
+        raise ValueError("Overlap must be less or equal to than the number of users and must not be negative.")
+
+    if overlap == len(users):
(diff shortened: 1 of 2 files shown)


sdiazlor:
Hi @kdutia, thanks for your PR. However, _assigning all records to all users_ is the default behavior you can use without the need for `assign_records`. If this is how you want to distribute your workload, you just need to create a dataset in a workspace and let the users with the role `annotator` access it, so they will be able to provide feedback for all the records in that dataset.

Let us know if that would work for your use case :)

kdutia:
@sdiazlor thanks for this advice! Does that mean that each record in one workspace can be annotated by multiple annotators by default? It could be useful to update the documentation around `assign_users` to give this advice for this specific case, or even add an instructional warning when someone tries to do this .

sdiazlor:
@kdutia Thanks for the feedback! Yes, if you allow your annotators to access and annotate the same dataset, you'll be able to get multiple annotations.

kdutia:
@sdiazlor Ok, I'd suggest a documentation update to make this clearer but I'll close this PR. Thanks for your help!

PR #4851, opened 2024-05-17
feat: UI Metadata info component

### LINKED ISSUE #4788: Metadata modal in records

leiyre:
This PR allows the visualization of the metadata clicking on the menu at the top of the record

closes #4788


argilla-frontend/v1/domain/usecases/get-records-by-criteria-use-case.ts
```diff
@@ -135,7 +138,10 @@ export class GetRecordsByCriteriaUseCase {
           null,
           [],
           0,
-          0
+          0,
+          "2024-04-24T14:05:37.802484",
```
damianpumar: I think you forgot some hard code.

damianpumar (changes requested):

argilla-frontend/translation/en.js
```diff
@@ -13,6 +13,7 @@ export default {
   copied: "Copied",
   copyClipboard: "Copy to clipboard",
   copyLink: "Copy link",
+  copyRecord: "Copy record",
```
damianpumar, line 16: Remember that now we have translations for German, so you should update the de.js file.

argilla-frontend/v1/domain/entities/record/Record.ts
```diff
@@ -20,7 +20,10 @@ export class Record {
     public answer: RecordAnswer,
     private readonly suggestions: Suggestion[],
     score: number,
-    public readonly page: number
+    public readonly page: number,
+    public readonly metadataInsertedAt: string,
```
damianpumar: the metadata inserted at should be a Date object right?

argilla-frontend/v1/domain/entities/record/Record.ts
```diff
@@ -20,7 +20,10 @@ export class Record {
...
+    public readonly metadataUpdatedAt: string,
+    public readonly metadata: Object
```
damianpumar: so, why the metadata is an "anonymous" object and why if you pass the metadata as an object you passed the metadataInsertedAt and metadataUpdatedAt in other params?

argilla-frontend/v1/domain/usecases/get-records-by-criteria-use-case.ts
```diff
@@ -97,7 +97,10 @@ export class GetRecordsByCriteriaUseCase {
           answer,
           suggestions,
           record.query_score,
-          recordPage
+          recordPage,
+          record.inserted_at,
```
damianpumar, line 102: I read before that you named this argument as "metadataInsertedAt" but the parameter is named "record.inserted_at" so, I think does not make sense... right ?

damianpumar (approved):

PR #4922, opened 2024-06-03
[CI] Review docs generation workflow

frascuchon:
This PR just applies some minimal changes for the argilla SDK docs generation:

- Freeze the default branch to `feat/v2.0.0`
- Launch a docs generation for docs branches starting with `docs/**`
- Remove PR-related events from triggers.

### DIFF
--- a/.github/workflows/argilla-sdk.docs.yml
+++ b/.github/workflows/argilla-sdk.docs.yml
@@ -14,7 +14,6 @@ on:
       # - "develop"
       - "feat/v2.0.0" # Fixing this branch until we merge everything into develop
       - "docs/**" # When we create a new branch only for docs
-      - "chore/review-docs-generation-workflowOD" # add this current branch for testing
     tags:
       - "**"
 
--- a/.github/workflows/argilla-sdk.docs.yml
+++ b/.github/workflows/argilla-sdk.docs.yml
@@ -7,27 +7,14 @@ concurrency:
 on:
   push:
     paths:
-      - "argilla-sdk/**"
-    branches:
-      # - "main"
-      - "feat/**"
-      - "gh-pages"
-    tags:
-      - "**"
-
-  pull_request:
-    paths:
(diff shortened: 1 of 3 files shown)


.github/workflows/argilla-sdk.docs.yml
```diff
@@ -7,28 +7,16 @@ concurrency:
 on:
   push:
     paths:
-      - "argilla-sdk/**"
+      - "argilla-sdk/docs/**"
```
davidberenstein1957, line 10:
```suggestion
      - "argilla-sdk/**"
```
I think we also need to take into account general code changes to include the potential API reference update?

.github/workflows/argilla-sdk.docs.yml
```diff
@@ -7,28 +7,16 @@ concurrency:
...
     branches:
       # - "main"
-      - "feat/**"
-      - "gh-pages"
+      # - "develop"
+      - "feat/v2.0.0" # Fixing this branch until we merge everything into develop
```
davidberenstein1957, line 15: I assumed we would like to include `feat/**` because we might branch off `feat/v2.0.0` into `feat/`-like branches?

.github/workflows/argilla-sdk.docs.yml
frascuchon: So, could we just add `argilla-sdk/src/**`?

.github/workflows/argilla-sdk.docs.yml
frascuchon: I've included the `docs/**` pattern if we want to have a specific doc version. Otherwise, we will finish with a lot of intermediate versions.

.github/workflows/argilla-sdk.docs.yml
davidberenstein1957: yes.

davidberenstein1957 (approved):

PR #4977, opened 2024-06-07
[FEATURE] SDK - Add support for response status

frascuchon:
# Description

This PR adds support to provide the status for record responses. This is quite important to admin users when restoring annotated data from external sources. Also, the migration script from the legacy dataset can benefit the change.

When several statuses are found for the same user before sending data to the server, a warning is shown to the users and the `draft` status is selected.

**Type of change**

**How Has This Been Tested**

- [ ] Test A
- [ ] Test B

**Checklist**

### DIFF
--- a/argilla-sdk/src/argilla_sdk/responses.py
+++ b/argilla-sdk/src/argilla_sdk/responses.py
@@ -11,23 +11,32 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
-from typing import Any, TYPE_CHECKING, List, Dict, Optional, Iterable
+import warnings
+from enum import Enum
+from typing import Any, TYPE_CHECKING, List, Dict, Optional, Iterable, Union
 from uuid import UUID
 
-from argilla_sdk._models import UserResponseModel, ResponseStatus
+from argilla_sdk._models import UserResponseModel, ResponseStatus as ResponseStatusModel
 from argilla_sdk._resource import Resource
 from argilla_sdk.settings import RankingQuestion
 
 if TYPE_CHECKING:
     from argilla_sdk import Argilla, Dataset, Record
 
-__all__ = ["Response", "UserResponse"]
+__all__ = ["Response", "UserResponse", "ResponseStatus"]
+
+
+class ResponseStatus(str, Enum):
+    """Enum for the status of a response"""
+
+    draft = "draft"
+    submitted = "submitted"
+    discarded = "discarded"
 
 
 class Response:
     """Class for interacting with Argilla Responses of records. Responses are answers to questions by a user.
-    Therefore, a recod question can have multiple responses, one for each user that has answered the question.
+    Therefore, a record question can have multiple responses, one for each user that has answered the question.
     A `Response` is typically created by a user in the UI or consumed from a data source as a label,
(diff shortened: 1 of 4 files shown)


argilla-sdk/src/argilla_sdk/responses.py
```diff
@@ -49,10 +59,14 @@ def __init__(
         if user_id is None:
             raise ValueError("user_id is required")
 
+        if isinstance(status, str):
```
burtenshaw, line 62: Should we warn the user when they create responses with a status of submitted and a different user_id?

burtenshaw (commented):
Just a low comment on warning.

argilla-sdk/src/argilla_sdk/responses.py
frascuchon: Not sure to understand your question. Do you mean setting a different user that connected one? I can see some migration scenarios where an admin user wants to provide several responses from different users. So, in that case makes sense to provide different user IDs. But again, not sure to understand your question correctly.

davidberenstein1957: @burtenshaw I would say no. I think people should consciously change these kinds of things and probably know what they are doing in that case.

argilla-sdk/src/argilla_sdk/responses.py
```diff
@@ -171,16 +184,34 @@ def to_dict(self) -> Dict[str, Any]:
...
+        statuses = set([answer.status for answer in answers if answer.status is not None])
+        if len(statuses) > 1:
+            warnings.warn(f"Multiple status found in user answers. Using {ResponseStatus.draft!r} as default.")
+        elif len(statuses) == 1:
+            return ResponseStatusModel(next(iter(statuses)))
+        return ResponseStatusModel.draft
+
+    def _compute_user_id_from_answers(self, answers: List[Response]) -> UUID:
+        user_ids = set([answer.user_id for answer in answers if answer.user_id is not None])
+        if len(user_ids) == 0:
```
davidberenstein1957: weren't we going to set the user ID to the client one by default? for me this makes the most sense in workflows where an AI engineer decides to add new content and might quickly want to register responses,.

davidberenstein1957: Potentially a warning could fit here.

frascuchon: In deed, this condition won't happen. I will remove it.

argilla-sdk/src/argilla_sdk/responses.py
frascuchon: this feat. is an admin one. I agree with @davidberenstein1957

burtenshaw (approved):

argilla-sdk/tests/integration/test_add_records.py
```diff
@@ -605,17 +604,17 @@ def test_add_records_objects_with_responses(client: Argilla):
     records = [
```
davidberenstein1957, line 604: should we include a test wihout status?

frascuchon: To check the default value, right?

davidberenstein1957: Yes

frascuchon: Done

davidberenstein1957 (approved):

Issue #4946, opened 2024-06-05
[BUG] Check services conectivity on server startup


frascuchon:
When running argilla server v2 the search engine connectivity is not properly checked.

PR #4979, opened 2024-06-07
chore: set logger level to error to reduce noise from Elasticsearch and OpenSearch client libraries

jfcalvo:
# Description

This PR includes some changes to reduce the logger level of Elasticsearch and OpenSearch to error. With this we are reducing a little bit the output noise of these client libraries.

I have improved a little bit the settings so only `elasticsearch` and `opensearch` are valid options for `search_engine` attribute. I have added some small useful functions too.

Refs #4946 

**Type of change**

- [ ] Documentation update

**How Has This Been Tested**

- [x] Manually testing that warning messages are not showed anymore using Elasticsearch and OpenSearch.

**Checklist**

### DIFF
--- a/argilla-server/src/argilla_server/settings.py
+++ b/argilla-server/src/argilla_server/settings.py
@@ -21,7 +21,7 @@ import os
 import re
 import warnings
 from pathlib import Path
-from typing import List, Literal, Optional, Union
+from typing import List, Optional
 from urllib.parse import urlparse
 
 from argilla_server.constants import (
@@ -99,10 +99,7 @@ class Settings(BaseSettings):
 
     es_mapping_total_fields_limit: int = 2000
 
-    search_engine: Union[
-        Literal[SEARCH_ENGINE_ELASTICSEARCH],
-        Literal[SEARCH_ENGINE_OPENSEARCH],
-    ] = SEARCH_ENGINE_ELASTICSEARCH
+    search_engine: str = SEARCH_ENGINE_ELASTICSEARCH
 
     vectors_fields_limit: int = Field(
         default=5,
@@ -229,13 +226,6 @@ class Settings(BaseSettings):
     @property
     def search_engine_is_opensearch(self) -> bool:
         return self.search_engine == SEARCH_ENGINE_OPENSEARCH
-
-    @property
-    def humanized_search_engine(self) -> str:
-        if self.search_engine_is_elasticsearch:
-            return "Elasticsearch"
-        elif self.search_engine_is_opensearch:
-            return "OpenSearch"
 
     def obfuscated_elasticsearch(self) -> str:
         """Returns configured elasticsearch url obfuscating the provided password, if any"""
--- a/argilla-server/src/argilla_server/settings.py
+++ b/argilla-server/src/argilla_server/settings.py
@@ -21,7 +21,7 @@ import os
 import re
 import warnings
 from pathlib import Path
-from typing import List, Optional
+from typing import List, Literal, Optional, Union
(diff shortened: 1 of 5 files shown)


argilla-server/src/argilla_server/settings.py
```diff
@@ -97,7 +99,10 @@ class Settings(BaseSettings):
 
     es_mapping_total_fields_limit: int = 2000
 
-    search_engine: str = "elasticsearch"
+    search_engine: Union[
+        Literal[SEARCH_ENGINE_ELASTICSEARCH],
+        Literal[SEARCH_ENGINE_OPENSEARCH],
+    ] = SEARCH_ENGINE_ELASTICSEARCH
```
frascuchon: Even if we can define constants for official engines, we shouldn't fix the available options. 

Creating new custom engine or specific engine connection could be helpful for the community but, by fixing this we’re avoiding this possibility. (See how engine [are registered](https://github.com/argilla-io/argilla/blob/chore/set-logger-level-to-error-for-search-engine/argilla-server/src/argilla_server/search_engine/elasticsearch.py#L40))

argilla-server/src/argilla_server/settings.py
```diff
@@ -217,6 +222,21 @@ def old_dataset_records_index_name(self) -> str:
...
+    @property
+    def search_engine_is_elasticsearch(self) -> bool:
+        return self.search_engine == SEARCH_ENGINE_ELASTICSEARCH
+
+    @property
+    def search_engine_is_opensearch(self) -> bool:
+        return self.search_engine == SEARCH_ENGINE_OPENSEARCH
+
+    @property
+    def humanized_search_engine(self) -> str:
```
frascuchon: The name of this method gives me chills...

"I've seen things you people wouldn't believe. Attack ships on fire off the shoulder of Orion. I watched C-beams glitter in the dark near the Tannhäuser Gate. All those moments will be lost in time like tears in rain..." 

Could we just call `human_readable_search_engine`? And also generalize a bit by using str.title() ? I think we don't need al this extra functions if we keep simple

frascuchon (changes requested):

argilla-server/src/argilla_server/settings.py
jfcalvo: "Humanize" is a very common concept and naming when formatting strings but I don't have any problem changing it to "human_readable_*". 🙂

Some examples:
* One Python library: https://python-humanize.readthedocs.io/en/latest/
* A Django module: https://docs.djangoproject.com/en/5.0/ref/contrib/humanize/
* Moment.js classic JS library: https://momentjs.com/docs/#/durations/humanize/
* Size: an open source library that I made some years ago https://github.com/jfcalvo/size?tab=readme-ov-file#humanize-file-sizes

argilla-server/src/argilla_server/settings.py
jfcalvo: But in the case that a PR from the community adds a new engine they should modify this `Union` as part of the PR right? I don't get the problem.

> (See how engine [are registered](https://github.com/argilla-io/argilla/blob/chore/set-logger-level-to-error-for-search-engine/argilla-server/src/argilla_server/search_engine/elasticsearch.py#L40))

I missed that one. I will use the constant there too.

jfcalvo:
@frascuchon I applied feedback from our discussion:
* Removed the `humanized_search_engine` property from settings.
* Removed the `Union` restricting the search_engine only to `elasticsearch` or `opensearch` values.

frascuchon (approved):

PR #5001, opened 2024-06-12
🎯 poc/string length

damianpumar:
Screenshot 2024-06-12 at 12 58 01

### DIFF
--- a/emojis.py
+++ b/emojis.py
@@ -1,0 +1,12 @@
+def jslen(string):
+    return int(len(string.encode(encoding='utf_16_le'))/2)
+
+
+array = [
+  'This is a message with a 👍',
+  'This is a message with a 👍🏾',
+  'This is a message with a 👍🏾👍',
+]
+
+for i in array:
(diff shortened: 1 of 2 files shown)


emojis.py
```diff
@@ -0,0 +1,12 @@
+def jslen(string):
+    return int(len(string.encode(encoding='utf_16_le'))/2)
```
jfcalvo, line 2: My only concern is that I don't understand why we need to specify the encoding to be `utf_16_le` when we are using utf-8 for everything.

frascuchon: Also, the problem to solve is to keep aligned string lengths between python and javascript. Spans boundaries are defined in python, and probably will be processed in python. We need to ensure that the same piece of text will be processed in both python and javascript. Here is an article about span processing 

https://betterprogramming.pub/slicing-strings-containing-emoji-differences-between-python-and-javascript-4716c419718f

damianpumar: Hi team, this feat does not work with utf 8

PR #5002, opened 2024-06-12
[DOCS] add migration notebook to docs format

burtenshaw:
This PR transfer the notebook guide by @frascuchon and @nataliaElv over to the documentation format, which makes it easier to follow.

You can find the page here: http://localhost:8000/argilla-python/guides/how_to_guides/migrate_from_legacy_datasets/

n.b. the page is not available via the menu. @davidberenstein1957 we should figure out navigation presentation for this page.

The guide looks like this:

image


davidberenstein1957:
Hi @burtenshaw, looks good. Adding it to guides makes sense to me too.


I have not taken the time to look at the migration guides yet but I think we might be able to make the migration a lot more user-friendly by providing code that maps everything to the Datasets using the 2.0 SDK automatically.

TextClassification -> TextField ("name=text") | MultiLabel|LabelQuestion ("name=label") + metadata configs
TokenClassification -> TextField ("name=text") | SpanQuestion ("name=span") + metadata configs
Text2Text -> TextField ("name=text") | TextQuestion ("name=text") + metadata configs

It takes us some more time to figure out the logic and mappings but will save a lot of community members a lot of time and effort too. @frascuchon @burtenshaw WDYT?

nataliaElv:
The document says that Feedback datasets do not need to be migrated, but if you have a server with legacy+feedback datasets I assume that one would like to move all of them over to the new server, right? @burtenshaw 

We may also want to discuss what to do with the migration of predictions to suggestions in multi-label text classification datasets. If I'm not mistaken, with the current code all labels would be selected as suggestions (as in legacy datasets you can have a prediction for each label). Maybe we can add a threshold that the user can modify so only labels where `score > threshold` are added to the suggestions?

davidberenstein1957:
> The document says that Feedback datasets do not need to be migrated, but if you have a server with legacy+feedback datasets I assume that one would like to move all of them over to the new server, right? @burtenshaw 

@nataliaElv I don't think the FeedbackDataset need to be moved. The underlying server logic is the same but the way of interacting with the server through the new SDK as API is different.

davidberenstein1957:
> We may also want to discuss what to do with the migration of predictions to suggestions in multi-label text classification datasets. If I'm not mistaken, with the current code all labels would be selected as suggestions (as in legacy datasets you can have a prediction for each label). Maybe we can add a threshold that the user can modify so only labels where score > threshold are added to the suggestions?

@nataliaElv Good catch. I think we need to provide a warning and on let the user choose to add the score processing mechanism, which might be as simple as choosing the highest or lowest one.

burtenshaw:
> TextClassification -> TextField ("name=text") | MultiLabel|LabelQuestion ("name=label") + metadata configs TokenClassification -> TextField ("name=text") | SpanQuestion ("name=span") + metadata configs Text2Text -> TextField ("name=text") | TextQuestion ("name=text") + metadata configs

This is a nice suggestion but I think it depends on use, and the metrics are declining on legacy datasets. 

Personally, I would implement this based on community engagement with this guide.

burtenshaw:
> > The document says that Feedback datasets do not need to be migrated, but if you have a server with legacy+feedback datasets I assume that one would like to move all of them over to the new server, right? @burtenshaw
> 
> @nataliaElv I don't think the FeedbackDataset need to be moved. The underlying server logic is the same but the way of interacting with the server through the new SDK as API is different.

@nataliaElv Is on to something here. The guide proposes that the v2 server is 'their new server'. So from a user persepctive, they would want their datasets migrated. I will add the a snippet that shows them how to `client.datasets` on all other datasets.

burtenshaw:
> > We may also want to discuss what to do with the migration of predictions to suggestions in multi-label text classification datasets. If I'm not mistaken, with the current code all labels would be selected as suggestions (as in legacy datasets you can have a prediction for each label). Maybe we can add a threshold that the user can modify so only labels where score > threshold are added to the suggestions?
> 
> @nataliaElv Good catch. I think we need to provide a warning and on let the user choose to add the score processing mechanism, which might be as simple as choosing the highest or lowest one.

I would add this to the tab with multilabel text classification as a note box.

argilla/docs/guides/how_to_guides/migrate_from_legacy_datasets.md
```diff
@@ -0,0 +1,200 @@
...
+!!! note
+    Legacy Datasets include: `DatasetForTextClassification`, `DatasetForTokenClassification`, and `DatasetForText2Text`.
+
+    `FeedbackDataset`'s do not need to be migrated as they are already in the Argilla V2 format.
+
+To follow this guide, you will need to have the following prerequisites:
+
+- An argilla 1.* server instance running with legacy datasets.
+- An argilla 2.* server instance running. If you don't have one, you can create one by following the [Argilla installation guide](../../getting_started/installation.md).
+- The `argilla` sdk package installed in your environment.
```
frascuchon: Indeed, the argilla v2 SDK is fully compatible with argilla-server >=1.29. So, maybe we can notify users if they want to use the same argilla-server instance in this process. 

So, users can still see both versions of the dataset (legacy and (Feedback)dataset). Once everything is fine, they can still upgrade the server to v2.

argilla/docs/guides/how_to_guides/migrate_from_legacy_datasets.md
```diff
@@ -0,0 +1,200 @@
...
+```
+
+Next, load the dataset settings and records from the Argilla V1 server:
+
+```python
+dataset_name = "news-programmatic-labeling"
+workspace = "demo"
+
+settings_v1 = rg_v1.load_dataset_settings(dataset_name, workspace)
+records_v1 = rg_v1.load(dataset_name, workspace, limit=100, query="_exists_:annotated_by")
```
frascuchon: This is just a sample query. The final version of this guide should remove the `limit` and the `query`
```suggestion
records_v1 = rg_v1.load(dataset_name, workspace)
```

argilla/docs/guides/how_to_guides/migrate_from_legacy_datasets.md
```diff
@@ -0,0 +1,200 @@
...
+            # The inputs field should be a dictionary with the same keys as the `fields` in the settings
+            metadata=data["metadata"],
+            # The metadata field should be a dictionary with the same keys as the `metadata` in the settings
+            vectors=vectors,
+            # The vectors field should be a dictionary with the same keys as the `vectors` in the settings
+            suggestions=suggestions,
+            responses=responses,
+        )
+    ```
+
```
frascuchon: I didn't check it but we can use it with a real dataset during the tutorial QA.
```suggestion
=== "For token classification"

def map_to_record_for_span(data: dict, users_by_name: dict, current_user: rg.User) -> rg.Record:
    suggestions = []
    responses = []

    if data.get("prediction"):
        # Prediction for token classification will be a list of tuple (label, start, end)
        spans = [
            {"start": start, "end": end, "label": label}
            for label, start, end in data["prediction"]
        ]
        agent = data.get("prediction_agent")
        suggestions.append(rg.Suggestion(question_name="labels", value=spans, agent=agent))

    if data.get("annotation"):
        # From data[annotation] and data[annotation_agent]
        spans = [
            {"start": start, "end": end, "label": label}
            for label, start, end in data["annotation"]
        ]
        user_id = users_by_name.get(data["annotation_agent"], current_user).id
        responses.append(rg.Response(question_name="label", value=spans, user_id=user_id))

    if data.get("vectors"):
        # From data["vectors"]
        vectors = [rg.Vector(name=name, values=value) for name, value in data["vectors"].items()]

    return rg.Record(
        id=data["id"],
        fields=data["inputs"],
        # The inputs field should be a dictionary with the same keys as the `fields` in the settings
        metadata=data["metadata"],
        # The metadata field should be a dictionary with the same keys as the `metadata` in the settings
        vectors=vectors,
        # The vectors field should be a dictionary with the same keys as the `vectors` in the settings
        suggestions=suggestions,
        responses=responses,
    )

```

argilla/docs/guides/how_to_guides/migrate_from_legacy_datasets.md
```diff
@@ -0,0 +1,272 @@
...
+```python
+import argilla.v1 as rg_v1
+
+# Initialize the API with an Argilla server less than 2.0
+old_client = rg.Argilla(old_server_api_url, old_server_api_key)
+new_client = rg.Argilla(new_server_api_url, new_server_api_key) 
+
+dataset_name = "feedback-dataset"
+old_dataset = old_client.datasets(name=dataset_name)
+new_dataset = new_client.datasets.add(dataset)
```
frascuchon: wrong naming or missing def?
```suggestion
new_dataset = new_client.datasets.add(old_dataset)
```

frascuchon (approved):

argilla/docs/how_to_guides/migrate_from_legacy_datasets.md
```diff
@@ -1 +1,287 @@
...
+        responses = []
+    
+        if prediction := data.get("prediction"):
+            scores = [span["score"] for span in prediction]
+            agent = data["prediction_agent"]
+            suggestions.append(rg.Suggestion(question_name="labels", value=prediction, score=scores, agent=agent))
+    
+        if annotation := data.get("annotation"):
+            user_id = users_by_name.get(data["annotation_agent"], current_user).id
+            responses.append(rg.Response(question_name="spans", value=annotation, user_id=user_id))
```
nataliaElv, line 190: `question_name` is different for `rg.Suggestion` and `rg.Response`, it should be the same, whether that's `labels` or `spans`.

argilla/docs/how_to_guides/migrate_from_legacy_datasets.md
```diff
@@ -1 +1,287 @@
...
+```python
+import argilla.v1 as rg_v1
+
+# Initialize the API with an Argilla server less than 2.0
+old_client = rg.Argilla(old_server_api_url, old_server_api_key)
+new_client = rg.Argilla(new_server_api_url, new_server_api_key) 
+
+dataset_name = "feedback-dataset"
+old_dataset = old_client.datasets(name=dataset_name)
+new_dataset = new_client.datasets.add(dataset)
```
nataliaElv: I guess this line should be `new_dataset = new_client.datasets.add(old_dataset)`? But it still throws an error even with the change.

argilla/docs/how_to_guides/migrate_from_legacy_datasets.md
frascuchon, line 190:
```suggestion
            responses.append(rg.Response(question_name="labels", value=annotation, user_id=user_id))
```

argilla/docs/how_to_guides/migrate_from_legacy_datasets.md
```diff
@@ -1 +1,287 @@
...
+dataset.records.log(records)
+```
+You have now successfully migrated your legacy dataset to Argilla V2. For more guides on how to use the Argilla SDK, please refer to the [How to guides](index.md).
+
+## Migrating Feedback Datasets on your Argilla 1.* server
+
+As mentioned above, `FeedbackDataset`'s are compatible with Argilla V2 and do not need to be reformatted. However, you may want to migrate your feedback datasets to the new server so that you can deprecate your Argilla 1.* server. Here is a guide on how to migrate your feedback datasets:
+
+```python
+import argilla.v1 as rg_v1
```
frascuchon:
```suggestion
import argilla as rg
```

argilla/docs/how_to_guides/migrate_from_legacy_datasets.md
```diff
@@ -1 +1,287 @@
...
+old_dataset = old_client.datasets(name=dataset_name)
+new_dataset = new_client.datasets.add(old_dataset)
+
+# Load the records from the old server
+new_dataset.records.log(
+    old_dataset.records(
+        with_responses=True, # (1)
+        with_suggestions=True,
+        with_vectors=True,
+        with_metadata=True,
```
frascuchon: This argument does not exists.
```suggestion
```

argilla/docs/how_to_guides/migrate_from_legacy_datasets.md
```diff
@@ -1 +1,286 @@
...
+new_dataset.records.log(
+    old_dataset.records(
+        with_responses=True, # (1)
+        with_suggestions=True,
+        with_vectors=True,
+    )
+)
+```
+
+1. The `with_responses`, `with_suggestions`, `with_vectors`, and `with_metadata` flags are used to load the records with the responses, suggestions, vectors, and metadata respectively.
```
frascuchon: This scenario has several problems. 

1. We cannot share resources between clients, or even create new resources from other resources. Those actions will modify the internal state of the original resource, so old_dataset settings will change when the new_dataset is created
2. Copy responses from one server to another requires a revision of response users. User IDs won't exist in the target server so we should provide a user mapping in the same way we do for legacy datasets.

Not sure how common would be this scenario. But I propose to skip this for now and think a bit more about how the SDK should work for this. cc @burtenshaw

frascuchon: Also, suggestions are linked to questions by a question ID. When we do:
```python
new_dataset.records.log(
    list(old_dataset.records(with_responses=False, with_suggestions=True, with_vectors=True)),
)
```
an error will be raised since the error contains wrong question info for the provided dataset.

PR #5009, opened 2024-06-13
[DOCS] Update and define all record mapping functions

frascuchon:
Helper functions for record mapping have been reviewed and tested with several datasets.


burtenshaw (approved):
Nice work @frascuchon . Thanks!

PR #5026, opened 2024-06-14
[ENHANCEMENT] Argilla SDK: Updating record fields and  vectors

frascuchon:
# Description

This PR reviews the record attributes and normalizes how to work with fields, vectors, and metadata. Now, all are treated as dictionaries and users can update in the same way that working with dictionaries or creating new attributes:

```python
record = Record(fields={"name": "John"})

record.fields.update({"name": "Jane", "age": "30"})
record.fields.new_field = "value"

record.vectors["new-vector"] = [1.0, 2.0, 3.0]
record.vectors.vector = [1.0, 2.0, 3.0]

record.metadata["new-key"] = "new_value"
record.metadata.key = "new_value"

```

Once this approach is approved, I will create a new PR changing the docs.

**Type of change**

**How Has This Been Tested**

- [ ] Test A
- [ ] Test B

**Checklist**

### DIFF
--- a/argilla/src/argilla/records/_dataset_records.py
+++ b/argilla/src/argilla/records/_dataset_records.py
@@ -412,7 +412,7 @@ class DatasetRecords(Iterable[Record], LoggingMixin):
         metadata: Dict[str, MetadataValue] = {}
 
         responses: List[Response] = []
-        suggestion_values = defaultdict(dict)
+        suggestion_values: Dict[str, dict] = defaultdict(dict)
 
         schema = self.__dataset.schema
 
--- a/argilla/src/argilla/records/_dataset_records.py
+++ b/argilla/src/argilla/records/_dataset_records.py
@@ -21,7 +21,7 @@ from tqdm import tqdm
 
 from argilla._api import RecordsAPI
 from argilla._helpers import LoggingMixin
-from argilla._models import RecordModel, MetadataValue
+from argilla._models import RecordModel, MetadataValue, VectorValue, FieldValue
 from argilla.client import Argilla
 from argilla.records._io import GenericIO, HFDataset, HFDatasetsIO, JsonIO
 from argilla.records._resource import Record
@@ -405,12 +405,14 @@ class DatasetRecords(Iterable[Record], LoggingMixin):
         Returns:
             A Record object.
         """
-        fields: Dict[str, str] = {}
-        responses: List[Response] = []
         record_id: Optional[str] = None
-        suggestion_values = defaultdict(dict)
-        vectors: List[Vector] = []
+
+        fields: Dict[str, FieldValue] = {}
+        vectors: Dict[str, VectorValue] = {}
         metadata: Dict[str, MetadataValue] = {}
+
+        responses: List[Response] = []
+        suggestion_values = defaultdict(dict)
 
         schema = self.__dataset.schema
 
@@ -466,7 +468,7 @@ class DatasetRecords(Iterable[Record], LoggingMixin):
                     {"value": value, "question_name": attribute, "question_id": schema_item.id}
                 )
             elif isinstance(schema_item, VectorField):
-                vectors.append(Vector(name=attribute, values=value))
+                vectors[attribute] = value
             elif isinstance(schema_item, MetadataPropertyBase):
                 metadata[attribute] = value
             else:
@@ -478,9 +480,9 @@ class DatasetRecords(Iterable[Record], LoggingMixin):
         return Record(
             id=record_id,
             fields=fields,
-            suggestions=suggestions,
-            responses=responses,
             vectors=vectors,
             metadata=metadata,
+            suggestions=suggestions,
+            responses=responses,
             _dataset=self.__dataset,
         )
--- a/argilla/src/argilla/records/_resource.py
+++ b/argilla/src/argilla/records/_resource.py
@@ -277,7 +277,7 @@ class RecordFields(dict):
         self[key] = value
 
     def to_dict(self) -> dict:
-        return {key: value for key, value in self.items()}
+        return dict(self.items())
 
 
 class RecordMetadata(dict):
@@ -293,7 +293,7 @@ class RecordMetadata(dict):
         self[key] = value
 
     def to_dict(self) -> dict:
-        return {key: value for key, value in self.items()}
+        return dict(self.items())
 
     def api_models(self) -> List[MetadataModel]:
         return [MetadataModel(name=key, value=value) for key, value in self.items()]
@@ -313,11 +313,11 @@ class RecordVectors(dict):
     def __setattr__(self, key: str, value: VectorValue):
         self[key] = value
 
+    def to_dict(self) -> Dict[str, List[float]]:
+        return dict(self.items())
+
     def api_models(self) -> List[VectorModel]:
         return [Vector(name=name, values=value).api_model() for name, value in self.items()]
-
-    def to_dict(self) -> Dict[str, List[float]]:
-        return {key: value for key, value in self.items()}
 
 
 class RecordResponses(Iterable[Response]):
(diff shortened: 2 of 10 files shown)


burtenshaw:
@frascuchon @davidberenstein1957 

Looks nice, just some in line comments and queries.

argilla/src/argilla/records/_dataset_records.py
```diff
@@ -405,13 +405,15 @@ def _infer_record_from_mapping(
...
         record_id: Optional[str] = None
-        suggestion_values = defaultdict(dict)
-        vectors: List[Vector] = []
+
+        fields: Dict[str, FieldValue] = {}
+        vectors: Dict[str, VectorValue] = {}
         metadata: Dict[str, MetadataValue] = {}
 
+        responses: List[Response] = []
+        suggestion_values = defaultdict(dict)
```
burtenshaw: Should we be defining types for suggestions as well?

argilla/src/argilla/records/_resource.py
```diff
@@ -258,27 +262,62 @@ def from_model(cls, model: RecordModel, dataset: "Dataset") -> "Record":
...
-    def __repr__(self) -> str:
-        return self.to_dict().__repr__()
+
+class RecordMetadata(dict):
+    """This is a container class for the metadata of a Record."""
+
+    def __init__(self, metadata: Optional[Dict[str, MetadataValue]] = None) -> None:
+        super().__init__(metadata or {})
+
+    def __getattr__(self, item: str):
```
burtenshaw, line 289: Are we definitely going to support both of these strategies?
```python
record.vectors["new-vector"] = [1.0, 2.0, 3.0]
record.vectors.vector = [1.0, 2.0, 3.0]

record.metadata["new-key"] = "new_value"
record.metadata.key = "new_value"
```

Personally, I'm not 100% sure about the attribute based approach because I don't like how the IDE behaves. Is there a way to improve the experience there?

argilla/src/argilla/records/_resource.py
```diff
@@ -258,27 +262,62 @@ def from_model(cls, model: RecordModel, dataset: "Dataset") -> "Record":
...
+        self[key] = value
+
+    def api_models(self) -> List[VectorModel]:
+        return [Vector(name=name, values=value).api_model() for name, value in self.items()]
+
+    def to_dict(self) -> Dict[str, List[float]]:
+        return {key: value for key, value in self.items()}
 
 
 class RecordResponses(Iterable[Response]):
```
burtenshaw, line 323: How will this class be used when users want to:
- add a first response after retrieving records?
- add subsequent responses after retrieving records?

Ideally, these should just be the same.

argilla/src/argilla/records/_dataset_records.py
davidberenstein1957: I would say so, yes. Perhaps can be covered in another issue, because it seems like a simple but time-consuming thing.

argilla/src/argilla/records/_resource.py
```diff
@@ -258,27 +262,62 @@ def from_model(cls, model: RecordModel, dataset: "Dataset") -> "Record":
...
 
-    def __iter__(self):
-        return iter(self.__fields)
+    def __setattr__(self, key: str, value: MetadataValue):
+        self[key] = value
 
-    def to_dict(self) -> Dict[str, Union[str, None]]:
-        return self.__fields
+    def to_dict(self) -> dict:
+        return {key: value for key, value in self.items()}
```
davidberenstein1957:
```suggestion
        return dict(self.items())
```

argilla/src/argilla/records/_resource.py
davidberenstein1957: I think it is clearer to only support one, because otherwise, we might end up with these too fragmented approaches for doing the same thing again.

argilla/src/argilla/records/_resource.py
```diff
@@ -258,27 +262,62 @@ def from_model(cls, model: RecordModel, dataset: "Dataset") -> "Record":
...
+        super().__init__(metadata or {})
+
+    def __getattr__(self, item: str):
+        return self[item]
+
+    def __setattr__(self, key: str, value: MetadataValue):
+        self[key] = value
+
+    def to_dict(self) -> dict:
+        return {key: value for key, value in self.items()}
```
davidberenstein1957:
```suggestion
        return dict(self.items())
```

argilla/src/argilla/records/_resource.py
```diff
@@ -258,27 +262,62 @@ def from_model(cls, model: RecordModel, dataset: "Dataset") -> "Record":
...
+        return self[item]
+
+    def __setattr__(self, key: str, value: VectorValue):
+        self[key] = value
+
+    def api_models(self) -> List[VectorModel]:
+        return [Vector(name=name, values=value).api_model() for name, value in self.items()]
+
+    def to_dict(self) -> Dict[str, List[float]]:
+        return {key: value for key, value in self.items()}
```
davidberenstein1957:
```suggestion
        return dict(self.items())
```

argilla/tests/unit/export/test_record_export_import_compatibillity.py
```diff
@@ -22,22 +22,19 @@
...
             rg.Suggestion("label", "positive", score=0.9),
             rg.Suggestion("topics", ["topic1", "topic2"], score=[0.9, 0.8]),
         ],
-        responses=[rg.Response("label", "positive", user_id=user_id)],
+        responses=[rg.Response("label", "positive", user_id=uuid.uuid4())],
         metadata={"source": "twitter", "language": "en"},
-        vectors=[rg.Vector("text", [0, 0, 0])],
-        id=str(uuid.uuid4()),
+        vectors={"text": [0, 0, 0]},
+
```
davidberenstein1957, line 37:
```suggestion
```

argilla/tests/unit/test_resources/test_records.py
```diff
@@ -56,7 +56,36 @@ def test_update_record_metadata_by_attribute(self):
...
+    def test_update_record_fields_by_attribute(self):
+        record = Record(fields={"name": "John"})
+
+        record.fields.name = "Jane"
+        record.fields.age = "30"
+        record.fields.new_field = "value"
+
+        assert record.fields == {"name": "Jane", "age": "30", "new_field": "value"}
+
+    def test_update_record_vectors(self):
```
davidberenstein1957, line 81: should we test for update with wrong vector length?

argilla/tests/unit/test_resources/test_records.py
```diff
@@ -56,7 +56,36 @@ def test_update_record_metadata_by_attribute(self):
...
+
+    def test_update_record_fields(self):
+        record = Record(fields={"name": "John"})
+
+        record.fields.update({"name": "Jane", "age": "30"})
+        record.fields["new_field"] = "value"
+
+        assert record.fields == {"name": "Jane", "age": "30", "new_field": "value"}
+
+    def test_update_record_fields_by_attribute(self):
```
davidberenstein1957, line 72: should we test for update with empty required field?

davidberenstein1957 (commented):
Left some comments.

argilla/src/argilla/records/_resource.py
frascuchon: Yes. I agree with this. I've aligned fields and vectors with metadata, supporting both approaches. Working with dict is fine for me, and is clearest to users since in the doc they are described as dict. Same feeling with IDE warnings when using attributes. I would say to keep the dict-like access and raise an error when accessing using attributes (I think the default behavior is do-nothing)

argilla/src/argilla/records/_resource.py
frascuchon: For me, both cases should be something like this:

```python
record.responses.add(rg.Response(...)..)

# or 
record.responses.add[rg.Response(...), rg.Response(...)])
```

argilla/src/argilla/records/_resource.py
frascuchon: A question here, should we do the same for responses and suggestions? Now, we support:

```python
record.responses.label 
record.suggestions.text
```

which have the same IDE problems.

argilla/tests/unit/test_resources/test_records.py
frascuchon: Validations are not executed at this point yet. Also, the record should have the inner dataset (or the settings) to know the context of validations. I would skip this for now. In the future, we should add validations at some point when log records.

argilla/tests/unit/test_resources/test_records.py
frascuchon: Same here. Validations are not happening here. Also, for now, fields cannot be updated. We can skip this test.

davidberenstein1957 (approved):

argilla/src/argilla/records/_resource.py
burtenshaw: Ok. So we'll open a new PR for suggestions and responses, where we'll use a dictionary style access as well? Nice.

burtenshaw (approved):

PR #5040, opened 2024-06-18
[ENHANCEMENTE] `argilla`: support python 3.12

frascuchon:
Relax the python requirements to support python 3.12

### DIFF
--- a/.github/workflows/argilla-frontend.yml
+++ b/.github/workflows/argilla-frontend.yml
@@ -10,12 +10,6 @@ on:
   pull_request:
     paths:
       - "argilla-frontend/**"
-    types:
-      - opened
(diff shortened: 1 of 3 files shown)


.github/workflows/argilla-frontend.yml
```diff
@@ -10,12 +10,6 @@ on:
   pull_request:
     paths:
       - "argilla-frontend/**"
-    types:
-      - opened
-      - edited
-      - reopened
-      - synchronize
-      - ready_for_review
```
frascuchon, line 18: This is just to avoid running the argilla-frontend workflow on every PR

gabrielmbmb (approved):

PR #5047, opened 2024-06-18
[CI] docs: Configure docs publish for releases

frascuchon:
This PR reviews the workflow for publishing docs:

- A merge to the `main`  (release/stable branch) will deploy docs `latest` version.
- A tag creation (tag pattern = v*) will deploy `vX.Y.Z` version. (we can iterate on this and keep just the minor version)

### DIFF
--- a/.github/workflows/argilla.docs.yml
+++ b/.github/workflows/argilla.docs.yml
@@ -11,11 +11,11 @@ on:
     paths:
       - "argilla/docs/**"
       - "argilla/src/**"
+      - "argilla/mkdocs.yml"
     branches:
       - "main"
       - "develop"
       - "feat/v2.0.0" # Fixing this branch until we merge everything into develop
-      - "docs/**" # When we create a new branch only for docs
 
 defaults:
   run:
@@ -68,11 +68,3 @@ jobs:
 
       - run: pdm run mike deploy ${{ github.ref_name }}
         if: startsWith(github.ref, 'refs/tags/')
-
-      - name: Extract branch name
-        shell: bash
-        run: echo "branch_name=${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}}" | sed 's/\//_/g'  >> $GITHUB_OUTPUT
-        id: extract_branch_name
-
-      - run: pdm run mike deploy ${{ steps.extract_branch_name.outputs.branch_name }} --push
-        if: startsWith(github.ref, 'refs/heads/docs') || startsWith(github.head_ref, 'docs/')
--- a/.github/workflows/argilla.docs.yml
+++ b/.github/workflows/argilla.docs.yml
@@ -6,12 +6,14 @@ concurrency:
 
 on:
   push:
+    tags:
+      - "v*"
     paths:
       - "argilla/docs/**"
       - "argilla/src/**"
     branches:
-      # - "main"
-      # - "develop"
+      - "main"
+      - "develop"
       - "feat/v2.0.0" # Fixing this branch until we merge everything into develop
       - "docs/**" # When we create a new branch only for docs
 
@@ -22,10 +24,10 @@ defaults:
 jobs:
   publish:
     runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
-      matrix:
-        python-version: ["3.10"]
+
+    env:
+      GH_ACCESS_TOKEN: ${{ secrets.GH_ACCESS_TOKEN }}
+
     steps:
       - name: checkout docs-site
         uses: actions/checkout@v4
@@ -37,10 +39,10 @@ jobs:
       - name: Setup Python
         uses: pdm-project/setup-pdm@v4
         with:
-          python-version: ${{ matrix.python-version }}
+          python-version: "3.10"
           cache: true
           cache-dependency-path: |
-            argilla/pyproject.toml
+            argilla/pdm.lock
 
       - name: Install dependencies
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/argilla.docs.yml
```diff
@@ -6,12 +6,14 @@ concurrency:
...
     paths:
       - "argilla/docs/**"
       - "argilla/src/**"
     branches:
-      # - "main"
-      # - "develop"
+      - "main"
+      - "develop"
       - "feat/v2.0.0" # Fixing this branch until we merge everything into develop
       - "docs/**" # When we create a new branch only for docs
```
davidberenstein1957:
```suggestion
```

.github/workflows/argilla.docs.yml
```diff
@@ -6,12 +6,14 @@ concurrency:
 
 on:
   push:
+    tags:
+      - "v*"
     paths:
       - "argilla/docs/**"
       - "argilla/src/**"
```
davidberenstein1957, line 13:
```suggestion
      - "argilla/docs/**"
      - "argilla/src/**"
      - "argilla/mkdocs.yml"
```

.github/workflows/argilla.docs.yml
```diff
@@ -55,16 +57,17 @@ jobs:
...
-          GH_ACCESS_TOKEN: ${{ secrets.GH_ACCESS_TOKEN }}
 
-      - run: pdm run mike deploy ${{ github.ref_name }} latest --update-aliases --push
+      - run: pdm run mike deploy ${{ github.ref_name }}
         if: startsWith(github.ref, 'refs/tags/')
-        env:
-          GH_ACCESS_TOKEN: ${{ secrets.GH_ACCESS_TOKEN }}
 
       - name: Extract branch name
         shell: bash
```
davidberenstein1957:
```suggestion
```

.github/workflows/argilla.docs.yml
```diff
@@ -73,5 +76,3 @@ jobs:
 
       - run: pdm run mike deploy ${{ steps.extract_branch_name.outputs.branch_name }} --push
         if: startsWith(github.ref, 'refs/heads/docs') || startsWith(github.head_ref, 'docs/')
```
davidberenstein1957:
```suggestion
```

.github/workflows/argilla.docs.yml
```diff
@@ -55,16 +57,17 @@ jobs:
...
           echo "${{ github.head_ref }}"
 
+      - run: |
+          pdm run mike deploy latest --push
+          pdm run mike set-default latest
+        if: github.ref == 'refs/heads/main'
+
       - run: pdm run mike deploy dev --push
         if: github.ref == 'refs/heads/feat/v2.0.0'
         # if: github.ref == 'refs/heads/develop'
```
davidberenstein1957, line 67:
```suggestion
```

.github/workflows/argilla.docs.yml
```diff
@@ -55,16 +57,17 @@ jobs:
...
-        env:
-          GH_ACCESS_TOKEN: ${{ secrets.GH_ACCESS_TOKEN }}
 
-      - run: pdm run mike deploy ${{ github.ref_name }} latest --update-aliases --push
+      - run: pdm run mike deploy ${{ github.ref_name }}
```
davidberenstein1957, line 69:
```suggestion
      - run: pdm run mike deploy ${{ github.ref_name }} latest --push
```
I believe we would still need to push as latstt if we add tags, to ensure we get directed to the latest version.

.github/workflows/argilla.docs.yml
```diff
@@ -55,16 +57,17 @@ jobs:
         run: echo "${{ github.ref }}"
           echo "${{ github.head_ref }}"
 
+      - run: |
+          pdm run mike deploy latest --push
+          pdm run mike set-default latest
```
davidberenstein1957, line 62:
```suggestion
          pdm run mike deploy develop --push
```
I would maybe just call the latest version of the docs/develop. I think we should try avoid the flow you had because merging new feature into main might then cause in doc updates without the supported feature.

davidberenstein1957 (commented):
Some small changes.
- let's not deplot docs/ now
- deploy a `develop` version with latest changes but not set this as default
- deploy new versions on release tags and set it to the latest tag by default.

.github/workflows/argilla.docs.yml
frascuchon: We still have 2 branches: `develop` and `main`. The one for continuous development is `develop` and the one with the stable version of argilla is `main`.  So `develop` should generate a `dev` version of the docs, and `main` update the latest release. We can also generate tagged versions when a new tag is created (should be created from the main branch). And we can define more rules to protect merges into the main branch.

.github/workflows/argilla.docs.yml
frascuchon: Since the latest version is built from `main` ( a release will be published when new commits are detected in the `main` branch), we can deploy it as the latest one.

Having `latest` linked to tags can introduce errors when creating patch releases for older versions

.github/workflows/argilla.docs.yml
frascuchon: I prefer to remove the feat/v2.0.0 once we merge the feat branch to develop

.github/workflows/argilla.docs.yml
frascuchon: This step applies only to the main branch (not develop)

.github/workflows/argilla.docs.yml
frascuchon: This would be the develop branch (once we have the v2 changes merged into develop). We should generate the `dev` version from here.

.github/workflows/argilla.docs.yml
frascuchon: Does it make sense to move mkdocs.yml under argilla/docs folder?

davidberenstein1957: Yes, we can and I think we should be able to via `mike deploy` params
`-F FILE, --config-file FILE the MkDocs configuration file to use`

frascuchon:
> Some small changes.
> 
> * let's not deplot docs/ now

:+1 I've removed conditions for that!

> * deploy a `develop` version with latest changes but not set this as default

Normally the `latest` docs refer to the latest release. We already have the `dev` version for this.

> * deploy new versions on release tags and set it to the latest tag by default.

We can have version problems publishing patch releases. The latest changes for docs will be on the last commit on `main`.

.github/workflows/argilla.docs.yml
frascuchon: Okay. I will let like this and we can move it to a different PR (I think we need to do some tests). Thanks

burtenshaw (approved):

davidberenstein1957 (approved):

Issue #5069, opened 2024-06-20
[TASK] Refresh records `status` column when a response is created/updated/upserted/deleted


jfcalvo:
The new `status` column created at `records` table should be refreshed every time an associated response is created/updated/upserted/deleted.

There are several approaches that we can explore to solve this issue:
1. Use a function that refresh the `status` column when some function inside our context modules execute a change over a response.
2. Use database triggers to refresh the `status` column.
3. Use SQLAchemy event callbacks to listen to changes in responses and refresh the `status` column.

We need to take into account that when records are created responses can be assigned too, in these cases we need to refresh the `status` column as well.

PR #5071, opened 2024-06-20
[SPIKE] feat: refresh record `status` column using context functions

jfcalvo:
# Description

This PR include changes to explore the possibility of refreshing the new records `status` column inside our Python context functions.

The following changes has been included in this PR:
* A new `distribution` context has been created.
* A new function `refresh_record_status` has been added to `distribution` context to be called when it's necessary to refresh the `status` column for a specific record.
* The function `refresh_record_status` is called when a response is created/updated/upserted/deleted.

## Changes that are not present in this PR
- [x] Refreshing of `status` column when records are created/upserted in bulk including responses.
- [x] Refresh modified records into the search engine.
  - [ ] We need to check if the current way of indexing is redundant (like indexing records and responses is redundant somehow). 
- [ ] Avoid possible concurrency problems. 
- [x] Include `status` into the search engine mapping.
- [ ] Modify the search endpoint to support filtering records with `status` value.
- [ ] Modify current user metrics.
- [ ] Modify dataset progress.
- [ ] Refresh `status` column when a user is deleted (only in the case we modify the user deletion to also delete their responses in the near future).

## Details that still need to be explored

The isolation level of SQLite transactions is `SERIALIZABLE` making concurrency problems almost impossible using this database.

On the other hand using PostgreSQL the default isolation level is `READ UNCOMMITTED`, a weak one, and the possibility of suffering concurrency problems is there. This is something that needs to be explored, for example trying to manually find a scenario where such concurrency problem is possible and avoiding changing it to a higher isolation level for our transactions.

Possible solutions to this:
* Change transaction to `SERIALIZABLE`.
* Use lock mechanism for the row or row column (this needs some research).

Refs #5069

**Type of change**

- [ ] Documentation update

**How Has This Been Tested**

- [ ] Test A
- [ ] Test B

**Checklist**


argilla-server/src/argilla_server/contexts/datasets.py
```diff
@@ -936,7 +936,9 @@ async def create_response(
         )
 
         await db.flush([response])
+        await distribution.refresh_record_status(db, record, autocommit=False)
         await _touch_dataset_last_activity_at(db, record.dataset)
+        await search_engine.index_records(record.dataset, [record])
         await search_engine.update_record_response(response)
```
jfcalvo, line 942: We need to check if only the record index is required and the one for response is not necessary. (the same for the rest of record updates into the search engine)

argilla-server/src/argilla_server/contexts/datasets.py
```diff
@@ -936,7 +936,9 @@ async def create_response(
         )
 
         await db.flush([response])
+        await distribution.refresh_record_status(db, record, autocommit=False)
         await _touch_dataset_last_activity_at(db, record.dataset)
+        await search_engine.index_records(record.dataset, [record])
```
jfcalvo, line 941: We need to do a partial update of the record to only refresh the `status` in elasticsearch but in the meantime we can do this:

```suggestion
        await _preload_records_relationships_before_index(self._db, record)
        await search_engine.index_records(record.dataset, [record])
```

argilla-server/src/argilla_server/search_engine/commons.py
```diff
@@ -712,6 +713,7 @@ def _configure_index_mappings(self, dataset: Dataset) -> dict:
             "properties": {
                 # See https://www.elastic.co/guide/en/elasticsearch/reference/current/explicit-mapping.html
                 "id": {"type": "keyword"},
+                "status": {"type": "text"},
```
jfcalvo, line 716:
```suggestion
                "status": {"type": "keyword"},
```

PR #5079, opened 2024-06-20
[BUGFIX] frontend: Change the default empty datasets view

frascuchon:
This is a quick fix and should change in future releases @damianpumar


frascuchon:
Fixed [here](https://github.com/argilla-io/argilla/pull/5080)

frascuchon:
Fixed in of https://github.com/argilla-io/argilla/pull/5080

PR #5088, opened 2024-06-21
[ENHANCEMENT] CI: Allow to publish hidden version for docs/ branches

frascuchon:


.github/workflows/argilla.docs.yml
```diff
@@ -64,3 +65,11 @@ jobs:
 
       - run: pdm run mike deploy ${{ github.ref_name }}
         if: startsWith(github.ref, 'refs/tags/')
+
+      - name: Extract branch name
+        id: extract_branch_name
+        shell: bash
+        run: echo "branch_name=${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}}" | sed 's/\//_/g'  >> $GITHUB_OUTPUT
+
+      - run: pdm run mike deploy ${{ steps.extract_branch_name.outputs.branch_name }} --prop-set hidden=true --push
```
davidberenstein1957: We can start with this, but I think we need one more thing for a useful MVP for us and non-technical people:
- more active way to publish the public URL / even if it is hidden from versioning, otherwise people still need to look into copy-pasting the branch name, replacing slashes with underscores and then pasting it after the source URL. This can be done with something, like this: https://github.com/marketplace/actions/create-or-update-comment or https://github.com/marketplace/actions/add-pr-comment

And one for a follow-up:
- cleanup workflow when releasing a new tag, otherwise we overload the [gh-page branch](https://github.com/argilla-io/argilla/tree/gh-pages) with different builds for each version of `docs/ `

davidberenstein1957: @frascuchon This was handled in my more recent commits, feel free to take a look.

.github/workflows/argilla.docs.yml
```diff
@@ -54,13 +55,46 @@ jobs:
...
+
+      - name: Delete Argilla docs (versions 'docs_*')
+        id: delete_deployment
+        run: |
+          versions=$(pdm run mike list)
+          formatted_versions=$(echo "$versions" | tr ',' '\n' | tr -d '[]')
+          for version in $versions; do
+            if [[ $version == docs_* ]]; then
+              echo "Deleting version: $version"
+              pdm run mike delete "$version"
```
frascuchon: do we need to add `--push`?

davidberenstein1957:
```suggestion
              pdm run mike delete "$version" --push
```

.github/workflows/argilla.docs.yml
```diff
@@ -54,13 +55,46 @@ jobs:
         run: echo "${{ github.ref }}"
           echo "${{ github.head_ref }}"
 
-      - run: |
-          pdm run mike deploy latest --push
+      - name: Deploy Argilla docs (branch /main)
+        run: |
+          pdm run mike set-default --push latest
```
frascuchon:
```suggestion
          pdm run mike deploy latest --push
```

.github/workflows/argilla.docs.yml
```diff
@@ -54,13 +55,46 @@ jobs:
...
+        id: delete_deployment
+        run: |
+          versions=$(pdm run mike list)
+          formatted_versions=$(echo "$versions" | tr ',' '\n' | tr -d '[]')
+          for version in $versions; do
+            if [[ $version == docs_* ]]; then
+              echo "Deleting version: $version"
+              pdm run mike delete "$version" --push
+            fi
+          done
```
frascuchon, line 82: Just keep in mind that this will delete ALL the existing `docs_*`. Some open work may be affected.

argilla/docs/how_to_guides/annotate.md
```diff
@@ -17,7 +17,6 @@ Argilla UI offers many functions to help you manage your annotation workflow, ai
 The UI is responsive with two columns for larger devices and one column for smaller devices. This enables you to annotate data using your mobile phone for simple datasets (i.e., not very long text and 1-2 questions) or resize your screen to get a more compact UI.
 
 === "Header"
-
```
frascuchon: these changes are not related to this PR, right?

.github/workflows/argilla.docs.yml
davidberenstein1957: Yes it was intended because it only executes `if: startsWith(github.ref, 'refs/tags/')` @frascuchon it does not in the dif manager, but you can look at the original file.

frascuchon (commented):
It's fine for me. Thanks

.github/workflows/argilla.docs.yml
```diff
@@ -54,13 +55,46 @@ jobs:
...
         if: github.ref == 'refs/heads/main'
 
-      - run: pdm run mike deploy dev --push
+      - name: Deploy Argilla docs (branch /develop)
+        run: pdm run mike deploy dev --push
         if: github.ref == 'refs/heads/develop'
 
-      - run: pdm run mike deploy ${{ github.ref_name }}
+      - name: Deploy Argilla docs (release vX.Y.Z)
+        run: pdm run mike deploy ${{ github.ref_name }}
```
frascuchon, line 69: Do we want this level for docs or should we apply this change in combination with https://github.com/argilla-io/argilla/pull/5092?

davidberenstein1957: @frascuchon I think it makes sense to deploy any minor (2.X) releases as part of the docs and update/overwrite patches to 2.x.

PR #5107, opened 2024-06-25
[ENHANCEMENT] [REFACTOR] optimise and refactor SDK ingestion methods 

burtenshaw:
This PR refactors refactors the ingestion flow in `DatasetRecords` by implementing a new class and module `IngestedRecordMapper` 

This PR supports mapping incoming columns/keys to dataset attributes in these two ways: 

- supports tuple values in the mapping parameter of the log method so that user can specify the two attributes as a tuple.
- refactors in the `_ingest_records` methods so that mapping is performed once before the ingestion loop instead of during.

This PR also optimises the `log` method so that it takes less time and is easier to work with:
- uses tqdm to log status 
- uses exception to show bad records
- iterates over the map not the data

Screenshot 2024-06-26 at 09 09 11

**How Has This Been Tested**
- tests have been modified, deprecated, and updated to support changes in the ingestion flow

**Checklist**


argilla/src/argilla/records/_dataset_records.py
```diff
@@ -239,7 +241,7 @@ def log(
         created_or_updated = []
         records_updated = 0
         for batch in tqdm(
-            iterable=range(0, len(records), batch_size), desc="Adding and updating records", unit="batch"
+            iterable=range(0, len(records), batch_size), desc="2/2: Adding and updating records", unit="batch"
```
davidberenstein1957:
```suggestion
            iterable=range(0, len(records), batch_size), desc="Adding and updating records", unit="batch"
```

davidberenstein1957: Not sure if 1/2 and 2/2 adds a lot or maybe makes it difficult to maintain?

argilla/src/argilla/records/_dataset_records.py
```diff
@@ -398,92 +410,103 @@ def _validate_vector_names(self, vector_names: Union[List[str], str]) -> None:
...
+    ) -> Dict[str, Tuple[Optional[str]]]:
+        """Renders a mapping from a list of records and a mapping dictionary, to a singular mapping dictionary."""
+        schema = self.__dataset.schema
+        mapping = mapping or {}
+        singular_mapping = defaultdict(dict)
+
+        # update the mapping with unmapped columns
+        for key, value in records[0].items():
+            if key not in schema and key not in mapping:
+                warnings.warn(
```
davidberenstein1957: not sure if we would skip in both cases? I think if it is not in the schema, I would say the user made a mistake right? If it it not present in the one record, I would assume it might by chance have an empty value.

argilla/src/argilla/records/_dataset_records.py
```diff
@@ -398,92 +410,103 @@ def _validate_vector_names(self, vector_names: Union[List[str], str]) -> None:
...
+        # create a singular mapping with destinations from the schema
+        for source_key, value in mapping.items():
+            destinations = [value] if isinstance(value, str) else list(value)
+
+            for attribute_mapping in destinations:
+                attribute_mapping = attribute_mapping.split(".")
+
+                attribute_name = attribute_mapping[0]
+                schema_item = schema.get(attribute_name)
+                attribute_type = attribute_mapping[1] if len(attribute_mapping) > 1 else None
```
davidberenstein1957: don't we want to check if all attribute types are the same and if not can't there be different attribute types? perhaps I want to map to metadata and field or something?

argilla/src/argilla/records/_dataset_records.py
```diff
@@ -494,3 +517,52 @@ def _infer_record_from_mapping(
...
+            A Suggestion object.
+
+        """
+        suggestion_values = defaultdict(dict)
+
+        for (attribute_name, sub_attribute), source_key in mapping.items():
+            value = data.get(source_key)
+            schema_item = self.__dataset.schema.get(attribute_name)
+            suggestion_values[attribute_name].update(
+                {sub_attribute: value, "question_name": attribute_name, "question_id": schema_item.id}
```
davidberenstein1957:
```suggestion
                {sub_attribute: value, "question_name": attribute_name, "question_id": schema_item.id}
```
maybe `sub_attribute` is not very specific and can be `attribute_value_key` or something?

nataliaElv:
I'd rather have the records ingested and pushed in batches and have an easy way to identify those that threw an error, fix them and try to import those again.
Otherwise it can take ages until I see any records in my dataset.
Captura de pantalla 2024-06-26 a las 9 23 42

davidberenstein1957 (commented):
Left some initial comments.

argilla/src/argilla/records/_dataset_records.py
burtenshaw: sub_attribute as yet only refers to the parameters of a suggestion. So score and agent. I don't think `_key` explains that, but I'm open to suggestions.

argilla/src/argilla/records/_dataset_records.py
frascuchon: I can imagine that there is a mechanism to handle this progress in `tqdm`. Something like task `i` of `n` progress.

argilla/src/argilla/records/_dataset_records.py
```diff
@@ -358,25 +360,35 @@ def to_datasets(self) -> HFDataset:
...
-                record.dataset = self.__dataset
-        else:
-            raise ValueError(
-                "Records should be a a list Record instances, "
-                "a Hugging Face Dataset, or a list of dictionaries representing the records."
-            )
-        return [record.api_model() for record in records]
+
+        ingested_records = []
+        for record in tqdm(records, desc="1/2: Ingesting records", unit="record"):
```
frascuchon: Yes, is a bit weird having this 1/2 in an isolated method without the whole progress picture.
```suggestion
        for record in tqdm(records, desc="Ingesting records", unit="record"):
```

argilla/src/argilla/records/_dataset_records.py
```diff
@@ -358,25 +360,35 @@ def to_datasets(self) -> HFDataset:
...
+            try:
+                if not isinstance(record, Record):
+                    rendered_mapping = self._render_record_mapping(records=records, mapping=mapping)
+                    record = self._infer_record_from_mapping(data=record, mapping=rendered_mapping, user_id=user_id)
```
frascuchon: I think we should merge both methods in once
```suggestion
                    record = self._infer_record_from_mapping(data=record, mapping=mapping, user_id=user_id)                    
```
and, internally, the `_infer_record_from_mapping` will have a call to                     rendered_mapping = self._render_record_mapping(records=records, mapping=mapping)

argilla/src/argilla/records/_dataset_records.py
```diff
@@ -208,7 +209,7 @@ def __repr__(self) -> str:
     def log(
         self,
         records: Union[List[dict], List[Record], HFDataset],
-        mapping: Optional[Dict[str, str]] = None,
+        mapping: Optional[Dict[str, Union[str, List[str], Tuple[str]]]] = None,
```
frascuchon: Why not just support list or tuple instead of both? Or just a Sequence[str]?

argilla/src/argilla/records/_dataset_records.py
burtenshaw: In fact, since step 1 is now completely local, we might as well skip the progress bar.

argilla/src/argilla/records/_dataset_records.py
burtenshaw: If a value is not present in the schema or mapping, then we assume that it's something that the user does not want to log to Argilla. We warn in case they've made a mistake.

argilla/src/argilla/records/_dataset_records.py
burtenshaw: Sorry, I don't understand this question. The user can map a value in their data by doing:

```python
mapping = {
        "my_prompt": ("prompt_field", "prompt_question"),
    }
```

The strings in the tuple could point to metadata etc.

argilla/src/argilla/records/_dataset_records.py
burtenshaw: I've updated this method so that we only call `_render_record_mapping` once before the loop. Do you think we should call it every time?

argilla/src/argilla/records/_dataset_records.py
frascuchon: Okay. Now I think about what these methods do. So the idea is to generate a normalized mapping which will be used to build the record. Maybe this mapping can be done once and used for all the records since we have the dataset schema which can be used as a default mapping. I have some comments about the implementation.

1. The `_render_record_mapping` can be executed  just once if we use the settings schema
2. I think it would be easier to follow if the "singular_mapping" is a class instead of a raw dict. This will be a normalized representation of the mapping so we can have something like:

```python
# this will apply the `_render_record_mapping` logic (I guess)
mapping = NormalizedMapping(mapping, schema=settings.schema)

# at some point inside _infer_record_from_mapping
# Parse suggestions and responses into objects aligned with questions
suggestions = self._parse_suggestion_from_mapping(data=data, mapping=mapping.suggestions)
responses = self._parse_response_from_mapping(data=data, mapping=mapping.responses, user_id=user_id)

suggestions = [Suggestion(**suggestion_dict) for suggestion_dict in suggestion_values.values()]
# Parse fields, metadata, and vectors into
fields = {attribute_name: data.get(source_key) for attribute_name, source_key in mapping.fields.items()}
metadata = {attribute_name: data.get(source_key) for attribute_name, source_key in mapping.metadata.items()}
vectors = {attribute_name: data.get(source_key) for attribute_name, source_key in mapping.vectors.items()}
```

or even more simple:
```python
mapping = NormalizedMapping(mapping, schema=settings.schema)
...

suggestions = mapping.suggestions(data)
responses = mapping.responses(data, user_id)
fields = mapping.fields(data)
metadata = mapping.metadata(data)
vectors = mapping.vectors(data)

```
or just in once:
```python
record = mapping.infer_record(data)
```

burtenshaw: @frascuchon I've reimplemented the PR using a class as suggested.

argilla/docs/how_to_guides/record.md
```diff
@@ -318,24 +318,31 @@ Suggestions refer to suggested responses (e.g. model predictions) that you can a
     You can add suggestions as a dictionary, where the keys correspond to the `name`s of the labels that were configured for your dataset. Remember that you can also use the `mapping` parameter to specify the data structure.
 
     ```python
-    # Add records to the dataset with the label 'my_label'
+        # Add records to the dataset with the label 'my_label'
```
frascuchon:
```suggestion
        # Add records to the dataset with the label question 'my_label'
```

argilla/src/argilla/records/_dataset_records.py
```diff
@@ -357,26 +357,36 @@ def to_datasets(self) -> HFDataset:
...
-                "Records should be a a list Record instances, "
-                "a Hugging Face Dataset, or a list of dictionaries representing the records."
-            )
-        return [record.api_model() for record in records]
+
+        ingested_records = []
+        record_mapper = IngestedRecordMapper(mapping=mapping, dataset=self.__dataset, user_id=user_id)
+        for record in records:
+            try:
+                if not isinstance(record, Record):
```
frascuchon: we can avoid unexpected errors trying to map some weird content.
```suggestion
                if isinstance(record, dict):
```

argilla/src/argilla/records/_mapping.py
```diff
@@ -0,0 +1,350 @@
...
+
+        return RecordAttributesMap(**schematized_map)
+
+    def _parse_dot_notation(self, attribute_mapping: str) -> Tuple[str, Optional[str], Optional[str]]:
+        """Parses a string in the format of 'attribute.type.parameter' into its parts using regex."""
+
+        # Get the available attributes, types, and parameters from the schema.
+        available_attributes = list(self._schema.keys()) + ["id"]
+        available_types = [type_.value for type_ in AttributeType]
+        available_parameters = [param.value for param in ParameterType]
```
frascuchon: Maybe we can move this list comprehension to the corresponding classes and define them as properties:
```python
class AtributeType(..):
   ...
   @property
   def values(self):
         return [t.value for t in self]

File: argilla/src/argilla/records/_mapping.py [hunk 11]
```diff
@@ -0,0 +1,350 @@
...
+
+class RecordAttributesMap(BaseModel):
+    """RecordAttributesMap is a representation of a record attribute mapping that is used to parse data into a record."""
+
+    suggestion: Dict[str, AttributeRoute]
+    response: Dict[str, AttributeRoute]
+    field: Dict[str, AttributeRoute]
+    metadata: Dict[str, AttributeRoute]
+    vector: Dict[str, AttributeRoute]
+    id: Dict[str, AttributeRoute]
```
frascuchon: Do we support multiple ID definitions in mapping? I would expect here a single `AttributeRoute` for id.
```python
   id: AttributeRoute
```

argilla/src/argilla/records/_mapping.py
```diff
@@ -0,0 +1,350 @@
...
+        Parameters:
+            data: A dictionary representing the record.
+            user_id: The user id to associate with the record responses.
+
+        Returns:
+            Record: The record object.
+
+        """
+
+        record_id = data.get(self.mapping.id["id"].source)
```
frascuchon: Could we do `self.mapping.id.source` instead?

argilla/src/argilla/records/_mapping.py
```diff
@@ -0,0 +1,350 @@
...
+        elif isinstance(schema_item, TextField):
+            attribute.type = AttributeType.FIELD
+        elif isinstance(schema_item, VectorField):
+            attribute.type = AttributeType.VECTOR
+        elif isinstance(schema_item, MetadataPropertyBase):
+            attribute.type = AttributeType.METADATA
+        elif attribute.name == "id":
+            attribute.type = AttributeType.ID
+        else:
+            warnings.warn(message=f"Record attribute {attribute.name} is not in the schema or mapping so skipping.")
```
frascuchon: In which scenario this will happen? Would be safe to return the attribute for those cases?

argilla/src/argilla/records/_mapping.py
```diff
@@ -0,0 +1,350 @@
...
+            if question.name not in mapping.suggestion:
+                mapping.suggestion[question.name] = AttributeRoute(
+                    source=question.name,
+                    name=question.name,
+                    type=AttributeType.SUGGESTION,
+                    parameters=[AttributeParameter(source=question.name)],
+                )
+
+            elif not mapping.suggestion[question.name].has_value:
+                mapping.suggestion[question.name].parameters.append(AttributeParameter(source=question.name))
```
frascuchon: which mapping will result in this? Should we avoid mappings without values instead?

argilla/src/argilla/records/_mapping.py
```diff
@@ -0,0 +1,350 @@
...
+            attribute.type = AttributeType.VECTOR
+        elif isinstance(schema_item, MetadataPropertyBase):
+            attribute.type = AttributeType.METADATA
+        elif attribute.name == "id":
+            attribute.type = AttributeType.ID
+        else:
+            warnings.warn(message=f"Record attribute {attribute.name} is not in the schema or mapping so skipping.")
+        return attribute
+
+    def _schematize_default_attributes(self, mapping: RecordAttributesMap) -> RecordAttributesMap:
```
frascuchon: Not sure because I'm missing some scenarios, but maybe  would be easier if we first build the "default" RecordAttributesMap from the schema and then update/override with content provided in `mapping`?
```python
def __init__(...):
        default_mapping = self.._schematize_default_attributes(schema=dataset.schema)
        self.mapping = self._schematize_default_attributes(default=default_mapping, mapping or {})
```

argilla/src/argilla/records/_mapping.py
burtenshaw: I can't implement them as properties without needing to init the class to access, so I've implemented them as classmethods instead.

Here's the attempt with properties:

```python
from enum import Enum
from typing import List

class ParameterType(str, Enum):
    """Parameter types are the different 'sub' values of a records attribute.
    For example, the value, score, or agent of a suggestion."""

    VALUE = "value"
    SCORE = "score"
    AGENT = "agent"
    
    @property
    def values(self) -> List[str]:
        return [param.value for param in ParameterType]

class AttributeType(str, Enum):
    """Attribute types are the different types of attributes a record can have."""

    FIELD = "field"
    SUGGESTION = "suggestion"
    RESPONSE = "response"
    METADATA = "metadata"
    VECTOR = "vector"
    ID = "id"
    
    @property
    def values(self) -> List[str]:
        return [attr.value for attr in AttributeType]

# Example usage:
print(ParameterType.VALUE.values)
print(AttributeType.FIELD.values)
```

argilla/src/argilla/records/_mapping.py
burtenshaw: This if statement is true if the use has mapped to a field name and that field name appears in the schema.

argilla/src/argilla/records/_mapping.py
burtenshaw: This is used when an attributes are mapped accross multiple columns. i.e. a suggestion's score, agent, and value:

```python
mock_mapping = {
        "my_prompt": "prompt",
        "my_label": "label.suggestion.value",
        "score": "label.suggestion.score",
        "model": "label.suggestion.agent",
    }
```

frascuchon: But for that case, the mapping.suggestion[question.name].has_value will return True, since the "default" mapping is computed after the first mapping normalization, right?

argilla/src/argilla/records/_mapping.py
frascuchon: In that case, the warning message is confusing:
```python
            warnings.warn(message=f"Record attribute {attribute.name} is not in the schema or mapping so skipping.")
```

argilla/src/argilla/records/_mapping.py
burtenshaw: I can see why that make sense, and the default method is easier to implement, but I think this is easier to follow.

I found that the `_schematize_mapped_attributes` became complex and needed to pick apart the `AttributeRoute` object to edit and add mapped attributes and sub attributes to the the routes.

argilla/src/argilla/records/_mapping.py
burtenshaw: I don't think this is relevant with the latest changes? to merge your PR.

frascuchon (approved):

argilla/src/argilla/records/_mapping.py
burtenshaw: I've replaced the warning with an error because in fact this should only happen with the user does something in valid. I've also changed the message.

PR #5112, opened 2024-06-26
[REFACTOR] Avoid autofetch when accessing settings

frascuchon:
Reviewing and improving records.log

Instead of: 

Captura de pantalla 2024-06-26 a las 12 48 14

for 50 records, records.log can log 1000:

Captura de pantalla 2024-06-26 a las 12 48 57


burtenshaw (approved):
🏃‍♂️

PR #5114, opened 2024-06-26
[REFACTOR] Redefine some property methods

frascuchon:
**Type of change**

**How Has This Been Tested**

**Checklist**

### DIFF
--- a/argilla/src/argilla/_api/_client.py
+++ b/argilla/src/argilla/_api/_client.py
@@ -31,7 +31,6 @@ from argilla._constants import _DEFAULT_API_KEY, _DEFAULT_API_URL
 
 __all__ = ["APIClient"]
 
-
 ARGILLA_API_URL = os.getenv(key="ARGILLA_API_URL", default=_DEFAULT_API_URL)
 ARGILLA_API_KEY = os.getenv(key="ARGILLA_API_KEY", default=_DEFAULT_API_KEY)
 DEFAULT_HTTP_CONFIG = HTTPClientConfig(api_url=ARGILLA_API_URL, api_key=ARGILLA_API_KEY)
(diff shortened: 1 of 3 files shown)


argilla/src/argilla/_api/_client.py
```diff
@@ -106,24 +105,19 @@ def __init__(
...
     ):
-        http_client_args = http_client_args or {}
-        http_client_args["timeout"] = timeout
-
         self.api_url = api_url
         self.api_key = api_key
-        self._http_client_args = http_client_args
 
-    @property
-    def http_client(self) -> httpx.Client:
```
frascuchon, line 117: Not sure if we need a new http_client each time we call the `client.api` property. We should keep 1-1 the rg.Argilla and the inner httpx client. This would be beneficial if we introduce some cache mechanisms

argilla/src/argilla/_api/_client.py
```diff
@@ -106,24 +105,19 @@ def __init__(
...
             api_url=self.api_url,  # type: ignore
             api_key=self.api_key,  # type: ignore
-            **self._http_client_args,
+            **http_client_args,
         )
 
-    @property
-    def api(self) -> "ArgillaAPI":
-        return ArgillaAPI(http_client=self.http_client)
+        self.api = ArgillaAPI(self.http_client)
```
frascuchon, line 120: This is not quite relevant, but it makes sense also having one single `api`  instance per Argilla client

argilla/src/argilla/client.py
```diff
@@ -72,10 +77,9 @@ def users(self) -> "Users":
         """A collection of users on the server."""
         return Users(client=self)
 
-    @property
+    @cached_property
```
frascuchon, line 80: Since one single client should be used (at least for now) for one credentials setup, we can define `me` as a `cached_property` (if we want a lazy load). We could also define it as an object attribute inside the __init__ method to check the client configuration. (I am not very comfortable with this, but it could make sense in this case)

argilla/src/argilla/_api/_client.py
burtenshaw: It makes the class consistent if theres 1==1==1 for class, http_client, api class

argilla/src/argilla/client.py
burtenshaw: A cached property makes the most sense to me.

burtenshaw (approved):

PR #5119, opened 2024-06-27
[ENHANCEMENT] Store count for submitted responses in search engine

frascuchon:


argilla-server/src/argilla_server/contexts/datasets.py
```diff
@@ -334,6 +334,8 @@ async def get_records_by_ids(
             query = query.outerjoin(
                 Response, and_(Response.record_id == Record.id, Response.user_id == user_id)
             ).options(contains_eager(Record.responses))
+    else:
+        query.options(joinedload(Record.responses))
```
frascuchon: This change should be removed when using a separate relationship to load all the responses

argilla-server/src/argilla_server/contexts/datasets.py
```diff
@@ -939,6 +941,18 @@ async def create_response(
         await _touch_dataset_last_activity_at(db, record.dataset)
         await search_engine.update_record_response(response)
 
+        record = await Record.get_or_raise(
+            db,
+            response.record_id,
+            options=[
+                selectinload(Record.responses),
+            ],
+        )
```
frascuchon: This won't be needed if records is loaded by default with all related responses

jfcalvo (approved):

PR #5132, opened 2024-06-28
feat: add `status` as persisted column for `records` table

### LINKED ISSUE #5058: [TASK] Add `status` column to `records` table
We should add `status` column to `records` table with the following details:
* Possible values for this `status` column are `pending` and `completed`.
* This new column will be non-nullable.
* It will have a database index so it will be faster to recover records with one specific `status` value.
* By default the `status` column will have the `pending` value. (Showing that the record is still pending of being annotated).
* Existent old records will be migrated in the following way:
  * Records with a least one response with `submitted` status will be updated from `pending` to `completed`.

jfcalvo:
# Description

Add `status` as a persisted database column for `records` table.


argilla-server/tests/unit/api/handlers/v1/test_list_dataset_records.py
```diff
@@ -821,7 +821,7 @@ async def test_list_current_user_dataset_records(
             "items": [
                 {
                     "id": str(record_a.id),
-                    "status": RecordStatus.completed,
+                    "status": RecordStatus.pending,
```
frascuchon, line 824: Why all these tests are changing the expected status?

argilla-server/src/argilla_server/bulk/records_bulk.py
```diff
@@ -67,6 +68,7 @@ async def create_records_bulk(self, dataset: Dataset, bulk_create: RecordsBulkCr
 
             await self._upsert_records_relationships(records, bulk_create.items)
             await _preload_records_relationships_before_index(self._db, records)
+            await distribution.update_records_status(self._db, records, autocommit=False)
```
frascuchon: Can we just remove the `autocommit=False` argument and handle this internally in the distribution context?

argilla-server/src/argilla_server/contexts/distribution.py
```diff
@@ -0,0 +1,42 @@
...
+    raise NotImplementedError(f"unsupported distribution strategy `{record.dataset.distribution_strategy}`")
+
+
+async def _update_record_status_with_overlap_strategy(db: AsyncSession, record: Record, autocommit: bool) -> Record:
+    if len(record.responses_submitted) >= record.dataset.distribution["min_submitted"]:
+        record.status = RecordStatus.completed
+    else:
+        record.status = RecordStatus.pending
+
+    return await record.save(db, autocommit)
```
frascuchon: We're not using `autocommit=True` anywhere. I suggest setting it to False by default and stopping exposing it to the outside.
```suggestion
    return await record.save(db, autocommit=False)
```

argilla-server/tests/unit/api/handlers/v1/test_list_dataset_records.py
jfcalvo: As commented in a call, the `status` is not calculated on-the-fly anymore (now is persisted on database) so factories are not modifying the `status` (only our handlers are doing that) so the values are now `pending` by default.

frascuchon (approved):

PR #5152, opened 2024-07-03
[ENHANCEMENT/REFACTOR] `argilla`: lazy resolution for dataset workspaces

frascuchon:
This PR removes the workspace resolution on `Dataset.__init__` and performs it lazily when creating datasets or accessing the `dataset.workspace`  property

**Type of change**

- Documentation update

**How Has This Been Tested**

**Checklist**

### DIFF
--- a/argilla/src/argilla/_helpers/_resource_repr.py
+++ b/argilla/src/argilla/_helpers/_resource_repr.py
@@ -32,7 +32,7 @@ RESOURCE_REPR_CONFIG = {
 class ResourceHTMLReprMixin:
     def _resource_to_table_row(self, resource) -> Dict[str, Any]:
         row = {}
-        dumped_resource_model = resource._model.model_dump()
+        dumped_resource_model = resource.api_model().model_dump()
         resource_name = resource.__class__.__name__
         config = RESOURCE_REPR_CONFIG[resource_name].copy()
         len_column = config.pop("len_column", None)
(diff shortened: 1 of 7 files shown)


burtenshaw (commented):
Looks good. One question: how would `Dataset.exist` behave if there's no workspace in the dataset resource? This method confuses a lot of users.

frascuchon:
The `.exists` method uses the resource ID internally, so a workspace is unnecessary for this query. Yes, I see the confusion of users setting datasets with a name and a workspace and checking the existence of the dataset. 

maybe this existence method should be defined at the client level:

```python
client.datasets.exists(dataset)
#or 
client.datasets.exists(name, workspace)
```

Just to force a direct client interaction to make more sense to the "exists" query.

burtenshaw:
> The `.exists` method uses the resource ID internally, so a workspace is unnecessary for this query. Yes, I see the confusion of users setting datasets with a name and a workspace and checking the existence of the dataset.
> 
> maybe this existence method should be defined at the client level:
> 
> ```python
> client.datasets.exists(dataset)
> #or 
> client.datasets.exists(name, workspace)
> ```
> 
> Just to force a direct client interaction to make more sense to the "exists" query.

Yeah. I think this would solve it because in most cases the user just wants to know 'can I use this name?' 

This could be implemented in a subsequent PR though.

burtenshaw (approved):

PR #5162, opened 2024-07-04
🚑 Remove resizable layout when no records and fix undefined prop

damianpumar:


leiyre (approved):

leiyre:
I see that we should change some styles for `wrapper--empty` on small screens, maybe remove the 80vh and add 16px padding is enough

Issue #5050, opened 2024-06-18
[TASK] Add restriction to update dataset distribution settings


jfcalvo:
We should add a restriction to only allow updating a dataset distribution settings when the associated records for the dataset has not responses.

As an alternative to this we could do the same but only allowing the update for datasets that have not records with `status` as `completed`.

We should do what is better from a performance point of view and return a easy to understand message to the user if there is an error.

frascuchon:
Fixed by #5171

PR #5171, opened 2024-07-05
[ENHANCEMENT]: `argilla-server`: allow update distribution for non annotated datasets

frascuchon:
# Description

This PR changes the current validator when updating the distribution task to allow updating the distribution task settings for datasets with records without ANY response.

cc @nataliaElv 

**Type of change**

**How Has This Been Tested**

**Checklist**

### DIFF
--- a/argilla-server/src/argilla_server/models/database.py
+++ b/argilla-server/src/argilla_server/models/database.py
@@ -361,7 +361,8 @@ class Dataset(DatabaseModel):
 
     responses_count: Mapped[int] = column_property(
         select(func.count(Response.id))
-        .where(Response.record_id.in_(select(Record.id).where(text("datasets.id == records.dataset_id"))))
+        .join(Record)
+        .filter(text("datasets.id == records.dataset_id"))
         .scalar_subquery(),
         deferred=True,
     )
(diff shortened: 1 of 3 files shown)


argilla-server/src/argilla_server/models/database.py
```diff
@@ -359,6 +359,13 @@ class Dataset(DatabaseModel):
         order_by=VectorSettings.inserted_at.asc(),
     )
 
+    responses_count: Mapped[int] = column_property(
+        select(func.count(Response.id))
+        .where(Response.record_id.in_(select(Record.id).where(text("datasets.id == records.dataset_id"))))
```
frascuchon: Maybe responses in draft status could be filtered for this count (or for the validation)

jfcalvo: Is not a join of records and responses a better (more efficient) way of doing this count?

jfcalvo: Btw I think we should not use associations for counting things. I think a helper function could better fit this.

jfcalvo (approved):

frascuchon:
Fixes https://github.com/argilla-io/argilla/issues/5050

PR #5184, opened 2024-07-09
[ENHANCEMENT] `argilla`: add record `status` property

### LINKED ISSUE #5141: [TASK] Add `status` attribute to records

frascuchon:
# Description

This PR adds the record status as a read-only property in the `Record` resource class.

Closes https://github.com/argilla-io/argilla/issues/5141

**Type of change**

**How Has This Been Tested**

**Checklist**


argilla/tests/unit/test_io/test_hf_datasets.py
```diff
@@ -46,6 +46,7 @@ def test_to_datasets_with_partial_values_in_records(self):
 
         ds = HFDatasetsIO.to_datasets(records)
         assert ds.features == {
+            "status": Value(dtype="string", id=None),
```
frascuchon, line 49: @burtenshaw  @davidberenstein1957  the status will be added to the HF dataset. Let me know if this makes sense

argilla/src/argilla/_models/_record/_record.py
```diff
@@ -36,6 +36,8 @@ class RecordModel(ResourceModel):
     responses: Optional[List[UserResponseModel]] = Field(default_factory=list)
     suggestions: Optional[Union[Tuple[SuggestionModel], List[SuggestionModel]]] = Field(default_factory=tuple)
 
+    status: Literal["pending", "completed"] = "pending"
+
     external_id: Optional[Any] = None
```
jfcalvo:
```suggestion
    status: Literal["pending", "completed"] = "pending"
    external_id: Optional[Any] = None
```

jfcalvo: Not in relationship with this PR but should not be `external_id` a string type?

jfcalvo: I would move `status` to be the first defined attribute too.

argilla/src/argilla/records/_resource.py
```diff
@@ -147,6 +147,10 @@ def metadata(self) -> "RecordMetadata":
     def vectors(self) -> "RecordVectors":
         return self.__vectors
 
+    @property
+    def status(self) -> "str":
```
jfcalvo:
```suggestion
    def status(self) -> str:
```

jfcalvo (approved):
Looks good to me. I have added some comments about code style.

argilla/tests/unit/test_io/test_hf_datasets.py
davidberenstein1957: perhaps the naming can be more explicit? as a random person's "status" might not be explicit enough. 

Also, what is the status without n-required responses? will it always be pending because this might lead people to filter on that assuming there are no responses? additionally it feels a bit for records to be always pending in that case.

davidberenstein1957 (commented):
small comments here https://github.com/argilla-io/argilla/pull/5184/files#r1671853645

argilla/tests/unit/test_io/test_hf_datasets.py
jfcalvo: `status` is a calculated value based on the responses so not sure if this value should be persisted in that way.

davidberenstein1957: I mean that the status for the record being done or not, is a bit unclear, and might have a more explicit name, like “required_n_annotation_status”. Also because we’ve got a so-called response-level status. 

Additionally, I noticed we will enforce a default 1 min_annotation per record but does this imply we will have a max_annotation at some point? I think n_required_responses might be a better name?

I thought it would be weird to have all records in pending status is we would not enforce a default, but given that it is set to 1, I guess it does not matter. 

Lastly, I think the status I recomputed when doing from_disk  or from_hub and you upload the records, right? So I would say the status for records is more of a state that lives within the specific Argilla instance and dataset that on the hub or disk.

frascuchon: Mmmm...the meaning of `status` is linked to the dataset distribution strategy. The record status will be available for filtering and also at some point as part of the record info in the UI. It's quite present to be clearly defined. Also, your naming proposal is attached to how the status is computed currently which may change in the future.

The record status is a read-only argument, and yes, depends on how the record will be computed by the Argilla server based on the task distribution settings. So, my feelings are to remove from the export. On the other hand, having this info here can be useful to filter or to snapshot data from Argilla. cc @burtenshaw @dvsrepo @nataliaElv

burtenshaw: I would not persist it or rename it when it's persisted, because it would have confusing effects that are difficult to get round and a bit opaque. As you say, It's result of a function of distributions settings & responses, and not attribute.

burtenshaw (approved):

PR #5192, opened 2024-07-09
improvement: add some SQLite pragma statement settings to improve performance

jfcalvo:
After suffering general slowness in HF spaces after merging the SQLite fine tuning changes I have reverted the changes so we can focus on distribution task changes.

After that we will investigate if these changes are the source of the slowness with HF Spaces using persistent storage or not.

Reverts argilla-io/argilla#5191


frascuchon:
We should check these improvements in HF Spaces removing all the WAL-related ones.

PR #5199, opened 2024-07-10
[ENHANCEMENT] `argilla`: Remove exists method and change client behavior

### LINKED ISSUE #5121: [BUG-python/deployment] Confusion about dataset_argilla.exists() / .delete() / .create()
**Describe the bug**

When I create a dataset with the following code:

```
dataset_argilla_name = "code-llm"
workspace_name = "admin"
reuse_existing_dataset = False  # for easier iterative testing

annotators = ["annotator-1", "annotator-2", "annotator-3"]

# Create the interface structure via an Argilla FeedbackDataset
settings = rg.Settings(
    # The overall annotation guidelines, which human annotators can refer back to inside of the interface
    guidelines=annotator_guidelines,
    # The fields on the left side of the interface
    fields=[
        rg.TextField(name="instruction", title="Instruction:", use_markdown=True, required=True),
        rg.TextField(name="generation_1", title="Response model 1:", use_markdown=True, required=True),
        rg.TextField(name="generation_2", title="Response model 2:", use_markdown=True, required=True),
    ],
    # The different questions on the right side of the interface 
    # These are the questions we ask annotators about the fields on the left of the interface
    # The available question types are documented here: https://docs.argilla.io/en/latest/getting_started/cheatsheet.html#configure-datasets
    questions=[
        rg.RatingQuestion(
            name="score_response_1",
            title="Your score for the response of model 1:",
            description=rating_tooltip,  #"1 = very bad\n2 = bad\n3 = mediocre\n4 = good\n5 = very good",
            values=[0, 1, 2, 3, 4, 5, 6, 7],
            required=True,
        ),
        rg.RatingQuestion(
            name="score_response_2",
            title="Your score for the response of model 2:",
            description=rating_tooltip,  #"1 = very bad\n2 = bad\n3 = mediocre\n4 = good\n5 = very good",
            values=[0, 1, 2, 3, 4, 5, 6, 7],
            required=True,
        ),
        rg.LabelQuestion(
            name="which_response_corrected",
            title="If both responses score below 4, select a response to correct:",
            description="Select the response you will correct in the text field below.",
            labels = ["Response 1", "Response 2", "Combination of both", "Neither"],
            required=False,
        ),
        rg.TextQuestion(
            name="correction",
            title="Paste the selected response below and correct it manually:",
            description="Your corrected response must fulfill all criteria from the annotation guidelines.",
            use_markdown=True,
            required=False
        ),
        rg.TextQuestion(
            name="comments",
            title="Annotator Comments",
            description="Add any additional comments here. E.g.: edge cases, issues with the interface etc.",
            use_markdown=True,
            required=False
        ),
    ],
    metadata = [
        rg.TermsMetadataProperty(
            name="annotator-groups",
            title="Annotator groups",
            options=annotators,
        ),
        rg.TermsMetadataProperty(
            name="source-dataset",
            title="Original dataset source",
        ),
    ],
    allow_extra_metadata = False
)


dataset_argilla = rg.Dataset(
    name=dataset_argilla_name,
    workspace=workspace_name,
    settings=settings,
    client=client,
)
```

And then I run dataset_argilla.exists() it returns `None` indicating that the dataset does not exist yet. 

However, when I then run dataset_argilla.create() again I get an error indicating that it does exist: `ConflictError: Argilla SDK error: ConflictError: Conflict with the server. Resource or entity already exists. Details: {"detail":"Dataset with name `code-llm` already exists for workspace with id `f6667c96-49cf-497e-821e-830b98e6e821`"}`

But running dataset_argilla.delete() returns UnprocessableEntityError: Argilla SDK error: UnprocessableEntityError: Unprocessable entity. The server cannot process the request. Details: {"detail":{"code":"argilla.api.errors::ValidationError","params":{"errors":[{"loc":["path","dataset_id"],"msg":"value is not a valid uuid","type":"type_error.uuid"}]}}}

**expected behaviour**

Does the dataset exist, or does it not exist?
dataset_argilla.exists() should probably return something indicating that the dataset exists (or .delete() / .create() should return something indicating that the dataset does not exist)

Or do I misunderstand something?

**Environment:**
!pip install argilla==2.0.0rc1 

with Argilla running in a HF space.

### LINKED ISSUE #5157: [ENHANCEMENT] Move the resource existence to  the `rg.Argilla` client
The `resource.exists` method creates a lot of confusion when users want to query if a name for a resource is available. The workflow can be simplified by moving this logic to the client directly. This makes the query more explicit.

```python
client.datasets.exists(name=...)
# or
dataset = rg:Dataset(name=...)
client.datasets.exists(dataset)

# also for other resources as users and workspaces
client.workspaces.exists(name=...)
client.users.exists(name=...)
```
_Originally posted by @burtenshaw in https://github.com/argilla-io/argilla/issues/5152#issuecomment-2208622302_

frascuchon:
# Description

This PR changes the resource existence flow. The `user.exists`, `workspace.exists`, and `dataset.exists` methods have been removed and the `client.datasets(...)`,  `client.workspaces(...)`,  and `client.users(...)` methods return `None` if resource is not found.

Closes https://github.com/argilla-io/argilla/issues/5157
Closes https://github.com/argilla-io/argilla/issues/5121

**Type of change**

- Documentation update

**How Has This Been Tested**

**Checklist**


argilla/src/argilla/datasets/_export.py
```diff
@@ -52,8 +52,8 @@ def to_disk(self: "Dataset", path: str) -> str:
         # Export the dataset model, settings and records
         self._persist_dataset_model(path=dataset_path)
         self.settings.to_json(path=settings_path)
-        if self.exists():
-            self.records.to_json(path=records_path)
+        self.records.to_json(path=records_path)
```
frascuchon, line 55: I've changed the records behavior to return an empty list in that cases instead of raising an error

argilla/src/argilla/datasets/_resource.py
```diff
@@ -216,8 +208,8 @@ def _resolve_workspace(self) -> Workspace:
         return workspace
 
     def _rollback_dataset_creation(self):
-        if self.exists() and not self._is_published():
+        if not self._is_published():
             self.delete()
 
     def _is_published(self) -> bool:
-        return self.exists() and self._model.status == "ready"
+        return self._model.status == "ready"
```
frascuchon, line 215: `status` is an internal status and will be updated after an API interaction. We can skip the existence check

argilla/src/argilla/datasets/_export.py
burtenshaw: Isn't this a bit mysterious? I'd expect a warning or error if I try to export nothing.

burtenshaw (commented):
So this PR would enforce that this interaction is obsolete?

```python
dataset = Dataset(name"existing_dataset")
# .. do anything
```

And the only way to interact with existing datasets is via `client.datasets`?

frascuchon:
> So this PR would enforce that this interaction is obsolete?
> 
> ```python
> dataset = Dataset(name"existing_dataset")
> # .. do anything
> ```
> And the only way to interact with existing datasets is via `client.datasets`?

Yes. If you want to check if a dataset already exists, you need to use `client.datasets(name=...)`

argilla/src/argilla/datasets/_export.py
frascuchon: The warning will be raised here https://github.com/argilla-io/argilla/pull/5199/files#diff-ebf7fdd040970f51cc33cfd6044d1187b30f0fae5369cdabf7fde482e092bd72R85

Even if the dataset exists, the number of records could be 0. So maybe we need a more complex mechanism inside the `records.to_*` methods.

burtenshaw (approved):

PR #5209, opened 2024-07-11
[BUGFIX] `argilla`: raise error adding record responses when a response with same question_name and user_id is found

frascuchon:
# Description

This PR prevents adding multiple question responses per user which result in a server error.

**Type of change**

**How Has This Been Tested**

**Checklist**


burtenshaw (approved):
We could potentially add a specific `Exception` for this so that it can be handled in logic. For example: `UserResponseException`.

PR #5235, opened 2024-07-16
Fix `Dataset` go to definition ambiguity

gabrielmbmb:
# Description

When using the "Go to definition" feature with `rg.Dataset` some IDEs (tested in VSCode and NVIM) displayed two results:

1. One to `argilla.workspaces._resource` because `Dataset` is imported from there
2. One to `argilla.datasets._resource` which is where `Dataset` class is defined and should be the only result.

This is caused because in `argilla.__init__` we're importing everything using `*` from `argilla.workspaces._resource`. As we're importing `from argilla import Dataset`, then the IDE it's not able to resolve this ambiguity. There were two ways to fix this issue:

- Use the full import `from argilla.datasets._resource import Dataset` from `argilla.workspace._resource` helps the IDE to resolve the ambiguity (this won't be needed with the other fix, but I think it's recommended to use the full import within the package code and not use the "user imports").
- Update the import from `workspaces` sub-package in `argilla.__init__` from `from argilla.workspaces._resource import *` to `from argilla.workspaces import *` (I guess we forgot to update it as the rest of the sub-packages are being imported like this). As `argilla.workspaces.__init__` has an `__all__` defined, then only `Workspace` is imported which is the expected thing to happen.

**Type of change**

**How Has This Been Tested**
After the updates, "Go to definition" in VSCode and NVIM works well.

**Checklist**

### DIFF
--- a/argilla/src/argilla/__init__.py
+++ b/argilla/src/argilla/__init__.py
@@ -15,8 +15,8 @@ from argilla._version import __version__  # noqa
 
 from argilla.client import *  # noqa
 from argilla.datasets import *  # noqa
-from argilla.workspaces._resource import *  # noqa
-from argilla.users._resource import *  # noqa
+from argilla.workspaces import *  # noqa
+from argilla.users import *  # noqa
 from argilla.settings import *  # noqa
 from argilla.suggestions import *  # noqa
 from argilla.responses import *  # noqa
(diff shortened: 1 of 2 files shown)


frascuchon (approved):
Thanks!

PR #5242, opened 2024-07-16
[ENHANCEMENT] `argilla`: add support to fetch by `id`

frascuchon:
# Description

This PR adds support to fetch users, workspaces, and datasets by id:

````python 
client.workspaces(id=...)

client.datasets(id=...)
````

When using the `id` and the resource is not found in the argilla server, an error will be raised 

**Type of change**

- Documentation update

**How Has This Been Tested**

**Checklist**

### DIFF
--- a/argilla/docs/how_to_guides/user.md
+++ b/argilla/docs/how_to_guides/user.md
@@ -138,7 +138,7 @@ import argilla as rg
 
 client = rg.Argilla(api_url="<api_url>", api_key="<api_key>")
 
-user = client.users("<uuid-or-uuid-string>")
+user = client.users(id="<uuid-or-uuid-string>")
 ```
 
 In that case, if the user does not exist for the given id, an `ArgillaError` will be raised.
--- a/argilla/docs/how_to_guides/user.md
+++ b/argilla/docs/how_to_guides/user.md
@@ -141,7 +141,7 @@ client = rg.Argilla(api_url="<api_url>", api_key="<api_key>")
 user = client.users(id="<uuid-or-uuid-string>")
 ```
 
-In that case, if the user does not exist for the given id, an `ArgillaError` will be raised.
+If the user does not exist for the given id, the call will return `None`
 
 ## List users in a workspace
 
--- a/argilla/docs/how_to_guides/user.md
+++ b/argilla/docs/how_to_guides/user.md
@@ -132,6 +132,17 @@ client = rg.Argilla(api_url="<api_url>", api_key="<api_key>")
 retrieved_user = client.users("my_username")
 ```
 
+You can also use the user `id` to fetch the user:
+```python
+import argilla as rg
+
+client = rg.Argilla(api_url="<api_url>", api_key="<api_key>")
+
+user = client.users("<uuid-or-uuid-string>")
+```
+
+In that case, if the user does not exist for the given id, an `ArgillaError` will be raised.
+
 ## List users in a workspace
 
 You can list all the users in a workspace by accessing the `users` attribute on the `Workspace` class and iterating over them. You can also use `len(workspace.users)` to get the number of users by workspace.
--- a/argilla/docs/how_to_guides/workspace.md
+++ b/argilla/docs/how_to_guides/workspace.md
@@ -84,6 +84,17 @@ client = rg.Argilla(api_url="<api_url>", api_key="<api_key>")
 retrieved_workspace = client.workspaces("my_workspace")
 ```
 
+You can also use the user `id` to fetch the workspace:
+```python
+import argilla as rg
+
+client = rg.Argilla(api_url="<api_url>", api_key="<api_key>")
+
+workspace = client.workspaces("<uuid-or-uuid-string>")
+```
+
+In that case, if the workspace does not exist for the given id, an `ArgillaError` will be raised.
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: argilla/src/argilla/client.py [hunk 1]
```diff
@@ -188,9 +206,26 @@ def __init__(self, client: "Argilla") -> None:
...
+            raise ArgillaError("One of 'name' or 'id' must be provided")
+
+        if name and id:
+            warnings.warn("Only one of 'name' or 'id' must be provided. Using 'id'")
+            name = None
+
+        if id is not None:
+            if not isinstance(id, UUID):
+                id = UUID(id)
+            return self._from_model(self._api.get(id))
```
davidberenstein1957: should we provide a warning if not found? Or perhaps use similar logic as with the `name`

argilla/src/argilla/client.py
```diff
@@ -260,11 +295,30 @@ def __init__(self, client: "Argilla") -> None:
...
+        if not (name or id):
+            raise ArgillaError("One of 'name' or 'id' must be provided")
+
+        if name and id:
+            warnings.warn("Only one of 'name' or 'id' must be provided. Using 'id'")
+            name = None
+
+        if id is not None:
+            if not isinstance(id, UUID):
+                id = UUID(id)
```
davidberenstein1957: Can't we move some of this code to a parent class to avoid having duplicate code?

argilla/src/argilla/client.py
davidberenstein1957: Similar code as below, perhaps we can simplify it.

argilla/docs/how_to_guides/workspace.md
```diff
@@ -84,6 +84,17 @@ client = rg.Argilla(api_url="<api_url>", api_key="<api_key>")
 retrieved_workspace = client.workspaces("my_workspace")
 ```
 
+You can also use the user `id` to fetch the workspace:
+```python
+import argilla as rg
+
+client = rg.Argilla(api_url="<api_url>", api_key="<api_key>")
+
+workspace = client.workspaces("<uuid-or-uuid-string>")
```
davidberenstein1957: does this work? it seems you are using two arguments `user` and `uuid` or is this covered by the `overload`? Also shouldn't `uuid-or-uuid-string` be `uuid-or-uuid-string-or-name`?

argilla/src/argilla/client.py
frascuchon: A wrong id will raise an error. I think it makes sense since IDs identify resources more accurately. But I can change the behavior

davidberenstein1957: I think we should then also raise and error when a non.existent workspace is provided.

argilla/docs/how_to_guides/workspace.md
frascuchon: No. You're right. I've updated the docs

argilla/src/argilla/client.py
frascuchon: Then I will align to the get-by-name behavior: return `None` if the resource does not exist by id

argilla/src/argilla/client.py
frascuchon: I wouldn't create a class hierarchy just for one single method. I will create a private helper function for this.

frascuchon: We should first align the inner Resource APIs to provide a more standard way of working with resources at the client level. This is an out-of-scope for this PR.

argilla/docs/how_to_guides/user.md
```diff
@@ -132,6 +132,17 @@ client = rg.Argilla(api_url="<api_url>", api_key="<api_key>")
 retrieved_user = client.users("my_username")
```
davidberenstein1957, line 132: should this be updated to a specific `username=my_username`?

davidberenstein1957: similar for the dataset flow

argilla/tests/integration/test_client.py
```diff
@@ -0,0 +1,61 @@
...
+
+import uuid
+
+import pytest
+
+from argilla import Argilla, Dataset, TextField, TextQuestion, Settings, User, Workspace
+from argilla._exceptions import ArgillaAPIError, ArgillaError
+
+
+class TestClient:
```
davidberenstein1957, line 46: should we not test for getting by name either? We do test this for users but not for workspaces and datasets.
Also, we don't seem to test for the warnings.

davidberenstein1957 (commented):
Some small comments

argilla/tests/integration/test_client.py
frascuchon: I've included more tests here. Anyway, this test suite should not be part of the integration tests, or at least not testing at this granularity, but it's hard to mock the current client.

I also added a check using the name but the rest of the integration tests are based on this functionality, so it would not be necessary as part of an integration test.

argilla/docs/how_to_guides/user.md
frascuchon: It's fine like this since is the default argument

frascuchon:
Merging this to prepare the v2.0 release

PR #5271, opened 2024-07-19
[BUGFIX] Run telemetry check o server startup

frascuchon:
# Description

After changes introduced in https://github.com/argilla-io/argilla/pull/5263, the telemetry message is shown when running CLI commands, since the server app is automatically created when importing the argilla_server module.

This PR fixes this by moving the telemetry check to the server startup lifespan.

Maybe we need to use uvicorn [app factories](https://www.uvicorn.org/#application-factories) instead of applications directly to avoid creating the server app when the module is imported.

**Type of change**

**How Has This Been Tested**

**Checklist**


argilla-server/src/argilla_server/_app.py
```diff
@@ -47,6 +47,7 @@
 @contextlib.asynccontextmanager
 async def app_lifespan(app: FastAPI):
     # See https://fastapi.tiangolo.com/advanced/events/#lifespan
+    check_telemetry()
```
jfcalvo: I think we should rename the function name to show better what's doing:

```suggestion
    show_telemetry_warning()
```

jfcalvo (approved):

PR #5301, opened 2024-07-23
[BUGFIX] `argilla`: support reading datasets without distribution

frascuchon:
# Description

This PR fixes errors when loading dataset settings and the optional distribution model is `None` (thanks @jfcalvo for reporting this).

Also, prevent mutating the default task distribution by defining a default factory instead of a default value.

**Type of change**

**How Has This Been Tested**

**Checklist**

### DIFF
--- a/argilla/src/argilla/settings/_resource.py
+++ b/argilla/src/argilla/settings/_resource.py
@@ -25,7 +25,7 @@ from argilla._resource import Resource
 from argilla.settings._field import TextField
 from argilla.settings._metadata import MetadataType, MetadataField
 from argilla.settings._question import QuestionType, question_from_model, question_from_dict
-from argilla.settings._task_distribution import DEFAULT_TASK_DISTRIBUTION, TaskDistribution
+from argilla.settings._task_distribution import TaskDistribution
 from argilla.settings._vector import VectorField
 
 if TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


argilla/tests/unit/test_settings/test_settings.py
```diff
@@ -154,6 +154,15 @@ def test_settings_access_by_out_of_range(self):
         with pytest.raises(IndexError):
             _ = settings.fields[10]
 
+    def test_settings_with_modified_default_task_distribution(self):
```
davidberenstein1957, line 160: @frascuchon this test case seems unrelated to this PR?

We want to test if the code is backward compatible with existing datasets without distributions and ensure we don't overwrite them or get an error, correct?

frascuchon: Yes. I missed uploading the test. Thanks!

argilla/src/argilla/settings/_resource.py
```diff
@@ -133,7 +133,7 @@ def allow_extra_metadata(self, value: bool):
 
     @property
     def distribution(self) -> TaskDistribution:
-        return self._distribution
+        return self._distribution or TaskDistribution.default()
```
frascuchon, line 136: This change prevents to mutate the default task distribution value

jfcalvo (approved):

PR #5306, opened 2024-07-24
✨ Catch unhandled exceptions

damianpumar:


frascuchon (approved):
I tests it locally by forcing some errors and it's working!!!

Thanks!

leiyre (approved):

PR #5313, opened 2024-07-25
[REFACTOR] `argilla-server`: rewiew OAuth owner condition

frascuchon:
# Description

This PR changes the behavior to detect the user role when an OAuth sign-in occurs.

- If the connected user matches the `USERNAME`, the connected user becomes an owner.
- The rest of the users will be defined as `annotator`. 
- All logic related to roles in ORG has been removed until finding a proper auth scope.

**Type of change**

**How Has This Been Tested**

**Checklist**


argilla-server/docker/argilla-hf-spaces/scripts/start.sh
```diff
@@ -2,4 +2,17 @@
...
+export OAUTH2_HUGGINGFACE_CLIENT_ID=$OAUTH_CLIENT_ID
+export OAUTH2_HUGGINGFACE_CLIENT_SECRET=$OAUTH_CLIENT_SECRET
+export OAUTH2_HUGGINGFACE_SCOPE=$OAUTH_SCOPES
+
+# Set the space author name as username if no provided.
+# See https://huggingface.co/docs/hub/en/spaces-overview#helper-environment-variables for more details
+ARGILLA_USERNAME="${ARGILLA_USERNAME:-$SPACE_AUTHOR_NAME}"
+
+export ARGILLA_USERNAME="${USERNAME:-$ARGILLA_USERNAME}"
+export ARGILLA_PASSWORD="${ARGILLA_PASSWORD:-$PASSWORD}"
```
frascuchon:
```suggestion
export ARGILLA_PASSWORD="${PASSWORD:-$ARGILLA_PASSWORD}"
```

argilla-server/src/argilla_server/api/handlers/v1/oauth2.py
```diff
@@ -74,32 +73,22 @@ async def get_access_token(
...
-        return Token(access_token=JWT.create(user_info))
-    except ValueError as e:
-        raise HTTPException(status_code=400, detail=str(e)) from e
-    # TODO: Create exception handler for AuthenticationError
-    except AuthenticationError as e:
-        raise HTTPException(status_code=401, detail=str(e)) from e
+    userinfo = UserInfo(await provider.get_user_data(request)).use_claims(provider.claims)
+
+    if not userinfo.username:
+        raise RuntimeError("OAuth error: Missing username")
```
frascuchon, line 79: This is an unexpected error and probably the OAuth config is wrong. It will raise a 500 error

argilla-server/src/argilla_server/api/errors/v1/exception_handlers.py
```diff
@@ -75,3 +75,7 @@ async def update_distribution_with_existing_responses(request, exc):
             status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
             content={"code": exc.code, "message": exc.message},
         )
+
+    @app.exception_handler(errors.AuthenticationError)
+    async def authentication_error(request, exc):
```
jfcalvo:
```suggestion
    async def unauthorized_error(request, exc):
```

argilla-server/src/argilla_server/api/errors/v1/exception_handlers.py
```diff
@@ -75,3 +75,7 @@ async def update_distribution_with_existing_responses(request, exc):
...
+        return JSONResponse(status_code=status.HTTP_401_UNAUTHORIZED, content={"detail": str(exc)})
```
jfcalvo:
```suggestion
        return JSONResponse(
            status_code=status.HTTP_401_UNAUTHORIZED,
            # TODO: Once we move to v2.0 we can remove the content using detail attribute
            # and use the new one using code and message.
            # content={"code": exc.code, "message": exc.message},
            content={"detail": str(exc)},
        )
```

argilla-server/src/argilla_server/api/errors/v1/exception_handlers.py
jfcalvo: I would move this exception handler to be the first one (with the rest of generic errors) following the order of the status code, in this case `401`.

argilla-server/src/argilla_server/api/handlers/v1/oauth2.py
jfcalvo: then do you think it's ok to don't check for that condition and let it crash on line 81?

jfcalvo (changes requested):

argilla-server/src/argilla_server/api/handlers/v1/oauth2.py
frascuchon: No, I prefer to raise this error instead of a validation one.

argilla-server/src/argilla_server/api/errors/v1/exception_handlers.py
frascuchon: We have all error handler's names aligned with the exception class names. I will keep this aligned too.

jfcalvo (approved):

PR #5351, opened 2024-07-31
fix: add a generic message to api error

burtenshaw:
This PR adds a generic message to ArgillaAPIError so that that is always satisfies error printing of message attribute. 

**Type of change**

### DIFF
--- a/argilla/src/argilla/_exceptions/_api.py
+++ b/argilla/src/argilla/_exceptions/_api.py
@@ -84,7 +84,7 @@ def api_error_handler(func):
             500: InternalServerError,
         }
         exception_class = switch.get(status_code, ArgillaAPIError)
-        raise exception_class(f"{exception_class.message}. Details: {error_detail}")
+        raise exception_class(f"{exception_class.message}. Details: {error_detail}", status_code=status_code)
 
     def _handler_wrapper(*args, **kwargs):
         try:
--- a/argilla/src/argilla/_exceptions/_api.py
+++ b/argilla/src/argilla/_exceptions/_api.py
@@ -19,6 +19,8 @@ from argilla._exceptions._base import ArgillaError
 
 
 class ArgillaAPIError(ArgillaError):
+    message = "Server error"
+
     def __init__(self, message: Optional[str] = None, status_code: int = 500):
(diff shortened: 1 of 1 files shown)


argilla/src/argilla/_exceptions/_api.py
```diff
@@ -19,6 +19,8 @@
 
 
 class ArgillaAPIError(ArgillaError):
+    message = "Server error"
```
frascuchon, line 22: This `message` is used by the API error hander, and the `ArgillaApiError` will be used when the incoming status code has no mapping defined in the error handler:

```python
        switch = {
            400: BadRequestError,
            401: UnauthorizedError,
            403: ForbiddenError,
            404: NotFoundError,
            409: ConflictError,
            422: UnprocessableEntityError,
            500: InternalServerError,
        }
        exception_class = switch.get(status_code, ArgillaAPIError)
        raise exception_class(f"{exception_class.message}. Details: {error_detail}")
```

Maybe we can also pass the `status_code`, otherwise we will lose it when raising the error (since it's already defined as an argument in the `ArgillaAPIError.__init__` method)
```python
        raise exception_class(f"{exception_class.message}. Details: {error_detail}", status_code=status_code)
```

burtenshaw: I've passed the status code as well. 

>This message is used by the API error hander, and the ArgillaApiError will be used when the incoming status code has no mapping defined in the error handler:

I'm not sure I understand this fully. Did you also want the message implemented in a different way? i.e. in `ArgillaAPIError`.

frascuchon: Sorry... I mean, since the base exception class `ArgillaAPIError` already supports the `status_code` as an init argument (with a default value of 500), I think we could pass it when creating the [exception instance](https://github.com/argilla-io/argilla/blob/bugfix/add-message-to-api-error/argilla/src/argilla/_exceptions/_api.py#L87):

```python
        raise exception_class(f"{exception_class.message}. Details: {error_detail}", status_code=status_code)
```

and probably show it if we're not doing it already

frascuchon (approved):

PR #5356, opened 2024-07-31
[ENHANCEMENT] catch record with no known keys and raise detailed error

### LINKED ISSUE #5281: [BUG-python/deployment] Error thrown by Argilla SDK : The reason of error thrown by Argilla SDK is different / ambigious from the cause of the error
**Describe the bug**
The reason of error thrown by Argilla SDK is different from actual error.
If we use name of fields while creating dataset settings different for the name of fields when creating records in dataset, the error thrown is 
```
Argilla SDK error: RecordsIngestionError: 
Failed to ingest record from dict {'question': 'Do you need oxygen to breathe?', 'answer': 'Yes'}:
 If fields are an empty dictionary, an id must be provided.
```

```python
import argilla as rg

client = rg.Argilla(api_url=...,api_key=...)

labels= [..,..,..]

settings = rg.Settings(
    guidelines="Classify individual tokens according to the specified categories, ensuring that any overlapping or nested entities are accurately captured.",
    fields=[
        rg.TextField(
            name="text", # Note the name of the field.
            title="Text",
            use_markdown=False,
        ),
    ],
    questions=[
        rg.SpanQuestion(
            name="span_label",
            field="text",
            labels=labels,
            title="Classify the tokens according to the specified categories.",
            allow_overlapping=False,
        )
    ],
)
dataset = rg.Dataset(
    name="error_dataset",
    settings=settings,
)
dataset.create()
## When creating data to be stored as records, I have used different field names (question,answer)
data = [
    {
        "question": "Do you need oxygen to breathe?",
    },
    {
        "question": "What is the boiling point of water?",
    },
]
dataset.records.log(data)  ## this line with throw error
#### Error thrown
## RecordsIngestionError: Argilla SDK error: RecordsIngestionError: Failed to ingest record from dict {'question': 'Do you need ## oxygen to breathe?', 'answer': 'Yes'}: If fields are an empty dictionary, an id must be provided.
```
Colab notebook with exact replication steps can be found [here](https://colab.research.google.com/drive/1Y5uE1iG4isXYpqmV03gTEcl55xly3W-9?usp=sharing)
The error says, I need to pass id if the fields are empty dictionary.  The name of field while creating setting and name of field while creating records are different. The error thrown makes it look like we will have to pass the id field.
**Expected behavior**
The error thrown should mention about the difference in names of fields vs names used at record creation and not mention about id.

**Environment:**

 - Argilla Version :2.0.0rc2
 - ElasticSearch Version : 8.8.2 ( mostly, not very sure. I am using HF hosted Argilla space)

**Additional context**

burtenshaw:
# Description

This PR responds to #5281 by catching records with no known keys and raising an ingestion error.

**Type of change**

### DIFF
--- a/argilla/src/argilla/_exceptions/_api.py
+++ b/argilla/src/argilla/_exceptions/_api.py
@@ -82,7 +82,7 @@ def api_error_handler(func):
             500: InternalServerError,
         }
         exception_class = switch.get(status_code, ArgillaAPIError)
-        raise exception_class(f"{exception_class.message}. Details: {error_detail}")
+        raise exception_class(f"{exception_class.message}. Details: {error_detail}", status_code=status_code)
 
     def _handler_wrapper(*args, **kwargs):
         try:
--- a/argilla/src/argilla/records/_mapping/_mapper.py
+++ b/argilla/src/argilla/records/_mapping/_mapper.py
@@ -15,6 +15,7 @@
 import re
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Sequence, Union, Tuple
 from uuid import UUID
+import warnings
 
 from argilla._exceptions import RecordsIngestionError
 from argilla.records._resource import Record
@@ -82,6 +83,13 @@ class IngestedRecordMapper:
         fields = self._map_attributes(data=data, mapping=self.mapping.field)
         metadata = self._map_attributes(data=data, mapping=self.mapping.metadata)
         vectors = self._map_attributes(data=data, mapping=self.mapping.vector)
+
+        if data and not record_id and not suggestions and not responses and not fields and not metadata and not vectors:
+            raise RecordsIngestionError(
+                message="""Record has no identifiable keys. If keys in source dataset 
+                do not match the names in `dataset.settings`, you should use a 
+                `mapping` with `dataset.records.log`."""
+            )
 
         return Record(
             id=record_id,
--- a/argilla/src/argilla/records/_mapping/_mapper.py
+++ b/argilla/src/argilla/records/_mapping/_mapper.py
@@ -88,8 +88,10 @@ class IngestedRecordMapper:
         if unknown_keys:
             warnings.warn(f"Keys {unknown_keys} in data are not present in the mapping and will be ignored.")
 
-        if len([k for k in data if k != self.mapping.id.source ]) == 0:
-            raise RecordsIngestionError(message=f"Record has no data. All records must have at least one attribute. Record id: {record_id}.")
+        if len([k for k in data if k != self.mapping.id.source]) == 0:
+            raise RecordsIngestionError(
+                message=f"Record has no data. All records must have at least one attribute. Record id: {record_id}."
+            )
 
         if data and not (record_id or suggestions or responses or fields or metadata or vectors):
             raise RecordsIngestionError(
--- a/argilla/src/argilla/records/_mapping/_mapper.py
+++ b/argilla/src/argilla/records/_mapping/_mapper.py
@@ -85,8 +85,8 @@ class IngestedRecordMapper:
 
         if data and not record_id and not suggestions and not responses and not fields and not metadata and not vectors:
             raise RecordsIngestionError(
-                message="""Record has no identifiable keys. If keys in source dataset 
-                do not match the names in `dataset.settings`, you should use a 
+                message="""Record has no identifiable keys. If keys in source dataset
+                do not match the names in `dataset.settings`, you should use a
                 `mapping` with `dataset.records.log`."""
             )
 
--- a/argilla/src/argilla/records/_mapping/_mapper.py
+++ b/argilla/src/argilla/records/_mapping/_mapper.py
@@ -15,7 +15,6 @@
 import re
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Sequence, Union, Tuple
 from uuid import UUID
-import warnings
 
 from argilla._exceptions import RecordsIngestionError
 from argilla.records._resource import Record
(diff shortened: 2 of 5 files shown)


argilla/src/argilla/records/_mapping/_mapper.py
```diff
@@ -83,6 +83,13 @@ def __call__(self, data: Dict[str, Any], user_id: Optional[UUID] = None) -> Reco
         metadata = self._map_attributes(data=data, mapping=self.mapping.metadata)
         vectors = self._map_attributes(data=data, mapping=self.mapping.vector)
 
+        if data and not record_id and not suggestions and not responses and not fields and not metadata and not vectors:
```
frascuchon:
```suggestion
        if data and not(record_id or suggestions or responses or fields or metadata or vectors):
```

argilla/src/argilla/records/_mapping/_mapper.py
```diff
@@ -83,6 +83,13 @@ def __call__(self, data: Dict[str, Any], user_id: Optional[UUID] = None) -> Reco
...
+            raise RecordsIngestionError(
+                message="""Record has no identifiable keys. If keys in source dataset
+                do not match the names in `dataset.settings`, you should use a
+                `mapping` with `dataset.records.log`."""
```
frascuchon: Besides this message, are we clearly showing which keys are not recognized?

burtenshaw: I've added a warning when no known keys are in data.

argilla/src/argilla/records/_mapping/_mapper.py
```diff
@@ -83,6 +84,19 @@ def __call__(self, data: Dict[str, Any], user_id: Optional[UUID] = None) -> Reco
         metadata = self._map_attributes(data=data, mapping=self.mapping.metadata)
         vectors = self._map_attributes(data=data, mapping=self.mapping.vector)
 
+        unknown_keys = [key for key in data.keys() if key not in self.mapping.keys()]
+        if unknown_keys:
+            warnings.warn(f"Keys {unknown_keys} in data are not present in the mapping and will be ignored.")
+
+        if data and not (record_id or suggestions or responses or fields or metadata or vectors):
```
frascuchon, line 96: Is there a way where `data` is empty? Should we add a check for that?

argilla/src/argilla/_exceptions/_api.py
```diff
@@ -82,7 +82,7 @@ def _error_switch(status_code: int, error_detail: str):
             500: InternalServerError,
         }
         exception_class = switch.get(status_code, ArgillaAPIError)
-        raise exception_class(f"{exception_class.message}. Details: {error_detail}")
+        raise exception_class(f"{exception_class.message}. Details: {error_detail}", status_code=status_code)
```
frascuchon, line 85: This is related to my comment in https://github.com/argilla-io/argilla/pull/5351?

burtenshaw: Yeah. Sorry the commit is on both branches now.

argilla/src/argilla/records/_mapping/_mapper.py
burtenshaw: Yes. `data` could be empty if the user has a ragged data structure where some records are empty. I would not check for that because it doesn't cause us a problem, and there's no data that we're missing. It's just their data.

frascuchon: Okay. In that case, the record creation will fail, right?

burtenshaw: Correct. Do you think an explicit error here would be better?

frascuchon: I would say yes. This would simplify error analysis for users.

burtenshaw: @frascuchon Ok. I've responded to that.

argilla/src/argilla/records/_mapping/_mapper.py
```diff
@@ -83,6 +84,24 @@ def __call__(self, data: Dict[str, Any], user_id: Optional[UUID] = None) -> Reco
         metadata = self._map_attributes(data=data, mapping=self.mapping.metadata)
         vectors = self._map_attributes(data=data, mapping=self.mapping.vector)
 
+        unknown_keys = [key for key in data.keys() if key not in self.mapping.keys()]
+        if unknown_keys:
+            warnings.warn(f"Keys {unknown_keys} in data are not present in the mapping and will be ignored.")
+
+        if len([k for k in data if k != self.mapping.id.source]) == 0:
```
frascuchon, line 91: can we rewrite the condition to:
```suggestion
        if not data:
```
? 

If data have some value there is the other check to raise the error ( Maybe I'm missing some scenarios)

burtenshaw: Here I'm also checking that the only value is not an id, because that will fail the same as `{}`

argilla/tests/unit/test_record_ingestion.py
```diff
@@ -50,6 +51,19 @@ def test_ingest_record_from_dict(dataset):
...
 
+def test_ingest_record_from_empty_dict_raises(dataset):
+    with pytest.raises(RecordsIngestionError):
+        dataset.records._ingest_records(
+            records=[
+                {
+                    "prompt": "What is the capital of France?",
+                    "label": "positive",
+                },
+                {},
```
frascuchon, line 64: This should also fail:
```suggestion
                {},
                { "id": "record_id" },
```

burtenshaw: Yes, but line 62 will fail first. I've added separate raises contexts for each case.

frascuchon: ahh sorry, I thought it was a test parameter.

frascuchon (approved):

PR #5358, opened 2024-07-31
[FEATURE] from_hub raise on existing dataset name

### LINKED ISSUE #5346: [FEATURE] review `Dataset.from_hub` -> `Dataset.create()` flow.

When a dataset with the same `name` and `workspace` is present, I would expect to sync the records or fail and not suddenly create a new dataset with a random UUID.
```python
dataset = rg.Dataset.from_hub(repo_id=f"argilla/{dataset}")
try:
    dataset.create()
except Exception as e:
    print(e)
```

An error message.

NA

**Additional context**

NA

burtenshaw:
# Description

This PR changes the behaviour of the `from_disk` method when a dataset of the same name already exists. Currently, a new dataset is create with the name + uuid. This change will:

- check id the dataset name exists, and if so
- warn that the dataset exists and that the `name` parameter could be used to create a new one
- try to push records to the existing dataset with a try except to add more context

Closes #5346 

**Type of change**

### DIFF
--- a/argilla/src/argilla/datasets/_export/_disk.py
+++ b/argilla/src/argilla/datasets/_export/_disk.py
@@ -21,7 +21,6 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Optional, Tuple, Type, Union
 
 from argilla._exceptions import RecordsIngestionError
-from argilla._exceptions._base import ArgillaError
 from argilla._models import DatasetModel
 from argilla.client import Argilla
 from argilla.settings import Settings
@@ -100,11 +99,11 @@ class DiskImportExportMixin(ABC):
         dataset_model.workspace_id = workspace_id
 
         if name:
-            logging.warning(f"Changing dataset name from {dataset_model.name} to {name}")
+            logging.info(f"Changing dataset name from {dataset_model.name} to {name}")
             dataset_model.name = name
 
         if client.api.datasets.name_exists(name=dataset_model.name, workspace_id=workspace_id):
-            logging.warning(
+            warnings.warn(
                 f"Loaded dataset name {dataset_model.name} already exists in the workspace so using it. To create a new dataset, provide a unique name to the `name` parameter."
             )
             dataset_model = client.api.datasets.get_by_name_and_workspace_id(
--- a/argilla/src/argilla/datasets/_export/_disk.py
+++ b/argilla/src/argilla/datasets/_export/_disk.py
@@ -21,6 +21,8 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Optional, Tuple, Type, Union
 from uuid import uuid4
 
+from argilla._exceptions import RecordsIngestionError
+from argilla._exceptions._base import ArgillaError
 from argilla._models import DatasetModel
 from argilla.client import Argilla
 from argilla.settings import Settings
@@ -114,7 +116,10 @@ class DiskImportExportMixin(ABC):
             dataset.create()
 
         if os.path.exists(records_path) and with_records:
-            dataset.records.from_json(path=records_path)
+            try:
+                dataset.records.from_json(path=records_path)
+            except RecordsIngestionError as e:
+                raise ArgillaError(message="Error importing dataset records from disk. Records and datasets settings are not compatible.") from e
         return dataset
 
     ############################
--- a/argilla/src/argilla/datasets/_export/_disk.py
+++ b/argilla/src/argilla/datasets/_export/_disk.py
@@ -90,24 +90,24 @@ class DiskImportExportMixin(ABC):
 
         # Get the relevant workspace_id of the incoming dataset
         if isinstance(workspace, str):
-            workspace_id = client.workspaces(workspace).id
-        elif isinstance(workspace, Workspace):
-            workspace_id = workspace.id
+            workspace = client.workspaces(workspace)
+            if not workspace:
+                raise ValueError(f"Workspace {workspace} not found.")
         else:
             warnings.warn("Workspace not provided. Using default workspace.")
-            workspace_id = client.workspaces.default.id
-        dataset_model.workspace_id = workspace_id
+            workspace = client.workspaces.default
+        dataset_model.workspace_id = workspace.id
 
         if name and (name != dataset_model.name):
             logging.info(f"Changing dataset name from {dataset_model.name} to {name}")
             dataset_model.name = name
 
-        if client.api.datasets.name_exists(name=dataset_model.name, workspace_id=workspace_id):
+        if client.api.datasets.name_exists(name=dataset_model.name, workspace_id=workspace.id):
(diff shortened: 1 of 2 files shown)


argilla/src/argilla/datasets/_export/_disk.py
```diff
@@ -19,8 +19,9 @@
 from abc import ABC
 from pathlib import Path
 from typing import TYPE_CHECKING, Optional, Tuple, Type, Union
-from uuid import uuid4
 
+from argilla._exceptions import RecordsIngestionError
+from argilla._exceptions._base import ArgillaError
```
davidberenstein1957: can't this be combined in one import?

burtenshaw:
@davidberenstein1957 @frascuchon 

I followed the recommendation here #5346 , but I'm a bit worried about pushing records to an unkown dataset, so I wrapped the push in a try/except. Let me know your thoughts?

We could also just raise when the dataset exists, and not try to push.

argilla/src/argilla/datasets/_export/_disk.py
burtenshaw: Sure. Tbh, I think I should just use `RecordsIngestionError` for both.

davidberenstein1957:
> @davidberenstein1957 @frascuchon
> 
> I followed the recommendation here #5346 , but I'm a bit worried about pushing records to an unkown dataset, so I wrapped the push in a try/except. Let me know your thoughts?
> 
> We could also just raise when the dataset exists, and not try to push.

@burtenshaw I think trying to push is fine and does what I would expect. In case you don't want records, you can just set `with_records=False`.

argilla/src/argilla/datasets/_export/_disk.py
```diff
@@ -98,20 +99,28 @@ def from_disk(
...
-            dataset_model.name = f"{dataset_model.name}_{uuid4()}"
-
-        # Create the dataset and load the settings and records
-        dataset = cls.from_model(model=dataset_model, client=client)
-        dataset.settings = Settings.from_json(path=settings_path)
-        dataset.create()
+
+        if client.api.datasets.name_exists(name=dataset_model.name, workspace_id=workspace_id):
+            logging.warning(
+                f"Loaded dataset name {dataset_model.name} already exists so using it. To create a new dataset, provide a unique name to the `name` parameter."
```
davidberenstein1957: Maybe specify it already exists within the specific workspace.

argilla/src/argilla/datasets/_export/_disk.py
```diff
@@ -98,20 +99,28 @@ def from_disk(
             workspace_id = client.workspaces.default.id
         dataset_model.workspace_id = workspace_id
 
-        # Get a relevant and unique name for the incoming dataset.
         if name:
             logging.warning(f"Changing dataset name from {dataset_model.name} to {name}")
```
davidberenstein1957: why is this a warning and not info? shouldn't we also only warn if the names and workspaces aren't equal?

argilla/src/argilla/datasets/_export/_disk.py
```diff
@@ -98,20 +98,29 @@ def from_disk(
             workspace_id = client.workspaces.default.id
         dataset_model.workspace_id = workspace_id
 
-        # Get a relevant and unique name for the incoming dataset.
         if name:
```
davidberenstein1957:
```suggestion
        if name and (name != dataset_model.name):
```

burtenshaw:
@davidberenstein1957 I've responded to comments on this and expanded testing.

davidberenstein1957 (approved):

argilla/src/argilla/datasets/_export/_disk.py
```diff
@@ -98,20 +98,29 @@ def from_disk(
...
             dataset_model.name = name
-        elif client.api.datasets.name_exists(name=dataset_model.name, workspace_id=workspace_id):
-            logging.warning(f"Loaded dataset name {dataset_model.name} already exists. Changing to unique UUID.")
-            dataset_model.name = f"{dataset_model.name}_{uuid4()}"
 
+        if client.api.datasets.name_exists(name=dataset_model.name, workspace_id=workspace_id):
+            warnings.warn(
+                f"Loaded dataset name {dataset_model.name} already exists in the workspace so using it. To create a new dataset, provide a unique name to the `name` parameter."
+            )
+            dataset_model = client.api.datasets.get_by_name_and_workspace_id(
```
frascuchon, line 109: If we're using the existing dataset here, the line [115](https://github.com/argilla-io/argilla/pull/5358/files#diff-05e119be93c43ed87ec50de3e71dbd6002e78b359d806d17214cd445022dd3f1L112) will fail, right?

argilla/src/argilla/datasets/_export/_disk.py
```diff
@@ -98,20 +98,29 @@ def from_disk(
...
+        if name and (name != dataset_model.name):
+            logging.info(f"Changing dataset name from {dataset_model.name} to {name}")
             dataset_model.name = name
-        elif client.api.datasets.name_exists(name=dataset_model.name, workspace_id=workspace_id):
-            logging.warning(f"Loaded dataset name {dataset_model.name} already exists. Changing to unique UUID.")
-            dataset_model.name = f"{dataset_model.name}_{uuid4()}"
 
+        if client.api.datasets.name_exists(name=dataset_model.name, workspace_id=workspace_id):
+            warnings.warn(
+                f"Loaded dataset name {dataset_model.name} already exists in the workspace so using it. To create a new dataset, provide a unique name to the `name` parameter."
```
frascuchon: Maybe we can add info about the workspace here
```suggestion
                f"Loaded dataset name {dataset_model.name} already exists in the workspace {workspace.name} so using it. To create a new dataset, provide a unique name to the `name` parameter."
```

argilla/src/argilla/datasets/_export/_disk.py
burtenshaw: Yes. Corrected.

argilla/src/argilla/datasets/_export/_disk.py
```diff
@@ -90,28 +90,39 @@ def from_disk(
 
         # Get the relevant workspace_id of the incoming dataset
         if isinstance(workspace, str):
-            workspace_id = client.workspaces(workspace).id
-        elif isinstance(workspace, Workspace):
-            workspace_id = workspace.id
+            workspace = client.workspaces(workspace)
+            if not workspace:
+                raise ValueError(f"Workspace {workspace} not found.")
```
frascuchon: Should we use the base ArgillaError instead?
```suggestion
                raise ArgillaError(f"Workspace {workspace} not found.")
```

frascuchon (approved):
Just a comment about errors. I'm not sure but maybe the SDK should raise its exceptions.

PR #5362, opened 2024-07-31
[BUGFIX] `argilla`: `TextField(required=false)` does not allow me to pass empty record entry

### LINKED ISSUE #5359: [BUG-python/deployment] `TextField(required=False)` does not allow me to pass empty record entry
**Describe the bug**

I defined with 2 required `TextField` and 4 optional ones but I get an error when uploading records without the optional fields.

https://huggingface.co/spaces/argilla/argilla-template-space
rag-embeddings-relevance-similarity

I've got issues with the following
```python
# this does not work
{
    "required" : "value",
    "optional" : None
}
{
    "required" : "value"
}
# this does work
{
    "required" : "value",
    "optional" : ""
}
```
```bash
argilla._exceptions._api.UnprocessableEntityError: Argilla SDK error: UnprocessableEntityError: Unprocessable entity. The server cannot process the request. Details: {"detail":"record at position 147 is not valid because missing required value for field: 'document-4'"}
```

**Expected behavior**

I would expect this to not pose problems.

**Environment:**

 - Argilla Version [e.g. 1.0.0]: 2.0.0
 - Docker Image (optional) [e.g. argilla:v1.0.0]: 2.0.0

**Additional context**

NA

frascuchon:
# Description

This PR prevents converting `required` to `True` when `False` is passed

Closes #5359

**Type of change**

**How Has This Been Tested**

**Checklist**

### DIFF
--- a/argilla-server/tests/unit/validators/test_records_bulk.py
+++ b/argilla-server/tests/unit/validators/test_records_bulk.py
@@ -29,6 +29,7 @@ class TestRecordsBulkValidators:
         dataset = await DatasetFactory.create(status="ready")
 
         await TextFieldFactory.create(name="text", dataset=dataset)
+        await TextFieldFactory.create(name="optional", dataset=dataset, required=False)
         await dataset.awaitable_attrs.fields
 
         await dataset.awaitable_attrs.metadata_properties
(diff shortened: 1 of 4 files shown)


argilla-server/tests/unit/validators/test_records_bulk.py
```diff
@@ -40,7 +41,9 @@ async def test_records_bulk_create_validator(self, db: AsyncSession):
 
         records_create = RecordsBulkCreate(
             items=[
-                RecordCreate(fields={"text": "hello world"}, metadata={"source": "test"}),
+                RecordCreate(fields={"text": "hello world", "optional": "optional"}, metadata={"source": "test"}),
+                RecordCreate(fields={"text": "hello world", "optional": ""}, metadata={"source": "test"}),
```
davidberenstein1957, line 45: @frascuchon @burtenshaw, I would perhaps also expect not to be allowed to pass an empty string to a required field. WDYT?

frascuchon: Yes, it makes sense. But I wouldn't include it here, since changes could have other side effects besides the bug fix. May you open an issue for this?

burtenshaw (approved):

PR #5376, opened 2024-08-05
[FEATURE] Chat field

### LINKED ISSUE #3338: [FEATURE] Add conversation support to fields in Argilla dataset (ChatField)
It is not clear how to use Feedback scheme to store multi-turn conversation.  It doesn't look like there is support in the web interface to annotate feedback for multi-turn conversations. A lot of Instruction Tuning data is multi-turn and the Feedback scheme only allows to record feedback to a single response.  Other schemes (i.e. Text2Text) also seem to not be suitable.

Native support for multi-turn, allowing to annotate feedback for each turn of the conversation.

Getting feedback on each turn individually and then providing context (full conversation) for each turn.  This is not great, as this results in the need to re-read each conversation N times, where N is the number of turns.  Slowing down the progress substantially.

Alternative is storing the complete conversation as text. And using the external tool, like Gradio to annotate.  This diminishes the value of Argilla, as it requires to add external/not integrated tool, disconnected from search / etc.  Storing the conversation as text, instead of structured data also reduces the ability to filter the conversation in the structured way (i.e. by last response).  Alternative is storing the complete conversation as a .json field.  This is also not great, as nearly all tooling of argilla do not have any support for this.

**Additional context**
Ideally this should be natively supported in the web UI.  But this may be implemented using a tool/plugin outside the current web/gui.  Using Gradio/Chat as a frontend may be a good option.  But either way, this still requires support at the schema level.

burtenshaw:
This is feature PR for the chat field.

It contains changes for:
- frontend
- sdk
- server

We need to work on:
- docs


argilla-frontend/v1/domain/entities/field/Field.ts
```diff
@@ -14,12 +14,17 @@ export class Field {
     public settings: any
   ) {
     this.initializeOriginal();
+    console.log(this);
```
jfcalvo:
```suggestion
```

argilla-server/CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ These are the section headers that we use:
 ### Added
 
 - Added new `image` type dataset field supporting URLs and Data URLs. ([#5279](https://github.com/argilla-io/argilla/pull/5279))
+- Added new `chat` type dataset field supporting chat-like conversations. ([#5375](https://github.com/argilla-io/argilla/pull/5375))
```
jfcalvo:
```suggestion
- Added new `chat` type dataset field supporting chat-like conversations. ([#5376](https://github.com/argilla-io/argilla/pull/5376))
```

argilla-server/src/argilla_server/api/schemas/v1/records.py
```diff
@@ -82,7 +82,7 @@ class Config:
 
 
 class RecordCreate(BaseModel):
-    fields: Dict[str, Union[StrictStr, None]]
+    fields: Dict[str, Union[StrictStr, None, List[Dict[StrictStr, StrictStr]]]]
```
jfcalvo: I believe we should define a Pydantic model specific for the chat field.

So we define the `fields` attribute with something similar to:

```suggestion
    fields: Dict[str, Union[StrictStr, List[RecordChatField], None]]
```

And with `RecordChatField` being:

```python
class RecordChatField(BaseModel):
    role: str
    content: str
```

You can add validations to `RecordChatField`, define a `RecordChatFieldCreate` and `RecordChatFieldUpdate` if necessary, etc.

If necessary an alternative to this is define a `RecordChatFieldValue` as a list of `RecordChatField`'s with validations:

```suggestion
    fields: Dict[str, Union[StrictStr, RecordChatFieldValue, None]]
```

And with `RecordChatFieldValue` being:

```python
RecordChatFieldValue = Annotated[conlist(item_type=RecordChatField, min_items=1), Field(..., description="...")] 
```

argilla-server/src/argilla_server/validators/records.py
```diff
@@ -126,6 +128,38 @@ def _validate_image_field(cls, field_name: str, field_value: Union[str, None]) -
...
+
+    @classmethod
+    def _validate_chat_field(cls, field_name: str, field_value: Union[str, None]) -> None:
+        if field_value is None:
+            return
+
+        if len(field_value) > CHAT_FIELD_MAX_LENGTH:
+            raise UnprocessableEntityError(
+                f"chat field {field_name!r} value is exceeding the maximum length of {CHAT_FIELD_MAX_LENGTH} characters"
+            )
```
jfcalvo: I don't get this. Is not the chat field value a list of objects? Then should the message be changed?

argilla-server/src/argilla_server/validators/records.py
```diff
@@ -126,6 +128,38 @@ def _validate_image_field(cls, field_name: str, field_value: Union[str, None]) -
         else:
             raise UnprocessableEntityError(f"image field {field_name!r} has an invalid URL value")
 
+    @classmethod
+    def _validate_chat_fields(cls, dataset: Dataset, fields: Dict[str, str]) -> None:
+        for field in filter(lambda field: field.is_chat, dataset.fields):
+            cls._validate_chat_field(field.name, fields.get(field.name))
+
+    @classmethod
+    def _validate_chat_field(cls, field_name: str, field_value: Union[str, None]) -> None:
```
jfcalvo: Maybe we don't need some of these validations if you add a Pydantic model.

jfcalvo (changes requested):

argilla-server/src/argilla_server/api/schemas/v1/records.py
burtenshaw: I agree that a Pydantic model would be ideal here. 

However, we want to allow the object to have more keys/values than just 'role' and 'content' because these structures aren't fixed. Let's discuss this.

argilla-server/src/argilla_server/validators/records.py
burtenshaw: Yes. This message is wrong. It should be 'max messages'. I will fix when we resolve the pydantic decision.

PR #5386, opened 2024-08-06
[FEATURE] Add retries to the internal `httpx.Client` used by the SDK

gabrielmbmb:
# Description

This PR adds a new argument `retries` that can be used to specify the number of times that an HTTP request performed by the internal `httpx.Client` used by the SDK should be retried before raising an exception.

This is useful as sometimes while using the `dataset.records.log` you can receive a `ConnectionError` or 5xx from the server and if you retry a few seconds later everything is fine.

**Type of change**

- Improvement

**How Has This Been Tested**

**Checklist**

### DIFF
--- a/argilla/src/argilla/_api/_http/_client.py
+++ b/argilla/src/argilla/_api/_http/_client.py
@@ -23,12 +23,8 @@ class HTTPClientConfig:
 
     api_url: str
     api_key: str
-    timeout: int = None
-
-    def __post_init__(self):
-        self.api_url = self.api_url
-        self.api_key = self.api_key
-        self.timeout = self.timeout or 60
+    timeout: int = 60
+    retries: int = 5
 
 
 def create_http_client(api_url: str, api_key: str, **client_args) -> httpx.Client:
@@ -37,5 +33,11 @@ def create_http_client(api_url: str, api_key: str, **client_args) -> httpx.Clien
 
     headers = client_args.pop("headers", {})
     headers["X-Argilla-Api-Key"] = api_key
+    retries = client_args.pop("retries", 0)
 
-    return httpx.Client(base_url=api_url, headers=headers, **client_args)
+    return httpx.Client(
+        base_url=api_url,
+        headers=headers,
+        transport=httpx.HTTPTransport(retries=retries),
+        **client_args,
+    )
--- a/argilla/src/argilla/client.py
+++ b/argilla/src/argilla/client.py
@@ -41,13 +41,7 @@ class Argilla(_api.APIClient):
         datasets: A collection of datasets.
         users: A collection of users.
         me: The current user.
-
     """
-
-    workspaces: "Workspaces"
-    datasets: "Datasets"
-    users: "Users"
-    me: "User"
 
     # Default instance of Argilla
     _default_client: Optional["Argilla"] = None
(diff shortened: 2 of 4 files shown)


argilla/src/argilla/_api/_http/_client.py
```diff
@@ -23,12 +23,8 @@ class HTTPClientConfig:
 
     api_url: str
     api_key: str
-    timeout: int = None
-
-    def __post_init__(self):
-        self.api_url = self.api_url
-        self.api_key = self.api_key
-        self.timeout = self.timeout or 60
+    timeout: int = 60
```
davidberenstein1957, line 26: @gabrielmbmb I think we can add this to the constants.py file.

davidberenstein1957 (commented):
@gabrielmbmb should we define upon which failures we retry. I think it only makes sense for the 500 series and perhaps on request timeout or anything.  https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses
if payload is invalid or something, we don't want to retry.

argilla/src/argilla/_api/_http/_client.py
burtenshaw: @gabrielmbmb The refactor of `timeout` and `retries` looks good to me. It's more consistent and avoids background passing with `**client_args`.

@davidberenstein1957 A few PRs have used constants in context lately. I think we should tackle this in a dedicated PR because using a single constant file is not universally good.

burtenshaw:
@gabrielmbmb Implementation looks good. What do you think about expanding the testing for this? I thought of these two but open to your suggestions:
- One unit for `retries` attributes in here `argilla/tests/unit/api/http/test_http_client.py`.
- One using `pytest_httpx` mock and checking the number of retries. Documented [here](https://colin-b.github.io/pytest_httpx/#check-sent-requests)

argilla/src/argilla/_api/_http/_client.py
```diff
@@ -37,5 +33,11 @@ def create_http_client(api_url: str, api_key: str, **client_args) -> httpx.Clien
...
     headers = client_args.pop("headers", {})
     headers["X-Argilla-Api-Key"] = api_key
-
-    return httpx.Client(base_url=api_url, headers=headers, **client_args)
+    retries = client_args.pop("retries", 0)
+
+    return httpx.Client(
+        base_url=api_url,
+        headers=headers,
+        transport=httpx.HTTPTransport(retries=retries),
```
frascuchon, line 41: Just a comment. The `retries` argument is used when establishing connections. This won't work when for example a 503 error response is returned. The docs reference another approach for that case. https://www.python-httpx.org/advanced/transports/#http-transport

frascuchon: @burtenshaw what do you think about this? Should we add a retry mechanism for some 50x errors before merge this?

burtenshaw: IMO, it's ok to only retry on error other than 50* because to get a 50* error the client needs to connect and make a valid request.

burtenshaw:
@frascuchon @davidberenstein1957 What's your view of this PR? I've added a test and the remaining comments are: 
- [variables in module or shared constants](https://github.com/argilla-io/argilla/pull/5386#discussion_r1714830438)
- [dealing with 50* errors](https://github.com/argilla-io/argilla/pull/5386#discussion_r1743776609)

davidberenstein1957:
> @frascuchon @davidberenstein1957 What's your view of this PR? I've added a test and the remaining comments are:
> 
> * [variables in module or shared constants](https://github.com/argilla-io/argilla/pull/5386#discussion_r1714830438)
> * [dealing with 50* errors](https://github.com/argilla-io/argilla/pull/5386#discussion_r1743776609)

I think there are several error that can occur and will never be fixed by retries, like authentication or wrong payload, so for me it doesn't make sense to retry during those cases. 

W.r.t. variables vs shared constants, I don't have a strict opinion besides having everything else in `constants.py` already.

argilla/src/argilla/_api/_http/_client.py
frascuchon: My comment was related to the PR description:

> This is useful as sometimes while using the dataset.records.log you can receive a ConnectionError or 5xx from the server and if you retry a few seconds later everything is fine.

Sometimes, 502 or 503 errors are temporal network problems or restarts.  But it's fine for me if we add retries when creating the connection.

frascuchon (approved):

argilla/src/argilla/client.py
```diff
@@ -57,9 +51,24 @@ def __init__(
...
+            api_url: the URL of the Argilla API. If not provided, then the value will try
+                to be set from `ARGILLA_API_URL` environment variable. Defaults to
+                `"http://localhost:6900"`.
+            api_key: the key to be used to authenticate in the Argilla API. If not provided,
+                then the value will try to be set from `ARGILLA_API_KEY` environment variable.
+                Defaults to `None`.
+            timeout: the maximum time in seconds to wait for a request to the Argilla API
+                to be completed before raising an exception. Defaults to `60`.
+            retries: the number of times to retry a failed HTTP request to the Argilla API
+                before raising an exception. Defaults to `5`.
```
frascuchon: Based on the[ httpx docs](https://www.python-httpx.org/advanced/transports/), we should change the value description
```suggestion
            retries: the number of times to retry the HTTP connection to the Argilla API
                before raising an exception. Defaults to `5`.
```

PR #5412, opened 2024-08-14
🌚 feat: App dark theme

### LINKED ISSUE #5371: [UI/UX] Implement dark theme
At the moment the Argilla app only has available the light theme, the idea is to implement the dark mode as well. The automatic mode will be the one that the user has in the preferences from the system settings and can also be changed through a toggle.

leiyre:
fix #5371


Josephrp:
thanks for this :-)

PR #5422, opened 2024-08-21
[FEATURE] custom-field-support

burtenshaw:
This PR implements a custom field on the server and in the SDK.

## Server

Fields response is:

```json
{
    "items": [
        {
            "id": "11aff30d-e5cf-4485-b646-07adab0bc5b7",
            "name": "name",
            "title": "name",
            "required": true,
            "settings": {
                "type": "text",
                "use_markdown": false
            },
            "dataset_id": "7cccb1e6-9053-4919-b824-43cdb380f7b8",
            "inserted_at": "2024-08-21T10:47:10.471676",
            "updated_at": "2024-08-21T10:47:10.471676"
        },
        {
            "id": "9650009e-e198-40e3-827f-8e0474287fd6",
            "name": "custom_field",
            "title": "custom_field",
            "required": true,
            "settings": {
                "type": "custom",
                "template": "<html></html>"
            },
            "dataset_id": "7cccb1e6-9053-4919-b824-43cdb380f7b8",
            "inserted_at": "2024-08-21T10:47:10.481962",
            "updated_at": "2024-08-21T10:47:10.481962"
        }
    ]
}
```

## SDK

```python
import argilla as rg

client = rg.Argilla()

settings = rg.Settings(
    fields = [
        rg.TextField('name'),
        rg.CustomField('custom_field', '<html></html>'),
    ],
    questions=[
        rg.TextQuestion('response'),
    ]
)

dataset = rg.Dataset(
    settings=settings,
    name="custom_4"
)

dataset.create()
````

### CODE REVIEW

File: argilla/src/argilla/settings/_field.py [hunk 1]
```diff
@@ -199,7 +207,74 @@ def use_markdown(self, value: bool) -> None:
...
-Field = Union[TextField, ImageField, ChatField]
+class CustomField(AbstractField):
+    """Custom field for use in Argilla `Dataset` `Settings`"""
+
+    def __init__(
+        self,
+        name: str,
+        template: Optional[str] = "",
+        advanced_mode: Optional[bool] = False,
+        title: Optional[str] = None,
```
davidberenstein1957:
```suggestion
        title: Optional[str] = None,
        template: Optional[str] = "",
        advanced_mode: Optional[bool] = False,
```

argilla-frontend/components/features/annotation/container/fields/custom-field/CustomField.vue
```diff
@@ -0,0 +1,145 @@
...
+  const currentHead = document.getElementsByTagName("head")[0];
+  const styles = parent.document.getElementsByTagName("style");
+  for (const style of styles) {
+    currentHead.appendChild(style.cloneNode(true));
+  }
+}
+<\/script>`;
+
+const BASIC_TEMPLATE = `
+${STYLES}
```
davidberenstein1957: @damianpumar @burtenshaw I think it would be more straightforward to just use `record` as naming instead of `record_object`. WDYT?

burtenshaw: Sounds good to me.

damianpumar: Yes,! agree. I change it now!

Thanks

argilla-server/src/argilla_server/api/schemas/v1/fields.py
```diff
@@ -88,11 +88,30 @@ class ChatFieldSettingsUpdate(BaseModel):
     use_markdown: bool
 
 
+class CustomFieldSettings(BaseModel):
+    type: Literal[FieldType.custom]
+    template: str
+    advanced_mode: bool
```
jfcalvo, line 94: What is `advanced_mode`?

argilla-server/src/argilla_server/validators/records.py
jfcalvo, line 1: I remember discussing about adding (handlerbars) template validation to the custom field settings? We don't want to do it at the end?

argilla-frontend/components/features/annotation/container/fields/custom-field/CustomField.vue
```diff
@@ -0,0 +1,119 @@
...
+/* eslint-disable */
+const BASIC_TEMPLATE = `
+<script>const record = #RECORD_OBJECT#;<\/script>
+<script src="./js/handlebars.min.js"><\/script>
+<div id="template">
+#TEMPLATE#
+</div>
+<script>
+  Handlebars.registerHelper("json", (context) => {
+    return JSON.stringify(context);
```
frascuchon:
```suggestion
    return JSON.stringify(context, null, 4);
```

argilla-server/src/argilla_server/api/schemas/v1/fields.py
frascuchon: Flag to define the template rendering to use. `advanced_mode=False` will use handlebars.

jfcalvo: So `advanced_mode=True` will not use any template at all?

argilla-server/src/argilla_server/api/schemas/v1/fields.py
```diff
@@ -88,11 +88,30 @@ class ChatFieldSettingsUpdate(BaseModel):
...
+
+
+class CustomFieldSettingsCreate(BaseModel):
+    type: Literal[FieldType.custom]
+    template: str
```
jfcalvo, line 99: We don't need to add any length validation to the template? I mean an empty string is a valid template here.

argilla-server/src/argilla_server/api/schemas/v1/fields.py
frascuchon: Yes. It will use the template defined in `template`. But it must be defined as a `handlebars` template. 

If `advanced_mode=True` users must define a low-level HTML template. You can take a look at the docs to understand this: https://docs.argilla.io/docs_custom-field/how_to_guides/custom_fields/#advanced-mode.

And your feedback will be welcome here: https://github.com/argilla-io/argilla/pull/5536

PR #5426, opened 2024-08-26
[FEATURE] Default settings templates

burtenshaw:
# Description

This PR brings back templates to the v2 SDK with a minimal selection of document_classification, response_ranking, and response_rating. When combined with https://github.com/argilla-io/argilla/pull/5418 , this PR will make it easy to load most datasets form the hub in the SDK.

Settings templates can be accessed like this: 

```python
settings = Settings.for_classification(labels=mock_labels)
settings = Settings.for_ranking()
settings = Settings.for_rating()
```

## Options

Settings templates can be used as a base that is added to.

```python
settings = Settings.for_ranking()
settings.fields.add(rg.TextField(name="prompt"))
settings.mapping["x"] = "prompt"
```

We could also 

**Type of change**

**How Has This Been Tested**

- Tests are updated


argilla/tests/unit/test_settings/test_settings_templates.py
```diff
@@ -0,0 +1,35 @@
+import pytest
+from unittest import mock
+from argilla.settings._templates import DefaultSettingsMixin
+
+
+class TestDefaultSettingsMixin:
+    def test_for_document_classification(self):
```
davidberenstein1957, line 21: maybe the phrasing `document_classification` is a bit weird? Also, I would also consider adding image cat and token cat as basic defaults.

davidberenstein1957 (commented):
I loved the concept myself but I am not 100% sure where this would fit in the docs, and if these config options still are an issue with the renewed singular docs.

argilla/tests/unit/test_settings/test_settings_templates.py
burtenshaw: Yeah, I can see that the naming is odd. We should iterate on those

My aim is to make a few generic enough methods that could be updated into variations, and avoid an exhaustive list of ML tasks. I.e. `document_classification` could have have alternative fields for image or chat. 

In a subsequent PR, I would like to add these:
- `mapping` in settings. This would make the default templates flexible to column naming. 
- `Features` >> `Settings` : we select and update a 'Settings` class based on the values and names of `datasets.Features`

With these features, the settings templates could be limited because we can get information from the dataset.

burtenshaw:
> I loved the concept myself but I am not 100% sure where this would fit in the docs, and if these config options still are an issue with the renewed singular docs.

Personally, I think there are two key factors here:
- keep the list really short/ generic so that it's more about capturing the 80% of Argilla usage rather than all ML tasks.
- focus on templates as an attribute of `Dataset.form_hub` rather than something that end users rely on to start new projects. I.e. they are tool for importing datasets, rather than creating them.

argilla/tests/unit/test_settings/test_settings_templates.py
davidberenstein1957: @burtenshaw , perhaps `content_classification` or simply `classification` would be better? Similarly, should we rename to `ranking` / `rating`?

davidberenstein1957: I agree that we would need a mapping to truly avoid the difficulties because otherwise, any mis-named columns/field/question should still be re-defined, right?

davidberenstein1957: @burtenshaw, I agree the features to setting would be very interesting.

davidberenstein1957:
> * keep the list really short/ generic so that it's more about capturing the 80% of Argilla usage rather than all ML tasks.

- I agree.
- Shouldn't we just keep it as an internal method for us to streamline the process in that case?

argilla/tests/unit/test_settings/test_settings_templates.py
burtenshaw: > @burtenshaw , perhaps `content_classification` or simply `classification` 

I think these overlap with `token_classification`

> would be better? Similarly, should we rename to `ranking` / `rating`?

This makes sense. What about `for_ranking`,`for_rating`, `for_labels`, `for_spans` ?

burtenshaw:
> > * keep the list really short/ generic so that it's more about capturing the 80% of Argilla usage rather than all ML tasks.
> 
> * I agree.
> * Shouldn't we just keep it as an internal method for us to streamline the process in that case?

For me, exposing the working makes it easier to use and understand what's going on. I think it would cause us problems if the `from_hub` method and `Features` integration seems to `automagic`. This way users can see that some batteries are included, and if they won't to reuse it, they just define their own `Settings` object.

argilla/src/argilla/settings/_templates.py
```diff
@@ -0,0 +1,159 @@
...
+        questions (List[QuestionType]): List of questions.
+        use_chat (bool): If True, the field will be replaced with a chat field.
+        """
+        from argilla import Settings, TextField, LabelQuestion
+
+        settings = Settings(
+            guidelines="Select a label for the document.",
+            fields=[TextField(name="text")],
+            questions=[LabelQuestion(name="label", labels=labels)],
+            mapping={"input": "text", "output": "label", "document": "text"},
```
frascuchon, line 57: What about parameterize names on template creation? Diving in the code is a bit hard to follow all this mappings at the settings level. 

````python
Settings.for_document_classification(
   fields=["text", "title",...],
....
)
```

frascuchon: Also, maybe this can be more flexible for other dataset structure.

argilla/src/argilla/settings/_templates.py
```diff
@@ -0,0 +1,159 @@
...
+        """
+        from argilla import Settings, TextField, RankingQuestion
+
+        settings = Settings(
+            guidelines="Rank the responses.",
+            fields=[TextField(name="instruction"), TextField(name="response1"), TextField(name="response2")],
+            questions=[RankingQuestion(name="ranking", values=["response1", "response2"])],
+            mapping={
+                "input": "instruction",
+                "prompt": "instruction",
```
frascuchon, line 90: Same comment here. Maybe users want to see both fields separately. 

```python
Settings.for_response_ranking(
   instruction=["input", "prompt"], 
   response1="chosen", 
   response2="rejected"
)
```

argilla/src/argilla/settings/_templates.py
```diff
@@ -0,0 +1,159 @@
...
+            fields=fields,
+            questions=questions,
+            use_chat=use_chat,
+        )
+
+    @classmethod
+    def for_response_rating(
+        cls: "Settings",
+        guidelines: Optional[str] = None,
+        fields: Optional[List["Field"]] = None,
```
frascuchon: What's the purpose of these `fields` argument? I can imagine some kind of template extension, but maybe easier if we pass just the names? Users can always change them after the settings creation

argilla/src/argilla/settings/_templates.py
```diff
@@ -0,0 +1,130 @@
...
+# limitations under the License.
+
+from typing import List, Optional, Literal, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from argilla.settings._resource import Settings
+    from argilla.settings._field import TextField
+
+
+def _get_field_type(field_type: str) -> "TextField":
```
frascuchon:
```suggestion
def _get_field_by_type(field_type: str) -> "TextField":
```

argilla/src/argilla/settings/_templates.py
```diff
@@ -0,0 +1,130 @@
...
+            questions=[LabelQuestion(name="label", labels=labels)],
+            mapping={"input": "text", "output": "label", "document": "text"},
+        )
+
+        return settings
+
+    @classmethod
+    def for_ranking(
+        cls: "Settings",
+        field_type: Optional[Literal["text", "image", "chat"]] = "text",
```
frascuchon: "chat" value shouldn't be here yet, right?

argilla/src/argilla/settings/_templates.py
```diff
@@ -0,0 +1,130 @@
...
+                "rejected": "response2",
+            },
+        )
+
+        return settings
+
+    @classmethod
+    def for_rating(
+        cls: "Settings",
+        field_type: Optional[Literal["text", "image", "chat"]] = "text",
```
frascuchon: ditto
```suggestion
        field_type: Optional[Literal["text", "image"]] = "text",
```

frascuchon (approved):

PR #5432, opened 2024-08-27
feat: add background processing jobs

### LINKED ISSUE #5431: [FEATURE] Add background processing
This feature includes the addition of a background processing framework to Argilla server that allow us to execute jobs in background.

Some details about the feature:
* We should add as first background job the updating of dataset records status when a dataset distribution strategy is updated.
* It should work as expected on HF spaces.
```[tasklist]
### Tasks
- [ ] https://github.com/argilla-io/argilla/issues/5443
```

jfcalvo:
# Description

This PR add the following changes:

- [x] Add `rq` to help us execute background jobs.
- [x] Add a background job to update all records for a dataset when the dataset distribution strategy is updated.
- [x] Change HuggingFace Dockerfile to install Redis and run `rq` workers inside honcho Procfile.
- [x] Add documentation about new `ARGILLA_REDIS_URL` environment variable.
- [x] Add ping to Redis so Argilla server is not started if Redis is not ready.
- [x] Change Argilla docker compose file to include a container with Redis and rq workers.
- [x] Update Argilla server `README.md` file adding Redis as dependency to install.
- [x] Add documentation about Redis being a new Argilla server dependency.
- [x] Add `BACKGROUND_NUM_WORKERS` environment variable to specify the number of workers in the HF Space container.
- [ ] ~~Modify `Dockerfile` template on HF to include the environment variable https://github.com/argilla-io/argilla/issues/5443:~~
```
# (since: v2.2.0) Uncomment the next line to specify the number of background job workers to run (default: 2).
# ENV BACKGROUND_NUM_WORKERS=2
```
- [ ] Remove some `TODO` sections before merging.
- [ ] Review K8s documentation (maybe delete it?).
- [ ] If we want to persist Redis data on HF Spaces we can change our `Procfile` Redis process to the following:
```
redis: /usr/bin/redis-server --dbfilename argilla-redis.rdb --dir ${ARGILLA_HOME_PATH}
```
- [ ] <del>Allow tests job workers synchronously (with pytest)</del> It's not working due to asyncio stuff (running an asynchronous loop inside another one, more info here: https://github.com/rq/rq/issues/1986).

Closes #5431

# Benchmarks

The following timings were obtained updating the distribution strategy of a dataset with 100 and 10.000 records, using a basic and an upgraded CPU on HF Spaces, with and without persistent storage and measuring how much time the background job takes to complete:

CPU basic: 2 vCPU, 16GB RAM
CPU upgrade: 8 vCPU, 32GB RAM

* CPU basic (with persistent storage):
  * 100 records dataset: ~8 seconds.
  * 10.000 records dataset: ~9 minutes.
* CPU upgrade (with persistent storage):
  * 100 records dataset: ~5 seconds.
  * 10.000 records dataset: ~6 minutes.
* CPU basic (no persistent storage):
  * 10.000 records dataset: ~101 seconds.
* CPU upgrade (no persistent storage):
  * 10.000 records dataset: ~62 seconds.

**Type of change**

**How Has This Been Tested**

- [x] Testing it on HF Spaces.

**Checklist**


examples/deployments/docker/docker-compose.yaml
```diff
@@ -1,27 +1,62 @@
...
+    image: argilladev/argilla-server:pr-5432
+    environment:
+      <<: *common-variables
+    networks:
+      - argilla
+    depends_on:
+      - postgres
+      - elasticsearch
+      - redis
+    command: sh -c 'rq worker --with-scheduler --url $${ARGILLA_REDIS_URL}'
```
jfcalvo: Change this to use rq worker-pool using and environment variable.

argilla-server/docker/argilla-hf-spaces/Procfile
```diff
@@ -1,2 +1,4 @@
 elastic: /usr/share/elasticsearch/bin/elasticsearch
+redis: /usr/bin/redis-server
+worker: sleep 30; rq worker-pool --num-workers ${BACKGROUND_NUM_WORKERS}
```
frascuchon, line 3: Reminder: Compute the queues to listen to.

argilla-server/src/argilla_server/jobs/dataset_jobs.py
```diff
@@ -0,0 +1,53 @@
...
+@job(default_queue, timeout=JOB_TIMEOUT_DISABLED, retry=Retry(max=3))
+async def update_dataset_records_status_job(dataset_id: UUID):
+    """This Job updates the status of all the records in the dataset when the distribution strategy changes."""
+
+    record_ids = []
+
+    async with AsyncSessionLocal() as db:
+        stream = await db.stream(
+            select(Record.id)
+            .where(Record.dataset_id == dataset_id)
```
frascuchon, line 43: Adding existing responses may reduce the number of records to evaluate.

```suggestion
            .where(Record.dataset_id == dataset_id)
            .join(Response)
```

argilla-server/src/argilla_server/jobs/queues.py
```diff
@@ -0,0 +1,24 @@
...
+import redis
+
+from rq import Queue
+
+from argilla_server.settings import settings
+
+
+redis_connection = redis.from_url(settings.redis_url)
+
+default_queue = Queue(connection=redis_connection)
```
frascuchon: Since it's a constant...Also, it would be nice to add the queue name explicitly.

```suggestion
DEFAULT_QUEUE = Queue(name="default", connection=redis_connection)
```

argilla-server/src/argilla_server/jobs/dataset_jobs.py
jfcalvo: This could work because we're not right now deleting responses, but if in the future we delete responses (something that can be useful for example when you don't want to have a response in any of the current "states") for some reason this optimization will not update the status of records correctly.

argilla-server/docker/argilla-hf-spaces/Procfile
jfcalvo: I think it's more flexible invoking it like this. For the Webhooks as example I want to have two different workers, one consuming the `high` queue and another one consuming the `low` queue. For the `high` queue maybe I want to have two workers so it should be something like that (we can also stop using the default queue):

```
worker_high: sleep 30; rq worker-pool --num-workers 2 high
worker_low: sleep 30; rq worker-pool --num-workers 1 low
```

This is specific for HF and doing programmatically the start of the workers is really strict about how the workers are configured.

In any case I did this in the case that we want to add a cli command with defaults:

```python
# filename: src/argilla_server/cli/worker.py

import typer

from typing import List

from argilla_server.jobs.queues import DEFAULT_QUEUE


def worker(
    queues: List[str] = typer.Option([DEFAULT_QUEUE.name], help="Name of queues to listen"),
    num_workers: int = typer.Option(2, help="Number of workers to start"),
) -> None:
    from rq.worker_pool import WorkerPool
    from argilla_server.jobs.queues import REDIS_CONNECTION

    worker_pool = WorkerPool(
        connection=REDIS_CONNECTION,
        queues=queues,
        num_workers=num_workers,
    )

    worker_pool.start()
```

What I think is not really useful.

For development we already have this `pdm` command where we can add additional queues if necessary:

```
worker = { cmd = "rq worker --with-scheduler" }
```

argilla-server/src/argilla_server/jobs/dataset_jobs.py
frascuchon: In those cases, the status computation doesn't need to be applied since the status is computed based on the record responses. Also, the number of min_submitted must be >= 1. 

We can review this optimization in the future if we support other acceptance criteria.

argilla-server/docker/argilla-hf-spaces/Procfile
jfcalvo: Done.

argilla-server/src/argilla_server/jobs/dataset_jobs.py
jfcalvo: Done.

frascuchon (approved):

argilla/docs/reference/argilla-server/configuration.md
```diff
@@ -80,6 +80,12 @@ The following environment variables are useful only when PostgreSQL is used:
 
 - `ARGILLA_ELASTICSEARCH_CA_PATH`: Path to CA cert for ES host. For example: `/full/path/to/root-ca.pem` (Optional)
 
+### Redis
+
+Redis is used by Argilla to store information about jobs to be processed on background. The following environment variables are useful to config how Argilla connects to Redis:
+
+- `ARGILLA_REDIS_URL`: A URL string that contains the necessary information to connect to a Redis instance (Default: `redis://localhost:6379/0`).
```
frascuchon, line 87:
```suggestion
- `ARGILLA_REDIS_URL`: A URL string that contains the necessary information to connect to a Redis instance (Default: `redis://localhost:6379`).
```

frascuchon (approved):

argilla/docs/reference/argilla-server/configuration.md
jfcalvo: The `0` it's the Redis database. We can remove it and it will use by default the db `0` but I though it was better to be explicit.

PR #5279, opened 2024-07-19
[FEATURE-BRANCH] ImageField: add support to new fields of type image

jfcalvo:
# Description

This PR is the first one related with the addition of a new field of type `image` including the following changes:
* Changes in the backend to support a new `image` field type supporting URLs and Data URLs.
* Changes in the frontend rendering this new `image` field for records.
* Changes in the SDK to define datasets with `image` fields and create records specifying images.
* Changes in the documentation with examples on how to use the SDK with this new type of field.

This new type will have the following characteristics:
* A new filed type `image` is supported (without additional settings).
* Records created with fields of type `image` are validated to have correct Web URLs or data URLs like:
  *  `https://argilla.io/image.jpeg`:
     * `HTTPS` schema is present
     * A host and a path is present.
     * The URL is not longer than `2038` characters.
  * `http://argilla.io/image.jpeg`:
     * `HTTP` schema is present
     * A host and a path is present.
     * The URL is not longer than `2038` characters.
  * `data:image/webp;base64,UklGRhgCAABXRUJQVlA4WA`:
     * `data` schema is present
     * A path is present.
     * The Data URL is not longer than `5` million characters.
     * A valid MIME type is present.

Closes #5276

**Type of change**

- New feature (non-breaking change which adds functionality)

**How Has This Been Tested**

- [x] Adding new tests to our suite checking this new field.

**Checklist**

- I added relevant documentation
- I followed the style guidelines of this project
- I did a self-review of my code
- I made corresponding changes to the documentation
- I confirm My changes generate no new warnings
- I have added tests that prove my fix is effective or that my feature works
- I have added relevant notes to the CHANGELOG.md file (See https://keepachangelog.com/)

PR #5433, opened 2024-08-27
[FEATURE] cast datasets pil objects to base64 in SDK

burtenshaw:
# Description

This PR introduces a utility into the SDK so that it automatically casts PIL objects to base64 based on the datasets `Features`

**Type of change**

- Documentation update

**How Has This Been Tested**

**Checklist**


argilla/src/argilla/media.py
```diff
@@ -0,0 +1,50 @@
...
+        from PIL import Image
+    except ImportError as e:
+        raise ImportError("The PIL library is required to convert PIL images for upload.") from e
+
+    if not isinstance(image_object, Image.Image):
+        raise ValueError("The image_object must be a PIL Image object.")
+
+    try:
+        buffered = io.BytesIO()
+        image_object.save(buffered, format="PNG")
```
jfcalvo: What's the motivation behind transforming it to PNG? I believe we should try not to process the image and use it as it is.

As a possible alternative we should try to get the mime-type (and use it inside the data-url) and if it's not supported by the backend raise an error.

argilla/src/argilla/records/_io/_datasets.py
```diff
@@ -58,11 +60,65 @@ def _record_dicts_from_datasets(dataset: HFDataset) -> List[Dict[str, Union[str,
...
             dataset: IterableDataset = dataset.to_iterable_dataset()
         except AttributeError:
             pass
+        record_dicts = []
         for example in dataset:
             record_dicts.append(example)
         return record_dicts
+
+    @staticmethod
+    def _get_media_features(dataset: "HFDataset") -> List[str]:
```
frascuchon: maybe rename it to `get_image_features` until we support other media types?

argilla/src/argilla/media.py
jfcalvo: You can get the list of supported mime types here: https://github.com/argilla-io/argilla/pull/5279/files#diff-05ee0287277b4f0bb73f7d03262800ff0cfb195ab90e9f11e29105eb88b51dcfR33

argilla/src/argilla/records/_io/_datasets.py
```diff
@@ -58,11 +60,65 @@ def _record_dicts_from_datasets(dataset: HFDataset) -> List[Dict[str, Union[str,
...
+            hf_dataset (HFDataset): The Hugging Face dataset to check.
+
+        Returns:
+            bool: True if the Hugging Face dataset contains image features, False otherwise.
+        """
+        media_features = []
+        for name, feature in dataset.features.items():
+            if isinstance(feature, Image):
+                media_features.append(name)
+        return media_features
```
frascuchon: list-comprehension approach. Feel free to apply this change.
```suggestion
        media_features = [
           name
           for name, feature in dataset.features.items()
           isinstance(feature, Image)
        ]
        return media_features
```

argilla/src/argilla/records/_io/_datasets.py
```diff
@@ -58,11 +60,65 @@ def _record_dicts_from_datasets(dataset: HFDataset) -> List[Dict[str, Union[str,
...
+                batched=True,
+                input_columns=[column],
+                remove_columns=[column],
+                fn_kwargs={"column": column},
+            )
+            hf_dataset = hf_dataset.rename_column(
+                original_column_name=f"column_{unique_identifier}", new_column_name=column
+            )
+
+        return hf_dataset
```
frascuchon: Since datasets supports modifying column values directly, what about this approach? 
```suggestion
        def batch_fn(batch, columns: List[str]):
            return {
                column: [pil_to_data_uri(sample[column]) for sample in batch]
                for column in columns 
            }

        return hf_dataset.map(
            function=batch_fn,
            batched=True,
            fn_kwargs={"columns": columns},
        )
```

argilla/src/argilla/records/_io/_datasets.py
```diff
@@ -58,11 +60,65 @@ def _record_dicts_from_datasets(dataset: HFDataset) -> List[Dict[str, Union[str,
...
+        """Check if the Hugging Face dataset contains image features.
+
+        Parameters:
+            hf_dataset (HFDataset): The Hugging Face dataset to check.
+
+        Returns:
+            bool: True if the Hugging Face dataset contains image features, False otherwise.
+        """
+        media_features = []
+        for name, feature in dataset.features.items():
```
davidberenstein1957: @burtenshaw I am not entirely sure how the `Image` feature for datasets works exactly. After [loading an image dataset](https://huggingface.co/docs/datasets/en/image_load#local-files), I've found different scenarios [working on dataset viber](https://github.com/davidberenstein1957/dataset-viber/blob/main/src/dataset_viber/_gradio/_mixins/_import_export.py#L277-L297).

burtenshaw: In the situation where the `Features` contain an `Image` but the values are not a pil the conversiont [function](https://github.com/argilla-io/argilla/blob/8917f5fc8af7b339426d137bd9ff8957c345cb2a/argilla/src/argilla/media.py#L23) with raise an error, so the user will need to cast the dataset themselves.

davidberenstein1957: @burtenshaw, don't we lose a lot of potential conversion cases only allowing for `PIL`? 

Also, I don't know what happens with the `Image` being in `features` when you load a dataset with `set decode=False` for example. According to the documentation, it still seems to be an `Image` feature but then containing this nested dict or just direct URL, which we could use to ease HF datasets upload IMO, given the code I shared above.

Additionally, I am not sure if `Image` might not also just contain a direct URL.

argilla/src/argilla/records/_io/_datasets.py
burtenshaw: @frascuchon I tested this out and an error is raise when we iterate over. I think this is because the dataset's `Features` need to match the samples.
 
I worked on this for a while trying a few approaches like `Dataset.cast_column` and found that this was the best I could get. 

Happy to try out more though.

argilla/src/argilla/records/_io/_datasets.py
burtenshaw: The datasets package converts the values with an `Image` `Feature` to PIL image see docs here: https://huggingface.co/docs/datasets/en/about_dataset_features#image-feature

`decode=False` is a parameter of the `Image` class which is passed to `cast_column` of `cast`, It tells `datasets` not to decode the value as a PIL object. I think we can say that we don't support images when the user casts the column with `decode=False`

>it still seems to be an Image feature but then containing this nested dict or just direct URL, which we could use to ease HF datasets upload IMO, given the code I shared above.

Is the direct URL in the object non-ephemeral?

argilla/src/argilla/records/_io/_datasets.py
frascuchon: I thought we could apply the image conversion for all the image columns at once. That was my motivation, but the code is fine as is.

frascuchon (approved):

PR #5436, opened 2024-08-28
feat: move image casting into record fields

burtenshaw:
# Description

This PR experiments with cast images within the `Record` class to abstract away data uri. This supports this flow:

```python
image_paths = [f"{my_image_dir}/{fn}" for fn in os.listdir(my_image_dir)]

records = [
    rg.Record(
        fields={
            "image": image_path,
        },
    )
    for image_path in image_paths
]


dataset.records.log(records)
```

**Type of change**

- Documentation update

**How Has This Been Tested**

**Checklist**


argilla/src/argilla/_helpers/_media.py
```diff
@@ -0,0 +1,120 @@
...
+            img_str = base64.b64encode(image_file.read()).decode()
+            mimetype = f"image/{file_path.suffix[1:]}"
+            data_uri = f"data:{mimetype};base64,{img_str}"
+    else:
+        raise FileNotFoundError(f"File not found at {file_path}")
+
+    return data_uri
+
+
+def cast_image(image: "Image") -> str:
```
frascuchon:
```suggestion
def cast_image(image: Union["Image", str, Path]) -> str:
```

argilla/src/argilla/records/_io/_datasets.py
```diff
@@ -19,7 +19,7 @@
 from datasets import IterableDataset, Image
 
 from argilla.records._io._generic import GenericIO
-from argilla.media import pil_to_data_uri
+from argilla._helpers._media import pil_to_data_uri
```
frascuchon, line 22: Since image processing is now at the record level? Do we need to process images for HF dataset before log them? 
Should not raise an error since the incoming image value is already a data:URI string?

argilla/src/argilla/records/_resource.py
```diff
@@ -272,11 +273,21 @@ class RecordFields(dict):
...
     """
 
-    def __init__(self, fields: Optional[Dict[str, FieldValue]] = None) -> None:
+    def __init__(self, record: Record, fields: Optional[Dict[str, FieldValue]] = None) -> None:
         super().__init__(fields or {})
+        self.record = record
 
     def to_dict(self) -> dict:
-        return dict(self.items())
+        return {key: cast_image(value) if self._is_image(key) else value for key, value in self.items()}
```
frascuchon, line 281: Just a comment here. The export to datasets will generate dataset URI values for images and the image feature will be a simple string feature. We can tackle this later, but it would be nice to generate the HF dataset with a proper image feature.

argilla/tests/unit/test_record_fields.py
```diff
@@ -0,0 +1,33 @@
...
+
+        fields = record.fields
+        assert fields["name"] == "John Doe"
+        assert record.metadata["age"] == 30
+
+    def test_create_record_image_path(self):
+        record = Record(fields={"image": "path/to/image.jpg"})
+
+        fields = record.fields
+        assert fields["image"] == "path/to/image.jpg"
```
frascuchon, line 60: This wouldn't be the expected result if the field `image` is an `ImageField`, right?

frascuchon: Okay, I understand now. This is valid when creating records before assigning them to an existing dataset, which makes sense.

argilla/src/argilla/records/_io/_datasets.py
burtenshaw: I think we should keep this integration because it takes advantage of datasets `map` function and batching to cast images to data uri. The `cast_image` method will then skip it if it gets a data uri.

argilla/src/argilla/records/_resource.py
burtenshaw: Yeah. This is a good point. I will test out what we get with this implementation and come back to it.

argilla/tests/unit/test_record_fields.py
burtenshaw: Correct. Do you think we need tests that include predefined settings?

frascuchon: Yes. I think it makes sense to include at least tests checking  a happy path and an error scenarios

frascuchon (approved):

PR #5437, opened 2024-08-28
[FEAT] argilla server: add basic endpoints telemetry support

frascuchon:
# Description

This PR adds a middleware component to track the API endpoint's usage.

**Type of change**

**How Has This Been Tested**

**Checklist**

### DIFF
--- a/argilla-server/src/argilla_server/_app.py
+++ b/argilla-server/src/argilla_server/_app.py
@@ -40,6 +40,8 @@ from argilla_server.models import User, Workspace
 from argilla_server.search_engine import get_search_engine
 from argilla_server.settings import settings
 from argilla_server.static_rewrite import RewriteStaticFiles
+from argilla_server.telemetry import get_telemetry_client
+from argilla_server.utils._fastapi import resolve_endpoint_path_for_request
 
 _LOGGER = logging.getLogger("argilla")
 
@@ -67,8 +69,9 @@ def create_server_app() -> FastAPI:
     )
 
     configure_logging()
-    configure_middleware(app)
+    configure_basic_middlewares(app)
     configure_api_router(app)
+    configure_telemetry(app)
     configure_app_statics(app)
     configure_api_docs(app)
 
@@ -92,7 +95,7 @@ def configure_api_docs(app: FastAPI):
         return RedirectResponse(url=f"{settings.base_url}api/v1/docs")
 
 
-def configure_middleware(app: FastAPI):
+def configure_basic_middlewares(app: FastAPI):
     """Configures fastapi middleware"""
 
     @app.middleware("http")
@@ -119,6 +122,32 @@ def configure_middleware(app: FastAPI):
 def configure_api_router(app: FastAPI):
     """Configures and set the api router to app"""
     app.mount("/api/v1", api_v1)
+
+
+def configure_telemetry(app: FastAPI):
+    """
+    Configures telemetry middleware for the app if telemetry is enabled
+
+    """
+    if not settings.enable_telemetry:
+        return
+
+    @app.middleware("http")
+    async def track_api_endpoint(request: Request, call_next):
+        response = await call_next(request)
+        try:
+            endpoint_path = resolve_endpoint_path_for_request(request)
+
+            if endpoint_path:
+                await get_telemetry_client().track_endpoint(
+                    endpoint_path,
+                    request,
+                    response,
+                )
+        except Exception as e:
+            _LOGGER.warning(f"Error tracking endpoint: {e}")
+        finally:
+            return response
 
 
 def configure_app_statics(app: FastAPI):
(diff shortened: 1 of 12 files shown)


argilla-server/tests/unit/test_endpoints_telemetry.py
```diff
@@ -0,0 +1,54 @@
...
+        settings.enable_telemetry = True
+
+        client = TestClient(create_server_app())
+
+        response = client.post("/api/v1/datasets")
+        assert response.status_code == 401
+
+        test_telemetry.track_endpoint.assert_called_once_with("/api/v1/datasets", ANY, ANY)
+
+    def test_not_track_endpoint_call_with_telemetry_disabled(self, test_telemetry: TelemetryClient):
```
frascuchon:
```suggestion
    def test_not_track_endpoint_call_with_disabled_telemetry(self, test_telemetry: TelemetryClient):
```

argilla-server/src/argilla_server/_app.py
```diff
@@ -121,6 +124,32 @@ def configure_api_router(app: FastAPI):
     app.mount("/api/v1", api_v1)
 
 
+def configure_telemetry(app: FastAPI):
+    """
+    Configures telemetry middleware for the app if telemetry is enabled
+
+    """
```
jfcalvo:
```suggestion
    """
```

argilla-server/src/argilla_server/_app.py
```diff
@@ -121,6 +124,32 @@ def configure_api_router(app: FastAPI):
...
+    if not settings.enable_telemetry:
+        return
```
jfcalvo, line 132: What do you think about moving this to the `configure_telemetry(app)` call so we avoid checking it on every request?

Something like:

```python
if settings.enable_telemetry:
    configure_telemetry(app)
```

argilla-server/src/argilla_server/_app.py
```diff
@@ -121,6 +124,32 @@ def configure_api_router(app: FastAPI):
...
+
+    @app.middleware("http")
+    async def track_api_endpoint(request: Request, call_next):
+        response = await call_next(request)
+        try:
+            endpoint_path = resolve_endpoint_path_for_request(request)
```
jfcalvo: Why not moving this logic to be inside the telemetry client? With that you can pass only `request` and `response` like:

```python
await get_telemetry_client().track_endpoint(request, response)
```

argilla-server/src/argilla_server/_app.py
```diff
@@ -121,6 +124,32 @@ def configure_api_router(app: FastAPI):
...
+def configure_telemetry(app: FastAPI):
+    """
+    Configures telemetry middleware for the app if telemetry is enabled
+
+    """
+    if not settings.enable_telemetry:
+        return
+
+    @app.middleware("http")
+    async def track_api_endpoint(request: Request, call_next):
```
jfcalvo: I believe a better naming could be something like:

```suggestion
    async def track_api_request(request: Request, call_next):
```

argilla-server/src/argilla_server/_app.py
```diff
@@ -121,6 +124,32 @@ def configure_api_router(app: FastAPI):
...
+        response = await call_next(request)
+        try:
+            endpoint_path = resolve_endpoint_path_for_request(request)
+
+            if endpoint_path:
+                await get_telemetry_client().track_endpoint(
```
jfcalvo:
```suggestion
                await get_telemetry_client().track_api_request(
```

jfcalvo (changes requested):

argilla-server/src/argilla_server/_app.py
frascuchon: The `configure_telemetry` function will be called once when FastAPI is created, not for every request. If telemetry is disabled, the telemetry middleware won't be loaded.

argilla-server/src/argilla_server/_app.py
frascuchon: Yes, I thought about this when developing it. It makes sense and the middleware keeps simpler. I will push the changes

argilla-server/tests/unit/api/handlers/v1/test_datasets.py
```diff
@@ -209,6 +209,7 @@ async def test_list_current_user_datasets_by_workspace_id(
         response_body = response.json()
         assert [dataset["name"] for dataset in response_body["items"]] == ["dataset-a"]
 
+    @pytest.mark.skip
```
jfcalvo, line 212: Why we need to skip these tests?

jfcalvo (approved):

argilla-server/tests/unit/api/handlers/v1/test_datasets.py
frascuchon: I needed to build the docker image. Those tests are failing in the base branch. I review all of them in a separate PR.

PR #5441, opened 2024-08-29
[DRAFT] feat: simplify telemetry

davidberenstein1957:
# Description

Closes NA


argilla-server/src/argilla_server/_app.py
```diff
@@ -239,4 +240,16 @@ async def ping_search_engine():
     await ping_search_engine()
 
 
+async def configure_telemetry(app):
+    from argilla_server.telemetry import get_telemetry_client
+
+    @app.on_event("startup")
```
frascuchon: We're using app lifespan, which is recommended for this. 

Both flows are not compatible: https://github.com/argilla-io/argilla/blob/c4085eefe442026f53417685f2cb07fcfc4baadc/argilla-server/src/argilla_server/_app.py#L48

https://fastapi.tiangolo.com/advanced/events/#lifespan

argilla-server/src/argilla_server/telemetry.py
```diff
@@ -144,138 +102,139 @@ def _process_dataset_setting_settings(
...
+
+    async def track_server_launch(self) -> None:
+        """
+        This method is used to track the launch of the server.
+
+        Returns:
+            None
+        """
+        topic = "startup"
+        await self.track_data(topic=topic, crud_action="create")
```
frascuchon: What does mean the crud action here? The startup and shutdown are action itself. I think it would be easy to understand just to pass the server startup/shutdown.

davidberenstein1957: I removed it.

argilla-server/src/argilla_server/_app.py
davidberenstein1957: done

argilla-server/src/argilla_server/telemetry.py
```diff
@@ -142,140 +82,113 @@ def _process_dataset_setting_settings(
...
-            user_agent.update(self._process_dataset_model(dataset=dataset))
-            user_agent.update(self._process_dataset_settings(dataset=dataset))
-        await self.track_data(topic=topic, user_agent=user_agent, count=count)
-
-        attributes: list[str] = ["fields", "questions", "vectors_settings", "metadata_properties"]
+        self, crud_action: str, dataset: Union[Dataset, None] = None, count: Union[int, None] = None
+    ) -> None:
         if dataset:
-            for attr in attributes:
+            for attr in _RELEVANT_ATTRIBUTES + ["distribution"]:
```
frascuchon: The `track_crud_dataset` method is used for dataset creation. In that case, no fields, questions, or other setting relationships will be provided, so this code is not necessary. Also, the distribution is not a relationship but a column, so the `is_relationship_loaded` is not needed.  You could track the distribution by accessing it directly. I think this simplifies the code readability

argilla-server/src/argilla_server/telemetry.py
```diff
@@ -142,140 +82,113 @@ def _process_dataset_setting_settings(
...
+                            crud_action=crud_action, setting_name=attr, setting=obtained_attr
                         )
 
+            for attr in _RELEVANT_ATTRIBUTES:
+                if dataset.is_relationship_loaded(attr):
+                    obtained_attr_list = getattr(dataset, attr)
+                    await self.track_resource_size(
+                        crud_action=crud_action, setting_name=f"dataset/{attr}", count=len(obtained_attr_list)
+                    )
+
```
frascuchon: This info won't be available on dataset creation, so I would say to remove it.

argilla-server/src/argilla_server/telemetry.py
```diff
@@ -142,140 +82,113 @@ def _process_dataset_setting_settings(
...
-            "type": error.__class__.__name__,
-        }
-
-        await self.track_data(topic=topic, user_agent=user_agent)
+    ) -> None:
+        """
+        This method is used to track the creation, update, and deletion of dataset settings. These
+        settings include fields, questions, vectors settings, and metadata properties.
+
+        Args:
```
frascuchon: I think we use Parameters format instead.

argilla-server/src/argilla_server/api/handlers/v1/datasets/datasets.py
```diff
@@ -71,77 +71,59 @@ async def _filter_metadata_properties_by_policy(
...
+    telemetry_client: TelemetryClient = Depends(get_telemetry_client),
 ):
     await authorize(current_user, DatasetPolicy.list(workspace_id))
 
     if not workspace_id:
         if current_user.is_owner:
             dataset_list = await datasets.list_datasets(db)
+            await telemetry_client.track_resource_size(
+                crud_action="read", setting_name="dataset", count=len(dataset_list)
+            )
```
frascuchon, line 85: I think adding the dataset count when listing the datasets for the owner is a bit of a convoluted way of computing the metric. It might make more sense to send this information somewhere else in the system. I think I should rethink these metrics and include them in a second iteration.

frascuchon: Also, not sure about the need for `crud_action="read"` here since is no related to a resource crud action (we compute a size)

argilla-server/src/argilla_server/api/handlers/v1/users.py
```diff
@@ -69,9 +60,7 @@ async def list_users(
 
     users = await accounts.list_users(db)
 
-    await telemetry_client.track_crud_user(action="list", user=None, is_oauth=False, count=len(users))
-    for user in users:
-        await telemetry_client.track_crud_user(action="read", user=user, is_oauth=False)
+    await telemetry_client.track_resource_size(crud_action="read", setting_name="user", count=len(users))
```
frascuchon, line 63: Similarly, computing this metric here can be not easy to understand later. Better if we reduce the number of metrics for the first iteration an think about these stats for a second iteration

argilla-server/src/argilla_server/api/handlers/v1/users.py
```diff
@@ -118,13 +97,15 @@ async def list_user_workspaces(
...
     current_user: User = Security(auth.get_current_user),
+    telemetry_client: TelemetryClient = Depends(get_telemetry_client),
 ):
     await authorize(current_user, UserPolicy.list_workspaces)
 
     user = await User.get_or_raise(db, user_id)
 
     if user.is_owner:
         workspaces = await accounts.list_workspaces(db)
+        await telemetry_client.track_resource_size(crud_action="read", setting_name="workspace", count=len(workspaces))
```
frascuchon, line 108: same comments here.

argilla-server/src/argilla_server/telemetry.py
```diff
@@ -142,140 +82,113 @@ def _process_dataset_setting_settings(
...
-            user_data["dimensions"] = setting.dimensions
-            user_data["type"] = "default"
-        else:
-            raise NotImplementedError("Expected a setting to be processed.")
-
-        return user_data
+            return {"type": "vector"}
+        elif isinstance(setting, dict):
+            return {"type": f"distribution_{setting['strategy']}_{setting['min_submitted']}"}
+        raise NotImplementedError("Expected a setting to be processed.")
```
frascuchon: Having this here is a bit scary. We don't want to break endpoint functionally because the telemetry module cannot extract info. Also, in general, we must be sure to not propagate errors when telemetry info.

argilla-server/src/argilla_server/telemetry.py
```diff
@@ -142,140 +82,113 @@ def _process_dataset_setting_settings(
...
-        if include_system_info:
-            user_agent.update(self._system_info)
+        user_agent = user_agent or {}
+        user_agent.update(self._system_info)
         if count is not None:
-            user_agent["count"] = count
-
+            user_agent["count"] = count or 1
+        if crud_action:
+            user_agent["request.endpoint.method"] = crud_action
```
frascuchon: what's the purpose of this attribute? We only are tracking  "create" actions, and my previous PR is already tracking endpoint method info.

argilla-server/src/argilla_server/telemetry.py
```diff
@@ -142,140 +82,113 @@ def _process_dataset_setting_settings(
...
-
-        return user_data
+            return {"type": "vector"}
+        elif isinstance(setting, dict):
+            return {"type": f"distribution_{setting['strategy']}_{setting['min_submitted']}"}
+        raise NotImplementedError("Expected a setting to be processed.")
 
-    async def track_data(self, topic: str, user_agent: dict, include_system_info: bool = True, count: int = 1):
-        library_name = "argilla"
+    async def track_data(self, topic: str, count: int = None, crud_action: str = None, user_agent: dict = None) -> None:
```
frascuchon: Since the track_data is the base method to send telemetry info, I would say to simplify the signature and move the count and the crud_action as part of the data.  Also, the name of the user agent reveals implementation details of how the data is sent and is not aligned with the method name.
```
async def track_data (self, topic, data:dict)
...

File: argilla-server/src/argilla_server/telemetry.py [hunk 12]
```diff
@@ -142,140 +82,113 @@ def _process_dataset_setting_settings(
...
                         )
 
+            for attr in _RELEVANT_ATTRIBUTES:
+                if dataset.is_relationship_loaded(attr):
+                    obtained_attr_list = getattr(dataset, attr)
+                    await self.track_resource_size(
+                        crud_action=crud_action, setting_name=f"dataset/{attr}", count=len(obtained_attr_list)
+                    )
+
     async def track_crud_dataset_setting(
```
frascuchon: Even if the telemetry library won't send data if disabled, we should prevent making unnecessary calls and data processing when telemetry is disabled. Also, if there is an error in our code, users have no way to disable the whole telemetry tracking.

PR #5445, opened 2024-09-02
[FEAT] `argilla server`: add user and server id on telemetry metrics

frascuchon:
# Description

This PR restores the server_id for telemetry purposes and also add the user.id and user.role when tracking API requests. 

**Type of change**

- Documentation update

**How Has This Been Tested**

**Checklist**


argilla-server/src/argilla_server/telemetry.py
```diff
@@ -16,26 +16,31 @@
...
     server_deployment_type,
 )
+from argilla_server.security.authentication.provider import get_request_user
 
 _LOGGER = logging.getLogger(__name__)
 
 
 @dataclasses.dataclass
 class TelemetryClient:
+    _server_id: int = uuid.getnode()
```
frascuchon: The `uuid.getnode` returns an integer value. We can use it "as is".

argilla-server/src/argilla_server/telemetry.py
```diff
@@ -47,16 +52,11 @@ def __post_init__(self):
...
-        library_name = "argilla/server"
-        topic = f"{library_name}/{topic}"
-
-        user_agent = {**data}
-        if include_system_info:
-            user_agent.update(self._system_info)
-        if count is not None:
-            user_agent["count"] = count
+    async def track_data(self, topic: str, data: dict):
+        library_name = "argilla-server"
```
frascuchon, line 57: This aligns the library name (argilla-server) to the package published in pypi.org

argilla-server/src/argilla_server/telemetry.py
jfcalvo: We should check that `uuid.getnode()` is returning the same value on HF Spaces even after a factory rebuild.

jfcalvo (approved):

jfcalvo (approved):

frascuchon:
> We should check that uuid.getnode() is returning the same value on HF Spaces even after a factory rebuild.

Changes introduced storing the server ID allow reuse of the server ID on restart when persistent storage is enabled. We can add also info about persistent storage on server startup.

Issue #4944, opened 2024-05-27
[REFACTOR] Simplify naming of serialize, to_dict, and to_json methods


burtenshaw:
Includes argilla-io/argilla-python#196

It is not clear for users how the could or should serialize resources for persistance.

- The method names should be made consistent
- API methods should be mad private
- Documentation should be added for persistence workflows.

frascuchon:
Will help this simplify the v2 adoption? cc @burtenshaw

nataliaElv:
I'd say that we should avoid changing naming after the release, so I'm going to add this to v2, but if in the end this feels like too much feel free to remove the milestone.

burtenshaw:
@nataliaElv These are mainly internal methods, so I think it can be tackled in 2.1

PR #5454, opened 2024-09-04
[ENHANCEMENT] Make exporting records with image consistent

burtenshaw:
# Description

This PR implements a casting step from data uri to pil in the export method for hf_datasets.

**Type of change**

- Documentation update

**How Has This Been Tested**

**Checklist**

### DIFF
--- a/argilla/src/argilla/records/_io/_datasets.py
+++ b/argilla/src/argilla/records/_io/_datasets.py
@@ -23,6 +23,7 @@ from argilla._helpers._media import pil_to_data_uri, uncast_image
 
 if TYPE_CHECKING:
     from argilla.records import Record
+    from argilla.datasets import Dataset
 
 
 class HFDatasetsIO:
@@ -39,7 +40,7 @@ class HFDatasetsIO:
         return isinstance(dataset, HFDataset)
 
     @staticmethod
-    def to_datasets(records: List["Record"], schema: Dict) -> HFDataset:
+    def to_datasets(records: List["Record"], dataset: "Dataset") -> HFDataset:
         """
         Export the records to a Hugging Face dataset.
 
@@ -47,11 +48,11 @@ class HFDatasetsIO:
             The dataset containing the records.
         """
         record_dicts = GenericIO.to_dict(records, flatten=True)
-        dataset = HFDataset.from_dict(record_dicts)
-        image_fields = HFDatasetsIO._get_image_fields(schema=schema)
+        hf_dataset = HFDataset.from_dict(record_dicts)
+        image_fields = HFDatasetsIO._get_image_fields(schema=dataset.schema)
         if image_fields:
-            dataset = HFDatasetsIO._cast_uris_as_images(hf_dataset=dataset, columns=image_fields)
-        return dataset
+            hf_dataset = HFDatasetsIO._cast_uris_as_images(hf_dataset=hf_dataset, columns=image_fields)
+        return hf_dataset
 
     @staticmethod
     def _record_dicts_from_datasets(dataset: HFDataset) -> List[Dict[str, Union[str, float, int, list]]]:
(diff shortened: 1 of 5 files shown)


davidberenstein1957 (commented):
@burtenshaw, looks fine to me. I wanted to ask how we will cover the flow when looping through record in a non-HF fashion because that feels a bit quirky now and exposes the datauri to the end user, right?

```python
for record in ds.records:
    record.fiels["image"] => data uri
```

argilla/src/argilla/records/_io/_datasets.py
```diff
@@ -39,7 +39,7 @@ def _is_hf_dataset(dataset: Any) -> bool:
         return isinstance(dataset, HFDataset)
 
     @staticmethod
-    def to_datasets(records: List["Record"]) -> HFDataset:
+    def to_datasets(records: List["Record"], schema: Dict) -> HFDataset:
```
frascuchon: Maybe more flexible and easier to extend if we pass the whole rg.Dataset, even if internally we only use the schema.

```suggestion
    def to_datasets(records: List["Record"], dataset: "Datasert") -> HFDataset:
```

argilla/src/argilla/records/_io/_datasets.py
```diff
@@ -48,6 +48,9 @@ def to_datasets(records: List["Record"]) -> HFDataset:
         """
         record_dicts = GenericIO.to_dict(records, flatten=True)
         dataset = HFDataset.from_dict(record_dicts)
+        image_fields = HFDatasetsIO._get_image_fields(schema=schema)
```
frascuchon:
```suggestion
        image_fields = HFDatasetsIO._get_image_fields(schema=dataset.schema)
```

argilla/src/argilla/records/_io/_datasets.py
```diff
@@ -48,6 +48,9 @@ def to_datasets(records: List["Record"]) -> HFDataset:
...
+        if image_fields:
+            dataset = HFDatasetsIO._cast_uris_as_images(hf_dataset=dataset, columns=image_fields)
```
frascuchon: What about moving this to an helper method?
```python
def _prepare_image_features(dataset, schema:dict) -> "HFDataset":
    image_fields = HFDatasetsIO._get_image_fields(schema=schema)
    if image_fields:
        dataset = HFDatasetsIO._cast_uris_as_images(hf_dataset=dataset, columns=image_fields)
    return dataset
```
```suggestion
        dataset = _prepare_image_features(dataset, dataset.schema)
```

frascuchon (approved):
I think as future code refactor, we should keep the same repr for resources and apply a different serializer depending on the serialization  is for.

````python

serialize_for_disk(dataset)

serialize_for_api(records)

...
```

burtenshaw:
> @burtenshaw, looks fine to me. I wanted to ask how we will cover the flow when looping through record in a non-HF fashion because that feels a bit quirky now and exposes the datauri to the end user, right?
> 
> ```python
> for record in ds.records:
>     record.fiels["image"] => data uri
> ```

In this case the value of the field is casted to PIL

burtenshaw:
> I think as future code refactor, we should keep the same repr for resources and apply a different serializer depending on the serialization is for.
> 
> ```python
> serialize_for_disk(dataset)
> 
> serialize_for_api(records)
> 
> ...
> ```
> ```

Yeah, this make sense. We should do this along with a general refactor of serialisation methods. Like here: https://github.com/argilla-io/argilla/issues/4944

argilla/src/argilla/records/_io/_datasets.py
burtenshaw: I tried this first, but I get `AttributeError: 'Dataset' object has no attribute 'schema'` because the property is not materialized.

argilla/src/argilla/records/_io/_datasets.py
burtenshaw: I feel like that would be too many short methods. I think that the class need to be refactored when we bring custom and chat field in, so that there's a methods for casting features.

frascuchon (approved):

PR #5463, opened 2024-09-05
[FEATURE] add different error handling strategies to the log method 

burtenshaw:
# Description

Closes #<issue_number>

**Type of change**

- Documentation update

**How Has This Been Tested**

**Checklist**

### DIFF
--- a/argilla/src/argilla/records/_dataset_records.py
+++ b/argilla/src/argilla/records/_dataset_records.py
@@ -15,6 +15,7 @@ import warnings
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Dict, Iterable, List, Optional, Sequence, Union
 from uuid import UUID
+from enum import Enum
 
 from tqdm import tqdm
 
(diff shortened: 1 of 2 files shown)


argilla/src/argilla/records/_dataset_records.py
```diff
@@ -207,6 +214,7 @@ def log(
         mapping: Optional[Dict[str, Union[str, Sequence[str]]]] = None,
         user_id: Optional[UUID] = None,
         batch_size: int = DEFAULT_BATCH_SIZE,
+        on_error: RecordErrorHandling = RecordErrorHandling.RAISE,
```
frascuchon, line 217: I think this `on_error` parameter should also skip failed errors when the error comes from the server, but we should show the user information about failed records.

frascuchon: Otherwise, can be confusing to users when set on_error="ignore" and the backend raises errors when sending records

frascuchon (approved):
We should have a wider issue to tackle backend error flows.

PR #5468, opened 2024-09-09
feat: add Webhooks notify events using background jobs

jfcalvo:
# Description

* Add first real Webhooks notifications for responses using some background jobs for them.
* Add a new RQ `high` queue where we will enqueue the Webhooks notification jobs.
* Add webhook events:
  * `response.created`
  * `response.updated`
  * `response.deleted`
  * `response.upserted`
  * `dataset.created`
  * `dataset.updated`
  * `dataset.deleted`
  * `dataset.published`

**Type of change**

**How Has This Been Tested**

- [x] Manually tested locally and in HF spaces.

**Checklist**


argilla-server/src/argilla_server/jobs/webhook_jobs.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+@job(HIGH_QUEUE, retry=Retry(max=3, interval=[10, 60, 180]))
+async def notify_event_job(webhook_id: UUID, event: str, timestamp: datetime, data: dict) -> httpx.Response:
+    async with AsyncSessionLocal() as db:
+        webhook = await Webhook.get_or_raise(db, webhook_id)
+
+    response = notify_event(webhook, event, timestamp, data)
+    if not response.is_success:
+        raise Exception(f"Failed to notify event `{event}` to webhook with id `{webhook.id}` and URL `{webhook.url}`")
```
frascuchon:
```suggestion
    response = notify_event(webhook, event, timestamp, data)
    response.raise_for_status()
```

jfcalvo: I didn't use `raise_for_status` because I wanted to show a custom message on the exception and be more explicit as well.

argilla-server/src/argilla_server/jobs/queues.py
```diff
@@ -21,4 +21,5 @@
 
 REDIS_CONNECTION = redis.from_url(settings.redis_url)
 
-DEFAULT_QUEUE = Queue("default", connection=REDIS_CONNECTION)
```
frascuchon, line 24: I was thinking about this. This means that the default queue must be empty when upgrading from 2.2.0 to 2.3.0. Otherwise, the pending jobs will never be processed, right?

frascuchon (approved):

argilla-server/src/argilla_server/jobs/queues.py
jfcalvo: Yes, I think it will not be a problem right?

jfcalvo: Specially because we are not persisting Redis databases and when we restart we are losing those jobs.

jfcalvo:
It's working fine the `response.upserted` webhook event on Spaces using bulk annotation.

I have also added a `version` attribute to the body of the webhook events in the case that you want to take a look @frascuchon: https://github.com/argilla-io/argilla/pull/5468/commits/2e7a5ff90cae8aad9a44594657936acaebf0755d

PR #5472, opened 2024-09-10
[FEATURE] Define `rg.Settings` based on `datasets.Features`

### LINKED ISSUE #5470: [FEATURE] expand from_hub method to interpret any dataset based on Features

burtenshaw:
# Description

This PR define a module that interacts with the datasets server API to create a Settings object based on  datasets features.

It can be used like this:

## Via `rg.Settings`

```python
from argilla import Settings, Dataset

settings = Settings.from_hub(repo_id= "yahma/alpaca-cleaned")
dataset = Dataset(name= "alpaca-cleaned", settings=settings)
dataset.create()
hf_dataset = load_dataset(path=repo_id, split="train[:10]")
dataset.records.log(hf_dataset)
```

## Via `rg.Dataset`

```python
from argilla import Dataset

dataset = Dataset.from_hub(repo_id="yahma/alpaca-cleaned")
```

from_hub has 3 uses:
- used with settings
- used on an argilla config
- used with nothing

## Things to note

- It's implemented as a helper module with functions, in case we want to migrate it to the server.
- It's pretty cautious in it's error handling from datasets-server, we might want to add a fall back, and more warnings.
- I think we should upgrade integration testing when we use this in the `from_hub` method.

## Next Steps
The next step is to integrate it with `from_hub`, so this works:

```python
dataset = rg.Dataset.from_hub(repo_id="yahma/alpaca-cleaned")
```

**Type of change**

**How Has This Been Tested**

**Checklist**


davidberenstein1957:
@burtenshaw, could you update the branch :)?

davidberenstein1957 (commented):
@burtenshaw, can you update the PR and only include the relevant commits?

burtenshaw:
> @burtenshaw, can you update the PR and only include the relevant commits?

Sorry @davidberenstein1957 I'll resolve that now.

burtenshaw:
@davidberenstein1957 It's now pointed at `develop`

argilla/src/argilla/_helpers/_datasets_server.py
frascuchon, line 1: Why not to move this under the _io/_datasets/ module?

argilla/src/argilla/_helpers/_datasets_server.py
```diff
@@ -0,0 +1,175 @@
...
+        config (str, optional): The configuration of the dataset. Defaults to None.
+
+    Returns:
+        Dict[str, Dict[str, Any]]: The features of the dataset.
+    """
+
+    url = f"{DATASETS_SERVER_BASE_URL}/info?dataset={repo_id}"
+
+    try:
+        response = requests.get(url, headers=DATASETS_SERVER_HEADERS, timeout=DATASETS_SERVER_TIMEOUT)
```
frascuchon: Since we're already using httpx for http request, maybe we could align this and use the same package here.

argilla/src/argilla/_helpers/_datasets_server.py
```diff
@@ -0,0 +1,175 @@
...
+
+    try:
+        response_json = response.json()
+    except Exception as e:
+        raise DatasetsServerException() from e
+
+    try:
+        dataset_info = response_json["dataset_info"]
+    except KeyError as e:
+        raise DatasetsServerException() from e
```
frascuchon: I think we can wrap all the caught exceptions in one `except` block:
```suggestion
    try:
        response = requests.get(url, headers=DATASETS_SERVER_HEADERS, timeout=DATASETS_SERVER_TIMEOUT)
        response.raise_for_status()    
        response_json = response.json()
        dataset_info = response_json["dataset_info"]    
    except (requests.exceptions.RequestException, requests.exceptions.HTTPError, KeyError, Exception) as e:
        raise DatasetsServerException() from e
```

argilla/src/argilla/_helpers/_datasets_server.py
```diff
@@ -0,0 +1,175 @@
...
+
+    available_configs = list(dataset_info.keys())
+
+    if len(available_configs) > 1 and config is None:
+        raise ValueError("Multiple configurations found. Please specify a config.")
```
frascuchon: I think is easier for users just warn and read the first available config. What do you think?

argilla/src/argilla/_helpers/_datasets_server.py
```diff
@@ -0,0 +1,175 @@
...
+    elif len(available_configs) == 1:
+        config = available_configs[0]
+    elif config not in available_configs:
+        raise ValueError(f"Configuration '{config}' not found.")
```
frascuchon: In which scenario this is possible? `config` is a required parameter for getting rows.

argilla/src/argilla/_helpers/_datasets_server.py
burtenshaw: I mainly put it there because of the format, but `datasets._io` works. But  `settings._io` could make the most sense.

argilla/src/argilla/_helpers/_datasets_server.py
```diff
@@ -0,0 +1,175 @@
...
+
+        if feature_type == FeatureType.CHAT:
+            # TODO: Implement chat support to create `rg.ChatField`
+            # fields.append(rg.ChatField(name=f"{name}_field"))
+            # mapping[name] = f"{name}_field"
+            pass
+
+        elif feature_type == FeatureType.TEXT:
+            fields.append(TextField(name=f"{name}_field"))
+            questions.append(TextQuestion(name=f"{name}_question"))
```
frascuchon: Do we want to add a required text question for each text feature found?

argilla/src/argilla/_helpers/_datasets_server.py
```diff
@@ -0,0 +1,175 @@
...
+            mapping[name] = (f"{name}_field", f"{name}_question")
+
+        elif feature_type == FeatureType.IMAGE:
+            fields.append(ImageField(name=f"{name}_field"))
+            mapping[name] = f"{name}_field"
+
+        elif feature_type == FeatureType.LABEL:
+            fields.append(TextField(name=f"{name}_field"))
+            questions.append(
+                LabelQuestion(
```
frascuchon: What about MultiLabelQuestion ? Do we have a mechanism to identify Multi vs Label questions? If not maybe is more flexible to create MultiLabelQuestion

argilla/src/argilla/_helpers/_datasets_server.py
```diff
@@ -0,0 +1,175 @@
...
+            fields.append(TextField(name=f"{name}_field"))
+            questions.append(TextQuestion(name=f"{name}_question"))
+            mapping[name] = (f"{name}_field", f"{name}_question")
+
+        elif feature_type == FeatureType.IMAGE:
+            fields.append(ImageField(name=f"{name}_field"))
+            mapping[name] = f"{name}_field"
+
+        elif feature_type == FeatureType.LABEL:
+            fields.append(TextField(name=f"{name}_field"))
```
frascuchon: Do we want to create a text field including label values? I was thinking of passing them as suggestionsvalues

argilla/src/argilla/_helpers/_datasets_server.py
```diff
@@ -0,0 +1,175 @@
...
+        feature_type = _map_feature_type(feature)
+
+        if feature_type == FeatureType.CHAT:
+            # TODO: Implement chat support to create `rg.ChatField`
+            # fields.append(rg.ChatField(name=f"{name}_field"))
+            # mapping[name] = f"{name}_field"
+            pass
+
+        elif feature_type == FeatureType.TEXT:
+            fields.append(TextField(name=f"{name}_field"))
```
frascuchon: Why we cannot use the name directly?
```python
fields.append(TextField(name=name))
```

In general, do we need to apply some name conversion to adapt feature names to argilla setting names? (lower case,...)

argilla/src/argilla/settings/_resource.py
```diff
@@ -257,6 +258,13 @@ def from_json(cls, path: Union[Path, str]) -> "Settings":
             settings_dict = json.load(file)
             return cls._from_dict(settings_dict)
 
+    @classmethod
+    def from_hub(cls, repo_id: str, version: str = "latest") -> "Settings":
+        """Load the settings from the Hub"""
+
+        settings = build_settings_from_repo_id(repo_id)
```
frascuchon: We're not using the `config` parameter here. I would say to remove and fetch the first configuration when multiple are found.

argilla/src/argilla/_helpers/_datasets_server.py
```diff
@@ -0,0 +1,175 @@
...
+
+    TEXT = "text"
+    CHAT = "chat"
+    IMAGE = "image"
+    LABEL = "label"
+    INT = "int"
+    FLOAT = "float"
+
+
+def _get_dataset_features(repo_id: str, config: Optional[str] = None) -> Dict[str, Dict[str, Any]]:
```
frascuchon: Regarding this comment https://github.com/argilla-io/argilla/pull/5472/files#r1751978426
```suggestion
def _get_dataset_features(repo_id: str) -> Dict[str, Dict[str, Any]]:
```

argilla/src/argilla/_helpers/_datasets_server.py
```diff
@@ -0,0 +1,175 @@
...
+        "content" in sub_features
+        and "role" in sub_features
+        and sub_features["content"]["_type"] == "Value"
+        and sub_features["role"]["_type"] == "Value"
+        and sub_features["content"].get("dtype") == "string"
+        and sub_features["role"].get("dtype") == "string"
+    )
+
+
+def _define_settings_from_features(features: Union[List[Dict], Dict[str, Any]], config=None) -> "Settings":
```
frascuchon: Regarding this comment https://github.com/argilla-io/argilla/pull/5472/files#r1751978426
```suggestion
def _define_settings_from_features(features: Union[List[Dict], Dict[str, Any]]) -> "Settings":
```

frascuchon: I think the `config` argument is not used on this function. We can remove it.

argilla/src/argilla/_helpers/_datasets_server.py
burtenshaw: If the user defines a non-existent config.

argilla/src/argilla/_helpers/_datasets_server.py
burtenshaw: Ideally not, but because datasets require at least one required question we need to do something here. Any suggestions?

argilla/src/argilla/_helpers/_datasets_server.py
burtenshaw: Because of the mapping, they would appear as suggestion like this. I can remove this if you think it's overkill.

argilla/src/argilla/_helpers/_datasets_server.py
burtenshaw: We discussed using values as both field and question so that it's as flexible as possible. Based on that, I need to differentiate the names for attributes and then map them.

argilla/src/argilla/_helpers/_datasets_server.py
burtenshaw: Moved it to `settings._io`

argilla/src/argilla/_helpers/_datasets_server.py
burtenshaw: Some datasets require a specific config to be used, but they're rare. Let's go with this for now.

argilla/src/argilla/_helpers/_datasets_server.py
burtenshaw: There is a separate representation for multilabel features in datasets, so we know that this is a `LabelQuestion` compatible field.

argilla/src/argilla/_helpers/_datasets_server.py
frascuchon: What about adding a default question if questions are not mapped at the end of the process? We need to create a dataset without the required questions, but we need to evaluate the implications of doing this.

argilla/src/argilla/settings/_io/_hub.py
```diff
@@ -0,0 +1,167 @@
...
+            )
+            mapping[name] = (f"{name}_field", f"{name}_question")
+
+        elif feature_type in [FeatureType.INT, FeatureType.FLOAT]:
+            # TODO: Implement metadata support to create `rg.FloatMetadata` or `rg.IntMetadata`
+            pass
+        else:
+            warnings.warn(f"Feature '{name}' has an unsupported type. Skipping. Feature type: {feature_type}")
+
+    settings = Settings(fields=fields, questions=questions, metadata=metadata, mapping=mapping)
```
frascuchon, line 236: Maybe we can add this if no questions were mapped (with a warning). So we can remove the TextQuestion for text features. What do you think?

```suggestion
    if not questions:
        questions.append(rg.TextQuestion(name="comment",...)
    settings = Settings(fields=fields, questions=questions, metadata=metadata, mapping=mapping)
```

argilla/src/argilla/settings/_io/_hub.py
```diff
@@ -0,0 +1,167 @@
...
+
+    return settings
+
+
+def build_settings_from_repo_id(repo_id: str):
```
frascuchon: >  dealing with  datasets columns as fields, metadata, and questions. When should we do this? When is it overkill?

Maybe as a first flexible step, allow passing an optional features mapping dict could help to customize and evolve this? For example for this [dataset](https://huggingface.co/datasets/AI-MO/NuminaMath-CoT?row=0)

```python

features_mapping = { "source": "metadata",  "solution": "question" }
build_settings_from_repo_id("AI-MO/NuminaMath-CoT", features_mapping)
```

argilla/src/argilla/_helpers/_datasets_server.py
burtenshaw: That would get us moving. We could go for something like a TextQuestion named 'comment'?

argilla/src/argilla/settings/_io/_hub.py
burtenshaw: Makes sense

argilla/src/argilla/settings/_io/_hub.py
```diff
@@ -0,0 +1,203 @@
...
+    )
+
+    fields = []
+    questions = []
+    metadata = []
+    mapping = defaultdict(list)
+    feature_mapping = feature_mapping or {}
+    mapped_questions = [key for key, value in feature_mapping.items() if value == "question"]
+    mapped_fields = [key for key, value in feature_mapping.items() if value == "field"]
+    mapped_metadata = [key for key, value in feature_mapping.items() if value == "metadata"]
```
frascuchon: Do we need these extra lists? Maybe we can access the feature_mapping dictionary directly:

```python
   defined_mapping = feature_mapping.get(name)
   if defined_mapping == "question" ...
```
with some default values depending on the extracted feature type? (For example, field for text features, and question for label features?

argilla/src/argilla/settings/_io/_hub.py
```diff
@@ -0,0 +1,203 @@
...
+        feature_type = _map_feature_type(feature)
+
+        if feature_type == FeatureType.CHAT:
+            # TODO: Implement chat support to create `rg.ChatField`
+            # fields.append(rg.ChatField(name=f"{name}_field"))
+            # mapping[name] = f"{name}_field"
+            pass
+
+        elif feature_type == FeatureType.TEXT:
+            if name not in mapped_questions:
```
frascuchon: Why not a positive check?
```python
   defined_mapping = feature_mapping.get(name, "field")
   if defined_mapping == "question":
       questions.append(...)
   elif defined_mapping == "field":
       fields.append(...)
   elif defined_mapping == "metadata": 
       # allow extra metadata = True?
```

argilla/src/argilla/settings/_io/_hub.py
```diff
@@ -0,0 +1,203 @@
...
+                questions.append(
+                    LabelQuestion(
+                        name=name,
+                        labels=names,
+                    )
+                )
+                mapping[name].append(name)
+            if name not in mapped_questions:
+                metadata.append(TermsMetadataProperty(name=f"{name}_metadata", options=names))
+                mapping[name].append(f"{name}_metadata")
```
frascuchon: same as the comment before:
```python
if feature_mappings.get(name) == "question":
    questions.append(...)
elif feature_mapping(name ) == "metadata":
    metadata.append(...)
```

argilla/src/argilla/settings/_io/_hub.py
```diff
@@ -0,0 +1,203 @@
...
+    if not fields:
+        raise SettingsError("No fields found in the dataset features. Argilla datasets require at least one field.")
+    settings = Settings(fields=fields, questions=questions, metadata=metadata, mapping=mapping)
+
+    return settings
+
+
+def build_settings_from_repo_id(repo_id: str, feature_mapping: Optional[Dict[str, str]] = None) -> "Settings":
+    dataset_info = _get_dataset_features(repo_id)
+    return _define_settings_from_features(dataset_info, feature_mapping)
```
frascuchon:
```suggestion
    dataset_features = _get_dataset_features(repo_id)
    return _define_settings_from_features(dataset_features, feature_mapping)
```

argilla/src/argilla/settings/_io/_hub.py
```diff
@@ -0,0 +1,203 @@
...
+            metadata.append(IntegerMetadataProperty(name=name))
+        elif feature_type == FeatureType.FLOAT:
+            metadata.append(FloatMetadataProperty(name=name))
+        else:
+            warnings.warn(f"Feature '{name}' has an unsupported type. Skipping. Feature type: {feature_type}")
+
+    mapping = {
+        key: value[0] if isinstance(value, list) and len(value) == 1 else tuple(value)
+        for key, value in feature_mapping.items()
+    }
```
frascuchon: I'm a bit confused with these lines. Can you share an example of the `feature_mapping` content?

argilla/src/argilla/settings/_resource.py
```diff
@@ -257,6 +258,15 @@ def from_json(cls, path: Union[Path, str]) -> "Settings":
             settings_dict = json.load(file)
             return cls._from_dict(settings_dict)
 
+    @classmethod
+    def from_hub(
+        cls, repo_id: str, feature_mapping: Optional[Dict[str, Literal["question", "field", "metadata"]]] = None
```
frascuchon, line 264: Based on this feature_mapping definition, I think the code above will fail, since it expects a dict[str, str]

argilla/src/argilla/settings/_io/_hub.py
burtenshaw: You're right that the logic is a bit weird, but your suggestion changes the logic. The implemented logic does this:

```
defined_mapping == "field" == "field"
defined_mapping =="question" == "question"
defined_mapping == None == "question" and "field"

defined_mapping =="question" == "question"
defined_mapping =="metadata" == "metadata"
defined_mapping == None == "question" and "metadata"
```

We can change logic if you thing that's wrong as well.

argilla/tests/unit/test_settings/test_settings_from_features.py
```diff
@@ -0,0 +1,120 @@
...
+    assert settings.questions[1].name == "label_column"
+
+
+def test_mapped_question():
+    features = {
+        "text_column": {"_type": "Value", "dtype": "string"},
+        "image_column": {"_type": "Image"},
+        "label_column": {"_type": "ClassLabel", "names": ["A", "B"]},
+    }
+    settings = _define_settings_from_features(features, feature_mapping={"text_column": "question"})
```
frascuchon, line 77: Does the resultant settings mapping match the expected one?

argilla/src/argilla/settings/_io/_hub.py
burtenshaw: This is a mistake, it should be on `mapping`:

```python
    mapping = {
        key: value[0] if isinstance(value, list) and len(value) == 1 else tuple(value)
        for key, value in mapping.items()
    }
```

argilla/tests/unit/test_settings/test_settings_from_features.py
```diff
@@ -0,0 +1,120 @@
...
+
+    assert len(settings.fields) == 1
+    assert isinstance(settings.fields[0], rg.ImageField)
+    assert settings.fields[0].name == "image_column"
+    assert len(settings.questions) == 2
+    assert isinstance(settings.questions[0], rg.TextQuestion)
+    assert settings.questions[0].name == "text_column_question"
+    assert isinstance(settings.questions[1], rg.LabelQuestion)
+    assert settings.questions[1].name == "label_column"
```
frascuchon, line 86: Should be this the expected mapping?
```suggestion
    assert settings.questions[1].name == "label_column"
    assert settings.mapping == {  "text_column":  "text_columnt_question" }
```

argilla/src/argilla/settings/_io/_hub.py
burtenshaw: In simple terms, the implementation says: if a feature is mapped, assign to one attribute, otherwise assign to both.

argilla/src/argilla/settings/_io/_hub.py
frascuchon: Okay, so this code is to normalize mapping values since we don't support tuples with one single value, right?

argilla/src/argilla/datasets/_io/_hub.py
```diff
@@ -249,3 +249,27 @@ def _get_dataset_split(hf_dataset: "HFDataset", split: Optional[str] = None, **k
...
+
+        if hf_dataset:
+            sample_huggingface_record = {}
+            for key, value in hf_dataset[0].items():
+                try:
+                    json.dumps(value)
+                    sample_huggingface_record[key] = value
+                except TypeError:
+                    try:
+                        sample_huggingface_record[key] = pil_to_data_uri(value)
```
davidberenstein1957, line 296: can't we check if it is an image specifically?

argilla/src/argilla/records/_io/_datasets.py
```diff
@@ -26,6 +27,125 @@
...
+        columns (List[str]): The names of the columns containing the class label features.
+
+    Returns:
+        HFDataset: The Hugging Face dataset with class label features cast as strings.
+    """
+    for column in columns:
+        features = hf_dataset.features.copy()
+        features[column] = Value("string")
+        hf_dataset = hf_dataset.map(
+            lambda x: {column: hf_dataset.features[column].int2str(x[column])}, features=features
```
davidberenstein1957, line 72: is `int2str` always defined? I thought that sometimes datasets don't contain the proper mapping.

argilla/src/argilla/records/_io/_datasets.py
```diff
@@ -26,6 +27,125 @@
...
+        casted_hf_dataset = hf_dataset.map(
+            function=lambda batch: {column: [uncast_image(sample) for sample in batch]},
+            with_indices=False,
+            batched=True,
+            input_columns=[column],
+            remove_columns=[column],
+            features=features,
+        )
+        try:
+            casted_hf_dataset[0]
```
davidberenstein1957, line 106: is this correct? when a `None` return, 
```suggestion
        
        try:
            casted_hf_dataset = hf_dataset.map(
                function=lambda batch: {column: [uncast_image(sample) for sample in batch]},
                with_indices=False,
                batched=True,
                input_columns=[column],
                remove_columns=[column],
                features=features,
            )
            casted_hf_dataset[0]
```

argilla/src/argilla/records/_io/_datasets.py
```diff
@@ -64,20 +182,18 @@ def _record_dicts_from_datasets(dataset: HFDataset) -> List[Dict[str, Union[str,
...
+        hf_dataset = HFDatasetsIO.to_argilla(hf_dataset=hf_dataset)
         try:
-            dataset: IterableDataset = dataset.to_iterable_dataset()
+            hf_dataset: IterableDataset = hf_dataset.to_iterable_dataset()
         except AttributeError:
             pass
         record_dicts = []
-        for example in dataset:
+        for example in hf_dataset:
             record_dicts.append(example)
```
davidberenstein1957:
```suggestion
        record_dicts.append([example for example in hf_dataset])
```

argilla/src/argilla/records/_io/_datasets.py
```diff
@@ -98,64 +222,15 @@ def _get_image_features(dataset: "HFDataset") -> List[str]:
...
-            HFDataset: The Hugging Face dataset with image features cast as URLs.
-        """
+        casted_features = defaultdict(list)
 
-        unique_identifier = uuid4().hex
+        for name, feature in hf_dataset.features.items():
+            if isinstance(feature, Image):
+                casted_features[Image].append(name)
+            if isinstance(feature, ClassLabel):
+                casted_features[ClassLabel].append(name)
```
davidberenstein1957, line 229: so, there is not differentiation between single/multi-label?

argilla/src/argilla/settings/_io/_hub.py
```diff
@@ -0,0 +1,203 @@
...
+DATASETS_SERVER_TIMEOUT = 30
+
+
+class FeatureType(Enum):
+    TEXT = "text"
+    CHAT = "chat"
+    IMAGE = "image"
+    LABEL = "label"
+    INT = "int"
+    FLOAT = "float"
```
davidberenstein1957, line 38: same here, no differentiation between single and multi label?

argilla/src/argilla/settings/_io/_hub.py
```diff
@@ -0,0 +1,203 @@
...
+
+from argilla._exceptions._hub import DatasetsServerException
+from argilla._exceptions._settings import SettingsError
+
+if TYPE_CHECKING:
+    from argilla import Settings
+
+DATASETS_SERVER_BASE_URL = "https://datasets-server.huggingface.co"
+DATASETS_SERVER_HEADERS = {"Accept": "application/json"}
+DATASETS_SERVER_TIMEOUT = 30
```
davidberenstein1957, line 29: do we have back-off retry policy somewhere else by chance?

argilla/src/argilla/settings/_io/_hub.py
```diff
@@ -0,0 +1,203 @@
...
+        with httpx.Client(timeout=DATASETS_SERVER_TIMEOUT) as client:
+            response = client.get(url, headers=DATASETS_SERVER_HEADERS)
+            response.raise_for_status()
+            response_json = response.json()
+
+            dataset_info = response_json["dataset_info"]
+            available_configs = list(dataset_info.keys())
+
+            if len(available_configs) > 1:
+                warnings.warn("Multiple configurations found. Using the first one.")
```
davidberenstein1957, line 72: should we create an issue to outline this for a next iteration / help wanted issue on GH?

argilla/src/argilla/settings/_io/_hub.py
```diff
@@ -0,0 +1,203 @@
...
+                warnings.warn(f"Feature '{name}' has no labels. Skipping.")
+                continue
+            if name not in mapped_metadata:
+                questions.append(
+                    LabelQuestion(
+                        name=name,
+                        labels=names,
+                    )
+                )
+                mapping[name].append(name)
```
davidberenstein1957: So, no support for MultiLabel? Might we infer this from the task tag?

argilla/src/argilla/settings/_io/_hub.py
```diff
@@ -0,0 +1,203 @@
...
+            if name not in mapped_questions:
+                metadata.append(TermsMetadataProperty(name=f"{name}_metadata", options=names))
```
davidberenstein1957: Wouldn't it be easier to potentially allow for all options? Similar to the approach with Integers and Floats?

argilla/src/argilla/settings/_io/_hub.py
```diff
@@ -0,0 +1,203 @@
...
+        else:
+            warnings.warn(f"Feature '{name}' has an unsupported type. Skipping. Feature type: {feature_type}")
+
+    mapping = {
+        key: value[0] if isinstance(value, list) and len(value) == 1 else tuple(value)
+        for key, value in feature_mapping.items()
+    }
+
+    if not questions:
+        questions.append(TextQuestion(name="comment", required=True))
```
davidberenstein1957, line 228: `required=True` by default right?
```suggestion
        questions.append(TextQuestion(name="comment"))
```

argilla/src/argilla/settings/_io/_hub.py
burtenshaw: That's correct.

argilla/src/argilla/records/_io/_datasets.py
frascuchon: or:
```python
record_dicts.extend(hf_dataset)
```
?

argilla/src/argilla/records/_io/_datasets.py
burtenshaw: If it's a class label it will have an int2str attribute.

argilla/tests/unit/test_settings/test_settings_from_features.py
burtenshaw: Only necessary Argilla dataset attributes are mapped. I.e. ones that get renamed to deal with multiple attributes for the same column.

argilla/src/argilla/records/_io/_datasets.py
burtenshaw: `casted_hf_dataset[0]` will fail if the image value is not compatible with dataset serialization. Why should the map execution be inside the exception?

argilla/src/argilla/records/_io/_datasets.py
burtenshaw: Dataset samples are serialised on iteration. I can use a comprehension if that prettier.

argilla/src/argilla/records/_io/_datasets.py
burtenshaw: `ClassLabel` == `rg.LabelQuestion` multi label in datasets is a `Sequence[ClassLabel]` but it's out of scope.

argilla/src/argilla/settings/_io/_hub.py
burtenshaw: Multilabel is out of scope.

argilla/src/argilla/settings/_io/_hub.py
burtenshaw: Not that I know of, for the datasets server API.

argilla/src/argilla/settings/_io/_hub.py
burtenshaw: Yes that could be useful.

argilla/src/argilla/settings/_io/_hub.py
burtenshaw: Multilabel is out of scope.

argilla/src/argilla/settings/_io/_hub.py
burtenshaw: So to leave the `options` param empty? I always thought defining options improved search filters UX, but I might be wrong.

argilla/src/argilla/settings/_io/_hub.py
burtenshaw: I believe we need at least 1 required question for valid Argilla settings.

argilla/src/argilla/settings/_io/_hub.py
davidberenstein1957: I am not entirely sure @frascuchon

burtenshaw: The backend supplies unique values if the options are not defined. So I've gone with this.

argilla/src/argilla/settings/_io/_hub.py
```diff
@@ -0,0 +1,235 @@
...
+        repo_id (str): The repository ID of the dataset.
+
+    Returns:
+        Dict[str, Dict[str, Any]]: The features of the dataset.
+    """
+
+    url = f"{DATASETS_SERVER_BASE_URL}/info?dataset={repo_id}"
+
+    try:
+        with httpx.Client(timeout=DATASETS_SERVER_TIMEOUT) as client:
```
frascuchon: Just a comment here: 

We can preset some values here to avoid repeat on each call (but we only have one, :_))
```suggestion
        with httpx.Client(base_url= DATASETS_SERVER_BASE_URL, headers=DATASETS_SERVER_HEADERS, timeout=DATASETS_SERVER_TIMEOUT) as client:
```

frascuchon (approved):

PR #5474, opened 2024-09-11
[FEATURE] encode class labels as names not ints

burtenshaw:
# Description

This PR add to the settings from_hub PR by supporting ClassLabels coming from datasets package. 

It renders a label question with names based on the ClassLabel.

It uncasts the 'label.suggestion' to a corresponding class label.

**Type of change**

### DIFF
--- a/argilla/src/argilla/records/_io/_datasets.py
+++ b/argilla/src/argilla/records/_io/_datasets.py
@@ -12,11 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from collections import defaultdict
 from typing import TYPE_CHECKING, Any, Dict, List, Union
-from uuid import uuid4
 
 from datasets import Dataset as HFDataset
-from datasets import IterableDataset, Image
+from datasets import IterableDataset, Image, ClassLabel, Value
 
 from argilla.records._io._generic import GenericIO
 from argilla._helpers._media import pil_to_data_uri, uncast_image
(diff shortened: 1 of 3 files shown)


argilla/src/argilla/records/_io/_datasets.py
```diff
@@ -26,6 +26,116 @@
...
+
+    Parameters:
+        hf_dataset (HFDataset): The Hugging Face dataset to cast.
+        columns (List[str]): The names of the columns containing the class label features.
+
+    Returns:
+        HFDataset: The Hugging Face dataset with class label features cast as strings.
+    """
+    for column in columns:
+        column = f"{column}.suggestion"
```
frascuchon, line 118: what about responses? Should we add this mapping too?

burtenshaw: Yes, but we need to decide how that would work. We define responses as a value if there's only one, and arrays of value, statuses, users. So we need two things:

- a convenient way for users to access the responses in the hf_dataset.
- a symmetrical `Features` representation of responses with status, user_id, value that user can filter and we can reingest.

See here:

```python
{'id': '1da848fa-6134-4948-91fc-47f53d286c39',
  'status': 'pending',
  '_server_id': 'aaef66f9-3ca1-463f-993c-481498fca44e',
  'text_field': 'I Am ...',
  'label_field': 'neg',
  'label_question.responses': ['neg'],
  'label_question.responses.users': ['f9a47be8-0f7e-4c2b-9170-41f3a5b470d4'],
  'label_question.responses.status': ['draft']},
```

- In a first iteration, we could implement a mapping from 'question_name' >> `ClassLabel`
- Later, we could implement 'question_name.responses' >> `Sequence(Features({"user_id": ...}))`.

frascuchon (approved):

PR #5496, opened 2024-09-16
[UI] Improve autofill behaviour in login page

leiyre:
this PR removes autofocus in login page inputs and enables the button when the form is autofilled


damianpumar:
I would like to review it for the next release... @leiyre

leiyre:
fixed with #5522

PR #5509, opened 2024-09-18
Feature/chat-in-feature-based-settings

burtenshaw:
# Description

```python
import argilla as rg

dataset = rg.Dataset.from_hub(
    repo_id="argilla/magpie-ultra-v0.1",
    name="magpie-ultra",
    split="train[:100]",
)
```

**Type of change**

- Documentation update

**How Has This Been Tested**

**Checklist**

### DIFF
--- a/argilla-server/src/argilla_server/api/schemas/v1/chat.py
+++ b/argilla-server/src/argilla_server/api/schemas/v1/chat.py
@@ -15,7 +15,7 @@
 from argilla_server.pydantic_v1 import BaseModel, Field
 
 MIN_MESSAGE_LENGTH = 1
-MAX_MESSAGE_LENGTH = 5000
+MAX_MESSAGE_LENGTH = 20000
 
 MIN_ROLE_LENGTH = 1
 MAX_ROLE_LENGTH = 20
--- a/argilla/src/argilla/datasets/_io/_hub.py
+++ b/argilla/src/argilla/datasets/_io/_hub.py
@@ -116,6 +116,7 @@ class HubImportExportMixin(DiskImportExportMixin):
         client: Optional["Argilla"] = None,
         with_records: bool = True,
         settings: Optional["Settings"] = None,
+        split: Optional[str] = None,
         **kwargs: Any,
     ):
         """Loads a `Dataset` from the Hugging Face Hub.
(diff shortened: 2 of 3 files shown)


argilla/src/argilla/datasets/_io/_hub.py
```diff
@@ -116,6 +116,7 @@ def from_hub(
         client: Optional["Argilla"] = None,
         with_records: bool = True,
         settings: Optional["Settings"] = None,
+        split: Optional[str] = None,
```
frascuchon, line 119: maybe add the argument in docstring

argilla-server/src/argilla_server/api/schemas/v1/chat.py
```diff
@@ -15,7 +15,7 @@
 from argilla_server.pydantic_v1 import BaseModel, Field
 
 MIN_MESSAGE_LENGTH = 1
-MAX_MESSAGE_LENGTH = 5000
+MAX_MESSAGE_LENGTH = 20000
```
frascuchon, line 18: Did you find content length up to this value? High values for this could result in performance problems

argilla/src/argilla/datasets/_io/_hub.py
```diff
@@ -164,13 +165,18 @@ def from_hub(
                     client=client,
                     with_records=with_records,
                     settings=settings,
+                    split=split,
```
frascuchon, line 170: I think we should also add the incoming **kwargs, right?
```suggestion
                    split=split,
                    **kwargs
```

argilla/src/argilla/datasets/_io/_hub.py
frascuchon, line 119: Reading the code below, do we need to materialize the split argument? Could not be supported inside the **kwargs?

argilla-server/src/argilla_server/api/schemas/v1/chat.py
burtenshaw: That's right, but 5000 characters is very limited for a mandatory parameter. Some of these datasets have samples with 32k tokens. I think in the next release we will need to expose this parameter to the user.

argilla/src/argilla/datasets/_io/_hub.py
burtenshaw: Yeah. I realised that from your comment. But I think exposing this parameter clearly is helpful to the user

argilla/src/argilla/settings/_io/_hub.py
```diff
@@ -225,6 +223,12 @@ def _define_settings_from_features(
         questions.append(TextQuestion(name="comment", required=True))
     if not fields:
         raise SettingsError("No fields found in the dataset features. Argilla datasets require at least one field.")
+
+    # validate that atleast one field is required
+    if not any(field.required for field in fields):
```
frascuchon: based on the code above, could we just check the number of fields?
```suggestion
    if len(fields) > 1:
```

frascuchon (commented):
> If the feature is a list (not sure in which scenarios could be this), the `feature.get(...)` will raise with an error. 

That's correct. I added an elif check for non dicts. 

> Should we extract the first element in those cases as we already for chat features?

There are other sequential Feature types which we don't yet support. I would keep it simple like this for now.

argilla/src/argilla/settings/_io/_hub.py
burtenshaw: We could also just not check and always make the first field required because it's a necessity.

frascuchon: Should we apply the same logic for questions? I think at least one required question is also needed.

frascuchon (approved):

PR #5523, opened 2024-09-20
[BUGFIX] `argilla`: Fix some `from_hub` method errors

frascuchon:
# Description

Trying this code:

```python
import argilla as rg

client = rg.Argilla(...)

dataset = rg.Dataset.from_hub("google-research-datasets/circa")
```
I found several errors ([Here the source dataset](https://huggingface.co/datasets/google-research-datasets/circa/viewer/default/train?f[goldstandard2][value]=0)):
- The mapping for records does not match since column names contain uppercase characters
- Some values for the ClassValue column are unlabelled,  making the uncast process fail (-1)
- Importing cached datasets without config files fails when trying to read the dataset JSON file.

So, this PR tackles all the problems commented below.

**Type of change**

**How Has This Been Tested**

**Checklist**

### DIFF
--- a/argilla/src/argilla/datasets/_io/_disk.py
+++ b/argilla/src/argilla/datasets/_io/_disk.py
@@ -20,7 +20,7 @@ from abc import ABC
 from pathlib import Path
 from typing import TYPE_CHECKING, Optional, Tuple, Type, Union
 
-from argilla._exceptions import RecordsIngestionError, ArgillaError
+from argilla._exceptions import RecordsIngestionError, ArgillaError, ImportDatasetError
 from argilla._models import DatasetModel
 from argilla.client import Argilla
 from argilla.settings import Settings
@@ -82,11 +82,15 @@ class DiskImportExportMixin(ABC):
 
         client = client or Argilla._get_default()
 
-        dataset_path, settings_path, records_path = cls._define_child_paths(path=path)
-        logging.info(f"Loading dataset from {dataset_path}")
-        logging.info(f"Loading settings from {settings_path}")
-        logging.info(f"Loading records from {records_path}")
-        dataset_model = cls._load_dataset_model(path=dataset_path)
+        try:
+            dataset_path, settings_path, records_path = cls._define_child_paths(path=path)
+            logging.info(f"Loading dataset from {dataset_path}")
+            logging.info(f"Loading settings from {settings_path}")
+            logging.info(f"Loading records from {records_path}")
+
+            dataset_model = cls._load_dataset_model(path=dataset_path)
+        except (NotADirectoryError, FileNotFoundError) as e:
+            raise ImportDatasetError(f"Error loading dataset from disk. {e}") from e
 
         # Get the relevant workspace_id of the incoming dataset
         if isinstance(workspace, str):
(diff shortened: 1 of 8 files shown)


argilla/src/argilla/datasets/_io/_disk.py
```diff
@@ -82,11 +82,15 @@ def from_disk(
...
-        dataset_model = cls._load_dataset_model(path=dataset_path)
+        try:
+            dataset_path, settings_path, records_path = cls._define_child_paths(path=path)
+            logging.info(f"Loading dataset from {dataset_path}")
+            logging.info(f"Loading settings from {settings_path}")
+            logging.info(f"Loading records from {records_path}")
+
+            dataset_model = cls._load_dataset_model(path=dataset_path)
+        except (NotADirectoryError, FileNotFoundError) as e:
+            raise ImportDatasetError(f"Error loading dataset from disk. {e}") from e
```
frascuchon, line 93: Wrapping this in a custom exception can be easier for method callers.

argilla/src/argilla/datasets/_io/_disk.py
```diff
@@ -112,6 +116,9 @@ def from_disk(
             dataset = cls.from_model(model=dataset_model, client=client)
         else:
             # Create a new dataset and load the settings and records
+            if not os.path.exists(settings_path):
+                raise ImportDatasetError(f"Settings file not found at {settings_path}")
```
frascuchon, line 120: Here we check the existence of the settings file. We raise the custom ImportDatasetError if it does not exist.

argilla/src/argilla/datasets/_io/_hub.py
```diff
@@ -192,8 +193,11 @@ def from_hub(
     @staticmethod
     def _log_dataset_records(hf_dataset: "HFDataset", dataset: "Dataset"):
         """This method extracts the responses from a Hugging Face dataset and returns a list of `Record` objects"""
+        # THIS IS REQUIRED SINCE NAME IN ARGILLA ARE LOWERCASE. The Settings BaseModel models apply this logic
+        # Ideally, the restrictions in Argilla should be removed and the names should be case-insensitive
+        hf_dataset = hf_dataset.rename_columns({col: col.lower() for col in hf_dataset.column_names})
```
frascuchon, line 198: Here we apply the same logic as for the settings names. I only included the change here, but we could have problems importing case-insensitive records.

But maybe the solution would be relaxing this constraint on the server backend cc @jfcalvo . Anyway, I think we currently have a risk applying the name.lower() for all the settings models and we should change it.

argilla/src/argilla/records/_io/_datasets.py
```diff
@@ -65,12 +65,23 @@ def _cast_classlabels_as_strings(hf_dataset: "HFDataset", columns: List[str]) ->
     Returns:
         HFDataset: The Hugging Face dataset with class label features cast as strings.
     """
+
+    def label_column2str(x: dict, column: str, features: dict) -> Dict[str, Union[str, None]]:
+        try:
+            value = features[column].int2str(x[column])
+        except Exception as ex:
+            warnings.warn(f"Could not cast {x[column]} to string. Error: {ex}")
+            value = None
```
frascuchon, line 74: The behavior here is to define a None value when conversion is not possible, but we could this and just check `x[column] >= 0` and fail otherwise.

burtenshaw (approved):

PR #5524, opened 2024-09-20
[BUGFIX] `argilla`: improve `from_hub` robustness

frascuchon:
# Description

This PR adds some changes in how settings are built from the dataset features:

- All questions are optional except the first one
- Avoid generating extra questions/metadata duplicated from the same feature
- A default question will be created if there is no question at all

Also, add robustness to the process with the following changes:

- Sanitize settings and dataset name according to the server validation
- Skip features with unsupported names (after sanitizing)
- Allow passing subset dataset configuration for HF datasets.

These changes result in datasets in Argilla with a data structure more similar to the original one.

**Type of change**

**How Has This Been Tested**

**Checklist**

### DIFF
--- a/argilla/src/argilla/settings/_io/_hub.py
+++ b/argilla/src/argilla/settings/_io/_hub.py
@@ -136,7 +136,7 @@ def _render_code_snippet(repo_id: str):
     from rich.syntax import Syntax
 
     message = """
-    No questions found in the dataset features. A default question 'quality' has been added. 
+    No questions found in the dataset features. A default question 'quality' has been added.
     If you want to customize the dataset differently, you can do the following:
     """
     code_block = f"""
--- a/argilla/src/argilla/settings/_io/_hub.py
+++ b/argilla/src/argilla/settings/_io/_hub.py
@@ -131,8 +131,35 @@ def _is_chat_feature(sub_features):
     )
 
 
+def _render_code_snippet(repo_id: str):
+    """Render the code snippet to use feature_mapping load a dataset and log its records."""
+    from rich.console import Console
+    from rich.syntax import Syntax
+
+    message = f"""
+    You can assign the dataset's columns to questions, fields, or metadata properties using the feature_mapping argument.
+    You can also add custom settings to the dataset by manipulting the settings object.
+    """
+    code_block = f"""
+    settings = rg.Settings(
+        repo_id="{repo_id}", 
+        feature_mapping={{{{"<column_name>": "question"}}}}
+    )
+    # add more custom settings here
+    # via settings.questions, settings.fields, settings.vectors, or settings.metadata
+    dataset = rg.Dataset(name="{repo_id}", settings=settings)
+    ds = load_dataset("{repo_id}")
+    dataset.records.log(ds)
+    """
+
+    console = Console()
+    code_block = Syntax(code_block, "python", theme="monokai", line_numbers=True)
+    console.print(f"[bold yellow]{message}[/bold yellow]")
+    console.print(code_block)
+
+
 def _define_settings_from_features(
-    features: Union[List[Dict], Dict[str, Any]], feature_mapping: Optional[Dict[str, str]]
+    features: Union[List[Dict], Dict[str, Any]], feature_mapping: Optional[Dict[str, str]], repo_id: str
 ) -> "Settings":
     """Define the argilla settings from the features of a dataset.
 
@@ -158,7 +185,6 @@ def _define_settings_from_features(
     fields = []
     questions = []
     metadata = []
-    mapping = defaultdict(list)
     feature_mapping = feature_mapping or {}
 
     for name, feature in features.items():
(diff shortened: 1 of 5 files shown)


davidberenstein1957 (approved):
It makes sense to me to only define these thing explicitly.

argilla/src/argilla/settings/_io/_hub.py
```diff
@@ -131,8 +130,34 @@ def _is_chat_feature(sub_features):
...
+    from rich.syntax import Syntax
+
+    message = """
+    You can assign the dataset's columns to questions, fields, or metadata properties using the feature_mapping argument.
+    You can also add custom settings to the dataset by manipulting the settings object.
+    """
+    code_block = f"""
+    # 1. map the dataset columns to question, field, vector, or metadata
+    settings = rg.Settings.from_hub(
+        repo_id="yahma/alpaca-cleaned",
```
dvsrepo: {repo_id} ?

argilla/src/argilla/settings/_io/_hub.py
```diff
@@ -223,24 +229,24 @@ def _define_settings_from_features(
         else:
             warnings.warn(f"Feature '{name}' has an unsupported type. Skipping. Feature type: {feature_type}")
 
-    mapping = {
-        key: value[0] if isinstance(value, list) and len(value) == 1 else tuple(value) for key, value in mapping.items()
-    }
-
     if not questions:
-        questions.append(TextQuestion(name="comment", required=True))
+        questions.append(LabelQuestion(name="quality", required=True, labels=["good", "bad"]))
```
dvsrepo: Just discussed with @frascuchon and give it another thought and maybe rating with 1 to 5 is even more general and widely applicable, what do you think @burtenshaw ? This way if I want to start "categorizing" the dataset at least I have a pretty standard likert-scale for "quality"

argilla/src/argilla/settings/_io/_hub.py
```diff
@@ -223,24 +229,24 @@ def _define_settings_from_features(
...
+        warnings.warn(
```
dvsrepo: Screenshot 2024-09-24 at 15 26 37

Could we try to make this message more integrated?

No questions found in the dataset features. A default question 'quality' has been added. If you want to customize the dataset differently, you can do the following:

[and here we show the code snippet, otherwise the message is fragmented and not clear what to do.]

Also the first recommendation should be how to add a question, rather than mapping /or add question.

burtenshaw: I made the message closer to the snippet and dropped the warning
Screenshot 2024-09-24 at 16 01 33

If I change the order of the options in the snippet, I will need to define `rg.Settings` twice. It might become a bit complex?

argilla/src/argilla/settings/_io/_hub.py
```diff
@@ -223,24 +230,32 @@ def _define_settings_from_features(
...
-    settings = Settings(fields=fields, questions=questions, metadata=metadata, mapping=mapping)
+    settings = Settings(fields=fields, questions=questions, metadata=metadata)
 
     return settings
 
 
 def build_settings_from_repo_id(repo_id: str, feature_mapping: Optional[Dict[str, str]] = None) -> "Settings":
     dataset_features = _get_dataset_features(repo_id)
-    return _define_settings_from_features(dataset_features, feature_mapping)
+    return _define_settings_from_features(dataset_features, feature_mapping, repo_id)
```
frascuchon: I have a weird feeling about passing this `repo_id` to just build the code snippet. Why not handle the no-questions scenario here?

```python
    dataset_features = _get_dataset_features(repo_id)
    settings = _define_settings_from_features(dataset_features, feature_mapping)
    if not settings.questions:
       question= RatingQuestion(...)
       warnings.warn(...)
       _render_code_snippet(repo_id)
    return settings
```

argilla/src/argilla/settings/_io/_hub.py
```diff
@@ -131,8 +130,33 @@ def _is_chat_feature(sub_features):
...
+    If you want to customize the dataset differently, you can do the following:
+    """
+    code_block = f"""
+    # 1. map the dataset columns to question, field, vector, or metadata
+    settings = rg.Settings.from_hub(
+        repo_id="{repo_id}",
+        feature_mapping={"{"}"<column_name>": "question"{"}"},
+    )
+    # 2. or, create new questions, fields, vectors, or metadata properties
+    settings.questions = [rg.TextQuestion(name="new_question", required=True)]
```
frascuchon: Would be `settings.questions.add(...)` more aligned to the doc references?

argilla/src/argilla/settings/_io/_hub.py
burtenshaw: I implemented this in the latest commit, but it impacts the tests. I noted out the failing assertions. What do you think we should do? I could leave it, or mock _get_dataset_features.

burtenshaw (approved):

PR #5540, opened 2024-09-25
Feat: 5530 feature updated at inserted at properties on retrieved records

### LINKED ISSUE #5530: [FEATURE] updated_at / inserted_at properties on retrieved Records
**Why do we want this feature**

When building and maintaining datasets, it is critical to trace **when** a data point has been generated or modified. As this can be used for:

* KPI definition
* Explainability and debugging of new models
* Better understanding of drifts
* ...

When storing the records in any Argilla dataset, we can see that the following record information is available in the server:
image

Nevertheless, when loading that record using the python client, we cannot access to the `updated_at` and `inserted_at` fields. 

We checked the Record model and saw that these properties are not available in the python client.

The properties of `updated_at` and `inserted_at` are retrieved from the server and can be retrieved from the Record model as properties.

Ideally, these two properties could also be used for filtering records.

Using metadata to that end. It might be a temporary fix but adds unnecessary redundancy (this information already exists in the record at server-side) and adding some server-generated timestamp will be more robust.

davidberenstein1957:
# Description

Closes #5530 

**Type of change**

**How Has This Been Tested**

**Checklist**

### DIFF
--- a/argilla/tests/integration/test_list_records.py
+++ b/argilla/tests/integration/test_list_records.py
@@ -106,3 +106,21 @@ def test_list_records_with_responses(client: Argilla, dataset: Dataset):
 
     assert records[1].responses["comment"][0].value == "The comment"
     assert records[1].responses["sentiment"][0].value == "negative"
+
+
+def test_list_records_with_updated_at_and_inserted_at(client: Argilla, dataset: Dataset):
+    dataset.records.log(
+        [
+            {"text": "The record text field", "id": 1},
+            {"text": "The record text field", "id": 2},
+        ]
+    )
+
+    records = list(dataset.records(with_responses=True))
+    assert len(records) == 2
+
+    assert records[0].inserted_at
+    assert records[0].updated_at
+
+    assert records[1].inserted_at
+    assert records[1].updated_at
--- a/argilla/tests/integration/test_list_records.py
+++ b/argilla/tests/integration/test_list_records.py
@@ -66,6 +66,93 @@ def test_list_records_with_start_offset(client: Argilla, dataset: Dataset):
     ]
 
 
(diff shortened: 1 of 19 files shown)


davidberenstein1957:
@frascuchon do you think it also makes sense to update the `__repr__` of all resources to include this info?

frascuchon:
> @frascuchon do you think it also makes sense to update the `__repr__` of all resources to include this info?

Yes, I think it makes sense

argilla/tests/integration/test_list_records.py
```diff
@@ -106,3 +106,21 @@ def test_list_records_with_responses(client: Argilla, dataset: Dataset):
 
     assert records[1].responses["comment"][0].value == "The comment"
     assert records[1].responses["sentiment"][0].value == "negative"
+
+
+def test_list_records_with_updated_at_and_inserted_at(client: Argilla, dataset: Dataset):
```
frascuchon, line 198: Should we expose these new attributes in the `to_dict` methods? cc @burtenshaw

davidberenstein1957: for me it makes sense too.

burtenshaw: That should be ok. They will be used in `records._io._generic` and exported in all formats, so we should validate serialisation.

frascuchon: I see the `.serialize` methods already serialize these new attributes. I would leave it as is and review the logic when unifying to_dict and serialize methods: Ref #4944 

(I added a todo commit https://github.com/argilla-io/argilla/pull/5540/commits/87cb1ebc3973f4b327b2dd2dc0e76f5db563dc1f)

frascuchon (approved):

PR #5541, opened 2024-09-25
[FEATURE] update token retrieval logic

### LINKED ISSUE #5543: automatically load token from collab secrets if it exists
## Describe the solution you'd like
if token exists in collab secrets automatically load it from there
# Describe alternatives you've considered**
manually loading the token from collab secrets

not-lain:
# Description
support token retrieval from colab secrets
this PR is inspired by huggingface_hub's [get_token](https://github.com/huggingface/huggingface_hub/blob/12ec449b2346cb115a23bd1628aeac4dae5014bd/src/huggingface_hub/utils/_token.py) function
if token exists in secrets it will prompt the user if they want to access it or not
image

#Fixes : https://github.com/argilla-io/argilla/issues/5543

**Type of change**

**How Has This Been Tested**

**Checklist**

### DIFF
--- a/argilla/src/argilla/_api/_client.py
+++ b/argilla/src/argilla/_api/_client.py
@@ -29,6 +29,8 @@ from argilla._api._users import UsersAPI
 from argilla._api._vectors import VectorsAPI
 from argilla._api._workspaces import WorkspacesAPI
 from argilla._exceptions import ArgillaError
+from argilla._constants import _DEFAULT_API_URL
+from argilla._api._token import get_secret
 
 __all__ = ["APIClient"]
 
--- a/argilla/src/argilla/_api/_client.py
+++ b/argilla/src/argilla/_api/_client.py
@@ -33,9 +33,9 @@ from argilla._api._token import get_token
 
 __all__ = ["APIClient"]
 
-tokens = get_token()
-ARGILLA_API_URL = tokens["ARGILLA_API_URL"]
-ARGILLA_API_KEY = tokens["ARGILLA_API_KEY"]
+
+ARGILLA_API_URL = get_secret("ARGILLA_API_URL") or _DEFAULT_API_URL
+ARGILLA_API_KEY = get_secret("ARGILLA_API_KEY")
 
 DEFAULT_HTTP_CONFIG = HTTPClientConfig(api_url=ARGILLA_API_URL, api_key=ARGILLA_API_KEY)
 
--- a/argilla/src/argilla/_api/_client.py
+++ b/argilla/src/argilla/_api/_client.py
@@ -29,7 +29,6 @@ from argilla._api._users import UsersAPI
 from argilla._api._vectors import VectorsAPI
 from argilla._api._workspaces import WorkspacesAPI
 from argilla._exceptions import ArgillaError
-from argilla._api._token import get_token
 
 __all__ = ["APIClient"]
 
--- a/argilla/src/argilla/_api/_client.py
+++ b/argilla/src/argilla/_api/_client.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 import logging
-import os
 from typing import Optional
 
 import httpx
@@ -29,13 +28,14 @@ from argilla._api._records import RecordsAPI
 from argilla._api._users import UsersAPI
 from argilla._api._vectors import VectorsAPI
 from argilla._api._workspaces import WorkspacesAPI
-from argilla._constants import _DEFAULT_API_URL
 from argilla._exceptions import ArgillaError
+from argilla._api._token import get_token
 
 __all__ = ["APIClient"]
 
-ARGILLA_API_URL = os.getenv(key="ARGILLA_API_URL", default=_DEFAULT_API_URL)
-ARGILLA_API_KEY = os.getenv(key="ARGILLA_API_KEY")
+tokens = get_token()
+ARGILLA_API_URL = tokens["ARGILLA_API_URL"]
+ARGILLA_API_KEY = tokens["ARGILLA_API_KEY"]
 
 DEFAULT_HTTP_CONFIG = HTTPClientConfig(api_url=ARGILLA_API_URL, api_key=ARGILLA_API_KEY)
 
--- a/argilla/src/argilla/_api/_token.py
+++ b/argilla/src/argilla/_api/_token.py
@@ -41,7 +41,7 @@ def _get_secret_from_environment(name: str) -> Optional[str]:
 
 
 
-def _get_token_from_google_colab() -> Optional[str]:
+def _get_secret_from_google_colab(name:str) -> Optional[str]:
     """Get token from Google Colab secrets vault using `google.colab.userdata.get(...)`.
 
     Token is read from the vault only once per session and then stored in a global variable to avoid re-requesting
--- a/argilla/src/argilla/_api/_token.py
+++ b/argilla/src/argilla/_api/_token.py
@@ -15,7 +15,7 @@
 import os
 import warnings
 from threading import Lock
-from typing import Optional
+from typing import Optional, Dict
 from huggingface_hub.utils._runtime import is_google_colab
 from argilla._constants import _DEFAULT_API_URL
 
--- a/argilla/src/argilla/_api/_token.py
+++ b/argilla/src/argilla/_api/_token.py
@@ -2,7 +2,7 @@ import os
 import warnings
 from threading import Lock
 from typing import Optional
-from huggingface_hub.utils import is_google_colab, is_colab_enterprise
+from huggingface_hub.utils._runtime import is_google_colab, is_colab_enterprise
 from argilla._constants import _DEFAULT_API_URL
 
 
--- a/argilla/src/argilla/_api/_token.py
+++ b/argilla/src/argilla/_api/_token.py
@@ -35,8 +35,8 @@ def get_token() -> Optional[dict]:
     return _get_token_from_google_colab() or _get_token_from_environment()
 
 
-def _get_token_from_environment() -> Dict[str,str]:
-    """ Get the argilla credentials from the environment """
+def _get_token_from_environment() -> Dict[str, str]:
+    """Get the argilla credentials from the environment"""
     ARGILLA_API_URL = _clean_token(os.getenv(key="ARGILLA_API_URL", default=_DEFAULT_API_URL))
     ARGILLA_API_KEY = _clean_token(os.getenv(key="ARGILLA_API_KEY"))
     return {"ARGILLA_API_URL": ARGILLA_API_URL, "ARGILLA_API_KEY": ARGILLA_API_KEY}
--- a/argilla/src/argilla/_api/_token.py
+++ b/argilla/src/argilla/_api/_token.py
@@ -70,12 +70,13 @@ def _get_secret_from_google_colab(name: str) -> Optional[str]:
         except userdata.NotebookAccessError:
             # Means the user has a secret call `ARGILLA_API_URL` and `ARGILLA_API_URL` and got a popup "please grand access to ARGILLA_API_URL" and refused it
             # => warn user but ignore error => do not re-request access to user
-            warnings.warn(
-                "\nAccess to the secret `ARGILLA_API_URL` and `ARGILLA_API_KEY` has not been granted on this notebook."
-                "\nYou will not be requested again."
-                "\nPlease restart the session if you want to be prompted again."
-            )
-            _GOOGLE_COLAB_SECRET = None
+            if not _IS_GOOGLE_COLAB_CHECKED:
+               warnings.warn(
+                   f"\nAccess to the secret {name} has not been granted on this 
+notebook."
+                   "\nYou will not be requested again."
+                   "\nPlease restart the session if you want to be prompted again."
+               )
         except userdata.SecretNotFoundError:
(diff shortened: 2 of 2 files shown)


not-lain:
here's a minimalistic colab code to try this PR before merging: 
```
!pip install -e "git+https://github.com/not-lain/argilla@update-token-retrieval#egg=argilla&subdirectory=argilla"
```
restart session then try this : 
```python
from argilla._api._token import get_token
get_token()
```

not-lain:
cc @frascuchon for review

argilla/src/argilla/_api/_token.py
```diff
@@ -0,0 +1,114 @@
...
+    """
+    Get token if present
+
+    Returns:
+        `dict` or `None`: The token, `None` if it doesn't exist.
+    """
+    return _get_token_from_google_colab() or _get_token_from_environment()
+
+
+def _get_token_from_environment() -> Optional[str]:
```
burtenshaw: Should this output type be a dict?
```suggestion
def _get_token_from_environment() -> Dict[str,str]:
```

argilla/src/argilla/_api/_token.py
```diff
@@ -0,0 +1,114 @@
...
+    ARGILLA_API_URL = _clean_token(os.getenv(key="ARGILLA_API_URL", default=_DEFAULT_API_URL))
```
burtenshaw:
```suggestion
    """ Get the argilla credentials from the environment """
    ARGILLA_API_URL = _clean_token(os.getenv(key="ARGILLA_API_URL", default=_DEFAULT_API_URL))
```

burtenshaw (commented):
Thanks for the PR. This looks really handy. I added some small comments. Two general things:

- On naming: Currently, we're referring to our server api credentials as 'credentials' and 'api_key', rather than 'token'. I think this code should use the same naming. There's been confusion between argilla api_key and hf token before. 
- The logic at the moment is to get both or no values (key and url) from `userdata` and fall back to `environ`. Do you think there's a use case where the user just wants the key from the userdata?

frascuchon:
Thanks @not-lain for the contribution. Yesterday I was thinking about this. Since `argilla` already has the `huggingface_hub` dependency, why not use the `get_token` function to configure the token? With this, we keep aligned with other HF packages. What do you think?

burtenshaw:
@frascuchon If you want to try it, I used this colab: https://colab.research.google.com/drive/1Qh1DFo7rZD_qR-FqCXYSRyv4m2_Z8qIG?usp=sharing.

not-lain:
thanks a lot for the review
> On naming: Currently, we're referring to our server api credentials as 'credentials' and 'api_key', rather than 'token'. I think this code should use the same naming. There's been confusion between argilla api_key and hf token before.

thanks for letting me know, will fix this in the next commit.
>The logic at the moment is to get both or no values (key and url) from userdata and fall back to environ. Do you think there's a use case where the user just wants the key from the userdata?

I can adapt my implementation to handle this case, I was a bit too skeptical because of the collab thread hanging forever issue (see https://github.com/huggingface/huggingface_hub/issues/1952 ) and I wanted to make something that is robust instead of flexible.
Will try to make it work and commit to this pr once it's ready.

burtenshaw:
> I can adapt my implementation to handle this case, I was a bit too skeptical because of the collab thread hanging forever issue (see [huggingface/huggingface_hub#1952](https://github.com/huggingface/huggingface_hub/issues/1952) )

@not-lain Understood. I think it's a nice to have anyway, so if it compromises the functionality feel free to ignore the suggestion.

burtenshaw:
@not-lain @frascuchon We should put this is 2.3.0!

argilla/src/argilla/_api/_token.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+def _get_token_from_environment() -> Dict[str, str]:
+    """Get the argilla credentials from the environment"""
+    ARGILLA_API_URL = _clean_token(os.getenv(key="ARGILLA_API_URL", default=_DEFAULT_API_URL))
+    ARGILLA_API_KEY = _clean_token(os.getenv(key="ARGILLA_API_KEY"))
+    return {"ARGILLA_API_URL": ARGILLA_API_URL, "ARGILLA_API_KEY": ARGILLA_API_KEY}
+
+
+def _get_token_from_google_colab() -> Optional[str]:
```
frascuchon: Maybe we can generalize this function to retrieve a secret from colab, an use it multiple times, instead of hardcoding the expected variables:

```
_get_secret_from_google_colab("ARGILLA_AP_KEY")
_get_secret_from_google_colab("ARGILLA_API_URL")
```
We could do something similar to the `get_token` and the `_get_token_from_environment` functions:

```
ARGILLA_API_KEY = get_secret("ARGILLA_API_KEY")
ARGILLA_API_URL = get_secret("ARGILLA_API_URL")
```

argilla/src/argilla/_api/_token.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+def get_token() -> Optional[dict]:
+    """
+    Get token if present
+
+    Returns:
+        `dict` or `None`: The token, `None` if it doesn't exist.
+    """
+    return _get_token_from_google_colab() or _get_token_from_environment()
```
frascuchon:
```suggestion
def get_secret(name: str) -> Optional[str]:
    """
    Get secret if present

    Returns:
        `str` or `None`: The secret value, `None` if it doesn't exist.
    """
    return _get_secret_from_google_colab(name) or _get_secret_from_environment(name)
```

argilla/src/argilla/_api/_token.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+def _get_token_from_environment() -> Dict[str, str]:
+    """Get the argilla credentials from the environment"""
+    ARGILLA_API_URL = _clean_token(os.getenv(key="ARGILLA_API_URL", default=_DEFAULT_API_URL))
+    ARGILLA_API_KEY = _clean_token(os.getenv(key="ARGILLA_API_KEY"))
+    return {"ARGILLA_API_URL": ARGILLA_API_URL, "ARGILLA_API_KEY": ARGILLA_API_KEY}
```
frascuchon:
```suggestion
def _get_secret_from_environment(name: str) -> Optional[str]:
    """Get the secret  from the environment"""
    return _clean_token(os.getenv(key=name))

```

argilla/src/argilla/_api/_client.py
```diff
@@ -29,13 +28,14 @@
...
 from argilla._exceptions import ArgillaError
+from argilla._api._token import get_token
 
 __all__ = ["APIClient"]
 
-ARGILLA_API_URL = os.getenv(key="ARGILLA_API_URL", default=_DEFAULT_API_URL)
-ARGILLA_API_KEY = os.getenv(key="ARGILLA_API_KEY")
+tokens = get_token()
+ARGILLA_API_URL = tokens["ARGILLA_API_URL"]
+ARGILLA_API_KEY = tokens["ARGILLA_API_KEY"]
```
frascuchon:
```suggestion

ARGILLA_API_URL = get_secret("ARGILLA_API_URL") or _DEFAULT_API_URL
ARGILLA_API_KEY = get_secret("ARGILLA_API_KEY")
```

argilla/src/argilla/_api/_token.py
frascuchon:
```suggestion
def _get_secret_from_google_colab(name:str) -> Optional[str]:
```

argilla/src/argilla/_api/_token.py
```diff
@@ -0,0 +1,115 @@
...
+            return None
+
+        try:
+            ARGILLA_API_URL = userdata.get("ARGILLA_API_URL")
+            ARGILLA_API_KEY = userdata.get("ARGILLA_API_KEY")
+
+            _GOOGLE_COLAB_SECRET = {
+                "ARGILLA_API_URL": _clean_token(ARGILLA_API_URL),
+                "ARGILLA_API_KEY": _clean_token(ARGILLA_API_KEY),
+            }
```
frascuchon:
```suggestion
    with _GOOGLE_COLAB_SECRET_LOCK:
        global _IS_GOOGLE_COLAB_CHECKED

        # we can add here some cache mechanism

        try:
            from google.colab import userdata
            from google.colab.errors import Error as ColabError
        except ImportError:
            return None

        try:
            secret_value = userdata.get(name)
```

argilla/src/argilla/_api/_token.py
```diff
@@ -0,0 +1,115 @@
...
+
+        except userdata.NotebookAccessError:
+            # Means the user has a secret call `HF_TOKEN` and got a popup "please grand access to HF_TOKEN" and refused it
+            # => warn user but ignore error => do not re-request access to user
+            warnings.warn(
+                "\nAccess to the secret `ARGILLA_API_URL` and `ARGILLA_API_KEY` has not been granted on this notebook."
+                "\nYou will not be requested again."
+                "\nPlease restart the session if you want to be prompted again."
+            )
+            _GOOGLE_COLAB_SECRET = None
```
frascuchon:
```suggestion
            if not _IS_GOOGLE_COLAB_CHECKED:
               warnings.warn(
                   f"\nAccess to the secret {name} has not been granted on this 
notebook."
                   "\nYou will not be requested again."
                   "\nPlease restart the session if you want to be prompted again."
               )
```

argilla/src/argilla/_api/_token.py
```diff
@@ -0,0 +1,115 @@
...
+            warnings.warn(
+                f"\nError while fetching `ARGILLA_API_URL` and `ARGILLA_API_KEY` secret value from your vault: '{str(e)}'."
+                "\nYou are not authenticated with the Argilla in this notebook."
+                "\nIf the error persists, please let us know by opening an issue on GitHub "
+                "(https://github.com/argilla-io/argilla//issues/new)."
+            )
+            _GOOGLE_COLAB_SECRET = None
+
+        _IS_GOOGLE_COLAB_CHECKED = True
+        return _GOOGLE_COLAB_SECRET
```
frascuchon:
```suggestion
            # Means the user did not define a name  secret => warn
            warnings.warn(f"\nThe secrets {name} and does not exist in your Colab secrets.")
        except ColabError as e:
            # Something happen but we don't know what => recommend to open a GitHub issue
            warnings.warn(
                f"\nError while fetching {name} secret value from your vault: '{str(e)}'."
                "\nYou are not authenticated with the Argilla in this notebook."
                "\nIf the error persists, please let us know by opening an issue on GitHub "
                "(https://github.com/argilla-io/argilla//issues/new)."
            )

        _IS_GOOGLE_COLAB_CHECKED = True

        return secret_value
```

argilla/src/argilla/_api/_token.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+def _clean_token(token: Optional[str]) -> Optional[str]:
```
frascuchon:
```suggestion
def _clean_secret_value(value: Optional[str]) -> Optional[str]:
```

not-lain:
hi @frascuchon I have tried your changes which can be tested via the following code (different branch from this one)
```
!pip install -e "git+https://github.com/not-lain/argilla@test-colab-lock#egg=argilla&subdirectory=argilla"
```
which produced some weird behavior when running the following code 
```python
from argilla._api._token import get_secret
key = get_secret("ARGILLA_API_KEY")
url = get_secret("ARGILLA_API_URL")
```
initialize your secrets and turn them off 
image
then click [ cancel, grant, cancel ] 
as you might have noticed **3 tabs popped up** which is quite strange and I might attribute this to the shared global lock
```python
        global _IS_GOOGLE_COLAB_CHECKED
```
not entirely sure of this but this is what I have discovered so far and I wanted to report my findings

not-lain:
Apologies I have been looking at this from a different angle, colab turns out to handle things a slightly bit differently fromthe the normal way I was looking at this.
after careful testing, it turns out that 
```python
from argilla._api._token import get_secret
ARGILLA_API_URL = get_secret("ARGILLA_API_URL") 
ARGILLA_API_KEY = get_secret("ARGILLA_API_KEY")
```
is **DIFFERENT** from 
```python
import argilla as rg
rg.client._api.APIClient() 
```
I have no clue how and why this happened, but I will report my findings to colab and let them figure this out.

I will now proceed to merge the changes that @frascuchon has recommended

not-lain:
This PR should be ready for review, you can test this out by using the following code 

1. install

```
!pip install -e "git+https://github.com/not-lain/argilla@update-token-retrieval#egg=argilla&subdirectory=argilla"
```
2. restart session
3. run the following code
```python
import argilla as rg
rg.client._api.APIClient()
```

frascuchon:
Thanks @not-lain for taking into account my comments, and apologize for the inconvenience. So, finally, did you manage the errors applying my changes? Does the env loading work as expected? The current code looks nice!

not-lain:
Hi @frascuchon, yes I can indeed confirm that everything is working perfectly.

frascuchon (approved):
Thanks @not-lain !

frascuchon:
@burtenshaw Does it make sense to add it somewhere in the docs?

burtenshaw:
> @burtenshaw Does it make sense to add it somewhere in the docs?

@frascuchon Yeah. I would just add a section to client API reference that mentions support for colab userdata. But let's merge this to develop, and I can open a docs PR.

PR #5547, opened 2024-09-27
feat: 5108 feature add method to deploy on spaces through huggingface hub

### LINKED ISSUE #5108: [FEATURE] add method to deploy on spaces through `huggingface_hub`
Argilla is cool but sometimes deployment can contain a lot of clickety steps with copy_pasting.

Some underlying code that goes over steps on https://huggingface.co/docs/hub/en/spaces-sdks-docker-argilla but through the `huggingface_hub`SDK. 
https://huggingface.co/docs/huggingface_hub/main/en/guides/manage-spaces#manage-your-space

Also, the client might easily be initialized with the HF key headers after.

```python
rg.deploy_on_spaces(
    repo_id="user/name",
    private=True,
    admin_username=...,
    admin_password=...,
    admin_apikey=...,
) 
```

show a message:

```
Argilla has been deployed on a private HF Space with url http://hf.co/spaces/user/name. You can now connect with 
Client(etc)
```

**Additional context**

davidberenstein1957:
# Description

Closes #5108 

- sets USERNAME to `hf_token["username"]`
- sets no PASSWORD and delegates this to the quickstart image
- sets no WORKSPACE to `argilla`
- sets API_KEY needs to be provided manually
- sets`org_name` to `hf_token["username"]`
- sets`repo_name` to `"argilla"`
- default spaces and hardware, i.e. cpu basic and no persistant storage
- if private, sets the correct connection headers.

```python
import argilla as rg

client = rg.Argilla.deploy_on_spaces(api_key="12345678")
```

```python
import argilla as rg

client = rg.Argilla.deploy_on_spaces(
    repo_name="my-example",
    org_name=None,
    token=None,
    space_storage=None,
    space_hardware=None,
    private=False,
)
```

image

**Type of change**

**How Has This Been Tested**

**Checklist**


argilla/src/argilla/_helpers/_deploy.py
```diff
@@ -0,0 +1,203 @@
...
+_FROM_REPO_ID = "argilla/argilla-template-space"
+
+
+class SpacesDeploymentMixin(LoggingMixin):
+    def _acquire_hf_token(self, token: Union[str, None]) -> str:
+        """Obtain the Hugging Face authentication token to deploy a space and authenticate."""
+        if token is None:
+            token = get_token()
+        if token is None:
+            token = os.getenv("HF_TOKEN")
```
not-lain:
```suggestion
```
already part of the `get_token` function

argilla/src/argilla/_helpers/_deploy.py
```diff
@@ -89,7 +89,7 @@ def deploy_on_spaces(
         For a full guide check our docs.
 
         Args:
-            repo_name (str): The ID of the repository where Argilla will be deployed.
+            repo_name (Optional[str]): The ID of the repository where Argilla will be deployed. Defaults to "argilla".
             org_name (Optional[str]): The name of the organization where Argilla will be deployed. Defaults to None.
```
burtenshaw:
```suggestion
            org_name (Optional[str]): The name of the organization where Argilla will be deployed. Defaults to None and Hugging Face user name space will be used.
```

argilla/src/argilla/_helpers/_deploy.py
```diff
@@ -108,14 +108,15 @@ def deploy_on_spaces(
...
+                repo_name="my-space",
+                org_name="my-org",
                 token="hf_ABC123",
                 space_storage="10GB",
                 space_hardware="cpu-basic",
                 private=True,
                 username="admin",
                 password="12345678",
-                overwrite=True,
+                overwrite=False,
```
burtenshaw: Do you think something like `on_exists` could be more user friendly here? So the user can say `on_exists=skip` and just leave it in their code?

argilla/src/argilla/_helpers/_deploy.py
```diff
@@ -0,0 +1,201 @@
...
+        # Get the org name from the repo name or default to the current user
+        token_username = api.whoami(token=hf_token)["name"]
+        org_name = org_name or token_username
+        repo_id = f"{org_name}/{repo_name}"
+
+        # Define the api_key for the space
+        secrets = [
+            {"key": "API_KEY", "value": api_key, "description": "The API key of the owner user."},
+            {"key": "USERNAME", "value": token_username, "description": "The username of the owner user."},
+            {"key": "PASSWORD", "value": api_key, "description": "The password of the owner user."},
```
frascuchon: You can also pass a default WORKSPACE name to be created.

davidberenstein1957: @frascuchon I agree, but I wanted to keep it as simple as possible, which is why I don't let users provide a `username` or `password` either. Or do you propose setting the WORKSPACE based on the username?

davidberenstein1957: @frascuchon WDYT of using the api_key as password? I think it would make it easier to simplify the flow given that we support OAuth. An alternative would be just creating a random `uuid`

frascuchon: or just a `default` or `argilla` workspace.

davidberenstein1957: @frascuchon, when we don't pass anything, what will it use as default? I assymed we were using a default: `default` or `argilla`, but perhaps that is not the case?

burtenshaw: I think the api_key==password is ok. The hf_key vs api_key is already complex enough for some users.

argilla/src/argilla/_helpers/_deploy.py
```diff
@@ -0,0 +1,195 @@
...
+
+from argilla._helpers._log import LoggingMixin
+
+if TYPE_CHECKING:
+    from huggingface_hub.hf_api import RepoUrl, SpaceHardware, SpaceStorage  # noqa
+
+    from argilla.client import Argilla
+
+_SLEEP_TIME = 10
+_FROM_REPO_ID = "argilla/argilla-template-space"
```
frascuchon:
```suggestion
_ARGILLA_SPACE_TEMPLATE_REPO = "argilla/argilla-template-space"
```

argilla/src/argilla/_helpers/_deploy.py
```diff
@@ -0,0 +1,195 @@
...
+
+        # Get the org name from the repo name or default to the current user
+        token_username = api.whoami(token=hf_token)["name"]
+        org_name = org_name or token_username
+        repo_id = f"{org_name}/{repo_name}"
+
+        # Define the api_key for the space
+        secrets = [
+            {"key": "API_KEY", "value": api_key, "description": "The API key of the owner user."},
+            {"key": "USERNAME", "value": token_username, "description": "The username of the owner user."},
```
frascuchon: If the token is related to an org, the created owner will be the org name, but this clashes with the changes related to the `SPACES_CREATOR_USER_ID`. What should be the expected behavior in that case?

argilla/src/argilla/_helpers/_deploy.py
```diff
@@ -0,0 +1,195 @@
...
+            import argilla as rg
+
+            client = rg.Argilla.deploy_on_spaces(api_key="12345678")
+            ```
+        """
+        hf_token = cls._acquire_hf_token(ht_token=hf_token)
+        api = HfApi(token=hf_token)
+
+        # Get the org name from the repo name or default to the current user
+        token_username = api.whoami(token=hf_token)["name"]
```
frascuchon: since the HfApi instance is created with a token, do we need to pass again on each `api` method call?

argilla/src/argilla/_helpers/_deploy.py
```diff
@@ -0,0 +1,195 @@
...
+        secrets = [
+            {"key": "API_KEY", "value": api_key, "description": "The API key of the owner user."},
+            {"key": "USERNAME", "value": token_username, "description": "The username of the owner user."},
+            {"key": "WORKSPACE", "value": "argilla", "description": "The workspace of the space."},
+        ]
+
+        # Check if the space already exists
+        if api.repo_exists(repo_id=repo_id, repo_type="space", token=hf_token):
+            if cls._check_if_runtime_can_be_build(api.get_space_runtime(repo_id=repo_id, token=hf_token)):
+                api.restart_space(repo_id=repo_id, token=hf_token)
```
frascuchon: What is the purpose of this line? Do we need to restart the space before adding the secrets? and the hardware and storage requests?

argilla/src/argilla/_helpers/_deploy.py
```diff
@@ -0,0 +1,195 @@
...
+            if cls._is_interactive():
+                notebook_login()
+            else:
+                login()
+            ht_token = get_token()
+        return ht_token
+
+    @classmethod
+    def _check_if_running(cls, runtime: SpaceRuntime) -> bool:
+        """Check the current stage of the space runtime. Simplified to return True when being built."""
```
frascuchon, line 147: Can you link the hf docs describing the deployment states?

argilla/src/argilla/_helpers/_deploy.py
```diff
@@ -0,0 +1,195 @@
...
+        if runtime.stage in ["RUNNING"]:
+            return False
+        elif runtime.stage in [
+            "RUNNING_APP_STARTING",
+            "RUNNING_BUILDING",
+            "BUILDING",
+            "PAUSED",
+            "STOPPED",
```
frascuchon: What happens when the space is PAUSED or STOPPED? The generated client won't work, right?

argilla/src/argilla/_helpers/_deploy.py
```diff
@@ -0,0 +1,195 @@
...
+class SpacesDeploymentMixin(LoggingMixin):
+    @classmethod
+    def deploy_on_spaces(
+        cls,
+        api_key: str,
+        repo_name: Optional[str] = "argilla",
+        org_name: Optional[str] = None,
+        hf_token: Optional[str] = None,
+        space_storage: Optional[Union[str, "SpaceStorage", None]] = None,
+        space_hardware: Optional[Union[str, "SpaceHardware"]] = "cpu-basic",
```
frascuchon: We could include the available values for these parameters.

argilla/src/argilla/_helpers/_deploy.py
davidberenstein1957: It checks if the space can be restarted, i.e. is in a valid state without errors and without it being in the process of being built. If it can, the space is just restarted/rebooted, to make the code more reusable.

argilla/src/argilla/_helpers/_deploy.py
davidberenstein1957: @frascuchon, there isn't any proper documentation. I looked into this and ended up copying the states from the internal Hub Pydantic model/enum. Also, this has originally been copied from the internal moonlanding repo.

argilla/src/argilla/_helpers/_deploy.py
davidberenstein1957:
```suggestion
```

davidberenstein1957: Then the space is restarted in the code above. It makes these values obsolete, so I will add them to another if statement to be more secure.

argilla/src/argilla/_helpers/_deploy.py
davidberenstein1957: There are no organizational-level tokens anymore, however, we can remove this logic since it is handled internally within the quickstart image now.

argilla/src/argilla/_helpers/_deploy.py
davidberenstein1957: I will

argilla/src/argilla/_helpers/_deploy.py
```diff
@@ -0,0 +1,185 @@
...
+        if ht_token is None:
+            if cls._is_interactive():
+                notebook_login()
+            else:
+                login()
+            ht_token = get_token()
+        return ht_token
+
+    @classmethod
+    def _check_if_running(cls, stage: SpaceStage) -> bool:
```
frascuchon: I thought this method returned True if the space is running. Maybe we can change the name
```suggestion
    def _is_building(cls, stage: SpaceStage) -> bool:
```

argilla/src/argilla/_helpers/_deploy.py
```diff
@@ -0,0 +1,185 @@
...
+        """Check the current stage of the space runtime. Simplified to return True when being built."""
+        if stage in ["RUNNING"]:
+            return False
+        elif stage in ["RUNNING_APP_STARTING", "RUNNING_BUILDING", "BUILDING", "APP_STARTING"]:
+            return True
+        elif stage in ["PAUSED", "STOPPED"]:
+            return False
+        else:
+            raise ValueError(f"Space configuration is wrong and in stage: {stage}")
```
frascuchon:
```suggestion
        if stage in ["RUNNING_APP_STARTING", "RUNNING_BUILDING", "BUILDING", "APP_STARTING"]:
            return True
        elif stage in ["RUNNING", "PAUSED", "STOPPED"]:
            return False
        else:
            raise ValueError(f"Space configuration is wrong and in stage: {stage}")
```

argilla/src/argilla/_helpers/_deploy.py
```diff
@@ -0,0 +1,185 @@
...
+
+    @classmethod
+    def _check_if_stage_can_be_build(cls, stage: SpaceStage) -> bool:
```
frascuchon: This method only checks if the space is paused or stopped. So, maybe change the name to what it checks instead of what flow will do when it is stopped or paused.
```suggestion
    def _is_space_stopped(cls, stage: SpaceStage) -> bool:
```

argilla/src/argilla/_helpers/_deploy.py
```diff
@@ -0,0 +1,185 @@
...
+        # Define the api_key for the space
+        secrets = [
+            {"key": "API_KEY", "value": api_key, "description": "The API key of the owner user."},
+            {"key": "WORKSPACE", "value": "argilla", "description": "The workspace of the space."},
+        ]
+
+        # Check if the space already exists
+        if hf_api.repo_exists(repo_id=repo_id, repo_type="space"):
+            if cls._check_if_stage_can_be_build(hf_api.get_space_runtime(repo_id=repo_id).stage):
+                hf_api.restart_space(repo_id=repo_id)
```
frascuchon: So, the logic here is to restart the space if is down (stopped or paused), right?

argilla/src/argilla/_helpers/_deploy.py
```diff
@@ -0,0 +1,185 @@
...
+
+            if overwrite:
```
frascuchon: To make this work for existing spaces, the api_key should be recreated. Otherwise, the argilla client could not connect properly to the server.

davidberenstein1957: Yes but for me that only makes sense when `overwrite=True` or do you want to enable that by default?

frascuchon: The thing is, with a different api_key, the generated client may fail if the credentials are not the same. Also even with override=True, since the owner API_KEY is not modified once the user is created, the flow may equally fail. 

Maybe the easier way is to skip the override flow for now until we understand it better. Keep the restart in case the space is already created, notifying the user that the proper credentials must be provided. Just to simplify the usage.

burtenshaw (approved):

PR #5554, opened 2024-10-01
Enhancement/improve-error-messaging-for-role-forbidden

burtenshaw:
# Description

Closes #<issue_number>

**Type of change**

- Documentation update

**How Has This Been Tested**

**Checklist**

### DIFF
--- a/argilla/src/argilla/_exceptions/_api.py
+++ b/argilla/src/argilla/_exceptions/_api.py
@@ -36,7 +36,7 @@ class BadRequestError(ArgillaAPIError):
 
 
 class ForbiddenError(ArgillaAPIError):
-    message = "User role is forbidden from performing this this action by server"
+    message = "User role is forbidden from performing this action by server"
 
 
 class NotFoundError(ArgillaAPIError):
--- a/argilla/src/argilla/_exceptions/_api.py
+++ b/argilla/src/argilla/_exceptions/_api.py
@@ -36,7 +36,7 @@ class BadRequestError(ArgillaAPIError):
 
 
 class ForbiddenError(ArgillaAPIError):
-    message = "Forbidden request to the server"
+    message = "User role is forbidden from performing this this action by server"
 
 
 class NotFoundError(ArgillaAPIError):
--- a/argilla/src/argilla/datasets/_resource.py
+++ b/argilla/src/argilla/datasets/_resource.py
@@ -16,7 +16,7 @@ from typing import Optional, Union
 from uuid import UUID, uuid4
 
 from argilla._api import DatasetsAPI
-from argilla._exceptions import NotFoundError, SettingsError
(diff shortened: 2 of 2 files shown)


davidberenstein1957 (commented):
I think this works great. I believe this is specifically for dataset creation but shouldn't we be covering the basic scenario of `rg.Argilla(api_url="correct", api_key="wrong")` as well?

burtenshaw:
> I think this works great. I believe this is specifically for dataset creation but shouldn't we be covering the basic scenario of `rg.Argilla(api_url="correct", api_key="wrong")` as well?

I've improved that with changes to the message in `argilla/src/argilla/_exceptions/_api.py` . We also already have a validation message in `argilla/src/argilla/_exceptions/_client.py` that says:  ""Credentials (api_key and/or api_url) are invalid"".

davidberenstein1957:
> > I think this works great. I believe this is specifically for dataset creation but shouldn't we be covering the basic scenario of `rg.Argilla(api_url="correct", api_key="wrong")` as well?
> 
> I've improved that with changes to the message in `argilla/src/argilla/_exceptions/_api.py` . We also already have a validation message in `argilla/src/argilla/_exceptions/_client.py` that says: ""Credentials (api_key and/or api_url) are invalid"".

but perhaps it would be nice if we could also share a similar message, urging the user to go to my-settings with their owner account?

argilla/src/argilla/datasets/_resource.py
```diff
@@ -157,7 +162,12 @@ def create(self) -> "Dataset":
         Returns:
             Dataset: The created dataset object.
         """
-        super().create()
+        try:
+            super().create()
+        except ForbiddenError as e:
+            settings_url = f"{self._client.api_url}/user-settings"
+            message = f"User is not authorized to create a dataset in workspace {self.workspace.name}. Go to {settings_url} to view your role."
```
frascuchon: What about showing the user role from client.me?

argilla/src/argilla/_exceptions/_api.py
```diff
@@ -36,7 +36,7 @@ class BadRequestError(ArgillaAPIError):
 
 
 class ForbiddenError(ArgillaAPIError):
-    message = "Forbidden request to the server"
+    message = "User role is forbidden from performing this this action by server"
```
frascuchon: What do you think about this message? The forbidden action could be raised not only when creating datasets but also when listing users (`client.users`) or getting datasets from workspaces without access (for admin users) 
```suggestion
    message = "Connected user has no privileges to perform this action on the server"
```

frascuchon: Extra `this`
```suggestion
    message = "User role is forbidden from performing this action by server"
```

frascuchon (approved):

PR #5592, opened 2024-10-10
[FIX] Prevent duplicated field text

leiyre:
This PR updates the conditional logic to ensure that the field text is not duplicated.


damianpumar (approved):

damianpumar:
Great! @leiyre just merge it when you want!

leiyre:
fixed #5629

PR #5601, opened 2024-10-16
feat: add support for class labels and casting rows

jfcalvo:
# Description

This PR adds the following features:
* Add support to dataset `ClassLabel` features casting them using `int2str` so we store string values on the Argilla imported dataset.
* Now the casting is done at the row level and discarding keys that are not part of the mapping sources. Once the values arrive to the different record create values these are already casted.
* We are casting `ClassLabel` features to string and `Image` features to data-url strings.

**Type of change**

**How Has This Been Tested**

- [x] Adding additional tests import real datasets from HF.

**Checklist**

### DIFF
--- a/argilla-server/src/argilla_server/api/schemas/v1/datasets.py
+++ b/argilla-server/src/argilla_server/api/schemas/v1/datasets.py
@@ -168,6 +168,15 @@ class HubDatasetMapping(BaseModel):
     suggestions: Optional[List[HubDatasetMappingItem]] = []
     external_id: Optional[str] = None
 
+    @property
+    def sources(self) -> List[str]:
+        fields_sources = [field.source for field in self.fields]
+        metadata_sources = [metadata.source for metadata in self.metadata]
+        suggestions_sources = [suggestion.source for suggestion in self.suggestions]
+        external_id_source = [self.external_id] if self.external_id else []
+
+        return list(set(fields_sources + metadata_sources + suggestions_sources + external_id_source))
+
 
 class HubDataset(BaseModel):
     name: str
--- a/argilla-server/src/argilla_server/contexts/hub.py
+++ b/argilla-server/src/argilla_server/contexts/hub.py
@@ -22,7 +22,6 @@ from PIL import Image
 from datasets import load_dataset
 from sqlalchemy.ext.asyncio import AsyncSession
 
-
 from argilla_server.models.database import Dataset
 from argilla_server.search_engine import SearchEngine
 from argilla_server.bulk.records_bulk import UpsertRecordsBulk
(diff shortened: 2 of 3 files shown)


argilla-server/src/argilla_server/contexts/hub.py
```diff
@@ -34,13 +33,21 @@
...
 
+FEATURE_TYPE_IMAGE = "Image"
+FEATURE_TYPE_CLASS_LABEL = "ClassLabel"
+
 
 class HubDataset:
     def __init__(self, name: str, subset: str, split: str, mapping: HubDatasetMapping):
         self.dataset = load_dataset(path=name, name=subset, split=split, streaming=True)
         self.mapping = mapping
+        self.mapping_sources = mapping.sources
```
frascuchon: can we rename this to something more intuitive as "mapped_columns" or "source_columns"?

argilla-server/src/argilla_server/api/schemas/v1/datasets.py
```diff
@@ -168,6 +168,15 @@ class HubDatasetMapping(BaseModel):
     suggestions: Optional[List[HubDatasetMappingItem]] = []
     external_id: Optional[str] = None
 
+    @property
+    def sources(self) -> List[str]:
```
frascuchon, line 172: `column_names` instead? I think it would be easier to understand.

argilla-server/src/argilla_server/contexts/hub.py
```diff
@@ -71,53 +78,68 @@ async def _import_batch_to(
...
             dataset,
             RecordsBulkUpsertSchema(items=items),
             raise_on_error=False,
         )
 
-    def _batch_row_to_record_schema(self, batch: dict, index: int, dataset: Dataset) -> RecordUpsertSchema:
+    def _batch_index_to_row(self, batch: dict, index: int) -> dict:
+        row = {}
+        for key, values in batch.items():
+            if not key in self.mapping_sources:
```
frascuchon: Maybe we can simplify this by removing the non-mapped columns before start batching:
```python
self.dataset = self.dataset.remove_columns([c for c in self.dataset.column_names if c not in self.mapping_sources])
```

argilla-server/tests/unit/contexts/hub/test_hub_dataset.py
```diff
@@ -127,6 +127,78 @@ async def test_hub_dataset_import_to_with_suggestions(self, db: AsyncSession, mo
...
+            },
+            dataset=dataset,
+        )
+
+        await dataset.awaitable_attrs.fields
+        await dataset.awaitable_attrs.questions
+        await dataset.awaitable_attrs.metadata_properties
+
+        hub_dataset = HubDataset(
+            name="stanfordnlp/imdb",
```
frascuchon, line 154: Maybe we can create a tiny copy of this dataset for tests purposes

argilla-server/src/argilla_server/contexts/hub.py
jfcalvo: Sadly `remove_columns` is not available for `IterableDataset` instances.

frascuchon: :-(

argilla-server/src/argilla_server/contexts/hub.py
```diff
@@ -71,53 +78,68 @@ async def _import_batch_to(
...
+        row = {}
+        for feature_name, values in batch.items():
+            if not feature_name in self.mapping_feature_names:
+                continue
+
+            value = values[index]
+            feature = self.features[feature_name]
+
+            if feature._type == FEATURE_TYPE_CLASS_LABEL:
+                row[feature_name] = feature.int2str(value)
```
frascuchon, line 99: -1 value is used for non-labeled. You should support this. Otherwise the conversion will fail

jfcalvo: Done here: https://github.com/argilla-io/argilla/pull/5607

frascuchon (approved):

PR #5609, opened 2024-10-17
[PERF] argilla server: Reduce general transaction time

frascuchon:
# Description

- Remove unnecessary extra SAVEPOINTS
- Move out of the commit all search index operations
- ~~the refresh parameter for es can be set to `wait_for` instead of `true` since index operations are out of transactions. This will help with the es performance.~~

**Type of change**

- Documentation update

**How Has This Been Tested**

**Checklist**

### DIFF
--- a/argilla-server/src/argilla_server/bulk/records_bulk.py
+++ b/argilla-server/src/argilla_server/bulk/records_bulk.py
@@ -170,7 +170,6 @@ class UpsertRecordsBulk(CreateRecordsBulk):
         self._db.add_all(records)
         await self._db.flush(records)
         await self._upsert_records_relationships(records, bulk_upsert.items)
-
         await distribution.unsafe_update_records_status(self._db, records)
 
         await self._db.commit()
--- a/argilla-server/src/argilla_server/bulk/records_bulk.py
+++ b/argilla-server/src/argilla_server/bulk/records_bulk.py
@@ -48,26 +48,25 @@ class CreateRecordsBulk:
     async def create_records_bulk(self, dataset: Dataset, bulk_create: RecordsBulkCreate) -> RecordsBulk:
         await RecordsBulkCreateValidator.validate(self._db, bulk_create, dataset)
 
-        async with self._db.begin_nested():
-            records = [
-                Record(
-                    fields=jsonable_encoder(record_create.fields),
-                    metadata_=record_create.metadata,
-                    external_id=record_create.external_id,
-                    dataset_id=dataset.id,
-                )
-                for record_create in bulk_create.items
-            ]
+        records = [
+            Record(
+                fields=jsonable_encoder(record_create.fields),
+                metadata_=record_create.metadata,
+                external_id=record_create.external_id,
+                dataset_id=dataset.id,
+            )
+            for record_create in bulk_create.items
+        ]
 
-            self._db.add_all(records)
-            await self._db.flush(records)
-
-            await self._upsert_records_relationships(records, bulk_create.items)
-            await _preload_records_relationships_before_index(self._db, records)
-            await distribution.unsafe_update_records_status(self._db, records)
-            await self._search_engine.index_records(dataset, records)
+        self._db.add_all(records)
+        await self._db.flush(records)
+        await self._upsert_records_relationships(records, bulk_create.items)
+        await distribution.unsafe_update_records_status(self._db, records)
 
         await self._db.commit()
+
+        await _preload_records_relationships_before_index(self._db, records)
+        await self._search_engine.index_records(dataset, records)
 
         return RecordsBulk(items=records)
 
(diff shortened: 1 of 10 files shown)


argilla-server/src/argilla_server/bulk/records_bulk.py
```diff
@@ -151,32 +151,33 @@ async def upsert_records_bulk(self, dataset: Dataset, bulk_upsert: RecordsBulkUp
...
+            elif self._metadata_is_set(record_upsert):
+                record.metadata_ = record_upsert.metadata
+                record.updated_at = datetime.utcnow()
+
+            records.append(record)
+
+        self._db.add_all(records)
+        await self._db.flush(records)
+        await self._upsert_records_relationships(records, bulk_upsert.items)
+
```
frascuchon:
```suggestion
```

argilla-server/src/argilla_server/contexts/datasets.py
```diff
@@ -35,6 +35,8 @@
     Union,
 )
 from uuid import UUID
+
+from attr import attributes
```
frascuchon:
```suggestion
```

argilla-server/src/argilla_server/contexts/datasets.py
```diff
@@ -814,26 +804,26 @@ async def update_record(
...
-                db, objects=vectors, constraints=[Vector.record_id, Vector.vector_settings_id], autocommit=False
+                db=db,
+                objects=vectors,
+                constraints=[Vector.record_id, Vector.vector_settings_id],
+                autocommit=False,
             )
             await db.refresh(record, attribute_names=["vectors"])
 
-        if needs_search_engine_update:
-            await record.dataset.awaitable_attrs.vectors_settings
```
jfcalvo, line 769: You didn't move this line. Is not needed anymore?

argilla-server/src/argilla_server/contexts/distribution.py
```diff
@@ -30,6 +31,12 @@
 
 
 async def unsafe_update_records_status(db: AsyncSession, records: List[Record]):
+    await db.execute(
+        select(Record)
+        .where(Record.id.in_([record.id for record in records]))
+        .options(selectinload(Record.dataset), selectinload(Record.responses_submitted))
+    )
```
jfcalvo, line 38: Why is this needed now?

argilla-server/src/argilla_server/models/mixins.py
```diff
@@ -185,9 +185,9 @@ async def delete(self, db: AsyncSession, autocommit: bool = True) -> Self:
 
     @classmethod
     async def delete_many(
-        cls, db: AsyncSession, params: List["BinaryExpression"], autocommit: bool = True
+        cls, db: AsyncSession, where: List["BinaryExpression"], autocommit: bool = True
```
jfcalvo: We're using `conditions` in other functions:

```suggestion
        cls, db: AsyncSession, conditions: List["BinaryExpression"], autocommit: bool = True
```

argilla-server/src/argilla_server/search_engine/elasticsearch.py
```diff
@@ -123,8 +123,8 @@ async def _delete_index_request(self, index_name: str):
         await self.client.indices.delete(index=index_name, ignore=[404], ignore_unavailable=True)
 
     async def _update_document_request(self, index_name: str, id: str, body: dict):
-        # https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-refresh.html#refresh-api-desc
-        await self.client.update(index=index_name, id=id, **body, refresh=True)
+        # https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-refresh.html
+        await self.client.update(index=index_name, id=id, **body, refresh="wait_for")
```
jfcalvo: Why not use `false` here?

> Take no refresh related actions. The changes made by this request will be made visible at some point after the request returns.

We can wait I guess.

jfcalvo (changes requested):

argilla-server/src/argilla_server/search_engine/elasticsearch.py
frascuchon: By setting `refresh=False` the front-end may fail since they have some internal logic assuming that the records are ready for searches immediately after (for example, after bulk responses).

We should to sync with them to apply the change safely.

argilla-server/src/argilla_server/contexts/datasets.py
frascuchon: Nope. This line was not necessary since this is done inside the `_preload_record_relation_ships_before_index` function

argilla-server/src/argilla_server/contexts/distribution.py
frascuchon: Since we're moving the [_preload_records_relationships_before_index call](https://github.com/argilla-io/argilla/pull/5609/files#diff-8b6a4a23212d32e0db897e0fd54b482d2246be5f98c24915ce4f93923f14e8d5L66) before the search index, the records info must be loaded before the new status is computed. Indeed, this worked because of the side effect of the preload for the index.

jfcalvo (approved):

PR #5613, opened 2024-10-18
feat: add support to use datasets using features with sequences of class labels

jfcalvo:
# Description

* Add casting for features using sequences of class labels (casting them using `int2str` function).
* Casting to string values for suggestions mapped to multi label questions (iterating over the values).

**Type of change**

**How Has This Been Tested**

- [x] Adding additional tests to our suite.

**Checklist**

### DIFF
--- a/argilla-server/src/argilla_server/contexts/hub.py
+++ b/argilla-server/src/argilla_server/contexts/hub.py
@@ -167,7 +167,10 @@ class HubDataset:
                 value = str(value)
 
             if question.is_multi_label_selection:
-                value = [str(v) for v in value]
+                if isinstance(value, list):
+                    value = [str(v) for v in value]
+                else:
+                    value = [str(value)]
 
             if question.is_rating:
                 value = int(value)
--- a/argilla-server/src/argilla_server/contexts/hub.py
+++ b/argilla-server/src/argilla_server/contexts/hub.py
@@ -102,6 +102,8 @@ class HubDataset:
                     row[feature_name] = None
                 else:
                     row[feature_name] = feature.int2str(value)
+            elif isinstance(feature, features.Sequence) and isinstance(feature.feature, features.ClassLabel):
+                row[feature_name] = [feature.feature.int2str(v) for v in value]
             elif isinstance(feature, features.Image) and isinstance(value, Image.Image):
                 row[feature_name] = pil_image_to_data_url(value)
             else:
@@ -163,6 +165,9 @@ class HubDataset:
 
             if question.is_text or question.is_label_selection:
                 value = str(value)
+
+            if question.is_multi_label_selection:
+                value = [str(v) for v in value]
 
             if question.is_rating:
                 value = int(value)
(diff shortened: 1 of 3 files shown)


argilla-server/src/argilla_server/contexts/hub.py
```diff
@@ -102,6 +102,8 @@ def _batch_index_to_row(self, batch: dict, index: int) -> dict:
                     row[feature_name] = None
                 else:
                     row[feature_name] = feature.int2str(value)
+            elif isinstance(feature, features.Sequence) and isinstance(feature.feature, features.ClassLabel):
```
jfcalvo: Here I understand that inside a feature that it's a `Sequence` we have inside another feature. But I'm not 100% sure.

argilla-server/src/argilla_server/contexts/hub.py
```diff
@@ -164,6 +166,9 @@ def _row_suggestions(self, row: dict, dataset: Dataset) -> list:
             if question.is_text or question.is_label_selection:
                 value = str(value)
 
+            if question.is_multi_label_selection:
+                value = [str(v) for v in value]
```
frascuchon: Here we should check if the value is a list. Otherwise, we should "list-ilize" it: `value = [value]`

argilla-server/src/argilla_server/contexts/hub.py
frascuchon: From the [docs](https://huggingface.co/docs/datasets/v3.0.1/en/package_reference/main_classes#datasets.Sequence):

> If your data type contains a list of objects, then you want to use the [Sequence](https://huggingface.co/docs/datasets/v3.0.1/en/package_reference/main_classes#datasets.Sequence) feature. Remember the SQuAD dataset?
>
>```python
>from datasets import load_dataset
>
>dataset = load_dataset('squad', split='train')
>dataset.features
>{'answers': Sequence(feature={'text': Value(dtype='string', id=None), 'answer_start': Value(dtype='int32', id=None)}, length=-1, id=None),
>'context': Value(dtype='string', id=None),
>'id': Value(dtype='string', id=None),
>'question': Value(dtype='string', id=None),
>'title': Value(dtype='string', id=None)}
>```
>The answers field is constructed using the Sequence feature because it contains two subfields, text and >answer_start, which are lists of string and int32, respectively. 

The feature could be whatever. But this condition is quite enough.

jfcalvo: Ok, then my code it's fine.

argilla-server/src/argilla_server/contexts/hub.py
jfcalvo: Done.

argilla-server/src/argilla_server/contexts/hub.py
```diff
@@ -102,6 +102,8 @@ def _batch_index_to_row(self, batch: dict, index: int) -> dict:
                     row[feature_name] = None
                 else:
                     row[feature_name] = feature.int2str(value)
+            elif isinstance(feature, features.Sequence) and isinstance(feature.feature, features.ClassLabel):
+                row[feature_name] = [feature.feature.int2str(v) for v in value]
```
frascuchon: we need to tackle the -1 also for sequence of class labels

frascuchon (approved):

PR #5616, opened 2024-10-21
feat: add dataset split to be used along with row idx when `external_id` is not provided on mapping

jfcalvo:
# Description

This PR add the dataset imported split to be used as `external_id` when there is no value for `external_id` specified on the import mapping.

If importing the split `train` for a dataset and no `external_id` is provided the `external_id` will be calculated like the following:
- `train_0`: first row of `train` split.
- `train_1`: second row of `train` split.
- ...

With this we are avoiding row duplications when another split is imported to the same dataset. So if later we import the `test` split for the same dataset we will have for `external_id`:
- `train_0`: first row of `train` split.
- `train_1`: second row of `train` split.
- ...
- `test_0`: first row of `test` split.
- `test_1`: second row of `test` split.
- ...

**Type of change**

**How Has This Been Tested**

- [x] Adding additional tests.

**Checklist**


frascuchon (approved):
Ref https://github.com/argilla-io/argilla/pull/5617

PR #5628, opened 2024-10-24
[IMPROVEMENT] feat(helm): add support for default storage class in PVCs

dme86:
# Description

I updated the Helm chart to allow the use of a default storage class when none is explicitly specified. This change is important because some Kubernetes environments, like k3d, do not have a 'standard' storage class by default.
Without this change, the setup could fail in such environments, as Kubernetes would attempt to use an undefined or missing storage class. By not explicitly setting a storage class, I allow Kubernetes to default to whatever storage class is configured for the cluster.

This ensures the chart works across various setups, including those that don't predefine a 'standard' storage class.

**Type of change**

**How Has This Been Tested**

Tested on a MacBook Pro M3 Max with k3d (v5.7.4) and Minikube (v1.34.0).

**Checklist**:

I added relevant documentation
I followed the style guidelines of this project
I did a self-review of my code
I made corresponding changes to the documentation
I confirm My changes generate no new warnings
I have added tests that prove my fix is effective or that my feature works
I have added relevant notes to the CHANGELOG.md file (See https://keepachangelog.com/)

### DIFF
--- a/argilla-server/CHANGELOG.md
+++ b/argilla-server/CHANGELOG.md
@@ -19,6 +19,7 @@ These are the section headers that we use:
 ### Changed
 
 - Changed supported values for terms metadata options to accept other than strings values. ([#5589](https://github.com/argilla-io/argilla/pull/5589))
+- Changed Helm chart to allow the use of a default storage class when none is explicitly specified. This ensures compatibility with environments like k3d that do not have a 'standard' storage class by default.
 
 ### Removed
 
--- a/examples/deployments/k8s/argilla-chart/tests/suite_test.yaml
+++ b/examples/deployments/k8s/argilla-chart/tests/suite_test.yaml
@@ -86,7 +86,6 @@ tests:
       argilla.persistence.enabled: true
       argilla.persistence.accessMode: ReadWriteOnce
       argilla.persistence.size: 5Gi
-      argilla.persistence.storageClass: standard
     templates:
       - templates/pvc.yaml
     asserts:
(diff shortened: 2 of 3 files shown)


frascuchon:
Thanks @dme86 for your contribution!

Sorry, I'm not a Kubernetes expert but would like to understand the change :-). This change will use the default configured storage class instead of the "standard" storage class which could not exist for some clusters, but users still can define [a specific storage class](https://github.com/dme86/argilla/blob/feature/default-storageclass-support/examples/deployments/k8s/argilla-chart/templates/pvc.yaml#L15), right?

dme86:
@frascuchon,

You're absolutely correct in your understanding. This change updates the Helm chart to use the cluster's **default** storage class _instead of explicitly specifying_ the "standard" storage class, which does not exist on all Kubernetes clusters. By doing so, we enhance the chart's compatibility across different environments.

Users still have the **flexibility** to define a specific storage class if needed. They can do this by setting the `argilla.persistence.storageClass` value in the `values.yaml` file **or** by using the `--set` flag during installation. For example:
```shell
helm install my-release . --set argilla.persistence.storageClass=custom-storage-class
```

I've noticed that the current Helm chart **fails** on clusters like `k3d` and `k3s`, where the default storage class is "**local-path**" rather than "**standard**". This mismatch causes deployment failures because the specified "standard" storage class doesn't exist in these environments. By defaulting to the cluster's default storage class, we ensure smoother deployments across a wider range of Kubernetes distributions.

Given that many users utilize `k3d`, `k3s`, or other lightweight Kubernetes solutions, I recommend testing the Helm chart beyond just minikube. This approach will help you make deployments as seamless and user-friendly as possible for a broader audience.

frascuchon:
Thanks for the clarification, @dme86!

frascuchon (approved):

PR #5666, opened 2024-11-04
[REFACTOR] `argilla server`: using pydantic v2

### LINKED ISSUE #4935: [CONFIG] Resolve python requirements for python version and dependencies with server.

frascuchon:
# Description

An alternative work of pydantic v2 upgrade with the latest changes from develop.

The main problems I found:
- The Optional values are now required. This will affect how schemas are returned since default values are sets, including them as part of the schema result (record without responses or suggestions, for example)
- I found some serialization errors when the response body (an error) contains a ValueError object. The same is working with pydantic <2
- Automatic type conversion is not working anymore. This means that many client requests may fail ( values are not automatically converted to str, or some complex structures (record fields may be affected by this). We can still define custom model_validators, but we must apply them carefully.

Closes https://github.com/argilla-io/argilla/issues/4935

Refs https://github.com/argilla-io/argilla/pull/5508

**Type of change**

**How Has This Been Tested**

**Checklist**


argilla-server/.env.dev
```diff
@@ -1,2 +1,2 @@
 ALEMBIC_CONFIG=src/argilla_server/alembic.ini
-ARGILLA_DATABASE_URL=sqlite+aiosqlite:///${HOME}/.argilla/argilla.db?check_same_thread=False
+ARGILLA_DATABASE_URL=sqlite+aiosqlite:///${HOME}/.argilla/argilla.dev.db?check_same_thread=False
```
jfcalvo: For `.env.test` we have the following:

```
ARGILLA_DATABASE_URL=sqlite+aiosqlite:///${HOME}/.argilla/argilla-test.db?check_same_thread=False
```

So I guess we should do the same here:

```suggestion
ARGILLA_DATABASE_URL=sqlite+aiosqlite:///${HOME}/.argilla/argilla-dev.db?check_same_thread=False
```

argilla-server/pyproject.toml
```diff
@@ -22,7 +22,8 @@ maintainers = [{ name = "argilla", email = "contact@argilla.io" }]
 dependencies = [
     # Basic dependencies
     "fastapi ~= 0.115.0",
-    "pydantic ~= 1.10.18",
+    "pydantic ~= 2.9.0",
```
jfcalvo, line 24: Can we upgrade to the latest version?

```suggestion
    "pydantic ~= 2.9.2",
```

Don't forget to do `pdm install` after change this, so the lock file is updated.

argilla-server/pyproject.toml
```diff
@@ -22,7 +22,8 @@ maintainers = [{ name = "argilla", email = "contact@argilla.io" }]
...
+    "pydantic-settings ~= 2.6.0",
```
jfcalvo, line 25: Something similar for `pydantic-settings`:

```suggestion
    "pydantic-settings ~= 2.6.1",
```

Don't forget to do `pdm install` after change this, so the lock file is updated.

argilla-server/src/argilla_server/api/schemas/v1/chat.py
```diff
@@ -12,16 +12,15 @@
...
 MAX_MESSAGE_LENGTH = 20000
 
 MIN_ROLE_LENGTH = 1
 MAX_ROLE_LENGTH = 20
-MAX_ROLE_REGEX = r"^\S+$"
 
 
 class ChatFieldValue(BaseModel):
-    role: str = Field(..., min_role_length=MIN_ROLE_LENGTH, max_length=MAX_ROLE_LENGTH, regex=MAX_ROLE_REGEX)
-    content: str = Field(..., min_message_length=MIN_MESSAGE_LENGTH, max_length=MAX_MESSAGE_LENGTH)
```
jfcalvo, line 27: This means we didn't have a test checking this right (for empty strings)? Should we add one?

argilla-server/src/argilla_server/api/schemas/v1/suggestions.py
```diff
@@ -87,15 +87,28 @@ class SuggestionCreate(BaseSuggestion):
...
         description="Agent used to generate the suggestion",
     )
-    score: Optional[Union[float, List[float]]] = Field(
-        None,
-        min_items=SCORE_MIN_ITEMS,
-        ge=SCORE_GREATER_THAN_OR_EQUAL,
-        le=SCORE_LESS_THAN_OR_EQUAL,
-        description="The score assigned to the suggestion",
-    )
+    score: Optional[Union[float, List[float]]] = Field(None, description="The score assigned to the suggestion")
```
jfcalvo, line 95: Instead of adding custom validations maybe we can define custom types like `SuggestionScore`. 

Take a look to what I did here: https://github.com/argilla-io/argilla/pull/5508/files#diff-a9277ada64762a8062c1e132567edba6ce2aec3e279d1ded0452b9557bfd6b26

argilla-server/src/argilla_server/api/schemas/v1/users.py
```diff
@@ -35,16 +36,15 @@ class User(BaseModel):
...
 
 
 class UserCreate(BaseModel):
-    first_name: constr(min_length=1, strip_whitespace=True)
-    last_name: Optional[constr(min_length=1, strip_whitespace=True)]
     username: str = Field(..., min_length=1)
-    role: Optional[UserRole]
     password: str = Field(min_length=USER_PASSWORD_MIN_LENGTH, max_length=USER_PASSWORD_MAX_LENGTH)
+    first_name: constr(min_length=1, strip_whitespace=True)
+    last_name: Optional[constr(min_length=1, strip_whitespace=True)] = None
```
jfcalvo, line 46: Maybe we should start using annotated `str` with [`StringConstraints`](https://docs.pydantic.dev/latest/api/types/#pydantic.types.StringConstraints) instead of using the old `constr`.

This applies to all the `constr` used in the PR.

So something like:

```python
last_name: Optional[Annotated[str, StringConstraints(min_length=1, strip_whitespace=True)]] = None
```

You can take a look to https://github.com/argilla-io/argilla/pull/5508/files#diff-0e72eee9f8cea03b100e007fe81d5a97c5ebea93955187f083bb908a4cee2e29R48

argilla-server/src/argilla_server/api/schemas/v1/workspaces.py
```diff
@@ -16,7 +16,8 @@
 from typing import List
 from uuid import UUID
 
-from argilla_server.pydantic_v1 import BaseModel, Field
+from pydantic import BaseModel, Field, ConfigDict
+from starlette.config import Config
```
jfcalvo: Why is this import needed?

jfcalvo (changes requested):
Masive effort! 💪

argilla-server/pyproject.toml
frascuchon: We don't need to point to the latest patch release. The `~=` operator will match the latest compatible version. See [here](https://peps.python.org/pep-0440/#compatible-release) for more info

argilla-server/src/argilla_server/api/schemas/v1/chat.py
frascuchon: Done. See the commit e0e95381b

argilla-server/src/argilla_server/api/schemas/v1/suggestions.py
frascuchon: I tried this, but I found some weird behavior when running tests. I decided on a custom validator since it gives more control.

argilla-server/src/argilla_server/api/schemas/v1/users.py
frascuchon: Yeah. I was expecting some deprecation warnings, but `constr` and others run like a charm without any warning. I suggest keeping them as is so as not to add more changes.

argilla-server/src/argilla_server/api/schemas/v1/workspaces.py
frascuchon: WE don't need it. I'll remove it.

argilla-server/src/argilla_server/contexts/datasets.py
```diff
@@ -266,7 +266,7 @@ async def create_metadata_property(
         db,
         name=metadata_property_create.name,
         title=metadata_property_create.title,
-        settings=metadata_property_create.settings.dict(),
+        settings=metadata_property_create.settings.model_dump(),
```
jfcalvo, line 269: @frascuchon I added `model_dump` thinking that it's gonna work but after taking another look the settings is an annotated type not using `model_config = ConfigDict(from_attributes=True)` so I reverted this change in another commit.

frascuchon: Not sure to understand this. The `from_attributes` key is not affecting the `model_dump` method. What's the problem, exactly?

jfcalvo (approved):

PR #5667, opened 2024-11-04
fix: improve logic for detecting ChatFields

leiyre:
Check the role and content properties to validate the chat field


argilla-frontend/v1/domain/entities/hub/Subset.ts
```diff
@@ -42,6 +45,8 @@ export class Subset {
               name: key,
               kindObject: value._type,
               type: value.dtype,
+              role,
+              content,
```
frascuchon, line 49: We're losing info by keeping only the first entry of value on [this line](https://github.com/argilla-io/argilla/pull/5667/files#diff-fd0b434b14fcae8f230919f19ecd1bdec8fe42dca827438a2c6b6b848e34e9b3L39) (which is a dictionary of feature). This can introduce errors later. (Not for handling on this PR) cc @damianpumar

argilla-frontend/v1/domain/entities/hub/Subset.ts
```diff
@@ -7,6 +7,8 @@ import { QuestionCreation } from "./QuestionCreation";
 type Structure = {
   name: string;
   options?: string[];
+  role?: string;
+  content?: string;
```
frascuchon, line 11: I don't fully understand this data structure but I think `role` and `content` are quite specific to be part of a general structure. Anyway, this is something to review with @damianpumar.

We can go on with this solution if everything is working well.

frascuchon (approved):

PR #5670, opened 2024-11-05
Dataset Creation UI fixes & Improvements

leiyre:
- [x] Rating values validate min / max
- [x] Question name format and renaming
- [x] Minor style adjustments 
- [x] Show required / not required in the form preview
- [x] Redirect from /new/repoId to /repoId
- [x] Redirect to error page when repoId is invalid


argilla-frontend/v1/infrastructure/repositories/HubRepository.ts
```diff
@@ -19,8 +23,11 @@ export class HubRepository {
       );
 
       return data.dataset_info;
-    } catch {
-      return {};
+    } catch (e) {
+      throw {
+        status: e.response.status,
```
frascuchon: Can we map the status to 404 if the error status is 401? As a user is a bit weird to see the big 401 number on the error page. We can iterate later on this when supporting private datasets, where the 401 makes more sense.
Captura de pantalla 2024-11-06 a las 16 17 10

argilla-frontend/components/features/dataset-creation/configuration/DatasetConfigurationForm.vue
```diff
@@ -139,21 +137,21 @@ export default {
...
     };
   },
   methods: {
     createDataset() {
       this.create(this.dataset);
     },
     addQuestion(type) {
-      const questionName = `${type} ${this.dataset.selectedSubset.questions.length}`;
+      const name = this.$t(`config.questionId.${type}`);
+      const questionName = `${name}-${this.dataset.selectedSubset.questions.length}`;
```
frascuchon: I think the idea in the discussion was to use the `_` instead of `-`. Let's make the change if there are no technical limitations to that. 

```suggestion
      const questionName = `${name}_${this.dataset.selectedSubset.questions.length}`;
```

argilla-frontend/v1/infrastructure/repositories/HubRepository.ts
leiyre: Sure!

frascuchon: Also, I think we should define a copy message instead of the `ERROR_FETCHING_DATASET`. What about "Cannot fetch the dataset"?

argilla-frontend/components/features/dataset-creation/configuration/DatasetConfigurationForm.vue
```diff
@@ -144,7 +145,7 @@ export default {
       return Math.max(
         ...this.dataset.selectedSubset.questions.map((question) => {
           const numberInName = question.name.split("_").pop();
-          return parseInt(numberInName);
+          return parseInt(numberInName) || 0;
```
frascuchon, line 147: How can a mapped column named "column_10" influence this?

leiyre: In that case the next created question name will be "questionType_11"

frascuchon (approved):

PR #5649, opened 2024-10-29
[CONFIG] `argilla server`: Review and update dependencies

frascuchon:
# Description

- [x] Use the newer version of python 
- [x] Use the latest version of core dependencies. (without code changes)
- [x] Upgrade docker images
- [x] Remove unnecessary dependencies.

**Type of change**

- Refactor (change restructuring the codebase without changing functionality)
- Improvement (change adding some improvement to an existing functionality)

**How Has This Been Tested**

**Checklist**

- I added relevant documentation
- I followed the style guidelines of this project
- I did a self-review of my code
- I made corresponding changes to the documentation
- I confirm My changes generate no new warnings
- I have added tests that prove my fix is effective or that my feature works
- I have added relevant notes to the CHANGELOG.md file (See https://keepachangelog.com/)

PR #5672, opened 2024-11-05
[BUGFIX]: argilla server: install default `psycopg2` driver used by alembic

frascuchon:
# Description

Alembic will use psycopy2 driver by default. We still need to install it  to avoid startup errors.

**Type of change**

**How Has This Been Tested**

**Checklist**

### DIFF
--- a/argilla-server/pdm.lock
+++ b/argilla-server/pdm.lock
@@ -5,7 +5,7 @@
 groups = ["default", "postgresql", "test"]
 strategy = ["inherit_metadata"]
 lock_version = "4.5.0"
-content_hash = "sha256:19b06d68785fa57090938b31dca41ba61e2919cad8cda6d728cb9f73d42f3d65"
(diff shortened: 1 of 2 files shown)


frascuchon:
Tested is working on the dev environment

jfcalvo:
@frascuchon looks like `psycopg` aka psycopg 3 is the way to go (instead of `psycopg2`). It's a modern rewritten of the original `psycopg2`. In fact looks like all benchmarks shows that it it fast: https://www.timescale.com/blog/psycopg2-vs-psycopg3-performance-benchmark/

frascuchon:
Yes. But that means changing all the default and env vars when using PostgreSQL. If no driver is defined in the SQLAlchemy database UR when using PostgreSQL, psycopg2 will be used by default, falling into a missing package error. I created this fix to avoid reverting all the dependencies upgrades included on the PR https://github.com/argilla-io/argilla/pull/5649

PR #5679, opened 2024-11-07
[BUGFIX] `argilla`: prevent enum literal validation errors

frascuchon:
# Description

In some environments, working with enun instances as string values may lead to some validation errors. See [this discord discussion](https://discord.com/channels/879548962464493619/1300936213531983882/1300936213531983882). 

This PR reviews how metadata settings classes are defined, and defines type directly as a literal string, in the same way we do fo questions or fields, instead of using the MetadataPropertyType class.

**Type of change**

**How Has This Been Tested**

**Checklist**


argilla/src/argilla/settings/_metadata.py
```diff
@@ -122,13 +121,12 @@ def __init__(
         super().__init__(client=client)
 
         try:
-            settings = TermsMetadataPropertySettings(values=options, type=MetadataPropertyType.terms)
+            settings = TermsMetadataPropertySettings(values=options)
```
frascuchon, line 124: We don't need the type here anymore. Each setting instance defines its type.

argilla/src/argilla/settings/_metadata.py
```diff
@@ -176,13 +174,12 @@ def __init__(
...
 
         try:
-            settings = FloatMetadataPropertySettings(min=min, max=max, type=MetadataPropertyType.float)
+            settings = FloatMetadataPropertySettings(min=min, max=max)
         except ValueError as e:
             raise MetadataError(f"Error defining metadata settings for {name}") from e
 
         self._model = MetadataFieldModel(
             name=name,
-            type=MetadataPropertyType.float,
```
frascuchon, line 185: This type is redundant. Model should use the one defined under the `settings` attribute

argilla/src/argilla/_models/_settings/_metadata.py
```diff
@@ -22,19 +21,12 @@
...
-
-
 class BaseMetadataPropertySettings(BaseModel):
-    type: MetadataPropertyType
     visible_for_annotators: Optional[bool] = True
 
 
 class TermsMetadataPropertySettings(BaseMetadataPropertySettings):
-    type: Literal[MetadataPropertyType.terms]
+    type: Literal["terms"] = "terms"
```
jfcalvo, line 29: Is necessary to define a default? Will something like the following works?:

```suggestion
    type: Literal["terms"]
```

jfcalvo, line 29: Do you think we can create constants (now that Enums are not possible) to define the types?

Something like `METADATA_PROPERTY_TYPE_TERMS = "terms"` or similar.

jfcalvo (approved):

argilla/src/argilla/_models/_settings/_metadata.py
frascuchon: Yes. We need to define a default. Otherwise, the model expects a literal value for the attribute. 
https://github.com/pydantic/pydantic/issues/6346

frascuchon: I think it's not needed

PR #5689, opened 2024-11-13
[REFACTOR][ENHANCEMENT] `argilla server`: better OAuth2 integration

frascuchon:
# Description

This PR improves the `social_core` integration, making setup backends simpler.

Improvements:
- Reading `SOCIAL_AUTH_*` environment variables
- Remove unnecessary settings attribute `enabled`
- Add more backends with no-code: `extra_backends`* (some backends may be not fully supported, but users can create issues for that)

**Type of change**

- Refactor (change restructuring the codebase without changing functionality)
- Improvement (change adding some improvement to an existing functionality)
- Documentation update

**How Has This Been Tested**

**Checklist**

- I added relevant documentation
- I followed the style guidelines of this project
- I did a self-review of my code
- I made corresponding changes to the documentation
- I confirm My changes generate no new warnings
- I have added tests that prove my fix is effective or that my feature works
- I have added relevant notes to the CHANGELOG.md file (See https://keepachangelog.com/)

PR #5691, opened 2024-11-15
Feature - SSO keycloak

paulbauriegel:
# Introduces a new SSO option using Keycloak

Enables a different SSO provider next to HuggingFace SSO

**Type of change**
- New feature (non-breaking change which adds functionality)

**How Has This Been Tested**
Local build of front-end  & backend. Keycloak deployment as described in the docs

**Checklist**
- I added relevant documentation
- I followed the style guidelines of this project
- I did a self-review of my code
- I made corresponding changes to the documentation
- I confirm My changes generate no new warnings
- I have added tests that prove my fix is effective or that my feature works
- TODO - I have added relevant notes to the CHANGELOG.md file (See https://keepachangelog.com/)

**How to test & use it**
- Follow the instructions for Keycloak: docs/_source/tutorials_and_integrations/integrations/sso_keycloak.md
- Build both front & backend 
- Then it should look like this:
Argilla_SSO

PR #5694, opened 2024-11-18
[DOCS] OAuth2 configuration

frascuchon:
# Description

Docs for changes included in https://github.com/argilla-io/argilla/pull/5689

**Type of change**

- Documentation update

**How Has This Been Tested**

**Checklist**

### DIFF
--- a/argilla/docs/reference/argilla-server/oauth2_configuration.md
+++ b/argilla/docs/reference/argilla-server/oauth2_configuration.md
@@ -1,12 +1,13 @@
 # OAuth2 configuration
 
-Argilla supports OAuth2 authentication for users. This allows users to authenticate using other services like Google, GitHub, or Hugging Face.
-Next sections will guide you through the configuration of the OAuth2 authentication.
+Argilla supports OAuth2 authentication for users. This allows users to authenticate using other services like Google,
+GitHub, or Hugging Face. Next sections will guide you through the configuration of the OAuth2 authentication.
 
 ## The OAuth2 configuration file
 
-The OAuth2 configuration file is a YAML file that contains the configuration for the OAuth2 providers that you want to enable.
-The default file name is `.oauth.yml` and it should be placed in the root directory of the Argilla server. You can also specify a different file name using the `ARGILLA_OAUTH_CONFIG_FILE` environment variable.
+The OAuth2 configuration file is a YAML file that contains the configuration for the OAuth2 providers that you want to
+enable. The default file name is `.oauth.yml` and it should be placed in the root directory of the Argilla server. You
+can also specify a different file name using the `ARGILLA_AUTH_OAUTH_CFG` environment variable.
 
 The file should have the following structure:
 
--- a/argilla/docs/reference/argilla-server/oauth2_configuration.md
+++ b/argilla/docs/reference/argilla-server/oauth2_configuration.md
@@ -78,7 +78,7 @@ The redirect URI should be set to the Argilla server URL, followed by `/oauth/<p
 if the Argilla server is running on `http://localhost:8000`, the redirect URI for provider application should
 be `http://localhost:8000/oauth/huggingface/callback`.
 
-##  Hugging Face OAuth2 configuration
+###  Hugging Face OAuth2 configuration
 
 Argilla supports Hugging Face OAuth2 authentication out of the box, and is already configured when running Argilla
 on Hugging Face Spaces (See the [Hugging Face Spaces settings](../../getting_started/how-to-configure-argilla-on-huggingface.md) for more information).
@@ -103,7 +103,7 @@ settings page.
 The minimal scope required for the Hugging Face OAuth2 provider is `openid profile`, so you don't need to
 change the `scope` when creating the application.
 
-## GitHub OAuth2 configuration
+### GitHub OAuth2 configuration
 
 Argilla also supports GitHub OAuth2 authentication out of the box. To configure the GitHub OAuth2 provider, you should
 define the following fields in the `.oauth.yml` file:
@@ -120,7 +120,7 @@ providers:
 
 To get your client ID and client secret, you need to register a new [OAuth application](https://github.com/settings/applications/new) in the GitHub settings page.
 
-## Google OAuth2 configuration
+### Google OAuth2 configuration
 
 Argilla also supports Google OAuth2 authentication out of the box. To configure the Google OAuth2 provider, you
 should define the following fields in the `.oauth.yml` file:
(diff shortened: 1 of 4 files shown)


argilla/docs/reference/argilla-server/oauth2_configuration.md
```diff
@@ -0,0 +1,134 @@
...
+Next sections will guide you through the configuration of the OAuth2 authentication.
+
+## The OAuth2 configuration file
+
+The OAuth2 configuration file is a YAML file that contains the configuration for the OAuth2 providers that you want to enable.
+The default file name is `.oauth.yml` and it should be placed in the root directory of the Argilla server. You can also specify a different file name using the `ARGILLA_OAUTH_CONFIG_FILE` environment variable.
+
+The file should have the following structure:
+
+```yaml
```
davidberenstein1957, line 14: should we include https://github.com/argilla-io/argilla/pull/5691?

argilla/docs/reference/argilla-server/oauth2_configuration.md
```diff
@@ -0,0 +1,134 @@
+# OAuth2 configuration
+
+Argilla supports OAuth2 authentication for users. This allows users to authenticate using other services like Google, GitHub, or Hugging Face.
+Next sections will guide you through the configuration of the OAuth2 authentication.
+
+## The OAuth2 configuration file
+
+The OAuth2 configuration file is a YAML file that contains the configuration for the OAuth2 providers that you want to enable.
+The default file name is `.oauth.yml` and it should be placed in the root directory of the Argilla server. You can also specify a different file name using the `ARGILLA_OAUTH_CONFIG_FILE` environment variable.
```
davidberenstein1957: I think this should be added to the server env vars.

argilla/docs/reference/argilla-server/oauth2_configuration.md
```diff
@@ -0,0 +1,134 @@
...
+```
+
+### Providers
+
+The `providers` key is a list of dictionaries, each dictionary represents a provider configuration, including the following fields:
+
+- `name`: The name of the provider. The available options by default are `huggingface`, `github` and `google-oauth2`. We will see later how to add more providers not supported by default.
+- `client_id`: The client ID provided by the OAuth2 provider. You can get this value by creating an application in the provider's developer console. This is a required field, but you can also use the `ARGILLA_OAUTH2_<PROVIDER_NAME>_CLIENT_ID` environment variable to set the value.
+- `client_secret`: The client secret provided by the OAuth2 provider. You can get this value by creating an application in the provider's developer console. This is a required field, but you can also use the `ARGILLA_OAUTH2_<PROVIDER_NAME>_CLIENT_SECRET` environment variable to set the value.
+- `scope`: The scope of the OAuth2 provider. This is an optional field, and normally you don't need to set it, but you can use it to request specific permissions from the user access.
```
davidberenstein1957: env var updates?

argilla/docs/reference/argilla-server/oauth2_configuration.md
```diff
@@ -0,0 +1,134 @@
...
+
+### Allowed Workspaces
+
+The `allowed_workspaces` key is a list of dictionaries, each dictionary represents a workspace configuration. For now, only the `name` field is defined, and it should be the name of the workspace that the user will be assigned to when logging in.
```
davidberenstein1957: what do you mean by "a list of dictionaries"? It seems the we just provide a list of `-name` right? or are we missing some info her"

argilla/docs/reference/argilla-server/oauth2_configuration.md
```diff
@@ -0,0 +1,134 @@
...
+
+```yaml
+
+providers:
+  - name: huggingface
+    client_id: "<client_id>" # You can use the ARGILLA_OAUTH2_HUGGINGFACE_CLIENT_ID environment variable
+    client_secret: "<client_secret>" # You can use the ARGILLA_OAUTH2_HUGGINGFACE_CLIENT_SECRET environment variable
+    scope: "openid profile" # This field is optional. But this value must be aligned your OAuth2 application created in Hugging Face.
+
+allowed_workspaces:
```
davidberenstein1957: the part of allowed workspaces is unrelated right?

argilla/docs/reference/argilla-server/oauth2_configuration.md
davidberenstein1957: what happens if we don't use allowed workspaces? will it fail or will people just get access to their own workspaces?

argilla/docs/reference/argilla-server/oauth2_configuration.md
davidberenstein1957: perhaps we can ask the creator of this PR for input?

argilla/docs/reference/argilla-server/oauth2_configuration.md
frascuchon: Those env vars are dynamic and will depend on which providers are defined. I think having a specific section for OAuth describing the process using env vars is easier to understand. Otherwise, we can have a growing server env section with a lot of env vars without clear context.

argilla/docs/reference/argilla-server/oauth2_configuration.md
frascuchon: > what do you mean by "a list of dictionaries"? It seems the we just provide a list of -name right? or are we missing some info her"

Yes. I'll review that part.

> what happens if we don't use allowed workspaces? will it fail or will people just get access to their own workspaces?

It will depend on the user's role. For owner roles using OAuth, the access will be fine even if there is no workspaces defined. Maybe we should force at least one allowed workspace.

davidberenstein1957: So, people with`write` access to an organisation to get the `owner` role and `admin` access the `admin` role? 

How do we deal with access from people that are not part of the organisation?

I think not providing a workspace should still allow people to get access to their own workspace, or does this introduce difficulties?

frascuchon: The user role mapping based on the user permission for an Org hadn't been implemented.  We still have some technical issues when accessing other info than the basic profile (commented on this [discord thread](https://discord.com/channels/879548962464493619/1304551384913281095/1305453111418163281)). For general OAuth integrations using out of an HF space, a user will defined as the owner only if the username matches with the [`USERNAME` variable](https://docs.argilla.io/latest/reference/argilla-server/configuration/#authentication).  The rest of the users will be defined as annotators, and the allowed_workspaces will define the workspaces where those "annotators" have access.

argilla/docs/reference/argilla-server/oauth2_configuration.md
```diff
@@ -0,0 +1,164 @@
...
+secret.
+
+A common step when creating an application in the provider's developer console is to set the redirect URI. The
+redirect URI is the URL where the OAuth2 provider will redirect the user after the authentication process.
+
+The redirect URI should be set to the Argilla server URL, followed by `/oauth/<provider_name>/callback`. For example,
+if the Argilla server is running on `http://localhost:8000`, the redirect URI for provider application should
+be `http://localhost:8000/oauth/huggingface/callback`.
+
+##  Hugging Face OAuth2 configuration
```
sdiazlor: Should they be h3?

sdiazlor (approved):

PR #5711, opened 2024-11-22
Add keycloak SSO

paulbauriegel:
# Add keycloak SSO 

Based on discussion in https://github.com/argilla-io/argilla/pull/5691

Points that need some feedback:
- A lot of configurations are set via env variables now. Not sure if that's ideal, error messages if something is not set correctly can be rather cryptic with social auth lib
- I added the Keycloak logo to the Oauth button id the provider is keycloak, generally the same could also be done for the HF logo not having a separate button
- Is the documentation to set-up a keycloak server sufficient?

**Type of change**
- Documentation update

**How Has This Been Tested**
Local build & Keycloak installation as described in the documentation.

**Checklist**

- TODO I have added relevant notes to the CHANGELOG.md file (See https://keepachangelog.com/)


davidberenstein1957:
Related https://github.com/argilla-io/argilla/pull/5691

argilla-server/src/argilla_server/security/authentication/oauth2/_backends.py
```diff
@@ -58,6 +58,7 @@ def load_supported_backends(extra_backends: list = None) -> Dict[str, Type[BaseO
         "argilla_server.security.authentication.oauth2._backends.HuggingfaceOpenId",
         "social_core.backends.github.GithubOAuth2",
         "social_core.backends.google.GoogleOAuth2",
+        "social_core.backends.keycloak.KeycloakOAuth2",
```
frascuchon: Maybe one thing that you can do to avoid weird configuration errors is to implement a custom KeyCloack backend as you do [here](https://github.com/argilla-io/argilla/pull/5691/files#diff-6b7a2afb57a1c291ec1384c39b7982276c1ea8ac81b0bc3dce7f02b402d9be26R28) which adds some specific validations, or env reading on init. 

This can help users to find solutions to configuration problems.

frascuchon: I will push a commit with my proposal here including also changes commented in https://github.com/argilla-io/argilla/pull/5691#issuecomment-2503963879

argilla-server/src/argilla_server/security/authentication/oauth2/_backends.py
```diff
@@ -48,6 +48,26 @@ class HuggingfaceOpenId(OpenIdConnectAuth):
     DEFAULT_SCOPE = ["openid", "profile"]
 
 
+class KeycloakOpenId(OpenIdConnectAuth):
+    """Huggingface OpenID Connect authentication backend."""
+
+    name = "keycloak"
+
+    def oidc_endpoint(self) -> str:
```
frascuchon, line 57: I think we only need this one. The rest of the required values (auth URL or access token URL) should be returned by the oicd_config endpoint based on the oidc endpoint

paulbauriegel: Yes this works, I updated the docs accordingly. Now configuring keycloak would work by just setting the `SOCIAL_AUTH_OIDC_ENDPOINT` environment variable in addition to the oauth yaml.

paulbauriegel:
I tested it a bit with a local set-up but it does not seem to work. I need some time to debug it, to understand why it is not working.

frascuchon:
> I tested it a bit with a local set-up but it does not seem to work. I need some time to debug it, to understand why it is not working.

I've pushed some missing tests to see the expected values https://github.com/argilla-io/argilla/blob/bf2a0f64a991731602cd521188dd95bb3896d7ea/argilla-server/tests/unit/api/handlers/v1/test_oauth2.py

bulatovv:
Just wanted to check if you need any help with this? Really looking forward to this feature

paulbauriegel:
@bulatovv @frascuchon I just started looking at it again after being on vacation for a while. What I see is the following behaviour:

- Existing users even if their role is changed have no changed role in the Argilla UI

- New users that are already added with new roles have a different problems and cannot login the first time getting a `HTTP_422_UNPROCESSABLE_ENTITY` the first time and with the second login time are just a annotator with basic rights

I'm looking at those things right now. Generally speaking the simple authentification with Keycloak works, what does not work are the roles & workspace assignments

paulbauriegel:
@frascuchon I updated the code for the cases that where failing for me:
1. Multiple conflicting roles set in keycloak
2. Creating users w. roles for which the workspaces do not exist
3. Changing the argilla role or workspace after the creation of the user

could you check specifically in the oauth2.py if the changes against the DB are correctly implemented.

frascuchon:
Thanks, @paulbauriegel, and sorry for the late response. I need to review your changes in deep since some of them could comprise other OAuth flows. Maybe the safer way would be a specific step syncing SSO roles and workspaces with the argilla DB.

paulbauriegel:
@frascuchon Can I help in any way to further improve this MR?. We are really needing this feature on our side to comply with different security policies. I'm not sure what you mean with:

> Maybe the safer way would be a specific step syncing SSO roles and workspaces with the argilla DB.

What would be the way to implement this? A separate function or something else?

argilla-server/src/argilla_server/security/authentication/userinfo.py
```diff
@@ -36,11 +36,19 @@ def username(self) -> str:
     def first_name(self) -> str:
         return self.get("first_name") or self.username
 
+    @property
+    def last_name(self) -> str:
+        return self.get("last_name") or ""
```
frascuchon:
```suggestion
        return self.get("last_name") or None
```

argilla-server/src/argilla_server/security/authentication/userinfo.py
```diff
@@ -36,11 +36,19 @@ def username(self) -> str:
     def first_name(self) -> str:
         return self.get("first_name") or self.username
 
+    @property
+    def last_name(self) -> str:
```
frascuchon:
```suggestion
    def last_name(self) -> Optional[str]:
```

frascuchon (approved):

PR #5716, opened 2024-11-26
fix: :bug: Send visible_options prop only when the questions has more…

### LINKED ISSUE #5713: [BUG-UI/UX]  Update question titles & descriptions in UI
**Describe the bug**

Question settings cannot be updated from the dataset settings page in the UI. The `Update` button isn't working.

This seems to happen to datasets that have been imported using the dataset import from the UI.

damianpumar:
Fix https://github.com/argilla-io/argilla/issues/5713

### DIFF
--- a/argilla-frontend/v1/infrastructure/repositories/QuestionRepository.ts
+++ b/argilla-frontend/v1/infrastructure/repositories/QuestionRepository.ts
@@ -80,7 +80,12 @@ export class QuestionRepository implements IQuestionRepository {
     return {
       title,
       description: newDescription,
-      settings: settings as any,
+      settings: {
+        ...(settings as any),
+        visible_options: settings.shouldShowVisibleOptions
+          ? settings.visible_options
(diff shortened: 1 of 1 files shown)


argilla-frontend/v1/infrastructure/repositories/QuestionRepository.ts
jfcalvo, line 1: Where is the bigger than 3 check that we saw this morning?

damianpumar: https://github.com/argilla-io/argilla/blob/f4f03bb5316591f295a63f124fecee0f59990c6f/argilla-frontend/v1/domain/entities/question/QuestionSetting.ts#L47

jfcalvo: Ok, thanks!

PR #5741, opened 2024-12-10
fix: UI - use `last_activity_at` in the dataset list

leiyre:
This PR uses `last_activity_At`instead of `updated_at` in the dataset card info and in the sorting filter on the homepage

### DIFF
--- a/argilla-frontend/components/features/home/dataset-list/DatasetList.vue
+++ b/argilla-frontend/components/features/home/dataset-list/DatasetList.vue
@@ -48,10 +48,10 @@ export default {
     return {
       querySearch: "",
       sortedOrder: "desc",
-      sortedByField: "updatedAt",
+      sortedByField: "lastActivityAt",
       sortOptions: [
         { value: "name", label: this.$t("home.name") },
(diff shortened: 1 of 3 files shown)


argilla-frontend/components/features/home/dataset-list/DatasetList.vue
```diff
@@ -48,10 +48,10 @@ export default {
     return {
       querySearch: "",
       sortedOrder: "desc",
-      sortedByField: "updatedAt",
+      sortedByField: "lastActivityAt",
       sortOptions: [
         { value: "name", label: this.$t("home.name") },
-        { value: "updatedAt", label: this.$t("home.updatedAt") },
+        { value: "lastActivityAt", label: this.$t("home.updatedAt") },
```
damianpumar, line 54: I was thinking about this change, should we still having the same label?

leiyre: Yes, we have mentioned keeping “updated” to express “last_activity_at”, we can maintain the label

damianpumar:
could you please add the changelog for this ?

PR #5778, opened 2025-01-08
Fix to Return Similarity Scores Along with Records

### LINKED ISSUE #5777: Unable to get score in Similarity search
### Describe the bug

The method used to fetch vector similarity scores is returning only a list of records, rather than a list of tuples containing both the records and their corresponding similarity scores.

### Stacktrace and code to create the bug

client = rg.Argilla(api_url, api_key)
dataset = client.datasets(name="datatset_name", workspace="workspace_name")
similar = rg.Similar(name="vector", value=vector)
records = dataset.records(query=rg.Query(similar=similar), with_vectors=True)

Iterate over the above records or just try to convert into to_list, to_json or to_dict methods, which doesnot return similarity score.

### Expected behavior

The method should return a list of tuples, where each tuple contains a record and its corresponding similarity score when fetching vector similarity.

### Environment

- Argilla Version [e.g. 1.0.0]:  2.6.0
- ElasticSearch Version [e.g. 7.10.2]: 8.15.1

### Additional context

_No response_

Saikiranbonu1661:
# Description

Closes #5777 

**Type of change**

**How Has This Been Tested**

Passes similar param in query method while fetching records and could see returning similarity score.

**Checklist**

### DIFF
--- a/argilla/CHANGELOG.md
+++ b/argilla/CHANGELOG.md
@@ -16,6 +16,10 @@ These are the section headers that we use:
 
 ## [Unreleased]()
 
+### Added
+
+- Return similarity score when searching by similarity. ([#5778](https://github.com/argilla-io/argilla/pull/5778))
+
 ## [2.6.0](https://github.com/argilla-io/argilla/compare/v2.5.0...v2.6.0)
 
 ### Fixed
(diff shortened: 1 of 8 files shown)


frascuchon (approved):
Thanks for your contribution @Saikiranbonu1661 

I pushed some changes to keep the backward compatibility as much as possible. The similarity score will be returned only when a similarity search is applied. The rest of the searches/filters won't return any score.

PR #5805, opened 2025-02-06
[RELEASES] v2.7.1

frascuchon:
# Description

Release v2.7.1


argilla-frontend/v1/infrastructure/services/useAxiosExtension.ts
```diff
@@ -15,6 +15,7 @@ export const useAxiosExtension = (context: Context) => {
   const makePublic = (config: PublicAxiosConfig) => {
     const $axios = context.$axios.create({
       withCredentials: false,
+      headers: { Authorization: undefined },
```
frascuchon, line 18: In any case, I think having a separate axios config to connect with the datasets viewer API would be easier.

Issue #5845, opened 2025-10-17
[BUG-python/deployment]HF space deployment triggle runtime error


RuiaRui:
### Describe the bug

I try to create a HF space to host argilla but got below runtime error. With several restart and rebuild the error is still there. 

<img width="1447" height="702" alt="Image" src="https://github.com/user-attachments/assets/1142b5a4-8c80-4186-a0be-7b879efb5eeb" />

### Environment

- Argilla Version [e.g. 1.0.0]:
- ElasticSearch Version [e.g. 7.10.2]:
- Docker Image (optional) [e.g. argilla:v1.0.0]:

PR #5851, opened 2025-12-29
fix: resolve HuggingFace Spaces runtime error during deployment

u-sheikh:
The issue was in the username retrieval logic in start.sh. When the HuggingFace API returns an error (e.g., user not found), jq would return the literal string "null" instead of triggering the fallback to SPACE_AUTHOR_NAME.

Changes:
- Use jq's '// empty' alternative operator to return empty string instead of "null" when the .user field is missing or null
- Use bash parameter expansion to properly fallback to SPACE_AUTHOR_NAME
- Add proper quoting around the curl URL

This fixes the runtime error reported in issue #5845.

# Description

Closes #<issue_number>

**Type of change**

- Documentation update

**How Has This Been Tested**

**Checklist**

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,7 +158,7 @@ Need more info, check out [our docs](https://argilla-io.github.io/argilla/latest
 
 ## 🥇 Contributors
 
-To help our community with the creation of contributions, we have created our [community](https://argilla-io.github.io/argilla/latest/community/) docs. 
+To help our community with the creation of contributions, we have created our [community](https://argilla-io.github.io/argilla/latest/community/) docs.
 
 <a  href="https://github.com/argilla-io/argilla/graphs/contributors">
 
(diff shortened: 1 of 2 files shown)


JiwaniZakir (commented):
The two-line split in `start.sh` correctly addresses the root cause: the original `|| echo` fallback only triggered on non-zero exit codes, but `jq -r '.user'` exits 0 even when the API returns an error object (producing the literal string `"null"`). Using `.user // empty` combined with the shell `${:-}` expansion is the right approach here.

One edge case to consider: if `curl` fails silently (the `-s` flag suppresses errors) and returns empty or malformed JSON, `jq` will exit with a parse error (exit code 2). If the script runs with `set -e` or `set -o pipefail`, the failed `jq` invocation inside the command substitution could terminate the script before `DEFAULT_USERNAME="${DEFAULT_USERNAME:-$SPACE_AUTHOR_NAME}"` on the next line ever executes. It would be worth verifying the top of `start.sh` for those flags, and potentially appending `|| true` to the `jq` command to make the failure explicitly non-fatal.

The URL quoting around `"https://huggingface.co/api/users/${SPACE_CREATOR_USER_ID}/overview"` is also a good hygiene fix — if `SPACE_CREATOR_USER_ID` were ever unset or empty, the unquoted form could behave unexpectedly depending on shell globbing rules.

PR #5865, opened 2026-08-13
fix: return plain dict from ElasticSearchEngine.info()

### LINKED ISSUE #5864: [BUG] GET /api/v1/status returns 422 on Elasticsearch backend (pydantic dict_type validation)
### Describe the bug

`GET /api/v1/status` returns HTTP 422 instead of 200 when running against
an Elasticsearch backend (reproduced with `argilla/argilla-server:latest`
+ `elasticsearch:8.17.0`, the exact versions pinned in the official
`examples/deployments/docker/docker-compose.yaml`):

```json
{
  "detail": "1 validation error for Status\nsearch_engine\n  Input should be a valid dictionary [type=dict_type, input_value=ObjectApiResponse({'name'...'You Know, for Search'}), input_type=ObjectApiResponse]\n    For further information visit https://errors.pydantic.dev/2.9/v/dict_type"
}
```

**Side effect in the UI**: dataset cards on the "Your datasets" page render
empty (`<li id="...">` with no content) for otherwise perfectly valid
datasets — `GET /api/v1/me/datasets` returns the correct data, but the
dataset-card component appears to depend on the same broken status/search-engine
data path and fails silently.

### Root cause

- [`Status.search_engine`](https://github.com/argilla-io/argilla/blob/main/argilla-server/src/argilla_server/api/schemas/v1/info.py#L24) is typed as a strict pydantic `dict`.
- [`ElasticSearchEngine.info()`](https://github.com/argilla-io/argilla/blob/main/argilla-server/src/argilla_server/search_engine/elasticsearch.py#L71-L72) returns `await self.client.info()` directly — this is an `elastic_transport.ObjectApiResponse`, which is dict-*like* (supports `__getitem__`/iteration) but is **not** an instance of `dict`. Pydantic v2's strict validation rejects it.
- `OpenSearchEngine.info()` is unaffected — `opensearch-py`'s transport already deserializes to a plain `dict`.

### To Reproduce

1. `docker compose up -d` using the official `examples/deployments/docker/docker-compose.yaml` (Elasticsearch backend).
2. `curl http://localhost:6900/api/v1/status` → 422.
3. Log in to the UI and check "Your datasets" for any dataset → card renders empty despite the dataset existing and being fully queryable via the SDK/API.

### Expected behavior

`/api/v1/status` returns 200 with cluster info, and dataset cards render normally.

### Environment

- `argilla/argilla-server:latest` (pip: `argilla-server==2.8.0`)
- `elasticsearch` server: `8.17.0`, client (`elastic-transport`): `8.17.0`
- `pydantic==2.9.2`
- Reproduced on Docker (x86_64 Linux), via the official docker-compose stack

### Fix

Verified fix: unwrap via `.body` (`elastic_transport` guarantees this is a plain `dict`) in `ElasticSearchEngine.info()`. One-line change, manually verified against a live ES 8.17.0 cluster (422 → 200), plus a regression test. Opening a PR right after this issue.

St4r4x:
## Problem
Fixes #5864

`GET /api/v1/status` returns 422 on the Elasticsearch backend, and dataset
cards silently fail to render in the UI as a side effect (both depend on
the same broken data path).

## Root cause
`ElasticSearchEngine.info()` returns `elasticsearch-py`'s raw
`elastic_transport.ObjectApiResponse` directly. It behaves like a dict
(supports `__getitem__`/iteration) but is not an actual `dict` instance,
so pydantic v2's strict `Status.search_engine: dict` validation rejects it.

`OpenSearchEngine.info()` is unaffected — checked `opensearch-py`'s
transport, which already deserializes JSON responses into plain dicts.

## Fix
Unwrap via `.body`, which `elastic_transport` guarantees is a plain dict.
One-line change.

## Verification
- Reproduced against the exact stack in `examples/deployments/docker/docker-compose.yaml` (`elasticsearch:8.17.0`).
- Before: `curl localhost:6900/api/v1/status` → 422 with the `dict_type` pydantic error.
- After (patched running container): same request → 200 with full cluster info; dataset list UI renders correctly again.
- Added `test_info_returns_plain_dict` in `test_elastisearch.py`, following the existing `search_engine` fixture pattern (runs against a live engine, not mocked).

## Type of change

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,7 +158,7 @@ Need more info, check out [our docs](https://argilla-io.github.io/argilla/latest
 
 ## 🥇 Contributors
 
-To help our community with the creation of contributions, we have created our [community](https://argilla-io.github.io/argilla/latest/community/) docs. 
+To help our community with the creation of contributions, we have created our [community](https://argilla-io.github.io/argilla/latest/community/) docs.
 
 <a  href="https://github.com/argilla-io/argilla/graphs/contributors">
 
--- a/argilla-server/CHANGELOG.md
+++ b/argilla-server/CHANGELOG.md
@@ -24,6 +24,7 @@ These are the section headers that we use:
 
 - Fixed error when computing user progress with PostgreSQL database. ([#5795](https://github.com/argilla-io/argilla/pull/5795))
 - Fixed error when updating records with PostgreSQL database. ([#5795](https://github.com/argilla-io/argilla/pull/5795))
+- Fixed `GET /api/v1/status` returning 422 on the Elasticsearch backend, which also caused dataset cards to render empty in the UI. ([#5865](https://github.com/argilla-io/argilla/pull/5865)). Contributed by @St4r4x.
 
 ## [2.7.1](https://github.com/argilla-io/argilla/compare/v2.7.0...v2.7.1)
 
--- a/argilla-server/src/argilla_server/search_engine/elasticsearch.py
+++ b/argilla-server/src/argilla_server/search_engine/elasticsearch.py
@@ -69,7 +69,7 @@ class ElasticSearchEngine(BaseElasticAndOpenSearchEngine):
         return await self.client.ping()
 
     async def info(self) -> dict:
-        return await self.client.info()
+        return (await self.client.info()).body
 
     def _mapping_for_vector_settings(self, vector_settings: VectorSettings) -> dict:
         return {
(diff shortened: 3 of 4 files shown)


St4r4x:
Closing this — argilla-io/argilla appears to no longer have an active maintainer team responding to community PRs (see the README's own notice), so this isn't likely to get reviewed. Leaving the fix available if anyone picks up maintenance again.

